PR #97, opened 2018-06-05
Process procedure declarations and derived types.

tskeith:
Add ObjectEntityDetails and ProcEntityDetails to distinguish between an entity
from an object-decl and one from a proc-decl. When we don't know, it stays as
EntityDetails until it is resolved. DeclareEntity in DeclarationVisitor creates
this kind of symbol.

Add flags to Symbol as a convenient place for boolean flags common to many
kinds of symbols. Use it to mark symbols known to be functions or subroutines
so that we can report errors when they are used incorrectly. Improve handling
of EXTERNAL statement.

Handle ProcDecl nodes and add symbols for them.

Partial processing of derived types. Data component declarations are processed
and added to the derived type. Define TypeBoundProc and TypeBoundGeneric in
type.h. Procedure components, type-bound procedures, etc. are not handled yet
and nothing is done with the derived type once it is created. Eliminate
DerivedTypeDefBuilder in favor of just setting fields in derivedTypeData_.

Add GetDeclTypeSpec to go with BeginDeclTypeSpec and EndDeclTypeSpec,
to avoid directly access the private variable.

Add tests in resolve20.f90 for errors related to procedure declarations.
Add missing copyrights to other tests.

### DIFF
--- a/lib/semantics/resolve-names.cc
+++ b/lib/semantics/resolve-names.cc
@@ -124,8 +124,6 @@ class DeclTypeSpecVisitor : public AttrsVisitor {
 public:
   using AttrsVisitor::Post;
   using AttrsVisitor::Pre;
-  void BeginDeclTypeSpec();
-  void EndDeclTypeSpec();
   bool Pre(const parser::IntegerTypeSpec &);
   bool Pre(const parser::IntrinsicTypeSpec::Logical &);
   bool Pre(const parser::IntrinsicTypeSpec::Real &);
@@ -148,12 +146,16 @@ public:
   void Post(const parser::ProcedureDeclarationStmt &);
 
 protected:
-  std::unique_ptr<DeclTypeSpec> declTypeSpec_;
+  std::unique_ptr<DeclTypeSpec> &GetDeclTypeSpec();
+  void BeginDeclTypeSpec();
+  void EndDeclTypeSpec();
+
   std::unique_ptr<DerivedTypeSpec> derivedTypeSpec_;
   std::unique_ptr<ParamValue> typeParamValue_;
 
 private:
   bool expectDeclTypeSpec_{false};  // should only see decl-type-spec when true
+  std::unique_ptr<DeclTypeSpec> declTypeSpec_;
   void MakeIntrinsic(const IntrinsicTypeSpec &intrinsicTypeSpec);
   void SetDeclTypeSpec(const DeclTypeSpec &declTypeSpec);
   static KindParamValue GetKindParamValue(
@@ -423,11 +425,11 @@ private:
   // Function result name from parser::Suffix, if any.
   const parser::Name *funcResultName_{nullptr};
 
-  bool BeginSubprogram(const parser::Name &,
+  bool BeginSubprogram(const parser::Name &, Symbol::Flag,
       const std::optional<parser::InternalSubprogramPart> &);
   void EndSubprogram();
   // Create a subprogram symbol in the current scope and push a new scope.
-  Symbol &PushSubprogramScope(const parser::Name &);
+  Symbol &PushSubprogramScope(const parser::Name &, Symbol::Flag);
   Symbol *GetSpecificFromGeneric(const parser::Name &);
 };
 
@@ -458,6 +460,25 @@ public:
   }
   void Post(const parser::TargetStmt &) { objectDeclAttr_ = std::nullopt; }
   void Post(const parser::DimensionStmt::Declaration &);
+  bool Pre(const parser::TypeDeclarationStmt &) { return BeginDecl(); }
+  void Post(const parser::TypeDeclarationStmt &) { EndDecl(); }
+  bool Pre(const parser::DerivedTypeDef &x);
+  void Post(const parser::DerivedTypeDef &x);
+  bool Pre(const parser::DerivedTypeStmt &x);
+  void Post(const parser::DerivedTypeStmt &x);
+  bool Pre(const parser::TypeAttrSpec::Extends &x);
+  bool Pre(const parser::PrivateStmt &x);
+  bool Pre(const parser::SequenceStmt &x);
+  bool Pre(const parser::ComponentDefStmt &) { return BeginDecl(); }
+  void Post(const parser::ComponentDefStmt &) { EndDecl(); }
+  void Post(const parser::ComponentDecl &x);
+  bool Pre(const parser::ProcedureDeclarationStmt &);
+  void Post(const parser::ProcedureDeclarationStmt &);
+  bool Pre(const parser::ProcComponentDefStmt &);
+  void Post(const parser::ProcComponentDefStmt &);
+  void Post(const parser::ProcInterface &x);
+  void Post(const parser::ProcDecl &x);
+  bool Pre(const parser::FinalProcedureStmt &x);
 
 protected:
   bool BeginDecl();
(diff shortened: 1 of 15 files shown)


lib/semantics/resolve-names.cc
```diff
@@ -420,18 +425,127 @@ class SubprogramVisitor : public InterfaceVisitor {
...
+  bool BeginSubprogram(const parser::Name &, Symbol::Flag,
       const std::optional<parser::InternalSubprogramPart> &);
   void EndSubprogram();
   // Create a subprogram symbol in the current scope and push a new scope.
-  Symbol &PushSubprogramScope(const parser::Name &);
+  Symbol &PushSubprogramScope(const parser::Name &, Symbol::Flag);
   Symbol *GetSpecificFromGeneric(const parser::Name &);
 };
 
+class DeclarationVisitor : public ArraySpecVisitor,
```
klausler: OO design point: is there an "is a" relationship between DeclarationVisitor and its base classes, or are they being used to mix in some capabilities?  Encapsulation might be clearer in the latter case.

lib/semantics/resolve-names.cc
```diff
@@ -420,18 +425,127 @@ class SubprogramVisitor : public InterfaceVisitor {
...
+  void Post(const parser::ProcInterface &x);
+  void Post(const parser::ProcDecl &x);
+  bool Pre(const parser::FinalProcedureStmt &x);
+
+protected:
+  bool BeginDecl();
+  void EndDecl();
+
+private:
+  // The attribute corresponding to the statement containing an ObjectDecl
```
klausler: Minor: Data members (apart from constants) typically come after member functions in each accessibility section in a class definition.  I don't think that this point is in the style guide directly, though.

lib/semantics/resolve-names.cc
```diff
@@ -420,18 +425,127 @@ class SubprogramVisitor : public InterfaceVisitor {
...
+      symbol.set_details(T{});
+    } else if (auto *details = symbol.detailsIf<EntityDetails>()) {
+      if (!std::is_same<EntityDetails, T>::value) {
+        symbol.set_details(T(*details));
+      }
+    }
+    if (T *details = symbol.detailsIf<T>()) {
+      // OK
+    } else if (std::is_same<EntityDetails, T>::value &&
+        (symbol.has<ObjectEntityDetails>() ||
```
klausler: I find "symbol.has<T>()" to be a readable idiom and I like it a lot.

lib/semantics/resolve-names.cc
```diff
@@ -1539,121 +1643,81 @@ Symbol *SubprogramVisitor::GetSpecificFromGeneric(const parser::Name &name) {
...
+  } else {
+    Symbol &symbol{DeclareEntity<EntityDetails>(name, attrs)};
+    if (auto &type = GetDeclTypeSpec()) {
+      SetType(name.source, symbol, *type);
     }
   }
 }
 
-bool ResolveNamesVisitor::Pre(const parser::AsynchronousStmt &x) {
+bool DeclarationVisitor::Pre(const parser::AsynchronousStmt &x) {
```
klausler: Be advised, ASYNCHRONOUS and VOLATILE have their own peculiar behavior in BLOCK constructs and contained subprograms -- when they are the only declaration statement used on a name that is already available through host association (or whatever the right term would be in the case of BLOCK), they don't introduce a new symbol, but just give it a scoped attribute.

lib/semantics/resolve-names.cc
```diff
@@ -1706,24 +1778,220 @@ void ResolveNamesVisitor::DeclareEntity(const parser::Name &name, Attrs attrs) {
...
+bool DeclarationVisitor::Pre(const parser::FinalProcedureStmt &x) {
+  for (const parser::Name &name : x.v) {
+    derivedTypeData_->finalProcs.push_back(name.source);
+  }
+  return false;
+}
+
+void DeclarationVisitor::SetType(
+    const SourceName &name, Symbol &symbol, const DeclTypeSpec &type) {
+  if (auto *details = symbol.detailsIf<EntityDetails>()) {
```
klausler: The flow here was a little tricky to grasp at first glance but I don't have a more easily readable paradigm to suggest -- using a flag seems no better, and splitting the function into a "get type if there is one" function and a second "set type or crash" function, calling both in a wrapper, seems worse.

lib/semantics/resolve-names.cc
```diff
@@ -1782,19 +2050,30 @@ void ModuleVisitor::SetAccess(const parser::Name &name, Attr attr) {
...
+    return false;
+  } else if (const auto *details = symbol.detailsIf<EntityDetails>()) {
+    return details->type().has_value();
+  } else if (const auto *details = symbol.detailsIf<ObjectEntityDetails>()) {
+    return details->type().has_value();
+  } else if (const auto *details = symbol.detailsIf<ProcEntityDetails>()) {
+    return details->interface().symbol() != nullptr ||
+        details->interface().type() != nullptr;
+  } else {
+    return true;  // doesn't need explicit type
```
klausler: Consider reversing the sense of this predicate, maybe to StillNeedsExplicitType(), to avoid the confusion of returning true for things that don't actually have types, just to avoid future confusion when somebody writes some code that crashes if HasExplicitType is true and yet later there is no type to be found in the object.

lib/semantics/symbol.h
```diff
@@ -169,12 +201,16 @@ class GenericDetails {
...
-      SubprogramDetails, SubprogramNameDetails, EntityDetails, UseDetails,
-      UseErrorDetails, GenericDetails>;
+      SubprogramDetails, SubprogramNameDetails, EntityDetails,
+      ObjectEntityDetails, ProcEntityDetails, UseDetails, UseErrorDetails,
+      GenericDetails>;
 std::ostream &operator<<(std::ostream &, const Details &);
 
 class Symbol {
 public:
+  ENUM_CLASS(Flag, Function, Subroutine);
```
klausler: Will these flags always be exclusive, or do you need to be able to have both flags set at some intermediate stages of processing?  If they're exclusive, a three-way state might be clearer (Func, Subr, AsYetUnknown).

Sometimes, flags on symbols can get out of hand in a compiler; the representation can be inverted into sets of symbols and possibly be clearer.  I.e., instead of having symbols with flags, you have some sets of symbols.

lib/semantics/type.h
```diff
@@ -361,57 +359,87 @@ class DataComponentDef {
...
+// interface), a decl-type-spec (representing the function return type),
+// or neither.
+class ProcInterface {
+public:
+  ProcInterface() = default;
+  ProcInterface(ProcInterface &&that)
+    : symbol_{that.symbol_}, type_{std::move(that.type_)} {}
+  ProcInterface(const ProcInterface &that) : symbol_{that.symbol_} {
+    if (that.type_) {
+      *this = *that.type_;
```
klausler: Should this assignment statement be "type_ = *that.type_;"?  I don't understand what's going on otherwise.

lib/semantics/resolve-names.cc
tskeith: Yes, that's much clearer.

lib/semantics/type.h
tskeith: It's calling the operator= for the class so the logic there doesn't have to be duplicated. But it is confusing, so I'll see if I can find a clearer way to do it.

lib/semantics/resolve-names.cc
schweitzpgi: LLVM coding convention: don't use if () { return; }  else if () { return; } else return; but rather if () { return; } if () return; ... return;

lib/semantics/type.h
klausler: Got it.  A short comment on the same line would be more than enough -- I always read `operator=` as setting the whole left-hand side, and didn't consider that possibility.

lib/semantics/symbol.h
tskeith: I definitely agree we shouldn't go overboard with flags. In this case I initially had isFunction() and isSubroutine() as member functions on Symbol. A set of symbols or a function mapping symbols to booleans is another choice. They all seem logically equivalent but I thought the flags approach was simplest. I think flags should be limited to things that apply to several different kinds of symbols (in this case, subprograms, generics, procedure entities).

lib/semantics/resolve-names.cc
klausler: Thanks, Eric.  We sometimes violate that convention when every branch of a conditional construct, including its final `else` block, has a `return` statement, since the whole construct can be viewed as a big expression determining the return value, rather than one that uses `return` for control flow purposes like `break`.

lib/semantics/resolve-names.cc
tskeith: HandleAttributeStmt() has special handling for ASYNCHRONOUS and VOLATILE for symbols that are use-associated. It will need the same when host association is implemented too.

PR #99, opened 2018-06-11
Add Dump() routines for debugging.

tskeith:
Add `Dump()` routines based on `operator<<` for the type so that they are
easy to call from the debugger. Overload for both pointer and reference
types and for dumping to `std::cerr` or a specific ostream.


schweitzpgi:
LLVM uses constant dump() member functions.

tskeith:
I like having the Dump routines isolated in one so that it is clear they are only for debugging.

PR #101, opened 2018-06-13
Basic operation evaluation framework

klausler:
Here's a templatized library for evaluating the intrinsic operations of Fortran's intrinsic types, as well as some of the intrinsic functions, in maximally portable code.  There is a tiny unit testing framework and code to exercise most of the implementation.  The intended next step is to extend this code to cover the rest of the intrinsics and to add a layer for arrays, enable usage of native hardware arithmetic, and then use this one library for both compile-time and run-time evaluation.


lib/evaluate/CMakeLists.txt
```diff
@@ -0,0 +1,20 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+add_library(FortranEvaluate
+  complex.cc
+  integer.cc
+  logical.cc
+  real.cc
+)
```
tskeith: Minor: that red icon indicates there is no newline on the last line.

lib/evaluate/common.h
```diff
@@ -0,0 +1,116 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef FORTRAN_EVALUATE_COMMON_H_
+#define FORTRAN_EVALUATE_COMMON_H_
+
+#include "../semantics/enum-set.h"
```
tskeith: enum-set.h should probably move to a new library like "basic". idioms.h could go there too.
evaluate should need to depend on semantics.

lib/evaluate/integer.h
```diff
@@ -0,0 +1,853 @@
...
+// types &/or partial fields in the parts.  The radix (i.e., the number
+// of possible values in a part), however, must be a power of two; this
+// template class is not generalized to enable, say, decimal arithmetic.
+// Member functions that correspond to Fortran intrinsic functions are
+// named accordingly so that they can be referenced easily in the
+// language standard.
+template<int BITS, bool IS_LITTLE_ENDIAN = IsHostLittleEndian,
+    int PARTBITS =
+        BITS<32 ? BITS : 32, typename PART = HostUnsignedInt<PARTBITS>,
+            typename BIGPART = HostUnsignedInt<PARTBITS * 2>> class Integer {
```
tskeith: clang-format is clearly baffled by this template. I suggest replacing `BITS<32` with `BITS<=32` and it will do a better job.

lib/evaluate/integer.h
```diff
@@ -0,0 +1,853 @@
...
+      } else {
+        result.value = result.value.IOR(Integer{field}.SHIFTL(j));
+        if (bits < j + 64 && (field >> (bits - j)) != 0) {
+          result.overflow = true;
+          break;
+        }
+      }
+    }
+    return result;
+  }
```
tskeith: Here is a more concise body for ValueWithOverflow. Whether it is clearer, I'm not sure. I don't like the `break`s in the original.
```
    std::uint64_t field{that.ToUInt64()};
    ValueWithOverflow result{field, false};
    if constexpr (bits < 64) {
      result.overflow = (field >> bits) != 0;
    }
    for (int j{64}; j < that.bits && !result.overflow; j += 64) {
      field = that.SHIFTR(j).ToUInt64();
      if (bits <= j) {
        result.overflow = field != 0;
      } else {
        result.value = result.value.IOR(Integer{field}.SHIFTL(j));
        result.overflow = bits < j + 64 && (field >> (bits - j)) != 0;
      }
    }
    return result;
```

lib/evaluate/integer.h
```diff
@@ -0,0 +1,853 @@
...
+  // the shift is leftward if count is nonnegative, rightward otherwise.
+  // Higher-order bits are unchanged.
+  constexpr Integer ISHFTC(int count, int size = bits) const {
+    if (count == 0 || size <= 0) {
+      return *this;
+    }
+    if (size > bits) {
+      size = bits;
+    }
+    if ((count %= size) == 0) {
```
tskeith: I think it's clearer to separate the assignment from the condition, so:
```
count %= size;
if (count == 0) {
```

lib/evaluate/real.h
```diff
@@ -0,0 +1,307 @@
...
+  }
+  constexpr bool IsDenormal() const {
+    return Exponent() == 0 && !GetSignificand().IsZero();
+  }
+
+  constexpr Real ABS() const {  // non-arithmetic, no flags returned
+    Real result;
+    result.word_ = word_.IBCLR(bits - 1);
+    return result;
+  }
```
tskeith: Why not `return Real(word_.IBCLR(bits - 1));`?

lib/evaluate/real.h
```diff
@@ -0,0 +1,307 @@
...
+
+  constexpr Real Negate() const {
+    Real result;
+    result.word_ = word_.IEOR(word_.MASKL(1));
+    return result;
```
tskeith: Same here.

lib/evaluate/integer.h
```diff
@@ -0,0 +1,853 @@
...
+    }
+    SetLEPart(parts - 1, n);
+  }
+
+  constexpr Integer &operator=(const Integer &) = default;
+
+  // Left-justified mask (e.g., MASKL(1) has only its sign bit set)
+  static constexpr Integer MASKL(int places) {
+    if (places <= 0) {
+      return {};
```
tskeith: I find using `{}` for zero disconcerting, though maybe you get used to it.
What do you think about adding `static constexpr Integer Zero() { return {}; }` so it can be written as `Zero()`?

lib/evaluate/integer.h
```diff
@@ -0,0 +1,853 @@
...
+    Integer result;
+    bool overflow{false};
+    const char *p{pp};
+    while (*p == ' ' || *p == '\t') {
+      ++p;
+    }
+    if (*p == '+') {
+      ++p;
+    }
+    Integer radix{base};
```
sscalpone: Do we need a warning comment for people in the future with non-linear character sets?  An assertion that base is less than 37 or something?

lib/evaluate/integer.h
```diff
@@ -0,0 +1,853 @@
...
+      return Ordering::Less;
+    } else if (IsZero()) {
+      return Ordering::Equal;
+    } else {
+      return Ordering::Greater;
+    }
+  }
+
+  // Count the number of contiguous most-significant bit positions
+  // that are clear.
```
sscalpone: If this is supposed to match the semantics of Fortran, you might say so in the comment.  Same for similar fcns.

lib/evaluate/type.h
```diff
@@ -0,0 +1,107 @@
...
+  using ValueType = std::uint8_t[kind];  // TODO: ?
+};
+
+// Default REAL just simply has to be IEEE-754 single precision today.
+// It occupies one numeric storage unit by definition.  The default INTEGER
+// and default LOGICAL intrinsic types also have to occupy one numeric
+// storage unit, so their kinds are also forced.  Default COMPLEX occupies
+// two numeric storage units.
+
+using DefaultReal = Real<4>;
```
sscalpone: How will this work with the command line option that changes default real to be Real<8>?

CMakeLists.txt
```diff
@@ -97,8 +97,16 @@ include_directories(BEFORE
...
 configure_file(
   ${FLANG_SOURCE_DIR}/include/flang/Config/config.h.cmake
   ${FLANG_BINARY_DIR}/include/flang/Config/config.h)
+
+enable_testing()
+add_test(NAME Leadz COMMAND leading-zero-bit-count-test)
+add_test(NAME PopPar COMMAND bit-population-count-test)
+add_test(NAME Integer COMMAND integer-test)
+add_test(NAME Logical COMMAND logical-test)
+add_test(NAME Real COMMAND real-test)
```
tskeith: Can the `add_test` commands go in `test/evaluate/CMakeLists.txt`? It seems like the more natural place for them.

tskeith: It seems to work if `enable_testing()` is moved before `add_subdirectory`.

lib/evaluate/CMakeLists.txt
klausler: Done, thanks.

lib/evaluate/common.h
klausler: "should need to depend on semantics" should be "should *not* need to depend on semantics" above, yes?

I'll collect likely candidates for a common utility package and sweep them together.  Headers that don't depend on anything else in f18 will go there.

tskeith: Yes, should not.

That sounds like a good plan.

lib/evaluate/integer.h
klausler: Thanks.

lib/evaluate/integer.h
klausler: Ok.

lib/evaluate/real.h
klausler: I didn't initially have a Real() constructor that accepted a Word argument.  Will reduce.

lib/evaluate/real.h
klausler: Right.

lib/evaluate/integer.h
klausler: I guess that I am used to reading `{}` as default construction.  How about `{0}`?

lib/evaluate/integer.h
klausler: disclaimer added

lib/evaluate/integer.h
klausler: There's a comment to that effect above the beginning of the class template definition.

lib/evaluate/type.h
klausler: We need to discuss that option and whether it's valid in the context of the current standard and what its consequences would be for storage association -- would it also change the byte size of a single numeric storage unit, and hence the size of default integer and logical?

lib/evaluate/integer.h
tskeith: Given that we can't do the same with Real (i.e. we can't use `{0.0}`), I think leave it as `{}`.

lib/evaluate/type.h
sscalpone: PGI, gfortran, and ifort have -r8 and similar options that control other default real and integer (and logical) sizes.

klausler: Yes, I know.  What I don't know is the extent to which those options pose problems for numeric storage association, which requires that default real and default integer and default logical all occupy a single numeric storage unit by definition.  I also don't know whether they affect explicitly kinded declarations (e.g. `REAL(KIND=4)` gets silently modified) or just default-kinded declarations and constants.  I won't be able to answer your original question without doing more research.

PR #107, opened 2018-06-19
Change how memory for Symbol instances is managed.

tskeith:
With this change, all instances `Symbol` are stored in `class Symbols`.
`Scope.symbols_`, which used to own the symbol memory, now maps names to
`Symbol*` instead. This causes a bunch of reference-to-pointer changes
because of the change in type of key-value pairs. It also requires a
default constructor for `Symbol`, which means `owner_` can't be a reference.

`Symbols` manages `Symbol` instances by allocating a block of them at a time
and returning the next one when needed. They are never freed.

The reason for the change is that there are a few cases where we need
to have a two symbols with the same name, so they can't both live in
the map in `Scope`. Those are:
1. When there is an erroneous re-declaration of a name we may delete the
   first symbol and replace it with a new one. If we have saved a pointer
   to the first one it is now dangling. This can be seen by running
   `f18 -fdebug-dump-symbols -fparse-only test/semantics/resolve19.f90`
   under valgrind. Subroutine s is declared twice: each results in a
   scope that contains a pointer back to the symbol for the subroutine.
   After the second symbol for s is created the first is gone so the
   pointer in the scope is invalid.
2. A generic and one of its specifics can have the same name. We currently
   handle that by moving the symbol for the specific into a `unique_ptr`
   in the generic. So in that case the symbol is owned by another symbol
   instead of by the scope. It is simpler if we only have to deal with
   moving the raw pointer around.
3. A generic and a derived type can have the same name. This case isn't
   handled yet, but it can be done like no. 2 above. It's more complicated
   because the derived type and the generic can be declared in either
   order.

### DIFF
--- a/lib/semantics/resolve-names.cc
+++ b/lib/semantics/resolve-names.cc
@@ -298,7 +298,7 @@ public:
     if (!symbol) {
       const auto pair = CurrScope().try_emplace(name, attrs, details);
       CHECK(pair.second);  // name was not found, so must be able to add
-      return pair.first->second;
+      return *pair.first->second;
     }
     symbol->add_occurrence(name);
     if (symbol->CanReplaceDetails(details)) {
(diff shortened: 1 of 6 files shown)


tskeith:
The "Fix bug in adding symbols to parse tree" commit isn't closely tied to the other one -- I just ran into while doing the refactoring. It can go in as a separate pull request if we decide not to do this one.

tskeith:
I considered the approach of moving deleted symbols to another collection, but their addresses would change so pointers would still have to be updated.

Any alternatives are welcome.

PR #109, opened 2018-06-22
Name resolution for derived types.

tskeith:
This consists of:
- a new kind of symbols to represent them with `DerivedTypeDetails`
- creating symbols for derived types when they are declared
- creating a new kind of scope for the type to hold component symbols
- resolving entity declarations of objects of derived type
- resolving references to objects of derived type and to components
- handling derived types with same name as generic

Type parameters are not yet implemented.

Refactor `DeclTypeSpec` to be a value class wrapping an IntrinsicTypeSpec
or a `DerivedTypeSpec` (or neither in the `TypeStar` and `ClassStar` cases).
Store `DerivedTypeSpec` objects in a new structure the current scope
using `MakeDerivedTypeSpec` so that `DeclTypeSpec` can just contain a pointer to
them, as it currently does for intrinsic types.

In `GenericDetails`, add `derivedType` field to handle case where generic
and derived type have the same name. The generic is in the scope and the
derived type is referenced from the generic, similar to the case where a
generic and specific have the same name. When one of these names is
mis-recognized, we sometimes have to fix up the 'occurrences' lists
of the symbols.

Assign implicit types as soon as an entity is encountered that requires
one. Otherwise implicit derived types won't work. When we see `x%y` we
have to know the type of `x` in order to resolve `y`. Add `Implicit` flag
to mark symbols that were implicitly typed

For symbols that introduce a new scope, include a pointer back to that
scope.

Add `CurrNonTypeScope()` for the times when we want the current scope but
ignoring derived type scopes. For example, that happens when looking for
types or parameters, or creating implicit symbols.

### DIFF
--- a/lib/semantics/resolve-names.cc
+++ b/lib/semantics/resolve-names.cc
@@ -148,7 +148,7 @@ protected:
   void BeginDeclTypeSpec();
   void EndDeclTypeSpec();
 
-  std::unique_ptr<DerivedTypeSpec> derivedTypeSpec_;
+  std::optional<DerivedTypeSpec> derivedTypeSpec_;
   std::unique_ptr<ParamValue> typeParamValue_;
 
 private:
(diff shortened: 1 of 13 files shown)


tskeith:
Implicit types are supposed to be set the first time a symbol is referenced (in a DataRef or ProcedureDesignator). My assumption is that its type can't be specific after that. Is that wrong?

klausler:
That's also my understanding.  I misunderstood the description of the changes, which seemed to me to say that the implicit type was being assigned earlier than the first reference; I see that it's not.

PR #111, opened 2018-06-22
Expressions

klausler:
Here's a representation for (some) Fortran expressions suitable for eventual use in the symbol table, in the parse tree as a product of semantic analysis, and in the action statements of the control flow graph.  It's type-safe, supports immutable value semantics as well as mutation in place, and supports both move construction and deep copies of subexpressions.

The classes are distinguished by the type categories (integer, real, &c.) and templatized over their kinds.

It's just operations and comparisons now, no variables or function references yet.


documentation/ControlFlowGraph.md
```diff
@@ -0,0 +1,157 @@
...
+
+### Expressions and Statements (Operations and Actions)
+Expressions are trees, not DAGs, of intrinsic operations,
+resolved function references, constant literals, and
+data designators.
+
+Expression nodes are represented in the compiler in a type-safe manner.
+There is a general `Expression` class template that is parameterized
+over type classification (e.g., `INTEGER`, `REAL`, &c.) and, for intrinsic
+types, kind type parameter values.
```
tskeith: I think this is out of date. It's now IntegerExpr, RealExpr, ComplexExpr, CharacterExpr.

klausler: Right, will update.

lib/evaluate/expression.cc
```diff
@@ -0,0 +1,289 @@
...
+}
+
+template<int KIND>
+std::ostream &IntegerExpr<KIND>::Dump(std::ostream &o) const {
+  std::visit(
+      common::visitors{[&](const Constant &n) { o << n.SignedDecimal(); },
+          [&](const Convert &c) { c.x->Dump(o); },
+          [&](const Parentheses &p) { p.x->Dump(o << '(') << ')'; },
+          [&](const Negate &n) { n.x->Dump(o << "(-") << ')'; },
+          [&](const Add &a) { a.y->Dump(a.x->Dump(o << '(') << '+') << ')'; },
```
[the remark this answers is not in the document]
klausler: Sure.  Will do, thanks.

lib/evaluate/common.h
```diff
@@ -83,7 +79,7 @@ template<typename A> struct ValueWithRealFlags {
   RealFlags flags;
 };
 
-enum class Rounding { TiesToEven, ToZero, Down, Up, TiesAwayFromZero };
+ENUM_CLASS(Rounding, TiesToEven, ToZero, Down, Up, TiesAwayFromZero)
 
 #if __BYTE_ORDER__ == __ORDER_BIG_ENDIAN__
```
sscalpone: Who defines __BYTE_ORDER__?  I would be nice to know that it is the host byte order and not the target byte order.  IsHostLIttleEndian is clear :)

klausler: It comes from the host C++ compilation environment.

PR #112, opened 2018-06-26
Support unparse with symbol annotations.

tskeith:
When `-fdebug-dump-symbols` is supplied with `-funparse`, include symbol
information in comments in the Fortran output. This will be used for
testing to verify that correct symbols are defined and references in
the right place.

In `UnparseWithSymbols()`, walk the parse tree and collect symbol
definitions and references, organized by statement. When a symbol is
defined across several statement it is associated with the first.
The definition of implicitly defined symbols is associated with the
first reference.

To write out the symbol information, a new optional argument is added to
`Unparse()`: it is a function that is called immediately before each
statement is unparsed. We pass in a function that prints out the symbol
information collected for that statement.

Add `Symbol::GetType()` to make it easier to write the symbol types
and add `Symbol::SetType()` for uniformity.

Here is an example. When the following is compiled with `-funparse -fdebug-dump-symbols`
the output is identical to the input (i.e. the comments are stripped out but the
symbol information is added back in during unparsing -- it's not just `cat`).
```
!DEF: /sub Subprogram
!DEF: /sub/y Entity REAL(8)
subroutine sub (y)
 real(kind=8) y
 !DEF: /sub/i ALLOCATABLE, VOLATILE ObjectEntity INTEGER
 allocatable :: i
 !DEF: /sub/j Entity INTEGER
 integer i, j
 volatile :: i
 !DEF: /sub/t DerivedType
 type :: t
  !DEF: /sub/t/i ObjectEntity INTEGER
  integer :: i
 end type
 !DEF: /sub/x Entity TYPE(t)
 !REF: /sub/t
 type(t) :: x
 !REF: /sub/x
 !REF: /sub/t/i
 x%i = 1
 !DEF: /sub/k (implicit) ObjectEntity INTEGER
 !REF: /sub/i
 !REF: /sub/j
 i = j+i+k
end subroutine
```

### DIFF
--- a/lib/semantics/unparse-with-symbols.cc
+++ b/lib/semantics/unparse-with-symbols.cc
@@ -1,0 +1,154 @@
+// Copyright (c) 2018, NVIDIA CORPORATION.  All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "unparse-with-symbols.h"
+#include "symbol.h"
+#include "../parser/parse-tree-visitor.h"
+#include "../parser/parse-tree.h"
+#include "../parser/unparse.h"
+#include <map>
+#include <ostream>
+#include <set>
+
+namespace Fortran::semantics {
+
+// Walk the parse tree and collection information about which statements
+// define and reference symbols. Then PrintSymbols outputs information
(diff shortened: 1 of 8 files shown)


lib/semantics/unparse-with-symbols.cc
```diff
@@ -0,0 +1,154 @@
...
+}
+
+void UnparseWithSymbols(std::ostream &out, const parser::Program &program,
+    parser::Encoding encoding) {
+  SymbolDumpVisitor visitor;
+  parser::Walk(program, visitor);
+  parser::preStatementType preStatement{
+      [&](const parser::CharBlock &location, std::ostream &out, int indent) {
+        visitor.PrintSymbols(location, out, indent);
+      }};
```
tskeith: It should be possible to define `preStatement` as a pointer to the member function with `std:bind`, but my attempt to do so didn't compile.

klausler:
Doesn't `-fdebug-dump-symbols` already do something else?  Maybe I'm misremembering.  But if I'm not, then it would be better to use a distinct command line option like `-funparse-with-symbols`.

tskeith:
OK, done.

PR #114, opened 2018-07-05
Check consistency of generic interfaces.

tskeith:
The specific procedures in a generic interface must be all functions or
all subroutines. If there is a derived type with the same name they can
only be functions.

### DIFF
--- a/lib/semantics/resolve-names.cc
+++ b/lib/semantics/resolve-names.cc
@@ -418,6 +418,7 @@ protected:
   void AddToGeneric(const Symbol &symbol);
   // Add to generic the symbol for the subprogram with the same name
   void SetSpecificInGeneric(Symbol &symbol);
+  void CheckGenericProcedures(Symbol &);
 
 private:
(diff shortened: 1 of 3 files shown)


lib/semantics/resolve-names.cc
```diff
@@ -1487,6 +1488,43 @@ void InterfaceVisitor::SetSpecificInGeneric(Symbol &symbol) {
   genericSymbol_->details<GenericDetails>().set_specific(symbol);
 }
 
+// Check that the specific procedures are all functions or all subroutines.
+// If there is a derived type with the same name they must be functions.
+// Set the corresponding flag on generic.
```
sscalpone: Do we want references to the fortran standard doc in comments like this?

tskeith: It would probably be useful. We need to decide how to write the reference to the standard. It should probably indicate which version of the standard we are referring to.

The requirement on line 1492 came from the 2018 draft standard, section 15.4.3.4.1, paragraph 4.

PR #117, opened 2018-07-09
Variables

klausler:
Define representation for variable accesses, hook it into integer expressions.  As is the case with expressions, the representation supports both copy and move semantics.  More to come, including unit testing.

### DIFF
--- a/documentation/FortranForCProgrammers.md
+++ b/documentation/FortranForCProgrammers.md
@@ -1,0 +1,254 @@
+<!--
+Copyright (c) 2018, NVIDIA CORPORATION.  All rights reserved.
+-->
+
+Fortran For C Programmers
+=========================
+
+This note is limited to essential information about Fortran so that
+a C or C++ programmer can get started quickly with the language,
+at least as a reader, and avoid some common pitfalls when starting
+to write or modify Fortran code.
+Please see other sources to learn about Fortran's rich history,
+current applications, and modern best practices in new code.
+
+Know This At Least
+------------------
+* There have been many implementations of Fortran, often from competing
+  vendors, and the standard language has been defined by U.S. and
+  international standards organizations.  The various editions of
+  the standard are known as the '66, '77, '90, '95, 2003, 2008, and
+  (real soon now) 2018 standards.
+* Forward compatibility is important.  Fortran has outlasted many
+  generations of computer systems hardware and software.  Standard
+  compliance notwithstanding, Fortran programmers generally expect that
+  code that has compiled successfully in the past will continue to
+  compile and work indefinitely.  The standards sometimes designate
+  features as being deprecated, obsolescent, or even deleted, but that
+  can be read only as discouraging their use in new code -- they'll
+  probably always work in any serious implementation.
+* Fortran has two source forms, which are typically distinguished by
+  filename suffixes.  `foo.f` is old-style "fixed-form" source, and
+  `foo.f90` is new-style "free-form" source.  All language features
+  are available in both source forms.  Neither form has reserved words
+  in the sense that C does.  Spaces are not required between tokens
+  in fixed form, and case is not significant in either form.
+* Variable declarations are optional by default.  Variables whose
+  names begin with the letters `I` through `N` are implicitly
+  `INTEGER`, and others are implicitly `REAL`.  These implicit typing
+  rules can be changed in the source.
+* Fortran uses parentheses for both array references and for function calls.
+  All arrays must be declared as such; other names followed by parenthesized
+  expressions are assumed to be function calls.
+* Fortran has a _lot_ of built-in "intrinsic" functions.  They are always
+  available without a need to declare or import them.  Their names reflect
+  the implicit typing rules, so you will encounter names that have been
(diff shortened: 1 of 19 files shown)


documentation/FortranForCProgrammers.md
```diff
@@ -0,0 +1,256 @@
+<!--
+Copyright (c) 2018, NVIDIA CORPORATION.  All rights reserved.
+-->
+
+Fortran For C Programmers
```
tskeith: I think this is really useful.

It's hard to know how much to include, but I think it would be worth saying more about modules and mentioning USE statements. E.g. it is surprising there are no "fully-qualified" names to reference entities in another module.

lib/common/indirection.h
```diff
@@ -18,14 +18,55 @@
...
+#include <type_traits>
 #include <utility>
 
 namespace Fortran::common {
 
-template<typename A> class Indirection {
+// The default case does not support (deep) copy construction and assignment.
+template<typename A, bool COPY = false> class Indirection {
+public:
+  using element_type = A;
```
tskeith: What is the purpose of element_type?

lib/evaluate/expression.cc
```diff
@@ -13,7 +13,9 @@
 // limitations under the License.
 
 #include "expression.h"
+#include "variable.h"
 #include "../common/idioms.h"
+#include "../parser/characters.h"
```
tskeith: Should characters.h move to common eventually?

tskeith:
Looks good

documentation/FortranForCProgrammers.md
```diff
@@ -0,0 +1,254 @@
...
+`CHARACTER` variables are fixed-length strings and they get padded out
+with space characters when not completely assigned.
+
+User-defined ("derived") data types can be synthesized from the intrinsic
+types and from previously-defined user types, much like a C `struct`.
+Derived types can be parameterized with integer values that either have
+to be constant at compilation time ("kind" parameters) or deferred to
+execution ("len" parameters).
+
+Derived types can inherit ("extend") from one other derived type, no more.
```
sscalpone: I read "no more" as in "this is not longer in the language".  Perhaps write "from exactly one other derived type" or something else.

documentation/FortranForCProgrammers.md
```diff
@@ -0,0 +1,254 @@
...
+They can have user-defined destructors (`FINAL` procedures).
+They can specify default initial values for their components.
+With some work, one can also specify a general constructor function,
+since Fortran allows a generic interface to have the same name as that
```
sscalpone: Perhaps define "generic" and "generic interface" somewhere above.

documentation/FortranForCProgrammers.md
```diff
@@ -0,0 +1,254 @@
...
+Variable initializers, e.g. `INTEGER :: J=123`, are _static_ initializers!
+They imply that the variable is stored in static storage, not on the stack,
+and the initialized value lasts only until the variable is assigned.
+One must use an assignment statement to implement a dynamic initializer
+that will apply to every fresh instance of the variable.
+(Derived type component initializers, however, do work as expected.)
+
+If you see an assignment to an array that's never been declared as such,
+it's probably a definition of a "statement function", which is like
+a parameterized macro definition, e.g. "A(X)=SQRT(X)**3".
```
sscalpone: It appears that "**" introduces bold-face in markup.

documentation/FortranForCProgrammers.md
```diff
@@ -0,0 +1,254 @@
...
+In the original Fortran language, this was the only means for user
+function definitions.
+Today, of course, one should use an external or internal function instead.
+
+Fortran expressions don't bind exactly like C's do.
+Watch out for exponentiation with `**`, which of course C lacks; it
```
sscalpone: And this ended the bolding.  Note that putting ** in code blocks does not cause bolding.  `**1`

lib/common/indirection.h
klausler: Reflection.  I generally save template type parameters as nested typedefs in case it's necessary to extract them later in another piece of code via `typename decltype(x)::element_type`.  I use the name `element_type` since the STL uses that name for this same purpose in many class templates, including `unique_ptr`, of which Indirection is meant to be a better version.

lib/evaluate/expression.cc
klausler: It could, yes.

documentation/FortranForCProgrammers.md
klausler: Can do.

documentation/FortranForCProgrammers.md
klausler: Rewording.

documentation/FortranForCProgrammers.md
klausler: Changed quotes.

documentation/FortranForCProgrammers.md
klausler: Ok.

PR #120, opened 2018-07-12
Support intent-stmt and resolve subprogram prefixes and suffixes.

tskeith:
Recognize `IntentStmt` and use `HandleAttributeStmt()` to implement it
as is done with other attribute statements. Add `Attr::INTENT_INOUT` as
a separate attribute for `INTENT(INOUT)`.

Collect attributes from the prefix and suffix of `FunctionStmt` and
`SubroutineStmt` (including `BIND(C)`) and set them on the subprogram
symbol.

Create a test for this using `test_symbol.sh`. It compiles with
`-funparse-with-symbols` and compares the output with the symbols in
comments in the input.

Change `test_errors.sh` to be similar to `test_symbol.sh`: check usage
and allow `F18` environment variable to override the path to the
compiler.


lib/semantics/attr.h
```diff
@@ -25,10 +25,10 @@ namespace Fortran::semantics {
 
 // All available attributes.
 ENUM_CLASS(Attr, ABSTRACT, ALLOCATABLE, ASYNCHRONOUS, BIND_C, CONTIGUOUS,
-    DEFERRED, ELEMENTAL, EXTERNAL, IMPURE, INTENT_IN, INTENT_OUT, INTRINSIC,
-    MODULE, NON_OVERRIDABLE, NON_RECURSIVE, NOPASS, OPTIONAL, PARAMETER, PASS,
-    POINTER, PRIVATE, PROTECTED, PUBLIC, PURE, RECURSIVE, SAVE, TARGET, VALUE,
-    VOLATILE)
+    DEFERRED, ELEMENTAL, EXTERNAL, IMPURE, INTENT_IN, INTENT_OUT, INTENT_INOUT,
```
sscalpone: Are we going for alphabetical order ?

tskeith: Yes, I'll fix it.

PR #122, opened 2018-07-12
Fix two problems described in issue#121

klausler:
Allow empty statements to appear as a language extension.
Improve error messages in error recovery situations.

### DIFF
--- a/documentation/FortranForCProgrammers.md
+++ b/documentation/FortranForCProgrammers.md
@@ -63,6 +63,7 @@ in particular ways that might be unfamiliar.
 | Deferred | Some attribute of a variable that is not known until an allocation or assignment |
 | Derived type | C++ class |
 | Dummy argument | C++ reference argument |
+| Final procedure | C++ destructor |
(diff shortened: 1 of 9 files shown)


tskeith:
To test the better error recovery I tried this example:
```
module foo
  call bar
end
```
The compiler output is:
```
foo.f90:2:5: error: expected declaration construct
    call bar
      ^
foo.f90:2:3: in the context: declaration construct
    call bar
    ^
foo.f90:2:3: in the context: specification part
foo.f90:1:1: in the context: module
  module foo
  ^
```
This is much better than before where it saw the "c" of "call" and expected "codimension".

But two questions:
1. Why is it pointing at column 5 for the error?
2. Can we omit the first "in the context" message? It doesn't seem to add anything.

klausler:
1. It got two letters into some possible keywords (CHARACTER, COMPLEX, CONTIGUOUS, CODIMENSION, maybe others) before failing an alternative.  I'll fix that with a backtracking point.

2. Will tweak, should be possible.

PR #126, opened 2018-07-16
Implement writing of .mod files.

tskeith:
Module file writing is implemented in `mod-file.cc`. They need to be
written after all semantic checking. Until then, for testing, write
them out whenever names are resolved.

There is a header comment in the .mod files but it is mostly a
placeholder until we can read them in and do something with it.

Rename `Symbol::details<D>` to `Symbol::get<D>`. This asserts that the
details of the symbol match D and returns that type. But we need a way
to access the details as a variant as well (not just one of its types).
`details()` is the best name for that, especially as we already have
`set_details()`. Renaming the old `details` to `get` also better matches
`has` which is used to check which variant is present.

Include tests and documentation.

### DIFF
--- a/documentation/mod-files.md
+++ b/documentation/mod-files.md
@@ -1,0 +1,124 @@
+# Module Files
+
+Module files hold information from a module that is necessary to compile 
+program units that depend on the module.
+
+## Name
+
+Module files must be searchable by module name. They are typically named
+`<modulename>.mod`. The advantage of using `.mod` is that it is consistent with
+other compilers so users will know what they are. Also, makefiles and scripts
+often use `rm *.mod` to clean up.
+
+The disadvantage of using the same name as other compilers is that it is not
+clear which compiler created a `.mod` file and files from multiple compilers
+cannot be in the same directory. This could be solved by adding something
+between the module name and extension, e.g. `<modulename>-f18.mod`.
+
+## Format
+
+The proposed format for module files is a Fortran source.
+Declarations of all visibile entities will be included, along with private
+entities that they depend on. Executable statements will be omitted.
+
+### Header
+
+There will be a header containing extra information that cannot be expressed
+in Fortran. This will take the form of a comment or directive
+at the beginning of the file.
+
+If it's a comment, the module file reader would have to strip it out and
+perform *ad hoc* parsing on it. If it's a directive the compiler could
+parse it like other directives as part of the grammar.
+Processing the header before parsing might result in better error messages
+when the `.mod` file is invalid.
+
+Regardless of whether the header is a comment or directive we can use the
+same string to introduce it: `!mod$`.
+
+Information in the header:
+- Magic string to confirm it is an f18 `.mod` file
+- Version information: to indicate the version of the file format, in case it changes,
+  and the version of the compiler that wrote the file, for diagnostics.
+- Checksum of the body of the current file
+- Modules we depend on and the checksum of their module file when the current
+  module file is created
+- Source file dependency information?
+- Compilation options?
+
+### Body
+
+The body will consist of minimal Fortran source for the required declarations.
+The order will match the order they first appeared in the source.
+
+Some normalization will take place:
+- extraneous spaces will be removed
+- implicit types will be made explicit
+- attributes will be written in a consistent order
+- entity declarations will be combined into a single declaration
+- function return types specified in a *prefix-spec* will be replaced by
+  an entity declaration
+- etc.
+
+#### Symbols included
+
+All public symbols from the module need to be included.
+
+In addition, some private symbols are needed:
+- private types that appear in the public API
+- private components of non-private derived types
+- private parameters used in non-private declarations (initial values, kind parameters)
+- others?
+
+It might be possible to anonymize private names if users don't want them exposed
+in the `.mod` file. (Currently they are readable in PGI `.mod` files.)
+
+#### USE associate
+
+A module that contains `USE` statements needs them represented in the
+`.mod` file.
+Each use-associated symbol will be written as a separate *use-only* statement,
+possibly with renaming.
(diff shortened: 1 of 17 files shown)


documentation/mod-files.md
```diff
@@ -0,0 +1,124 @@
...
+often use `rm *.mod` to clean up.
+
+The disadvantage of using the same name as other compilers is that it is not
+clear which compiler created a `.mod` file and files from multiple compilers
+cannot be in the same directory. This could be solved by adding something
+between the module name and extension, e.g. `<modulename>-f18.mod`.
+
+## Format
+
+The proposed format for module files is a Fortran source.
```
klausler: s/ a //

lib/semantics/mod-file.cc
```diff
@@ -0,0 +1,356 @@
...
+using namespace parser::literals;
+
+class ModFileWriter {
+public:
+  // The initial characters of a file that identify it as a .mod file.
+  static constexpr auto magic{"!mod$"};
+  static constexpr auto extension{".mod"};
+
+  // The .mod file format version number.
+  void set_version(int version) { version_ = version; }
```
klausler: These might be better expressed as arguments to a constructor.

tskeith: Yes, I think we should revisit the API when we know more about how it will be used.

documentation/mod-files.md
```diff
@@ -0,0 +1,124 @@
...
+## Name
+
+Module files must be searchable by module name. They are typically named
+`<modulename>.mod`. The advantage of using `.mod` is that it is consistent with
+other compilers so users will know what they are. Also, makefiles and scripts
+often use `rm *.mod` to clean up.
+
+The disadvantage of using the same name as other compilers is that it is not
+clear which compiler created a `.mod` file and files from multiple compilers
+cannot be in the same directory. This could be solved by adding something
```
sscalpone: GCC implements -J, which tells the compiler where to put mod files.  -I sets up the search path for modules.  Directories specified in -J are also added to the search path.  PGI has -module, which adds to the search path; the docs didn't say that -module is also the destination directory for modules; I did not try it.

documentation/mod-files.md
```diff
@@ -0,0 +1,124 @@
...
+between the module name and extension, e.g. `<modulename>-f18.mod`.
+
+## Format
+
+The proposed format for module files is Fortran source.
```
sscalpone: This is the actual format, not the proposed format.  You might want to say something like "The format for representing information in a module file is stylized Fortran source."

documentation/mod-files.md
```diff
@@ -0,0 +1,124 @@
...
+- Checksum of the body of the current file
+- Modules we depend on and the checksum of their module file when the current
+  module file is created
+- Source file dependency information?
+- Compilation options?
+
+### Body
+
+The body will consist of minimal Fortran source for the required declarations.
+The order will match the order they first appeared in the source.
```
sscalpone: Why?

documentation/mod-files.md
```diff
@@ -0,0 +1,124 @@
...
+
+Some normalization will take place:
+- extraneous spaces will be removed
+- implicit types will be made explicit
+- attributes will be written in a consistent order
+- entity declarations will be combined into a single declaration
+- function return types specified in a *prefix-spec* will be replaced by
+  an entity declaration
+- etc.
```
sscalpone: If an compiler-generated msg wants to refer to something in the module, it would be nice to say "foo at line 12 in module foo, file ../foo.f90"

documentation/mod-files.md
```diff
@@ -0,0 +1,124 @@
...
+
+## Reading and writing module files
+
+A command-line option (e.g. `-module`) will specified a directory to
+search for `.mod` files and to write them to.
+If not specified it defaults to the current directory.
+
+### Writing modules files
+
+When writing a module file, if the existing one matches what would be written,
```
sscalpone: If someone changes there command-line options, e..g adds -v, and the command line is recorded in the module file, does that constitute a change?  I'd say not, even if tools or eagle-eyed users might miss that the contents of the file changed without changing the date.  Alt. is that we do not keep per-compilation metadata in the mod file.

documentation/mod-files.md
```diff
@@ -0,0 +1,124 @@
...
+the timestamp is not updated.
+
+Module files will be written after semantics, i.e. after the compiler has
+determined the module is valid Fortran.<br>
+**NOTE:** PGI does create `.mod` files sometimes even when the module has a
```
sscalpone: Should we remove the mod file if the source file contains error?  Probably not.

documentation/mod-files.md
```diff
@@ -0,0 +1,124 @@
...
+compilation error.
+
+When the compiler can get far enough to determine it is compiling a module
+but then encounters an error, it will delete the existing `.mod` file
```
sscalpone: Oh, ok.  Does a C compiler delete the .o if there's a syntax error?

documentation/mod-files.md
tskeith: OK

documentation/mod-files.md
tskeith: I'll figure out what -module does and add some information about the options.

documentation/mod-files.md
tskeith: The order of declarations matters in some cases. For example, you can't use type(t) before t is declared.

documentation/mod-files.md
tskeith: OK. PGI also writes the source file name into the .mod file. Do we want that? I think it would only be for informational purposes, or maybe error messages. But if .mod files get copied around it may no longer be accurate.

PR #127, opened 2018-07-17
Use brace initialization with auto

tskeith:
Like PR #119 but applied to lib/semantics and tools/f18.

### DIFF
--- a/lib/semantics/mod-file.cc
+++ b/lib/semantics/mod-file.cc
@@ -268,7 +268,8 @@ void ModFileWriter::PutUse(const Symbol &symbol) {
 
 // We have "USE local => use" in this module. If attr was added locally
 // (i.e. on local but not on use), also write it out in the mod file.
-void ModFileWriter::PutUseExtraAttr(Attr attr, const Symbol &local, const Symbol &use) {
+void ModFileWriter::PutUseExtraAttr(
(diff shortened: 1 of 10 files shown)


lib/semantics/mod-file.cc
```diff
@@ -268,7 +268,8 @@ void ModFileWriter::PutUse(const Symbol &symbol) {
 
 // We have "USE local => use" in this module. If attr was added locally
 // (i.e. on local but not on use), also write it out in the mod file.
-void ModFileWriter::PutUseExtraAttr(Attr attr, const Symbol &local, const Symbol &use) {
+void ModFileWriter::PutUseExtraAttr(
```
sscalpone: Why would this change?

tools/f18/dump.cc
```diff
@@ -27,7 +27,10 @@
   } \
   void Dump(std::ostream &os, const ns::name &x) { os << x << '\n'; } \
   void Dump(std::ostream &os, const ns::name *x) { \
-    if (x == nullptr) os << "null\n"; else Dump(os, *x); \
+    if (x == nullptr) \
```
sscalpone: Needs braces?

lib/semantics/mod-file.cc
tskeith: I reformatted after making my changes and this line had not been correctly formatted.

PR #132, opened 2018-07-19
Support writing interfaces in module files.

tskeith:
Write symbols for external subprogram interfaces as interface-stmts.
Those go in the decls part of the module file, as opposed to contained
subprograms which go in the contains part. See modfile06.f90.

Write symbols with GenericDetails to module files. The specific
procedures of a generic interface are always written as procedure-stmts.
If they also have specific interfaces those are written in a separate
interface-stmt. See modfile07.f90.

Fix a bug where `real, external :: f` was not written like
`real f; external f`. We have to notice the EXTERNAL attribute on the
type-declaration-stmt and convert the entity to a procedure entity.
See modfile08.f90.

Fix a bug where a use-associated symbol is referenced in a
procedure-designator. We were not resolving that correctly.

Change ModFileWriter::PutEntity to include the kind of Details when
it reports an internal error due to a kind it can't handle.
Make DetailsToString public to support that.

Change test_errors.sh to fail if the f18 command exits due to a signal.
We were missing bugs where the correct errors were written out but then
module file writing crashed (due to failure to handle generics mentioned
above). Non-zero exit status is okay because we are expecting
compilation errors.

Change test_modfile.sh to allow for the expected module file contents to
be indented so the tests are easier to read.

### DIFF
--- a/test/semantics/modfile07.f90
+++ b/test/semantics/modfile07.f90
@@ -1,0 +1,69 @@
+! Copyright (c) 2018, NVIDIA CORPORATION.  All rights reserved.
+!
+! Licensed under the Apache License, Version 2.0 (the "License");
+! you may not use this file except in compliance with the License.
+! You may obtain a copy of the License at
+!
+!     http://www.apache.org/licenses/LICENSE-2.0
+!
+! Unless required by applicable law or agreed to in writing, software
+! distributed under the License is distributed on an "AS IS" BASIS,
+! WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+! See the License for the specific language governing permissions and
+! limitations under the License.
+
+! Check modfile generation for generic interfaces
+module m
+  interface foo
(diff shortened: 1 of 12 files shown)


test/semantics/modfile07.f90
```diff
@@ -0,0 +1,69 @@
...
+    integer x
+  end
+end
+
+!Expect: m.mod
+!module m
+! interface foo
+!  procedure::s1
+!  procedure::s2
+! end interface
```
tskeith: FYI, I'm thinking about changing this generic interface to be written as:
`generic::foo=>s1,s2`
It's more compact and allows for an access-spec.

klausler: Size savings probably don't matter much, but the `GENERIC` statement does seem to be the most general way to write such things.  I'm pretty sure that it can be used for `OPERATOR` and I/O overloads, but you might want to confirm that.

PR #144, opened 2018-07-24
Checkpoint development of expression analysis

klausler:
A couple weeks worth of iterative refinement here that I'm checkpointing now that everything is compiling and working again.  Many prior structs are now classes; the `Expr` class template is parameterized now on a type rather than on a category and a kind value; Integer and most of Real scalar expressions can be folded, using common Unary and Binary base classes that will soon be extended to operate over arrays.


lib/semantics/resolve-names.cc
```diff
@@ -2138,7 +2138,7 @@ const Symbol *ResolveNamesVisitor::ResolveDataRef(const parser::DataRef &x) {
                 ApplyImplicitRules(y.source, *symbol);
               }
             }
-            return static_cast<const Symbol *>(symbol);
+            return const_cast<const Symbol *>(symbol);
           },
```
tskeith: You can use either `static_cast` or `const_cast` to add const-ness, but neither seems ideal. I chose to avoid `const_cast` because I think it should be a red flag indicating const-ness is being removed. But `static_cast` means more than an implicit conversion too (e.g. down-casting).

Some googling turned up `implicit_cast`, which only does implicit conversions. I propose we add it to idioms.h and use it in this situation.
```
template<typename T> struct _identity { using type = T; };
template<typename T> T implicit_cast(typename _identity<T>::type t) {
  return t;
}
```

klausler: I changed `static_cast` to `const_cast` here because `const_cast` is far weaker -- it can only change the const-qualification of the type of a value, and that's all that was needed in this situation.

I do not worry about `const_cast<const T*>` being susceptible to misinterpretation as *removing* const-qualification from a pointer type.  It's more clear than a non-standard `implicit_cast` call would be.

If `const_cast<const T*>` is not acceptable, consider wrapping it within an abstraction with a very clear name, e.g. `common::force_const<const T*>()`, with a `static_assert(std::is_const_v())` on the result type.  But consider first whether `const_cast<const T*>` is really so bad that is needs to papered over.

PR #145, opened 2018-07-25
Implement reading of module files

tskeith:
When a use-stmt is encountered for a module that isn't in the global
scope, search for and read the appropriate `.mod` file. To perform the
search, pass the search directories in to ResolveNames.

For modules that were read from `.mod` files, we have to keep the cooked
source from being deleted so that the names so that references to names
stay valid. So we store the cooked source in the Scope of the module as
a `unique_ptr`.

Add `Symbol::Flag::ModFile` to distinguish module symbols that were read
from a `.mod` file rather than from the current compilation. Use it to
prevent writing those back out.

Fix `test_errors.sh` to run the compiler in the temp subdirectory --
otherwise tests could be affected by `.mod` files left from previous
tests.

### DIFF
--- a/lib/semantics/mod-file.cc
+++ b/lib/semantics/mod-file.cc
@@ -15,7 +15,11 @@
 #include "mod-file.h"
 #include "scope.h"
 #include "symbol.h"
+#include "../parser/grammar.h"
 #include "../parser/message.h"
+#include "../parser/openmp-grammar.h"
+#include "../parser/preprocessor.h"
+#include "../parser/prescan.h"
 #include <algorithm>
 #include <cerrno>
 #include <fstream>
@@ -23,18 +27,28 @@
 #include <ostream>
 #include <set>
 #include <sstream>
+#include <sys/stat.h>
+#include <sys/types.h>
 #include <vector>
 
 namespace Fortran::semantics {
 
 using namespace parser::literals;
 
+// The extension used for module files.
+static constexpr auto extension{".mod"};
+// The initial characters of a file that identify it as a .mod file.
+static constexpr auto magic{"!mod$"};
+// Construct the path to a module file.
+static std::string ModFilePath(const std::string &, const std::string &);
+// Helpers for creating error messages.
+static parser::Message Error(
+    const SourceName &, parser::MessageFixedText, const std::string &);
+static parser::Message Error(const SourceName &, parser::MessageFixedText,
+    const std::string &, const std::string &);
+
 class ModFileWriter {
 public:
-  // The initial characters of a file that identify it as a .mod file.
-  static constexpr auto magic{"!mod$"};
-  static constexpr auto extension{".mod"};
-
   // The .mod file format version number.
   void set_version(int version) { version_ = version; }
   // The directory to write .mod files in.
@@ -103,7 +117,9 @@ bool ModFileWriter::WriteAll() {
   for (const auto &scope : Scope::globalScope.children()) {
     if (scope.kind() == Scope::Kind::Module) {
       auto &symbol{*scope.symbol()};  // symbol must be present for module
-      WriteOne(symbol);
+      if (!symbol.test(Symbol::Flag::ModFile)) {
+        WriteOne(symbol);
+      }
     }
   }
   return errors_.empty();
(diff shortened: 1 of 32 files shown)


tskeith:
I haven't added any tests for this yet. When we implement a driver in its final form, we should add tests for options and that should ensure that `-I` and `-module` (once it's implemented) cause `.mod` files to be found correctly.

Here is an example of an error reported when a module file can't be found. It lists each place tried and why it is no good.

```
$ f18 -fdebug-resolve-names -fparse-only -I/tmp t.f90
t.f90:2:7: error: Cannot find module file for 'm1'
    use m1
        ^^
t.f90:2:7: m1.mod: Permission denied
    use m1
        ^^
t.f90:2:7: /tmp/m1.mod: No such file or directory
    use m1
        ^^
```

lib/semantics/mod-file.cc
```diff
@@ -15,26 +15,40 @@
...
 using namespace parser::literals;
 
+// The extension used for module files.
+static constexpr auto extension{".mod"};
+// The initial characters of a file that identify it as a .mod file.
+static constexpr auto magic{"!mod$"};
+// Construct the path to a module file.
+static std::string ModFilePath(const std::string &, const std::string &);
+// Helpers for creating error messages.
+static parser::Message Error(
```
klausler: These could be in message.h and be more generally useful.

lib/semantics/mod-file.cc
```diff
@@ -396,4 +412,103 @@ std::string ModFileWriter::CheckSum(const std::string &str) {
...
+  }
+  auto &modSymbol{*it->second};
+  modSymbol.scope()->set_cookedSource(std::move(cooked_));
+  modSymbol.set(Symbol::Flag::ModFile);
+  return true;
+}
+
+// Look for the .mod file for this module in the search directories.
+// Add to errors_ if not found.
+std::optional<std::string> ModFileReader::FindModFile(
```
klausler: There's already code in lib/parser/provenance.cc for locating and opening source and include files.  Would it make sense to extend that code to locate modules?

lib/semantics/mod-file.cc
```diff
@@ -396,4 +412,103 @@ std::string ModFileWriter::CheckSum(const std::string &str) {
...
+  parser::Preprocessor preprocessor{allSources_};
+  parser::Messages messages;
+  parser::Prescanner prescanner{messages, *cooked_, preprocessor, {}};
+  parser::ProvenanceRange range{
+      allSources_.AddIncludedFile(*sourceFile, parser::ProvenanceRange{})};
+  prescanner.Prescan(range);
+  if (!messages.empty()) {
+    errors_.push_back(
+        Error(modName, "Module file for '%s' is corrupt: %s"_err_en_US,
+            modName.ToString(), path));
```
klausler: It might be valuable debugging information to preserve or emit some or all of the parsing error messages.

lib/semantics/mod-file.cc
```diff
@@ -396,4 +412,103 @@ std::string ModFileWriter::CheckSum(const std::string &str) {
...
+bool ModFileReader::Prescan(
+    const SourceName &modName, const std::string &path) {
+  std::stringstream fileError;
+  const auto *sourceFile{allSources_.Open(path, &fileError)};
+  if (sourceFile == nullptr) {
+    errors_.push_back(
+        Error(modName, "Cannot read %s: %s"_err_en_US, path, fileError.str()));
+    return false;
+  }
+  parser::Preprocessor preprocessor{allSources_};
```
klausler: Maybe a TODO here about checksum calculation and verification on nested USEs?

lib/semantics/mod-file.cc
tskeith: I'll look into that and the provenance.cc suggestion below.

lib/semantics/mod-file.cc
tskeith: Maybe under a debug option, but by default I don't think we should be saying much more than this if there is a problem. You can always see exactly what the problem is by compiling the .mod file as a Fortran file.

lib/semantics/mod-file.cc
tskeith: OK, I'll add that.

lib/semantics/mod-file.cc
klausler: Sure.  I was thinking about the situation where a bug report comes in from the field with insufficient attachments.

PR #149, opened 2018-07-26
Refactor ModFileReader to use parser::Parsing

tskeith:
parser::Parsing does most of the work needed to prescan and parse a file.
The only problem is that for module files we need to take ownership of
the cooked character stream so that it lives as long as the module scope
does.

To achieve this, change Parsing::cooked_ to be a unique_ptr and add a
member function to allow the caller to take ownership of it.

### DIFF
--- a/lib/parser/parsing.h
+++ b/lib/parser/parsing.h
@@ -50,7 +50,8 @@ public:
 
   bool consumedWholeFile() const { return consumedWholeFile_; }
   const char *finalRestingPlace() const { return finalRestingPlace_; }
-  CookedSource &cooked() { return cooked_; }
+  CookedSource &cooked() { return *cooked_; }
+  std::unique_ptr<CookedSource> MoveCooked() { return std::move(cooked_); }
   Messages &messages() { return messages_; }
   std::optional<Program> &parseTree() { return parseTree_; }
 
@@ -64,7 +65,7 @@ public:
   void EmitMessage(std::ostream &o, const char *at, const std::string &message,
       bool echoSourceLine = false) const {
     allSources_.EmitMessage(
-        o, cooked_.GetProvenanceRange(at).start(), message, echoSourceLine);
+        o, cooked_->GetProvenanceRange(at).start(), message, echoSourceLine);
   }
 
   bool ForTesting(std::string path, std::ostream &);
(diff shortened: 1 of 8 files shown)


lib/parser/parsing.h
```diff
@@ -50,7 +50,8 @@ class Parsing {
 
   bool consumedWholeFile() const { return consumedWholeFile_; }
   const char *finalRestingPlace() const { return finalRestingPlace_; }
-  CookedSource &cooked() { return cooked_; }
+  CookedSource &cooked() { return *cooked_; }
+  std::unique_ptr<CookedSource> MoveCooked() { return std::move(cooked_); }
```
tskeith: I'm looking for ideas for a better name for this function.

klausler: Transfer or Relinquish?

tskeith:
I've changed it to save just the vector of characters as that is all that is needed for the CharBlock objects in the symbol table.

I didn't want to save the entire Parsing instance in the module Scope because I think the Scope should not be tied to classes related to parsing.

klausler:
I worry that those CharBlocks aren't going to have any provenance mappings, but we can revisit their exclusion later if it turns out that we really need them.

lib/parser/provenance.cc
```diff
@@ -303,19 +303,19 @@ const AllSources::Origin &AllSources::MapToOrigin(Provenance at) const {
...
+  ProvenanceRange last{provenanceMap_.Map(cookedRange.end() - &(*data_)[0])};
   return {first.start(), last.start() - first.start()};
 }
 
 void CookedSource::Marshal() {
   CHECK(provenanceMap_.size() == buffer_.size());
   provenanceMap_.Put(allSources_.AddCompilerInsertion("(after end of source)"));
-  data_.resize(buffer_.size());
-  char *p{&data_[0]};
+  data_ = std::make_unique<std::vector<char>>(buffer_.size());
```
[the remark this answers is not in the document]
tskeith: I've removed the unique_ptr.

tskeith:
You are correct that provenance is a problem if we need to refer to source positions in the .mod file. We need to prevent such references or figure out a way to handle them.

PR #151, opened 2018-07-27
Tweaks to source provenance and messaging in support of module files

klausler:


lib/parser/char-buffer.cc
```diff
@@ -53,4 +53,42 @@ void CharBuffer::Put(const char *data, std::size_t n) {
...
+  std::size_t bytes{bytes_};
+  result.reserve(bytes);
+  for (const Block &block : blocks_) {
+    std::size_t chunk{std::min(bytes, Block::capacity)};
+    for (std::size_t j{0}; j < chunk; ++j) {
+      result += block.data[j];
+    }
+    bytes -= chunk;
+  }
+  result.shrink_to_fit();
```
tskeith: Is shrink_to_fit going to do anything after having already reserved the right capacity?

lib/parser/message.h
```diff
@@ -207,6 +208,7 @@ class Messages {
 
   void Incorporate(Messages &);
   void Copy(const Messages &);
+  void ResolveProvenances(const CookedSource &);
```
tskeith: A comment about what this does would be helpful. As I understand it, it converts message locations to ProvenanceRange from CharBlock when it can.

lib/parser/char-buffer.cc
klausler: Probably not, but it can't hurt in that case.  It's present here to release any overallocation that `reserve` may have thought was prudent for growth, and as a signal to the reader that the string is complete at this point.

lib/parser/provenance.cc
```diff
@@ -149,15 +149,20 @@ ProvenanceRange AllSources::AddCompilerInsertion(std::string text) {
   return covers;
 }
 
-void AllSources::EmitMessage(std::ostream &o, ProvenanceRange range,
-    const std::string &message, bool echoSourceLine) const {
-  CHECK(IsValid(range));
-  const Origin &origin{MapToOrigin(range.start())};
+void AllSources::EmitMessage(std::ostream &o,
+    const std::optional<ProvenanceRange> &range, const std::string &message,
```
tskeith: Now we can emit messages that aren't associated with a source location. Should there be a way to create such messages? For example, in `ModFileWriter` errors are saved as `vector<MessageFormattedText>` because they have no source location. It would be nice to be able to treat them as normal `Messages`.

lib/semantics/mod-file.cc
```diff
@@ -371,7 +374,8 @@ bool ModFileReader::Read(const SourceName &modName) {
     return false;
   }
   auto &modSymbol{*it->second};
-  modSymbol.scope()->set_chars(parsing.cooked().MoveChars());
+  // TODO: Preserve the CookedSource rather than a copy of its string.
+  modSymbol.scope()->set_chars(std::string{parsing.cooked().data()});
```
tskeith: Reading module files won't work until this TODO is resolved. The CharBlocks will have references to the old cooked characters, not the saved ones.

klausler: fixing

lib/parser/message.h
klausler: adding one

lib/parser/provenance.cc
klausler: There's a mechanism in place already for describing text in the AllSources that didn't come from anywhere with a meaningful location, such as the occasional spaces or punctuation that get inserted by preprocessing.  Provenances that map to such text are left empty when messages that have such provenances are formatted.  We might be able to use that facility in this situation.

PR #160, opened 2018-08-02
 Add support for submodules

tskeith:
Symbols for submodules have `ModuleDetails` with `isSubmodule` set.
Scopes for submodules have `Module` kind and have a parent scope that
is also `Module` kind.

Scopes for modules now contain a mapping of submodule name to scope
so that we can find them without having to search the scope tree or
re-read their `.mod` file.

The module file for submodule `s` with ancestor module `m` is named `m-s.mod`.
The tree structure of scopes means module file writing is now recursive.
Similarly, reading the module file for a submodule may require reading
the module files of its parent and ancestor. `ResolveNames` now requires
the parent scope to be passed in -- it is not always the global scope.

`test_modfiles.sh` now handles an argument that is a filename glob so
that the test can involve multiple files. This allows `modfile09` to
test reading of `.mod` files for modules and submodules.

Make module files read-only after writing them.

### DIFF
--- a/lib/semantics/mod-file.cc
+++ b/lib/semantics/mod-file.cc
@@ -24,6 +24,9 @@
 #include <ostream>
 #include <set>
 #include <sstream>
+#include <sys/stat.h>
+#include <sys/types.h>
+#include <unistd.h>
 #include <vector>
 
 namespace Fortran::semantics {
@@ -54,6 +57,7 @@ static std::ostream &PutLower(std::ostream &, const Symbol &);
 static std::ostream &PutLower(std::ostream &, const DeclTypeSpec &);
 static std::ostream &PutLower(std::ostream &, const std::string &);
 static std::string CheckSum(const std::string &);
+static bool MakeReadonly(const std::string &);
 
 bool ModFileWriter::WriteAll() {
   WriteChildren(Scope::globalScope);
@@ -75,6 +79,7 @@ void ModFileWriter::WriteOne(const Scope &scope) {
     auto *ancestor{symbol->get<ModuleDetails>().ancestor()};
     auto ancestorName{ancestor ? ancestor->name().ToString() : ""s};
     auto path{ModFilePath(dir_, symbol->name(), ancestorName)};
+    unlink(path.data());
     std::ofstream os{path};
     PutSymbols(scope);
     std::string all{GetAsString(*symbol)};
@@ -85,6 +90,10 @@ void ModFileWriter::WriteOne(const Scope &scope) {
       errors_.emplace_back(
           "Error writing %s: %s"_err_en_US, path.c_str(), std::strerror(errno));
       return;
+    }
+    if (!MakeReadonly(path)) {
+      errors_.emplace_back("Error changing permissions on %s: %s"_en_US,
+          path.c_str(), std::strerror(errno));
     }
   }
   WriteChildren(scope);  // write out submodules
@@ -476,6 +485,16 @@ static std::string ModFilePath(const std::string &dir, const SourceName &name,
   }
   PutLower(path, name.ToString()) << extension;
   return path.str();
+}
+
+static bool MakeReadonly(const std::string &path) {
+  struct stat statbuf;
+  if (stat(path.c_str(), &statbuf) != 0) {
+    return false;
+  }
+  auto mode{statbuf.st_mode};
+  mode &= S_IRUSR | S_IRGRP | S_IROTH;
+  return chmod(path.data(), mode) == 0;
 }
 
 static parser::Message Error(const SourceName &location,
--- a/lib/semantics/mod-file.cc
+++ b/lib/semantics/mod-file.cc
@@ -71,32 +71,36 @@ void ModFileWriter::WriteChildren(const Scope &scope) {
 }
 
 void ModFileWriter::WriteOne(const Scope &scope) {
-  auto *symbol{scope.symbol()};
-  if (scope.kind() != Scope::Kind::Module) {
+  if (scope.kind() == Scope::Kind::Module) {
+    auto *symbol{scope.symbol()};
+    if (!symbol->test(Symbol::Flag::ModFile)) {
+      Write(*symbol);
+    }
+    WriteChildren(scope);  // write out submodules
+  }
+}
+
+// Write the module file for symbol, which must be a module or submodule.
+void ModFileWriter::Write(const Symbol &symbol) {
+  auto *ancestor{symbol.get<ModuleDetails>().ancestor()};
+  auto ancestorName{ancestor ? ancestor->name().ToString() : ""s};
+  auto path{ModFilePath(dir_, symbol.name(), ancestorName)};
+  unlink(path.data());
+  std::ofstream os{path};
+  PutSymbols(*symbol.scope());
+  std::string all{GetAsString(symbol)};
+  auto header{GetHeader(all)};
+  os << header << all;
+  os.close();
+  if (!os) {
+    errors_.emplace_back(
+        "Error writing %s: %s"_err_en_US, path.c_str(), std::strerror(errno));
     return;
   }
-  if (!symbol->test(Symbol::Flag::ModFile)) {
-    auto *ancestor{symbol->get<ModuleDetails>().ancestor()};
-    auto ancestorName{ancestor ? ancestor->name().ToString() : ""s};
-    auto path{ModFilePath(dir_, symbol->name(), ancestorName)};
-    unlink(path.data());
-    std::ofstream os{path};
-    PutSymbols(scope);
-    std::string all{GetAsString(*symbol)};
-    auto header{GetHeader(all)};
-    os << header << all;
-    os.close();
-    if (!os) {
-      errors_.emplace_back(
-          "Error writing %s: %s"_err_en_US, path.c_str(), std::strerror(errno));
-      return;
-    }
-    if (!MakeReadonly(path)) {
-      errors_.emplace_back("Error changing permissions on %s: %s"_en_US,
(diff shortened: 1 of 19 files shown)


lib/semantics/mod-file.cc
```diff
@@ -52,43 +57,65 @@ static std::ostream &PutLower(std::ostream &, const Symbol &);
...
+}
+
+void ModFileWriter::WriteOne(const Scope &scope) {
+  auto *symbol{scope.symbol()};
+  if (scope.kind() != Scope::Kind::Module) {
+    return;
   }
-  return true;
+  if (!symbol->test(Symbol::Flag::ModFile)) {
+    auto *ancestor{symbol->get<ModuleDetails>().ancestor()};
```
klausler: The body of this `if` might make a good subroutine, even though it would be called only once, since the implementation of the writing of the module file is independent of the traversal context around it.

lib/semantics/mod-file.cc
```diff
@@ -363,56 +402,99 @@ bool ModFileReader::Read(const SourceName &modName) {
...
-  const auto &it{Scope::globalScope.find(modName)};
-  if (it == Scope::globalScope.end()) {
-    return false;
+  Scope *parentScope;
+  if (!ancestor) {
+    // module: goes into global scope
+    parentScope = &Scope::globalScope;
+  } else {
+    // submodule: goes into parent module/submodule
+    auto *parent{GetSubmoduleParent(*parseTree)};
```
klausler: Might be more readable without the conditional expression, instead using `if (auto *parent{...}) { parentScope = Read(...); } else { parentScope = ancestor; }`.  Maybe not.

lib/semantics/resolve-names.cc
```diff
@@ -404,13 +407,15 @@ class ModuleVisitor : public virtual ScopeHandler {
...
-  const Scope *FindModule(const SourceName &);
   void AddUse(const parser::Rename::Names &);
   void AddUse(const parser::Name &);
   // Record a use from useModuleScope_ of useName as localName. location is
   // where it occurred (either the module or the rename) for error reporting.
   void AddUse(const SourceName &location, const SourceName &localName,
       const SourceName &useName);
+  Symbol &BeginModule(const SourceName &, bool isSubmodule,
+      const std::optional<parser::ModuleSubprogramPart> &);
+  Scope *FindModule(const SourceName &, Scope * = nullptr);
```
klausler: A name for that last optional argument would be informative.

lib/semantics/scope.cc
```diff
@@ -48,6 +48,18 @@ Scope::size_type Scope::erase(const SourceName &name) {
...
+  auto it{submodules_.find(name)};
+  if (it == submodules_.end()) {
+    return nullptr;
+  } else {
+    return it->second;
+  }
+}
+Scope *Scope::AddSubmodule(const SourceName &name, Scope *submodule) {
+  auto pair{submodules_.emplace(name, submodule)};
+  return !pair.second ? pair.first->second : nullptr;
```
klausler: What's being returned here?  A pointer to an extant Scope for a submodule of the same name that was already present in the `submodules_` map for the scope?

lib/semantics/symbol.cc
```diff
@@ -23,6 +23,28 @@ std::ostream &operator<<(std::ostream &os, const parser::CharBlock &name) {
   return os << name.ToString();
 }
 
+const Scope *ModuleDetails::parent() const {
+  return isSubmodule_ ? &scope_->parent() : nullptr;
+}
+const Scope *ModuleDetails::ancestor() const {
```
klausler: "ancestor" is a term that could apply to intermediate submodules, too.  If this member function locates the top-level module, a more specific name would be better, I think.

lib/semantics/type.cc
```diff
@@ -269,34 +269,22 @@ std::ostream &operator<<(std::ostream &o, const DeclTypeSpec &x) {
...
-}
-
 void ProcInterface::set_symbol(const Symbol &symbol) {
   CHECK(!type_);
   symbol_ = &symbol;
 }
 void ProcInterface::set_type(const DeclTypeSpec &type) {
   CHECK(!symbol_);
-  type_ = std::make_unique<DeclTypeSpec>(type);
+  type_ = type;
```
klausler: If symbol_ and type_ are mutually exclusive, consider using a std::variant.

lib/semantics/type.cc
```diff
@@ -269,34 +269,22 @@ std::ostream &operator<<(std::ostream &o, const DeclTypeSpec &x) {
...
 
 std::ostream &operator<<(std::ostream &o, const ProcDecl &x) {
   return o << x.name_.ToString();
 }
 
 ProcComponentDef::ProcComponentDef(
-    const ProcDecl &decl, Attrs attrs, ProcInterface &&interface)
-  : decl_{decl}, attrs_{attrs}, interface_{std::move(interface)} {
+    const ProcDecl &decl, Attrs attrs, const ProcInterface &interface)
+  : decl_{decl}, attrs_{attrs}, interface_{interface} {
```
schweitzpgi: The move semantics seems to be going away in a lot of places. Is that to fix bugs? Or copies of these various data are now needed?

lib/semantics/mod-file.cc
tskeith: Agreed

lib/semantics/mod-file.cc
tskeith: Yes, especially as it is really a three-way if whose purpose is to set parentScope.

lib/semantics/resolve-names.cc

lib/semantics/scope.cc
tskeith: Yes, but I'm only using it as a boolean now so I'll change it to return `pair.second`. I.e. `true` if the insertion took place.

lib/semantics/symbol.cc
tskeith: This is tricky. The standard refers to "the ancestor module" and "an ancestor module or submodule". So far the code has only dealt with the former, so I've been using "ancestor" for that. But it will be a problem it we do end up needing a name for the latter.

lib/semantics/type.cc
tskeith: This is still evolving, but that's probably a good idea.

lib/semantics/mod-file.cc
```diff
@@ -24,6 +24,9 @@
 #include <ostream>
 #include <set>
 #include <sstream>
+#include <sys/stat.h>
+#include <sys/types.h>
+#include <unistd.h>
```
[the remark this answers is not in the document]
tskeith: I haven't been, but I'm sure we will need to. The POSIX-specific stuff is isolated in `MakeReadonly` so hopefully it won't be hard to make a Windows-specific version in the future.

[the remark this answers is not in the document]
sscalpone: Yes we care about Windows but no quite yet; VS C++ still can't compile f18.

lib/semantics/mod-file.cc
```diff
@@ -34,14 +37,16 @@ using namespace parser::literals;
...
+
 // Helpers for creating error messages.
 static parser::Message Error(
     const SourceName &, parser::MessageFixedText, const std::string &);
 static parser::Message Error(const SourceName &, parser::MessageFixedText,
     const std::string &, const std::string &);
 
+static const SourceName *GetSubmoduleParent(const parser::Program &);
+static std::string ModFilePath(
+    const std::string &, const SourceName &, const std::string &);
```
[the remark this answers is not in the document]
tskeith: These are forward declarations for the benefit of the compiler rather than the reader. They can look at the definition of the functions further down the file where there is a comment and argument names.

If this were in a .h file I would put the comment there and argument names if they help document it (like in this case).

lib/semantics/mod-file.cc
```diff
@@ -52,43 +57,65 @@ static std::ostream &PutLower(std::ostream &, const Symbol &);
...
+    std::string all{GetAsString(*symbol)};
+    auto header{GetHeader(all)};
+    os << header << all;
+    os.close();
+    if (!os) {
+      errors_.emplace_back(
+          "Error writing %s: %s"_err_en_US, path.c_str(), std::strerror(errno));
+      return;
+    }
+    if (!MakeReadonly(path)) {
```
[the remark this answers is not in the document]
tskeith: I did this because `.mod` files are Fortran source and we want to discourage users from editing them. Why do you think it's a problem?

lib/semantics/mod-file.cc
```diff
@@ -52,43 +57,65 @@ static std::ostream &PutLower(std::ostream &, const Symbol &);
...
+          path.c_str(), std::strerror(errno));
+    }
+  }
+  WriteChildren(scope);  // write out submodules
 }
 
 // Return the entire body of the module file
 // and clear saved uses, decls, and contains.
-std::string ModFileWriter::GetAsString(const std::string &name) {
+std::string ModFileWriter::GetAsString(const Symbol &symbol) {
```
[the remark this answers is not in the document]
tskeith: I'll look into that. I'm not familiar with it.

lib/semantics/type.cc
tskeith: I did this because it was simpler in a few places to make copies and there was no real benefit to using move semantics for the Details classes.

lib/semantics/mod-file.cc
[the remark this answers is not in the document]
schweitzpgi: It doesn't solve a technical problem; it adds one to any tools/users, including the compiler, that have to manage the .mod files.

PR #162, opened 2018-08-03
Roll up work on folding, descriptors, runtime, RESHAPE

klausler:
Some disconnected changes here related to evaluation and execution.  The C++ interface to the ISO standard descriptor is now is much better shape, and can serve to implement a working RESHAPE intrinsic.  More to come, but it was past time to checkpoint and merge this work area.

As a readability improvement (I think), I've replaced many instances of `return {};` with `return std::nullopt;` in contexts where the function's type is an instantiation of `std::optional<>`.

Also, a GNU C++ compiler bug workaround in `lib/parser/basic-parsers.h` is no longer needed with GNU 8.2.0, and the conditional compilation tests were adjusted accordingly.

### DIFF
--- a/lib/evaluate/expression.cc
+++ b/lib/evaluate/expression.cc
@@ -621,9 +621,51 @@ auto ComplexExpr<KIND>::Fold(FoldingContext &context) -> std::optional<Scalar> {
 }
 
 template<int KIND>
+auto CharacterExpr<KIND>::Concat::FoldScalar(FoldingContext &context,
+    const Scalar &a, const Scalar &b) -> std::optional<Scalar> {
+  if constexpr (KIND == 1) {
+    return {a + b};
+  }
+  return std::nullopt;
+}
+
+template<int KIND>
+auto CharacterExpr<KIND>::Max::FoldScalar(FoldingContext &context,
+    const Scalar &a, const Scalar &b) -> std::optional<Scalar> {
+  if (Compare(a, b) == Ordering::Less) {
+    return {b};
+  }
+  return {a};
+}
+
+template<int KIND>
+auto CharacterExpr<KIND>::Min::FoldScalar(FoldingContext &context,
+    const Scalar &a, const Scalar &b) -> std::optional<Scalar> {
+  if (Compare(a, b) == Ordering::Greater) {
+    return {b};
+  }
+  return {a};
+}
+
+template<int KIND>
 auto CharacterExpr<KIND>::Fold(FoldingContext &context)
     -> std::optional<Scalar> {
-  return std::nullopt;  // TODO
+  return std::visit(
+      [&](auto &x) -> std::optional<Scalar> {
+        using Ty = typename std::decay<decltype(x)>::type;
+        if constexpr (std::is_same_v<Ty, Scalar>) {
+          return {x};
+        }
+        if constexpr (evaluate::FoldableTrait<Ty>) {
+          auto c{x.Fold(context)};
+          if (c.has_value()) {
+            u_ = *c;
+            return c;
+          }
+        }
+        return std::nullopt;
+      },
+      u_);
 }
 
 template<typename A>
--- a/lib/evaluate/expression.cc
+++ b/lib/evaluate/expression.cc
@@ -657,7 +657,24 @@ auto Comparison<A>::FoldScalar(FoldingContext &c,
     case Relation::Unordered: return std::nullopt;
     }
   }
-  // TODO complex and character comparisons
+  if constexpr (A::category == Category::Complex) {
+    bool eqOk{opr == RelationalOperator::LE || opr == RelationalOperator::EQ ||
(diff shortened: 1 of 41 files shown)


lib/evaluate/expression.cc
```diff
@@ -471,7 +472,13 @@ template<int KIND>
 auto RealExpr<KIND>::IntPower::FoldScalar(FoldingContext &context,
```
tskeith: Why use `auto` for these functions? They don't seem to obey the "Don't use `auto` functions unless..." rule.

lib/evaluate/int-power.h
```diff
@@ -0,0 +1,60 @@
...
+
+namespace Fortran::evaluate {
+
+template<typename REAL, typename INT>
+ValueWithRealFlags<REAL> IntPower(const REAL &base, const INT &power,
+    Rounding rounding = Rounding::TiesToEven) {
+  REAL one{REAL::FromInteger(INT{1}).value};
+  ValueWithRealFlags<REAL> result;
+  result.value = one;
+  if (base.IsNotANumber()) {
```
tskeith: We seem to be using "NaN" everywhere else, so why not "IsNaN" for this function?
(I know this wasn't added in this change -- sorry.)

lib/evaluate/expression.cc
klausler: This is an outlined member function definition whose result type uses the name of a type declared in the class.  Written without `auto`, the result type would have to be something like `std::optional<typename RealExpr<KIND>::Scalar>`.  This is exactly the situation in which the "unless..." exceptional language in the style guide is meant to apply.

lib/evaluate/int-power.h
klausler: I'll rename some of the others to avoid the needless use of an obscure term of art.

runtime/descriptor.cc
```diff
@@ -12,39 +12,166 @@
...
+}
+
+void Descriptor::Establish(TypeCode t, std::size_t elementBytes, void *p,
+    int rank, const SubscriptValue *extent, ISO::CFI_attribute_t attribute,
+    bool addendum) {
+  CHECK(ISO::CFI_establish(&raw_, p, attribute, t.raw(), elementBytes, rank,
+            extent) == CFI_SUCCESS);
+  raw_.f18Addendum = addendum;
+  if (addendum) {
+    new (Addendum()) DescriptorAddendum{};
```
tskeith: If any of these placement news fail because the expression is nullptr, does that mean there is a bug in the logic of Descriptor? If it could be due to an error in how the client of this class is using it, I think an explicit assertion would be helpful. Otherwise, this is fine.

runtime/type-code.cc
```diff
@@ -0,0 +1,62 @@
...
+    break;
+  case TypeCategory::Logical:
+    switch (kind) {
+    case 1: raw_ = CFI_type_Bool; break;
+    case 2: raw_ = CFI_type_int16_t; break;
+    case 4: raw_ = CFI_type_int32_t; break;
+    case 8: raw_ = CFI_type_int64_t; break;
+    }
+    break;
+  case TypeCategory::Derived: raw_ = CFI_type_struct; break;
```
schweitzpgi: clang will warn about missing default cases

test/evaluate/real.cc
```diff
@@ -19,37 +19,30 @@
...
-    { 0x00000001, "0x0.000002p-127" },
-    { 0, nullptr }
-  };
+  } table[] = {{0x7f876543, "NaN 0x7f876543"}, {0x7f800000, "Inf"},
+      {0xff800000, "-Inf"}, {0x00000000, "0.0"}, {0x80000000, "-0.0"},
+      {0x3f800000, "0x1.0p0"}, {0xbf800000, "-0x1.0p0"},
+      {0x40000000, "0x1.0p1"}, {0x3f000000, "0x1.0p-1"},
+      {0x7f7fffff, "0x1.fffffep127"}, {0x00800000, "0x1.0p-126"},
+      {0x00400000, "0x0.8p-127"}, {0x00000001, "0x0.000002p-127"},
+      {0, nullptr}};
```
schweitzpgi: Does this project use clang-format?

tskeith: Yes, unfortunately it sometimes makes a mess of the code.

In this case, adding a comma after `{0, nullptr}` would get it back the way it was. I think that would be an improvement.

runtime/descriptor.cc
klausler: In this case, Addendum must be non-null, but it can be made more explicitly so.

runtime/type-code.cc
klausler: It doesn't.

test/evaluate/real.cc
klausler: Ok.

PR #164, opened 2018-08-08
Module file improvements

tskeith:
Verify that checksums are correct when reading a module file.

Don't write a module file if its current contents are correct.

Don't make .mod files read-only. It doesn't do much to prevent
users from editing them, checksum verification will detect when
it happens, and having them read-only causes problems if you then
compile with another compiler (e.g. PGI).

### DIFF
--- a/lib/semantics/mod-file.cc
+++ b/lib/semantics/mod-file.cc
@@ -19,14 +19,13 @@
 #include "../parser/parsing.h"
 #include <algorithm>
 #include <cerrno>
+#include <cstring>
 #include <fstream>
 #include <functional>
 #include <ostream>
-#include <set>
 #include <sstream>
 #include <sys/stat.h>
 #include <sys/types.h>
-#include <unistd.h>
 #include <vector>
 
 namespace Fortran::semantics {
@@ -36,7 +35,7 @@ using namespace parser::literals;
 // The extension used for module files.
 static constexpr auto extension{".mod"};
 // The initial characters of a file that identify it as a .mod file.
-static const auto magic{"!mod$ v1 sum:"s};
+static constexpr auto magic{"!mod$ v1 sum:"};
 
 // Helpers for creating error messages.
 static parser::Message Error(
@@ -384,7 +383,7 @@ static bool FileContentsMatch(std::fstream &stream, const std::string &header,
       return false;
     }
   }
-  return true;
+  return !stream.get(c);
 }
 
 // Compute a simple hash of the contents of a module file and
@@ -409,10 +408,11 @@ static bool VerifyHeader(const std::string &path) {
   std::fstream stream{path};
   std::string header;
   std::getline(stream, header);
-  if (header.compare(0, magic.size(), magic) != 0) {
+  auto magicLen{strlen(magic)};
+  if (header.compare(0, magicLen, magic) != 0) {
     return false;
   }
-  std::string expectSum{header.substr(magic.size(), 16)};
+  std::string expectSum{header.substr(magicLen, 16)};
   std::string actualSum{CheckSum(std::istreambuf_iterator<char>(stream),
       std::istreambuf_iterator<char>())};
   return expectSum == actualSum;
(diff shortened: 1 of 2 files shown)


lib/semantics/mod-file.cc
```diff
@@ -85,21 +88,10 @@ void ModFileWriter::Write(const Symbol &symbol) {
   auto *ancestor{symbol.get<ModuleDetails>().ancestor()};
   auto ancestorName{ancestor ? ancestor->name().ToString() : ""s};
   auto path{ModFilePath(dir_, symbol.name(), ancestorName)};
-  unlink(path.data());
```
sscalpone: Do you still need the system include files?

lib/semantics/mod-file.cc
```diff
@@ -362,12 +347,53 @@ std::ostream &PutLower(std::ostream &os, const std::string &str) {
...
+    if (!stream.get(c) || c != contents[i]) {
+      return false;
+    }
+  }
+  return true;
+}
+
 // Compute a simple hash of the contents of a module file and
 // return it as a string of hex digits.
 // This uses the Fowler-Noll-Vo hash function.
```
schweitzpgi: Out of curiosity, is there anything special to note about this choice of hash?

lib/semantics/mod-file.cc
```diff
@@ -36,7 +36,7 @@ using namespace parser::literals;
 // The extension used for module files.
 static constexpr auto extension{".mod"};
 // The initial characters of a file that identify it as a .mod file.
-static constexpr auto magic{"!mod$"};
+static const auto magic{"!mod$ v1 sum:"s};
```
klausler: I think this might now be an instance of a static variable with a runtime constructor, and LLVM forbids their use.

lib/semantics/mod-file.cc
```diff
@@ -362,12 +347,53 @@ std::ostream &PutLower(std::ostream &os, const std::string &str) {
...
+  }
+  if (!stream.get(c) || c != '\n') {
+    return false;
+  }
+  for (std::size_t i{0}; i < contents.size(); ++i) {
+    if (!stream.get(c) || c != contents[i]) {
+      return false;
+    }
+  }
+  return true;
```
klausler: What if the file is longer than the new content?  I think you'll return a false positive true here.

lib/semantics/mod-file.cc
```diff
@@ -362,12 +347,53 @@ std::ostream &PutLower(std::ostream &os, const std::string &str) {
...
 // Compute a simple hash of the contents of a module file and
 // return it as a string of hex digits.
 // This uses the Fowler-Noll-Vo hash function.
-std::string CheckSum(const std::string &str) {
+template<typename Iter> static std::string CheckSum(Iter begin, Iter end) {
   std::uint64_t hash{0xcbf29ce484222325ull};
-  for (char c : str) {
+  for (auto it{begin}; it != end; ++it) {
+    char c{*it};
     hash ^= c & 0xff;
```
schweitzpgi: Do you need this mask if you use "unsigned char" as the type in the line above?

klausler: braced initialization with `unsigned char` would fail to compile if the signedness of default char was signed.  Using the explicit masking is most portable.

lib/semantics/mod-file.cc
tskeith: I've removed unistd.h and set.

lib/semantics/mod-file.cc
tskeith: OK, I've changed it back to a C string.

lib/semantics/mod-file.cc
tskeith: Right -- fixed.

lib/semantics/mod-file.cc
tskeith: Not really, just simple to compute and reasonably effective (based on my limited research).

PR #165, opened 2018-08-08
Expression analysis and message clean-up

klausler:
Continuation of work with expression analysis, plus a big diversion to improve the message APIs so that it's possible to invoke `Say` on a message text with automatic formatting without having to explicitly construct a `MessageFormattedText`.  Requesting review now to make sure that the message API changes are fine and that the changes made to messaging in name resolution and module file I/O are acceptable (I converted a vector of formatted messages into a instance of the `Messages` class).

### DIFF
--- a/lib/evaluate/CMakeLists.txt
+++ b/lib/evaluate/CMakeLists.txt
@@ -13,6 +13,7 @@
 # limitations under the License.
 
 add_library(FortranEvaluate
+  common.cc
   complex.cc
   expression.cc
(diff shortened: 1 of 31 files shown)


tskeith:
I've been wondering if there is a way to have an overloading of `Say` that takes `std::string` for the `%s` parameters instead of `const char *`. There are a lot of places we have to call `.c_str()` or `.data()` to pass strings in.

PR #167, opened 2018-08-23
Implement host association and import statements

tskeith:
Host association is done by search for symbols using `Scope::FindSymbol()`
which looks for the the name in the parent scope if the import rules
permit it.

Import statements are implemented using `importKind_` and `importNames_`
in class `Scope`. Most of the constraints are checked when the
`ImportStmt` is encountered. `CheckImports()` is called at the end of
the `SpecificationPart` to verify the names mentioned in the IMPORT
statement. That has to happen then so that we can detect if an imported
name is hidden by a declaration in the current scope.

### DIFF
--- a/lib/semantics/resolve-names.cc
+++ b/lib/semantics/resolve-names.cc
@@ -293,7 +293,7 @@ private:
 class ScopeHandler : public virtual ImplicitRulesVisitor {
 public:
   void set_rootScope(Scope &scope) { PushScope(scope); }
-  Scope &CurrScope() { return *scopes_.top(); }
+  Scope &CurrScope() { return *currScope_; }
   // Return the enclosing scope not corresponding to a derived type:
   Scope &CurrNonTypeScope();
 
@@ -375,8 +375,7 @@ protected:
   std::optional<const DeclTypeSpec> GetImplicitType(Symbol &);
 
 private:
-  // Stack of containing scopes; memory referenced is owned by parent scopes
-  std::stack<Scope *, std::list<Scope *>> scopes_;
+  Scope *currScope_{nullptr};
 };
 
 class ModuleVisitor : public virtual ScopeHandler {
@@ -1154,11 +1153,11 @@ Scope &ScopeHandler::PushScope(Scope::Kind kind, Symbol *symbol) {
   return scope;
 }
 void ScopeHandler::PushScope(Scope &scope) {
-  scopes_.push(&scope);
+  currScope_ = &scope;
   ImplicitRulesVisitor::PushScope();
 }
 void ScopeHandler::PopScope() {
-  scopes_.pop();
+  currScope_ = &currScope_->parent();
   ImplicitRulesVisitor::PopScope();
 }
 
--- a/lib/semantics/resolve-names.cc
+++ b/lib/semantics/resolve-names.cc
@@ -1330,7 +1330,7 @@ bool ModuleVisitor::Pre(const parser::Submodule &x) {
   }
   PushScope(*parentScope);  // submodule is hosted in parent
   auto &symbol{BeginModule(name, true, subpPart)};
-  if (!ancestor->AddSubmodule(name, &currScope())) {
+  if (!ancestor->AddSubmodule(name, currScope())) {
     Say(name, "Module '%s' already has a submodule named '%s'"_err_en_US,
         ancestorName, name);
   }
@@ -2151,7 +2151,7 @@ void ResolveNamesVisitor::Post(const parser::CallStmt &) {
 bool ResolveNamesVisitor::Pre(const parser::ImportStmt &x) {
   auto kind{MapImportKind(x.kind)};
   auto &scope{currScope()};
-  // Check C896 and C899
+  // Check C896 and C899: where IMPORT statements are allowed
   switch (scope.kind()) {
   case Scope::Kind::Module:
     if (!scope.symbol()->get<ModuleDetails>().isSubmodule()) {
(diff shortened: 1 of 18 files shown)


lib/semantics/scope.h
```diff
@@ -114,6 +125,16 @@ class Scope {
   // that are referenced by SourceName objects.
   void set_chars(std::string &&chars) { chars_ = std::move(chars); }
 
+  ImportKind importKind() const;
+  // Names appearing in IMPORT statements in this scope
+  std::set<SourceName> importNames() const { return importNames_; }
+
+  // Set the kind of imports from host into this scope.
+  // Return an error message for incompatible kinds.
+  std::optional<parser::MessageFixedText> set_importKind(ImportKind);
```
sscalpone: Does this follow our naming convention?

tskeith: Yes, at least for a normal setter it's `set_` and the data member name.

This one isn't quite normal though, because it can return an error. So maybe it should be `SetImportKind` to indicate it doesn't follow the standard pattern. Either way is okay with me.

lib/semantics/resolve-names.cc
```diff
@@ -293,12 +294,12 @@ class ArraySpecVisitor {
 class ScopeHandler : public virtual ImplicitRulesVisitor {
 public:
   void set_rootScope(Scope &scope) { PushScope(scope); }
-  Scope &CurrScope() { return *scopes_.top(); }
+  Scope &currScope() { return *currScope_; }
```
klausler: As we discussed, this member function isn't really an accessor.

lib/semantics/resolve-names.cc
```diff
@@ -293,12 +294,12 @@ class ArraySpecVisitor {
...
   // Return the enclosing scope not corresponding to a derived type:
   Scope &CurrNonTypeScope();
 
   // Create a new scope and push it on the scope stack.
-  Scope &PushScope(Scope::Kind kind, Symbol *symbol);
+  void PushScope(Scope::Kind kind, Symbol *symbol);
```
klausler: Was the returned reference never used?  Or did it become infeasible to produce?

One of the nice things about functions that return references is that their users know that they can't fail, since there's nothing like a null pointer for references.

lib/semantics/resolve-names.cc
```diff
@@ -1334,7 +1330,7 @@ bool ModuleVisitor::Pre(const parser::Submodule &x) {
   }
   PushScope(*parentScope);  // submodule is hosted in parent
   auto &symbol{BeginModule(name, true, subpPart)};
-  if (!ancestor->AddSubmodule(name, &CurrScope())) {
+  if (!ancestor->AddSubmodule(name, &currScope())) {
```
klausler: Maybe AddSubmodule() should take a reference argument rather than a pointer, since it can't be null.

lib/semantics/resolve-names.cc
```diff
@@ -2152,6 +2148,43 @@ void ResolveNamesVisitor::Post(const parser::CallStmt &) {
   expectedProcFlag_ = std::nullopt;
 }
 
+bool ResolveNamesVisitor::Pre(const parser::ImportStmt &x) {
+  auto kind{MapImportKind(x.kind)};
+  auto &scope{currScope()};
+  // Check C896 and C899
```
klausler: Could you summarize C896/C899 briefly here, please?

lib/semantics/scope.cc
```diff
@@ -64,6 +74,63 @@ DerivedTypeSpec &Scope::MakeDerivedTypeSpec(const SourceName &name) {
...
+
+std::optional<parser::MessageFixedText> Scope::set_importKind(ImportKind kind) {
+  if (!importKind_.has_value()) {
+    importKind_ = kind;
+    return std::nullopt;
+  }
+  std::optional<parser::MessageFixedText> error;
+  bool hasNone{kind == ImportKind::None || *importKind_ == ImportKind::None};
+  bool hasAll{kind == ImportKind::All || *importKind_ == ImportKind::All};
+  // Check C8100 and C898
```
klausler: A brief summary of these two constraints would be great, if they're not too complicated.

lib/semantics/resolve-names.cc
tskeith: It wasn't necessary to return it. The caller can get the current scope by calling currScope() like is done everywhere else.

lib/semantics/resolve-names.cc
tskeith: OK. The specific constraints are obvious from the error messages, but I can add a one-line summary.

lib/semantics/scope.cc
tskeith: OK

lib/semantics/resolve-names.cc
tskeith: This still feels like a simple accessor to me. The data member should be a `Scope &` but the language doesn't allow that, so we have to work around it by using a never-null `Scope *`.

lib/semantics/scope.h
```diff
@@ -33,6 +37,7 @@ class Scope {
   static Scope globalScope;  // contains program-units
 
   ENUM_CLASS(Kind, System, Global, Module, MainProgram, Subprogram, DerivedType)
+  ENUM_CLASS(ImportKind, Default, Only, None, All);
```
tskeith: This could be `using ImportKind = parser::ImportStmt::Kind` but I didn't want to cause `parse-tree.h` to be included wherever `scope.h` is. Maybe there could be a shared definition in a different `.h` file.

klausler: I've started putting a few universal enum classes like this into lib/common/fortran.h.

PR #169, opened 2018-08-27
Add support for BLOCK construct

### LINKED ISSUE #71: IMPLICIT NONE is not inherited by nested scope
In the example below, the compiler should report an error on the use of `x`:
```
program test
  implicit none
contains
  subroutine foo
    x = 666 ! wrong since x has no implicit rule
  end subroutine
end program
```

tskeith:
A `BLOCK` statement opens a new scope. It is different from other scopes
in that implicitly typed entities are defined in the enclosing non-block
scope, not immediately in the block. This means that `IMPORT` statements
can cause them to be hidden.

Check that blocks can't have `IMPLICIT` statements in them. It is simpler
for the parser not to deal with the different between a
specification-part and a block-specification-part.

Change `ImplicitRules` to have a parent that is consulted when there isn't
an answer in the current one. For an interface body that does not happen
but for all other nested scopes it does. This parent link eliminates the
need for the `implicitRules_` stack. Make `isImplicitNoneType_` and
`isImplicitNoneExternal_` optional: not set means look in parent.
Fixes #71.

Remove `CurrNonTypeScope()` and put the logic in `Symbol::FindSymbol`.

### DIFF
--- a/lib/semantics/resolve-names.cc
+++ b/lib/semantics/resolve-names.cc
@@ -37,8 +37,10 @@ class MessageHandler;
 
 static GenericSpec MapGenericSpec(const parser::GenericSpec &);
 
-// ImplicitRules maps initial character of identifier to the DeclTypeSpec*
-// representing the implicit type; nullptr if none.
+// ImplicitRules maps initial character of identifier to the DeclTypeSpec
+// representing the implicit type; std::nullopt if none.
+// It also records the presence of IMPLICIT NONE statements.
+// When inheritFromParent is set, defaults come from the parent rules.
 class ImplicitRules {
 public:
   ImplicitRules(MessageHandler &messages)
--- a/lib/semantics/resolve-names.cc
+++ b/lib/semantics/resolve-names.cc
@@ -41,27 +41,35 @@ static GenericSpec MapGenericSpec(const parser::GenericSpec &);
 // representing the implicit type; nullptr if none.
 class ImplicitRules {
 public:
-  ImplicitRules(MessageHandler &messages);
-  bool isImplicitNoneType() const { return isImplicitNoneType_; }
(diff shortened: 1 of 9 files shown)


lib/semantics/resolve-names.cc
```diff
@@ -41,27 +41,35 @@ static GenericSpec MapGenericSpec(const parser::GenericSpec &);
...
-  void AddDefaultRules();
 
 private:
   static char Incr(char ch);
 
+  std::unique_ptr<ImplicitRules> parent_;
   MessageHandler &messages_;
-  bool isImplicitNoneType_{false};
-  bool isImplicitNoneExternal_{false};
+  std::optional<bool> isImplicitNoneType_;
```
sscalpone: I think you need a comment here or near here that explains these booleans and what conditions affect inheritFromParent.

tskeith: OK, done

lib/semantics/resolve-names.cc
```diff
@@ -37,31 +37,41 @@ class MessageHandler;
...
-  ImplicitRules(MessageHandler &messages);
-  bool isImplicitNoneType() const { return isImplicitNoneType_; }
-  bool isImplicitNoneExternal() const { return isImplicitNoneExternal_; }
+  ImplicitRules(MessageHandler &messages)
+    : messages_{messages}, inheritFromParent_{false} {}
+  ImplicitRules(std::unique_ptr<ImplicitRules> &&parent)
+    : messages_{parent->messages_}, inheritFromParent_{true} {
+    parent_.swap(parent);
+  }
+  std::unique_ptr<ImplicitRules> &&parent() { return std::move(parent_); }
```
klausler: I have never before seen a function return an rvalue reference.  I wonder whether there are other interfaces in f18 code that return lvalue references that should be rvalues?  (No action req'd)

PR #170, opened 2018-08-27
Add label resolution initial pass implementation and tests

schweitzpgi:


tskeith:
Some overall comments:
Run clang-format on all of the changes so the formatting is consistent.
Local variables and formal parameters should start with a lower-case letter.
The tests should be run if you go into the build directory and run `ctest`.

lib/semantics/resolve-labels.cc
```diff
@@ -0,0 +1,832 @@
...
+using namespace Fortran;
+using namespace parser::literals;
+using ParseTree_t = parser::Program;
+using CookedSource_t = parser::CookedSource;
+using Index_t = parser::CharBlock;
+using IndexList = std::vector<std::pair<Index_t, Index_t>>;
+using Scope_t = unsigned;
+using LblStmt_t = std::tuple<Scope_t, Index_t, unsigned>;
+using ArcTrgt_t = std::map<parser::Label, LblStmt_t>;
+using ArcBase_t = std::vector<std::tuple<parser::Label, Scope_t, Index_t>>;
```
tskeith: I think that comments about what these types represent would be helpful. Also, it's better to use full words rather than abbreviations. I.e. 'Lbl' -> 'Label', 'Trgt' -> 'Target', etc.

lib/semantics/resolve-labels.cc
```diff
@@ -0,0 +1,832 @@
...
+  void Post(const parser::CaseConstruct& Case) { PopName(Case); }
+  void Post(const parser::SelectRankConstruct& SelRk) { PopName(SelRk); }
+  void Post(const parser::SelectTypeConstruct& SelTy) { PopName(SelTy); }
+
+  //  Named constructs without block scope
+  void Post(const parser::WhereConstruct& W) { PopNonBlockConstructName(W); }
+  void Post(const parser::ForallConstruct& F) { PopNonBlockConstructName(F); }
+
+  //  Statements with label references
+  void Post(const parser::LabelDoStmt& Do) { AddDoBase(std::get<1>(Do.t)); }
```
tskeith: `std::get<parser::Label>(Do.t)` is more readable than the index form of get.

lib/semantics/resolve-labels.cc
```diff
@@ -0,0 +1,832 @@
...
+  void CheckName(const parser::CriticalConstruct& C) { ChkNm(C, "CRITICAL"); }
+  /// \brief Check <i>do-construct-name</i>, constraint C1131
+  void CheckName(const parser::DoConstruct& Do) { ChkNm(Do, "DO"); }
+  /// \brief Check <i>forall-construct-name</i>, constraint C1035
+  void CheckName(const parser::ForallConstruct& F) { ChkNm(F, "FORALL"); }
+  /// \brief Common code for ASSOCIATE, CRITICAL, DO, and FORALL
+  template<typename A> void ChkNm(const A& V, const char *const Con) {
+    if (!BothEqOrNone(std::get<0>(std::get<0>(V.t).statement.t),
+		      std::get<2>(V.t).statement.v)) {
+      EH.Report(Index, "%s construct name mismatch"_err_en_US, Con);
```
tskeith: You can attach an information message to the error that points to the name you are trying to match. See `Say2` in `resolve-names.cc` for an example. At minimum I think the message should include both names.

lib/semantics/resolve-labels.cc
```diff
@@ -0,0 +1,832 @@
...
+  }
+  void Post(const parser::ExitStmt& Exit) {
+    if (Exit.v.has_value())
+      CheckLabelContext("EXIT", Exit.v.value().ToString());
+  }
+
+  // Getters for the results
+  const std::vector<UnitAnalysis>& GetProgramUnits() const { return PUnits; }
+  ErrorHandler& GetEH() { return EH; }
+  bool HasNoErrors() const { return NoErrors; }
```
tskeith: Can you implement this with `EH.messages.empty()` ? Then you wouldn't have to worry about setting clearing `NoErrors` whenever an error is reported.

lib/semantics/resolve-labels.cc
```diff
@@ -0,0 +1,832 @@
...
+  /// \brief Check constraint <i>construct-name</i> in scope (C1134 and C1166)
+  /// \param SStr  a string to specify the statement, \c CYCLE or \c EXIT
+  /// \param Label the name used by the \c CYCLE or \c EXIT
+  template<typename A> void CheckLabelContext(const char* const SStr,
+					      const A& Name) {
+    auto E{Names.crend()};
+    for (auto I{Names.crbegin()}; I != E; ++I) {
+      if (*I == Name)
+	return;
+    }
```
tskeith: Can you replace this loop with `std::find()`?

lib/semantics/resolve-labels.cc
```diff
@@ -0,0 +1,832 @@
...
+	   std::get_if<common::Indirection<parser::ExitStmt>>(P) ||
+	   std::get_if<common::Indirection<parser::StopStmt>>(P) ||
+	   std::get_if<common::Indirection<parser::GotoStmt>>(P) ||
+	   std::get_if<common::Indirection<parser::ReturnStmt>>(P));
+}
+
+/// \brief Is this a FORMAT stmt?
+/// Pattern match for FORMAT statement
+template<typename A> constexpr bool IsFormat(const parser::Statement<A>&) {
+  return false;
```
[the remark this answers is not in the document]
schweitzpgi: Thanks. Done.

lib/semantics/resolve-labels.cc
schweitzpgi: OK, done. (Doesn't reduce lines of code, though.)

lib/semantics/resolve-labels.cc
schweitzpgi: Not all the messages are errors.

tskeith: `Messages` also has `AnyFatalError()` which might match what you need.

lib/parser/message.cc
```diff
@@ -35,6 +35,19 @@ std::ostream &operator<<(std::ostream &o, const MessageFixedText &t) {
...
+  SetMessageFormattedText(text, ap);
+  va_end(ap);
+}
+
+MessageFormattedText::MessageFormattedText(MessageFixedText text, va_list ap)
+  : isFatal_{text.isFatal()} {
+  SetMessageFormattedText(text, ap);
+}
+
+void MessageFormattedText::SetMessageFormattedText(MessageFixedText text,
```
klausler: I get where this member function name came from but it might be more clear if it were shorter -- we don't need to use the name of the class redundantly in the name of the member function if it's not a constructor or destructor.  Even "vsnprintf" might be better.

lib/semantics/resolve-labels.h
```diff
@@ -0,0 +1,40 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef FORTRAN_SEMANTICS_RESOLVE_LABELS_H_
+#define FORTRAN_SEMANTICS_RESOLVE_LABELS_H_
+
+namespace Fortran {
```
klausler: C++17 allows `namespace Fortran::parser` and `namespace Fortran::semantics`, and it would be better if the code were consistent here with other usage like that instead of using nested `namespace`.

lib/semantics/resolve-labels.h
```diff
@@ -0,0 +1,40 @@
...
+/// \param ParseTree  the parse tree
+/// \param Source     the cooked source
+/// \return true, iff the program's labels pass semantics checks
+bool ValidateLabels(const parser::Program &ParseTree,
+		    const parser::CookedSource &Source);
+} // semantics
+} // Fortran
+
+#endif // FORTRAN_SEMANTICS_RESOLVE_LABELS_H_
+
```
klausler: Let's not have this extra cruft at the end of the file.

lib/semantics/resolve-labels.h
```diff
@@ -0,0 +1,40 @@
...
+class CookedSource;
+} // parser
+
+namespace semantics {
+
+/// \brief Validate the labels in the program
+/// \param ParseTree  the parse tree
+/// \param Source     the cooked source
+/// \return true, iff the program's labels pass semantics checks
+bool ValidateLabels(const parser::Program &ParseTree,
```
klausler: Where are the messages accounted for in this interface?  Maybe you should return a `Messages` and let the caller check it for fatal errors.

test/semantics/label01.F90
```diff
@@ -0,0 +1,226 @@
...
+           else
+              exit loopthree
+           end if condthree
+        end do loopthree
+     end if condone
+  end do loopone
+111 print *, "done"
+end subroutine do_loop08
+
+#ifndef STRICT_F18
```
klausler: Should the f18 preprocessor automatically define this macro?  Probably not, but something to consider.

tools/f18/f18.cc
```diff
@@ -215,6 +216,13 @@ std::string CompileFortran(
     }
     Fortran::semantics::ResolveNames(Fortran::semantics::Scope::globalScope,
         parseTree, parsing.cooked(), directories);
+    const auto& Cook = parsing.cooked();
```
klausler: For consistency: `const auto &cook{parsing.cooked()};`

We write '&' and '*' next to the identifier; we name variables, constants, and data members with lower-case initial letters; and we use braced initialization now that C++17 has fixed it to work with `auto`.

lib/semantics/resolve-labels.cc
```diff
@@ -0,0 +1,832 @@
...
+#include <cstdarg>
+#include <iostream>
+#include <cctype>
+#include <cassert>
+
+namespace {
+
+using namespace Fortran;
+using namespace parser::literals;
+using ParseTree_t = parser::Program;
```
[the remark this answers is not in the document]
schweitzpgi: Emacs like _t suffixes.

lib/semantics/resolve-labels.cc
```diff
@@ -0,0 +1,832 @@
...
+    Flags |= BRANCH_TARGET_FLAG;
+  if (IsFormat(S))
+    Flags |= FORMAT_STMT_FLAG;
+  return Flags;
+}
+
+/// \brief \p opt1 and \p opt2 must be either present and identical or absent
+/// \param opt1  an optional construct-name (opening statement)
+/// \param opt2  an optional construct-name (ending statement)
+template<typename A> inline bool BothEqOrNone(const A& opt1, const A& opt2) {
```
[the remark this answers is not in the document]
schweitzpgi: I'm leveraging type variables and type inference, as the concrete representation is not as important.

lib/semantics/resolve-labels.cc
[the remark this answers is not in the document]
sscalpone: Is it prohibited by llvm or google coding standards?  If not, and we want to adopt _t suffix, then let's do it *everywhere* and be consistent please.

lib/semantics/resolve-labels.cc
schweitzpgi: Right. I didn't spend very much time trying to craft "supermodel" error messages. I was thinking of doing that as another refactoring pass.

lib/semantics/resolve-labels.cc
schweitzpgi: OK

lib/semantics/resolve-labels.cc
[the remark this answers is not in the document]
klausler: The usage isn't prohibited by the public Google C++ coding style guide, but `_t` is never used for new type names.  C++ class names should start with capital letters and that's sufficient for making them stand out as such.  If we're going to start appending `_t` to them, let's do that as a massive naming-only edit over the whole code base later.

lib/parser/message.cc
schweitzpgi: OK

lib/semantics/resolve-labels.h
schweitzpgi: OK

lib/semantics/resolve-labels.cc
```diff
@@ -0,0 +1,817 @@
...
+using ParseTree_t = parser::Program;
+using CookedSource_t = parser::CookedSource;
+using Index_t = parser::CharBlock;
+using IndexList = std::vector<std::pair<Index_t, Index_t>>;
+using Scope_t = unsigned;
+using LblStmt_t = std::tuple<Scope_t, Index_t, unsigned>;
+using ArcTrgt_t = std::map<parser::Label, LblStmt_t>;
+using ArcBase_t = std::vector<std::tuple<parser::Label, Scope_t, Index_t>>;
+
+const bool StrictF18 = false; // FIXME - make a command-line option
```
klausler: `static constexpr` is better than declaring a variable.

Booleans stand out better when named like `isStrictF18`.

Please use lower-case letters when naming variables, constants, data members, &c.

We use braced initialization uniformly now that C++17 has fixed it for `auto`.

lib/semantics/resolve-labels.cc
```diff
@@ -0,0 +1,817 @@
...
+
+const unsigned DO_TERM_FLAG = 1u;
```
klausler: Take a look at lib/common/enum-set.h.

lib/semantics/resolve-labels.cc
```diff
@@ -0,0 +1,817 @@
...
+  const parser::CookedSource& cookedSource;
+  parser::Messages messages;
+};
+
+/// \brief Is this a legal DO terminator?
+/// Pattern match dependent on the standard we're enforcing
+/// F18:R1131 (must be CONTINUE or END DO)
+template<typename A> constexpr bool IsLegalDoTerm(const parser::Statement<A>&) {
+  using EDS = parser::EndDoStmt;
+  return std::disjunction<std::is_same<A,common::Indirection<EDS>>,
```
klausler: `std::disjunction` is a useful tool, but it's overkill here vs. the more readable `std::is_same_v<A,...> || std::is_same_v<A, ...>`

lib/semantics/resolve-labels.cc
```diff
@@ -0,0 +1,817 @@
...
+			  std::is_same<A,EDS>>::value;
+}
+template<> constexpr bool IsLegalDoTerm(const parser::Statement<parser::
+					ActionStmt>& A) {
+  if (std::get_if<parser::ContinueStmt>(&A.statement.u)) {
```
klausler: Use `std::holds_alternative<>()` rather than `get_if` and an implicit null pointer test.

lib/semantics/resolve-labels.cc
```diff
@@ -0,0 +1,817 @@
...
+			  std::is_same<A,parser::ForallStmt>,
+			  std::is_same<A,parser::WhereConstructStmt>,
+			  std::is_same<A,parser::EndFunctionStmt>,
+			  std::is_same<A,parser::EndMpSubprogramStmt>,
+			  std::is_same<A,parser::EndProgramStmt>,
+			  std::is_same<A,parser::EndSubroutineStmt>>::value;
+}
+template<> constexpr bool IsLegalBranchTarget(const parser::Statement<parser::
+					      ActionStmt>& A) {
+  if (!StrictF18)
```
klausler: Always use braces around nested statements.

lib/semantics/resolve-labels.cc
```diff
@@ -0,0 +1,817 @@
...
+public:
+  struct UnitAnalysis {
+  public:
+    ArcBase_t DoArcBases;      ///< bases of label-do-stmts
+    ArcBase_t FmtArcBases;     ///< bases of all other stmts with labels
+    ArcBase_t ArcBases;        ///< bases of all other stmts with labels
+    ArcTrgt_t ArcTrgts;        ///< unique map of labels to stmt info
+    std::vector<Scope_t> Scopes; ///< scope stack model
+
+    explicit UnitAnalysis() { Scopes.push_back(0); }
```
klausler: `explicit` should be used only when needed.

lib/semantics/resolve-labels.cc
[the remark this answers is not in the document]
schweitzpgi: It's done in LLVM code, but not extensively. e.g.,
./unittests/Support/Path.cpp:  typedef std::vector<std::string> v_t;

lib/semantics/resolve-labels.h
schweitzpgi: I assume this means you don't want any Emacs variables in the code.

klausler:
Not in every single source file, no.  I think that there are ways to put this kind of emacs configuration boilerplate in a single centralized location in a project that have been used by previous programmers on this project who use emacs.

lib/semantics/resolve-labels.cc
```diff
@@ -0,0 +1,832 @@
...
+template<typename A>
+constexpr unsigned ConsTrgtFlags(const parser::Statement<A>& S) {
+  unsigned Flags{0u};
+  if (IsLegalDoTerm(S))
+    Flags |= DO_TERM_FLAG;
+  if (IsLegalBranchTarget(S))
+    Flags |= BRANCH_TARGET_FLAG;
+  if (IsFormat(S))
+    Flags |= FORMAT_STMT_FLAG;
+  return Flags;
```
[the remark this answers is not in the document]
schweitzpgi: OK

lib/parser/message.cc
schweitzpgi: Removed this

test/semantics/label01.F90
schweitzpgi: Good idea. I think I'll defer for now, but I'll make a note.

schweitzpgi:
I believe all the comments have been addressed (or indicated to be deferred in a couple cases).

tskeith:
When I go into my build directory and do `ctest -V -R label`, I get this:
```
60: Test command: /local/home/tkeith/f18/easddd/test/semantics/test_any.sh "label*.[Ff]90"
60: Test timeout computed to be: 10000000
60: cat: label*.[Ff]90: No such file or directory
60: /local/home/tkeith/f18/easddd/test/semantics/test_any.sh: eval: line 56: syntax error near unexpected token `)'
60: /local/home/tkeith/f18/easddd/test/semantics/test_any.sh: eval: line 56: `(  )'
60: test label*.[Ff]90 failed
1/1 Test #60: label*.[Ff]90 ....................   Passed    0.02 sec
```

So it looks like the tests aren't running correctly.

lib/semantics/resolve-labels.cc
```diff
@@ -0,0 +1,969 @@
...
+
+const bool isStrictF18{false};  // FIXME - make a command-line option
+
+inline bool HasScope(ScopeProxy scope) { return scope != ScopeProxy{0u}; }
+
+inline bool HasNoScope(ScopeProxy scope) { return !HasScope(scope); }
+
+parser::Message &Report(ErrorHandler &eh, const parser::CharBlock &ip,
+    parser::MessageFormattedText &&msg) {
+  return eh.Say(parser::Message{ip, msg});
```
tskeith: This can just be `return eh.Say(ip, msg);`.

lib/semantics/resolve-labels.cc
```diff
@@ -0,0 +1,969 @@
...
+
+  /// \brief References to labeled statements as (possible) branches
+  void AddBase(const std::list<parser::Label> &labels) {
+    for (const parser::Label &label : labels) {
+      AddBase(label);
+    }
+  }
+
+  std::vector<UnitAnalysis> programUnits_;  ///< results for each program unit
+  ErrorHandler eh;  ///< error handler, collects messages
```
tskeith: `eh` should be `eh_`.

lib/semantics/resolve-labels.cc
```diff
@@ -0,0 +1,969 @@
...
+namespace {
+
+using namespace Fortran;
+using namespace parser::literals;
+
+ENUM_CLASS(TargetStatementEnum, Do, Branch, Format)
+using TargetStmtType = common::EnumSet<TargetStatementEnum, 8>;
+
+using IndexList = std::vector<std::pair<parser::CharBlock, parser::CharBlock>>;
+using ScopeProxy = unsigned;
```
tskeith: What is a `ScopeProxy`? I can't tell from the type so a comment would be helpful.

lib/semantics/resolve-labels.cc
```diff
@@ -0,0 +1,969 @@
...
+  if (IsFormat(statement)) {
+    targetStmtType.set(TargetStatementEnum::Format);
+  }
+  return targetStmtType;
+}
+
+/// \brief \p opt1 and \p opt2 must be either present and identical or absent
+/// \param a  an optional construct-name (opening statement)
+/// \param b  an optional construct-name (ending statement)
+template<typename A> inline bool BothEqOrNone(const A &a, const A &b) {
```
tskeith: Why not just define this for `std::optional<Name>`? That seems to be the only way it is used.

lib/semantics/resolve-labels.cc
```diff
@@ -0,0 +1,969 @@
...
+  /// Constraint per section 6.2.5., paragraph 2
+  void AddTrgt(parser::Label label, TargetStmtType targetStmtType) {
+    CheckLabelInRange(label);
+    const auto pair{programUnits_.back().targetStmts_.insert(
+        {label, {currentScope_, currentPosition_, targetStmtType}})};
+    if (!pair.second) {
+      // this is an error: labels must be pairwise distinct
+      Report(eh, currentPosition_,
+          parser::MessageFormattedText{
+              "label '%lu' is not distinct"_err_en_US, label});
```
tskeith: Can this error point to the duplicate?

lib/semantics/resolve-labels.cc
```diff
@@ -0,0 +1,969 @@
...
+  parser::CharBlock currentPosition_{
+      nullptr};  ///< current location in parse tree
+  ScopeProxy currentScope_{0};  ///< current scope in the model
+  std::vector<std::string> constructNames_;
+};
+
+template<typename A, typename B>
+bool InInclusiveScope(const A &scopes, B tail, const B &head) {
+  assert(HasScope(head));
+  assert(HasScope(tail));
```
tskeith: Elsewhere we have been using `CHECK` from `common/idioms.h` rather than `assert`.

lib/semantics/resolve-labels.cc
```diff
@@ -0,0 +1,969 @@
...
+}
+
+/// \brief Iterates over parse tree, creates the analysis result
+/// As a side-effect checks the constraints for the usages of
+/// <i>construct-name</i>.
+struct ParseTreeAnalyzer {
+public:
+  struct UnitAnalysis {
+  public:
+    SourceStmtList doStmtSources_;  ///< bases of label-do-stmts
```
klausler: If class data members are public, use a `struct` and don't affix `_` to their names.  But for most cases apart from simple tuple-like containers, use `class`, make the data members private, and retain the `_`.

lib/semantics/resolve-labels.cc
```diff
@@ -0,0 +1,969 @@
...
+    SourceStmtList
+        formatStmtSources_;  ///< bases of all other stmts with labels
+    SourceStmtList otherStmtSources_;  ///< bases of all other stmts with labels
+    TargetStmtMap targetStmts_;  ///< unique map of labels to stmt info
+    std::vector<ScopeProxy> scopeModel_;  ///< scope stack model
+
+    UnitAnalysis() { scopeModel_.push_back(0); }
+    UnitAnalysis(UnitAnalysis &&) = default;
+    ~UnitAnalysis() = default;
+    UnitAnalysis(const UnitAnalysis &) = delete;
```
klausler: Deletion of the copy constructor is implied by the C++ programming language in the presence of a move constructor.

lib/semantics/resolve-labels.cc
```diff
@@ -0,0 +1,969 @@
...
+    SourceStmtList doStmtSources_;  ///< bases of label-do-stmts
+    SourceStmtList
+        formatStmtSources_;  ///< bases of all other stmts with labels
+    SourceStmtList otherStmtSources_;  ///< bases of all other stmts with labels
+    TargetStmtMap targetStmts_;  ///< unique map of labels to stmt info
+    std::vector<ScopeProxy> scopeModel_;  ///< scope stack model
+
+    UnitAnalysis() { scopeModel_.push_back(0); }
+    UnitAnalysis(UnitAnalysis &&) = default;
+    ~UnitAnalysis() = default;
```
klausler: Implied by the language.

lib/semantics/resolve-labels.cc
```diff
@@ -0,0 +1,969 @@
...
+    const SourceStmtList &GetBranches() const { return otherStmtSources_; }
+    const TargetStmtMap &GetLabels() const { return targetStmts_; }
+    const std::vector<ScopeProxy> &GetScopes() const { return scopeModel_; }
+  };
+
+  ParseTreeAnalyzer() {}
+  ~ParseTreeAnalyzer() = default;
+  ParseTreeAnalyzer(ParseTreeAnalyzer &&) = default;
+  ParseTreeAnalyzer(const ParseTreeAnalyzer &) = delete;
+  ParseTreeAnalyzer &operator=(const ParseTreeAnalyzer &) = delete;
```
klausler: You delete the copy assignment, but not the move assignment.  If you don't want default assignments, delete both.  If you want just move assignment, define it and ignore the copy assignment.  Also applies to above.

lib/semantics/resolve-labels.cc
```diff
@@ -0,0 +1,969 @@
...
+
+  // Default Pre() and Post()
+  template<typename A> constexpr bool Pre(const A &) { return true; }
+  template<typename A> constexpr void Post(const A &) {}
+
+  // Specializations of Pre() and Post()
+
+  /// \brief Generic handling of all statements
+  template<typename A> bool Pre(const parser::Statement<A> &Stmt) {
```
klausler: Please don't capitalize local variable names.

lib/semantics/resolve-labels.cc
```diff
@@ -0,0 +1,969 @@
...
+  bool Pre(const parser::CriticalConstruct &criticalConstruct) {
+    return PushName(criticalConstruct);
+  }
+  bool Pre(const parser::DoConstruct &doConstruct) {
+    return PushName(doConstruct);
+  }
+  bool Pre(const parser::IfConstruct &ifConstruct) {
+    return PushName(ifConstruct);
+  }
+  bool Pre(const parser::IfConstruct::ElseIfBlock &) { return SwScope(); }
```
klausler: I think that `SwScope` could benefit from a more self-explanatory name.

lib/semantics/resolve-labels.cc
```diff
@@ -0,0 +1,969 @@
...
+    if (const auto *P{std::get_if<0>(&format.u)}) {
+      AddFmtBase(parser::Label{std::get<0>(std::get<parser::IntLiteralConstant>(
+          std::get<parser::LiteralConstant>((*P->thing).u).u)
+                                               .t)});
+    }
+#endif
+  }
+  void Post(const parser::CycleStmt &cycleStmt) {
+    if (cycleStmt.v.has_value()) {
+      CheckLabelContext("CYCLE", cycleStmt.v.value().ToString());
```
klausler: `.value().` can be much more clearly written as just `->`, here and elsewhere.

lib/semantics/resolve-labels.cc
```diff
@@ -0,0 +1,969 @@
...
+    if (exitStmt.v.has_value()) {
+      CheckLabelContext("EXIT", exitStmt.v.value().ToString());
+    }
+  }
+
+  // Getters for the results
+  const std::vector<UnitAnalysis> &GetProgramUnits() const {
+    return programUnits_;
+  }
+  ErrorHandler &GetEH() { return eh; }
```
klausler: Eh?

Typically, simple accessors to data members like `foo_` can be named just `foo()`.  In this case, consider `Messages &messages() { return messages_; }`.

lib/semantics/resolve-labels.cc
```diff
@@ -0,0 +1,969 @@
...
+
+private:
+  bool PushScope() {
+    programUnits_.back().scopeModel_.push_back(currentScope_);
+    currentScope_ = programUnits_.back().scopeModel_.size() - 1;
```
klausler: Maybe `scopeModel_` could be a `std::stack` and then you could just mess with its top element rather than a `currentScope_` index?

lib/semantics/resolve-labels.cc
```diff
@@ -0,0 +1,969 @@
...
+  /// \brief Check <i>select-construct-name</i> and pop the scope
+  /// Constraint C1165 - opening and ending name must match if present, and
+  /// <i>type-guard-stmt</i> must either match or be unnamed
+  void PopName(const parser::SelectTypeConstruct &selectTypeConstruct) {
+    CheckName(selectTypeConstruct, "TYPE", "TYPE ");
+    PopScope();
+    SelectivePopBack(selectTypeConstruct);
+  }
+
+  // -----------------------------------------------
```
klausler: We generally don't expend screen space on horizontal rules.

lib/semantics/resolve-labels.cc
```diff
@@ -0,0 +1,969 @@
...
+                "ELSE IF statement name mismatch"_err_en_US});
+      }
+    }
+    if (std::get<std::optional<parser::IfConstruct::ElseBlock>>(ifConstruct.t)
+            .has_value()) {
+      if (!PresentAndEq(
+              std::get<parser::Statement<parser::ElseStmt>>(
+                  std::get<std::optional<parser::IfConstruct::ElseBlock>>(
+                      ifConstruct.t)
+                      .value()
```
klausler: If you're grepping and replacing `.value().` with `->` as suggested above, don't miss this site with the extra newline.

lib/semantics/resolve-labels.cc
```diff
@@ -0,0 +1,969 @@
...
+      }
+    }
+  }
+
+  /// \brief Check constraint <i>construct-name</i> in scope (C1134 and C1166)
+  /// \param SStr  a string to specify the statement, \c CYCLE or \c EXIT
+  /// \param Label the name used by the \c CYCLE or \c EXIT
+  template<typename A>
+  void CheckLabelContext(const char *const stmtString, const A &constructName) {
+    const auto I{std::find(
```
klausler: Please avoid capital letters as local variable names.

lib/semantics/resolve-labels.cc
```diff
@@ -0,0 +1,969 @@
...
+          parser::MessageFormattedText{
+              "%s construct-name '%s' is not in scope"_err_en_US, stmtString,
+              constructName.c_str()});
+    }
+  }
+
+  /// \brief Check label range
+  /// Constraint per section 6.2.5, paragraph 2
+  void CheckLabelInRange(parser::Label label) {
+    if ((label < 1) || (label > 99999)) {
```
klausler: Redundant parentheses.

lib/semantics/resolve-labels.cc
```diff
@@ -0,0 +1,969 @@
...
+          parser::MessageFormattedText{
+              "label '%lu' is out of range"_err_en_US, label});
+    }
+  }
+
+  /// \brief Add a labeled statement (label must be distinct)
+  /// Constraint per section 6.2.5., paragraph 2
+  void AddTrgt(parser::Label label, TargetStmtType targetStmtType) {
+    CheckLabelInRange(label);
+    const auto pair{programUnits_.back().targetStmts_.insert(
```
klausler: `emplace` is more common usage than `insert` when the new element is being constructed, assuming that this is a `std::map` or other STL(-ish) container class.

lib/semantics/resolve-labels.cc
klausler: Are you sure that `label` is `long`?  I bet it isn't, since labels are in [1..99999].  The format `%lu` may only be working by accident.

lib/semantics/resolve-labels.cc
```diff
@@ -0,0 +1,969 @@
...
+          parser::MessageFormattedText{
+              "label '%lu' is not distinct"_err_en_US, label});
+    }
+    // Don't enforce a limit to the cardinality of labels
+  }
+
+  /// \brief Reference to a labeled statement from a DO statement
+  void AddDoBase(parser::Label label) {
+    CheckLabelInRange(label);
+    programUnits_.back().doStmtSources_.push_back(
```
klausler: `emplace_back` is probably preferable to `push_back` when the new element is being constructed.  Other instances follow.

schweitzpgi:
Not sure what happened to the tests. It was working for me.

lib/semantics/resolve-labels.cc
schweitzpgi: The trailing underscore serves no purpose in my opinion.

lib/semantics/resolve-labels.cc
```diff
@@ -0,0 +1,969 @@
...
+      nullptr};  ///< current location in parse tree
+  ScopeProxy currentScope_{0};  ///< current scope in the model
+  std::vector<std::string> constructNames_;
+};
+
+template<typename A, typename B>
+bool InInclusiveScope(const A &scopes, B tail, const B &head) {
+  assert(HasScope(head));
+  assert(HasScope(tail));
+  while (HasScope(tail) && (tail != head)) {
```
klausler: Redundant parentheses.  Also, the flow is somewhat unclear.  Assuming that `HasScope` is pure, consider this:

```
for (; tail != head; tail = scopes[tail]) {
  if (!HasScope(tail)) {
    return false;
  }
}
return true;
```

lib/semantics/resolve-labels.cc
```diff
@@ -0,0 +1,969 @@
...
+  }
+  /// \brief Common code for ASSOCIATE, CRITICAL, DO, and FORALL
+  template<typename A>
+  void CheckName(const char *const constructTag, const A &a) {
+    if (!BothEqOrNone(
+            std::get<std::optional<parser::Name>>(std::get<0>(a.t).statement.t),
+            std::get<2>(a.t).statement.v)) {
+      Report(eh, currentPosition_,
+          parser::MessageFormattedText{
+              "%s construct name mismatch"_err_en_US, constructTag});
```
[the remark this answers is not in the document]
schweitzpgi: The verb say is so generic in English as to be virtually invisible.

lib/semantics/resolve-labels.cc
```diff
@@ -0,0 +1,969 @@
...
+}
+
+ParseTreeAnalyzer LabelAnalysis(const parser::Program &program) {
+  ParseTreeAnalyzer analysis;
+  Walk(program, analysis);
+  return analysis;
+}
+
+template<typename A, typename B>
+inline bool InBody(const A &position, const B &pair) {
```
klausler: 'inline' probably not necessary.

lib/semantics/resolve-labels.cc
```diff
@@ -0,0 +1,969 @@
...
+  assert(pair.first.begin() < pair.second.begin());
```
klausler: `CHECK` is in lib/common/idioms.h.

lib/semantics/resolve-labels.cc
schweitzpgi: Why not let type inference do it?

lib/semantics/resolve-labels.cc
```diff
@@ -0,0 +1,969 @@
...
+  return (position.begin() >= pair.first.begin()) &&
```
klausler: Code that's implementing predicates over contiguous intervals might benefit from use of `lib/common/interval.h`.

Is this template really instantiated over multiple types for both arguments?

Also, there's redundant parentheses.

lib/semantics/resolve-labels.cc
```diff
@@ -0,0 +1,969 @@
...
+    if (HasScope(std::get<ScopeProxy>(branchTarget))) {
+      const auto &fromPosition{std::get<parser::CharBlock>(branch)};
+      const auto &toPosition{std::get<parser::CharBlock>(branchTarget)};
+      for (const auto body : loopBodies) {
+        if (!InBody(fromPosition, body) && InBody(toPosition, body)) {
+          // this is an error: branch into labeled DO body
+          if (isStrictF18) {
+            Report(eh, fromPosition,
+                parser::MessageFormattedText{
+                    "branch into '%s' from another scope"_err_en_US,
```
klausler: Does this have to be an error?  Maybe.  See comment above.

lib/semantics/resolve-labels.cc
```diff
@@ -0,0 +1,969 @@
...
+
+/// \brief Check that DO loops properly nest
+template<typename A>
+inline void CheckDoNesting(const A &loopBodies, ErrorHandler &eh) {
+  for (auto i1{loopBodies.cbegin()}; i1 != loopBodies.cend(); ++i1) {
+    const auto &v1{*i1};
+    for (auto i2{i1 + 1}; i2 != loopBodies.cend(); ++i2) {
+      const auto &v2{*i2};
+      assert(v1.first.begin() != v2.first.begin());
+      if ((v2.first.begin() < v1.second.end()) &&
```
klausler: Might be more clear as something like `Overlap(v1, v2) && !Contains(v1, v2) && !Contains(v2, v1)`.

lib/semantics/resolve-labels.cc
```diff
@@ -0,0 +1,969 @@
...
+}
+
+/// \brief Advance \p Pos past any label and whitespace
+/// Want the statement without its label for error messages, range checking
+template<typename A> inline A SkipLabel(const A &position) {
+  const long maxPosition{position.end() - position.begin()};
+  if (maxPosition && (position[0] >= '0') && (position[0] <= '9')) {
+    long i{1l};
+    for (; (i < maxPosition) && std::isdigit(position[i]); ++i)
+      ;
```
klausler: Please use empty braces for empty bodies rather than `;` or `continue;`.

lib/semantics/resolve-labels.cc
```diff
@@ -0,0 +1,969 @@
...
+  for (const auto stmt : dos) {
+    const auto &label{std::get<parser::Label>(stmt)};
+    const auto &scope{std::get<ScopeProxy>(stmt)};
+    const auto &position{std::get<parser::CharBlock>(stmt)};
+    auto doTarget{GetLabel(labels, label)};
+    if (HasNoScope(std::get<ScopeProxy>(doTarget))) {
+      // C1133: this is an error: label not found
+      Report(eh, position,
+          parser::MessageFormattedText{
+              "label '%lu' cannot be found"_err_en_US, label});
```
klausler: Please check that `label` is indeed `long`, or change the format.

lib/semantics/resolve-labels.cc
```diff
@@ -0,0 +1,969 @@
...
+      if (isStrictF18) {
+        Report(eh, position,
+            parser::MessageFormattedText{
+                "label '%lu' is not in scope"_err_en_US, label});
+      } else {
+        Report(eh, position,
+            parser::MessageFormattedText{
+                "label '%lu' is not in scope"_en_US, label});
+      }
+    } else if ((std::get<TargetStmtType>(doTarget) &
```
klausler: Use `.test(TargetStatementEnum::Do)` when checking for a nonempty intersection with a singleton.

lib/semantics/resolve-labels.cc
```diff
@@ -0,0 +1,969 @@
...
+    const auto &position{std::get<parser::CharBlock>(stmt)};
+    auto target{GetLabel(labels, label)};
+    if (HasNoScope(std::get<ScopeProxy>(target))) {
+      // this is an error: label not found
+      Report(eh, position,
+          parser::MessageFormattedText{
+              "label '%lu' was not found"_err_en_US, label});
+    } else if (!InInclusiveScope(scopes, scope, std::get<ScopeProxy>(target))) {
+      // this is an error: label not in scope
+      if (isStrictF18) {
```
klausler: See previous comments about errors vs warnings and the use of `%lu`.

(Pausing review here, will look at the rest of this code later.)

lib/semantics/resolve-labels.cc
klausler: Abstraction over types is like charging a local readability tax in the hope that there's a *net* gain in readability &/or correctness when the abstraction is applied to multiple instantiations.  When there's but a single instantiation of a template and no reasonable expectation of others in the future, concrete types will probably be more readable.

lib/semantics/resolve-labels.cc
tskeith: That's just the convention for private data members.

lib/semantics/resolve-labels.cc
schweitzpgi: deleted

lib/semantics/resolve-labels.cc
schweitzpgi: added comment

lib/semantics/resolve-labels.cc
schweitzpgi: added

lib/semantics/resolve-labels.cc
schweitzpgi: sure

lib/semantics/resolve-labels.cc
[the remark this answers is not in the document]
schweitzpgi: Report was deleted along with other functions

lib/semantics/resolve-labels.cc
schweitzpgi: Concrete types were pushed everywhere possible

lib/semantics/resolve-labels.cc
schweitzpgi: Probably

lib/semantics/resolve-labels.cc
```diff
@@ -0,0 +1,969 @@
...
+#include "resolve-labels.h"
+#include "../common/enum-set.h"
+#include "../parser/message.h"
+#include "../parser/parse-tree-visitor.h"
+#include <cassert>
+#include <cctype>
+#include <cstdarg>
+#include <iostream>
+
+namespace {
```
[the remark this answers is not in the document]
schweitzpgi: It's an LLVM programmer thing. deleted

lib/semantics/resolve-labels.cc
```diff
@@ -0,0 +1,969 @@
...
+
+using namespace Fortran;
```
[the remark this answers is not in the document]
schweitzpgi: deleted

lib/semantics/resolve-labels.cc
```diff
@@ -0,0 +1,969 @@
...
+using namespace parser::literals;
+
+ENUM_CLASS(TargetStatementEnum, Do, Branch, Format)
+using TargetStmtType = common::EnumSet<TargetStatementEnum, 8>;
```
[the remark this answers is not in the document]
schweitzpgi: sure

lib/semantics/resolve-labels.cc
```diff
@@ -0,0 +1,969 @@
...
+
+using IndexList = std::vector<std::pair<parser::CharBlock, parser::CharBlock>>;
+using ScopeProxy = unsigned;
+using LabelStmtInfo = std::tuple<ScopeProxy, parser::CharBlock, TargetStmtType>;
```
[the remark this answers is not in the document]
schweitzpgi: I used the tuple methods

lib/semantics/resolve-labels.cc
```diff
@@ -0,0 +1,969 @@
...
+using TargetStmtMap = std::map<parser::Label, LabelStmtInfo>;
+using SourceStmtList =
+    std::vector<std::tuple<parser::Label, ScopeProxy, parser::CharBlock>>;
+using ErrorHandler = parser::Messages;
```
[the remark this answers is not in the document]
schweitzpgi: ErrorHandler deleted

lib/semantics/resolve-labels.cc
```diff
@@ -0,0 +1,969 @@
...
+
+const bool isStrictF18{false};  // FIXME - make a command-line option
+
+inline bool HasScope(ScopeProxy scope) { return scope != ScopeProxy{0u}; }
+
+inline bool HasNoScope(ScopeProxy scope) { return !HasScope(scope); }
```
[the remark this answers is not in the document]
schweitzpgi: deleted

lib/semantics/resolve-labels.cc
```diff
@@ -0,0 +1,969 @@
...
+
+parser::Message &Report(ErrorHandler &eh, const parser::CharBlock &ip,
+    parser::MessageFormattedText &&msg) {
+  return eh.Say(parser::Message{ip, msg});
+}
+
+inline bool HasNoErrors(const ErrorHandler &eh) { return !eh.AnyFatalError(); }
```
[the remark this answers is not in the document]
schweitzpgi: deleted

lib/semantics/resolve-labels.cc
```diff
@@ -0,0 +1,969 @@
...
+
+/// \brief Is this a legal DO terminator?
+/// Pattern match dependent on the standard we're enforcing
+/// F18:R1131 (must be CONTINUE or END DO)
+template<typename A>
+constexpr bool IsLegalDoTerm(const parser::Statement<A> &) {
+  return std::disjunction_v<
```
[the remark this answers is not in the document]
schweitzpgi: syntax expanded

lib/semantics/resolve-labels.cc
```diff
@@ -0,0 +1,969 @@
...
+/// Pattern match for FORMAT statement
+template<typename A> constexpr bool IsFormat(const parser::Statement<A> &) {
+  return std::is_same_v<A, common::Indirection<parser::FormatStmt>>;
+}
+
+/// \brief Is this a legal branch target?
+/// Pattern match dependent on the standard we're enforcing
+template<typename A>
+constexpr bool IsLegalBranchTarget(const parser::Statement<A> &) {
+  return std::disjunction_v<std::is_same<A, parser::AssociateStmt>,
```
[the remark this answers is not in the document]
schweitzpgi: syntax expanded

lib/semantics/resolve-labels.cc
```diff
@@ -0,0 +1,969 @@
...
+            actionStmt.statement.u) ||
+        std::holds_alternative<common::Indirection<parser::AssignedGotoStmt>>(
+            actionStmt.statement.u) ||
+        std::holds_alternative<common::Indirection<parser::PauseStmt>>(
+            actionStmt.statement.u));
+  }
+}
+
+template<typename A>
+constexpr TargetStmtType ConsTrgtFlags(const parser::Statement<A> &statement) {
```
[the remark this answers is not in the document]
schweitzpgi: changed to a long name

lib/semantics/resolve-labels.cc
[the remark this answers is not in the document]
schweitzpgi: renamed

lib/semantics/resolve-labels.cc
```diff
@@ -0,0 +1,969 @@
...
+    targetStmtType.set(TargetStatementEnum::Format);
+  }
+  return targetStmtType;
+}
+
+/// \brief \p opt1 and \p opt2 must be either present and identical or absent
+/// \param a  an optional construct-name (opening statement)
+/// \param b  an optional construct-name (ending statement)
+template<typename A> inline bool BothEqOrNone(const A &a, const A &b) {
+  return (a.has_value() == b.has_value())
```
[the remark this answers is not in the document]
schweitzpgi: expanded all conditional logic to if cascades

lib/semantics/resolve-labels.cc
```diff
@@ -0,0 +1,969 @@
...
+      ? (a.has_value() ? (a.value().ToString() == b.value().ToString()) : true)
+      : false;
+}
+
+/// \brief \p opt1 must either be absent or identical to \p b
+/// \param a  an optional construct-name for an optional constraint
+/// \param b  an optional construct-name (opening statement)
+template<typename A> inline bool PresentAndEq(const A &a, const A &b) {
+  return (!a.has_value()) ||
```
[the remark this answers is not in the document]
schweitzpgi: same

lib/semantics/resolve-labels.cc
```diff
@@ -0,0 +1,969 @@
...
+      (b.has_value() && (a.value().ToString() == b.value().ToString()));
+}
+
+/// \brief Iterates over parse tree, creates the analysis result
+/// As a side-effect checks the constraints for the usages of
+/// <i>construct-name</i>.
+struct ParseTreeAnalyzer {
+public:
```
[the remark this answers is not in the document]
schweitzpgi: ok

lib/semantics/resolve-labels.cc
```diff
@@ -0,0 +1,969 @@
...
+  struct UnitAnalysis {
```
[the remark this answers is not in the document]
schweitzpgi: de-nested

lib/semantics/resolve-labels.cc
```diff
@@ -0,0 +1,969 @@
...
+  public:
```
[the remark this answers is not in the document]
schweitzpgi: deleted

lib/semantics/resolve-labels.cc
schweitzpgi: _ added and _ deleted

lib/semantics/resolve-labels.cc
schweitzpgi: expanded to user-defined move constructor

lib/semantics/resolve-labels.cc
schweitzpgi: delete the other ctors and dtors

lib/semantics/resolve-labels.cc
```diff
@@ -637,21 +656,21 @@ class ParseTreeAnalyzer {
   void addLabelReferenceFromDoStmt(parser::Label label) {
     CheckLabelInRange(label);
     programUnits_.back().doStmtSources.emplace_back(
-        std::tuple<parser::Label, ScopeProxy, parser::CharBlock>{
+        std::tuple<parser::Label, ProxyForScope, parser::CharBlock>{
             label, currentScope_, currentPosition_});
```
tskeith: Because you're using emplace_back, I don't think you need the type and braces. Just:
`emplace_back(label, currentScope_, currentPosition_)`

tskeith:
The tests still don't work for me. I changed label02.f90 so it should fail, but ctest still reports that all the tests pass.

lib/semantics/resolve-labels.cc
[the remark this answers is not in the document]
sscalpone: https://llvm.org/docs/CodingStandards.html#anonymous-namespaces

lib/semantics/resolve-labels.cc
```diff
@@ -29,22 +29,32 @@ using LabeledStmtClassificationSet =
...
 
 const bool isStrictF18{false};  // FIXME - make a command-line option
 
-bool HasScope(ScopeProxy scope) { return scope != ScopeProxy{0u}; }
+bool HasScope(ProxyForScope scope) {
+  if (scope != ProxyForScope{0u}) {
+    return true;
+  } else {
+    return false;
+  }
```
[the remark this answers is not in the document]
schweitzpgi: I don't disagree.  The comment was made that cascading if-trees were easier to read.

tskeith:
The tests are working for me now, thanks.

lib/semantics/resolve-labels.cc
```diff
@@ -0,0 +1,918 @@
...
+
+ENUM_CLASS(TargetStatementEnum, Do, Branch, Format)
+using LabeledStmtClassificationSet =
+    common::EnumSet<TargetStatementEnum, TargetStatementEnum_enumSize>;
+
+using IndexList = std::vector<std::pair<parser::CharBlock, parser::CharBlock>>;
+// A ProxyForScope is an integral proxy for a Fortran scope. This is required
+// because the parse tree does not actually have the scopes required.
+using ProxyForScope = unsigned;
+struct LabeledStatementInfoTuplePOD {
```
tskeith: I think `LabeledStatementInfo` would be a better name. Same with the next struct.

lib/semantics/resolve-labels.cc
```diff
@@ -0,0 +1,918 @@
...
+  LabeledStatementInfoTuplePOD(const ProxyForScope &_proxyForScope,
```
tskeith: No need for the underscore prefix.

lib/semantics/resolve-labels.cc
```diff
@@ -0,0 +1,918 @@
...
+              "label '%" PRIu64 "' is out of range"_err_en_US, label});
+    }
+  }
+
+  // 6.2.5., paragraph 2
+  void addTargetLabelDefinition(parser::Label label,
+      LabeledStmtClassificationSet labeledStmtClassificationSet) {
+    CheckLabelInRange(label);
+    const auto pair{
+        programUnits_.back().targetStmts.emplace(std::make_pair(label,
```
tskeith: I don't think make_pair is needed with emplace.

lib/semantics/resolve-labels.cc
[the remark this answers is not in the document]
schweitzpgi: changed the syntax to PODs

lib/semantics/resolve-labels.cc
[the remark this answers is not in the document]
schweitzpgi: squished some back

lib/semantics/resolve-labels.cc
schweitzpgi: deleted

lib/semantics/resolve-labels.cc
schweitzpgi: I was at a conference about 30 years ago (OOPSLA?) when Alex Stepanov gave a keynote. In that keynote, he (paraphrasing) said, he regretted having been coerced into adding std::stack to the STL. In summary, "no one should ever use std::stack".

lib/semantics/resolve-labels.cc
schweitzpgi: deleted

lib/semantics/resolve-labels.cc
schweitzpgi: syntax changes made

lib/semantics/resolve-labels.cc
schweitzpgi: fixed

lib/semantics/resolve-labels.cc
schweitzpgi: deleted

lib/semantics/resolve-labels.cc
schweitzpgi: fixed

lib/semantics/resolve-labels.cc
schweitzpgi: uint64_t

lib/semantics/resolve-labels.cc
schweitzpgi: done

lib/semantics/resolve-labels.cc
schweitzpgi: syntax changes made

lib/semantics/resolve-labels.cc
schweitzpgi: deleted

lib/semantics/resolve-labels.cc
schweitzpgi: deleted

lib/semantics/resolve-labels.cc
schweitzpgi: done

lib/semantics/resolve-labels.cc
schweitzpgi: part of the TuplePODing changes

lib/semantics/resolve-labels.cc
schweitzpgi: Yeah, but my compiler was wringing its hands about these simple PODs. The extra garbage lets me find them a little easier. Deferring.

lib/semantics/resolve-labels.cc
schweitzpgi: The miracles of modern C++

lib/semantics/resolve-labels.cc
schweitzpgi: ? ... ok, fixed

lib/semantics/resolve-labels.cc
schweitzpgi: reworked

lib/semantics/resolve-labels.cc
```diff
@@ -0,0 +1,969 @@
...
+
+using IndexList = std::vector<std::pair<parser::CharBlock, parser::CharBlock>>;
+using ScopeProxy = unsigned;
+using LabelStmtInfo = std::tuple<ScopeProxy, parser::CharBlock, TargetStmtType>;
+using TargetStmtMap = std::map<parser::Label, LabelStmtInfo>;
+using SourceStmtList =
+    std::vector<std::tuple<parser::Label, ScopeProxy, parser::CharBlock>>;
+using ErrorHandler = parser::Messages;
+
+const bool isStrictF18{false};  // FIXME - make a command-line option
```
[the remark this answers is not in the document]
schweitzpgi: deleted

tools/f18/f18.cc
schweitzpgi: done

lib/semantics/resolve-labels.cc
klausler: The arguments to `emplace` (and related STL member functions) are forwarded to the constructor of the new element.  If one uses `emplace` with a already-constructed temporary object, the new element's move constructor will be used, assuming that it has one.  The whole point of `emplace` is that it omits the use of an intermediate temporary object.  The runtime savings is probably not going to affect the compiler's performance significantly, but `emplace` should be used anyway when constructing a new object in a container that provides the interface.

schweitzpgi: right. The make_pair was a transient edit that was finally removed but mistakenly pushed in the midst of revisions.

lib/semantics/resolve-labels.cc
```diff
@@ -0,0 +1,917 @@
...
+              constructName.c_str()});
+    }
+  }
+
+  // 6.2.5, paragraph 2
+  void CheckLabelInRange(parser::Label label) {
+    if (label < 1 || label > 99999) {
+      errorHandler_.Say(currentPosition_,
+          parser::MessageFormattedText{
+              "label '%" PRIu64 "' is out of range"_err_en_US, label});
```
klausler: Rather than using type-specific formats that need to be found and updated when/if the `parser::Label` type alias ever changes, consider using just `%d` for the format and `static_cast<int>(label)` for the argument.  That would be future-proof and maybe even more readable.

lib/semantics/resolve-labels.cc
```diff
@@ -0,0 +1,917 @@
...
+
+  void addLabelReference(const std::list<parser::Label> &labels) {
+    for (const parser::Label &label : labels) {
+      addLabelReference(label);
+    }
+  }
+
+  std::vector<UnitAnalysis> programUnits_;
+  parser::Messages errorHandler_;
+  parser::CharBlock currentPosition_{nullptr};
```
klausler: `CharBlock` has default initialization to `{nullptr,0}`.

lib/semantics/resolve-labels.cc
```diff
@@ -0,0 +1,917 @@
...
+
+ParseTreeAnalyzer LabelAnalysis(const parser::Program &program) {
+  ParseTreeAnalyzer analysis;
+  Walk(program, analysis);
+  return analysis;
+}
+
+bool InBody(const parser::CharBlock &position,
+    const std::pair<parser::CharBlock, parser::CharBlock> &pair) {
+  if (position.begin() >= pair.first.begin()) {
```
klausler: This code might be more clear if `CharBlock` and/or `Interval` were extended so that the endpoints of the pair could be combined into a single interval and then used for testing containment.

lib/semantics/resolve-labels.cc
```diff
@@ -0,0 +1,917 @@
...
+    const auto &label{branch.parserLabel};
+    auto branchTarget{GetLabel(labels, label)};
+    if (HasScope(branchTarget.proxyForScope)) {
+      const auto &fromPosition{branch.parserCharBlock};
+      const auto &toPosition{branchTarget.parserCharBlock};
+      for (const auto body : loopBodies) {
+        if (!InBody(fromPosition, body) && InBody(toPosition, body)) {
+          errorHandler.Say(fromPosition,
+              parser::MessageFormattedText{
+                  "branch into '%s' from another scope"_en_US,
```
klausler: `scope` might be better if replaced with `outside the body of the loop` in this message.

lib/semantics/resolve-labels.cc
```diff
@@ -0,0 +1,917 @@
...
+            parser::MessageFormattedText{"'%s' doesn't properly nest"_err_en_US,
+                v1.first.ToString().c_str()});
+      }
+    }
+  }
+}
+
+parser::CharBlock SkipLabel(const parser::CharBlock &position) {
+  const long maxPosition{position.end() - position.begin()};
+  if (maxPosition && (position[0] >= '0') && (position[0] <= '9')) {
```
klausler: Consider using `parser::CharBlock::empty()` and `parser::CharBlock::size()`.
Consider using `parser::IsDecimalDigit()`.

lib/semantics/resolve-labels.cc
```diff
@@ -0,0 +1,917 @@
...
+    const auto &label{stmt.parserLabel};
+    auto branchTarget{GetLabel(labels, label)};
+    if (HasScope(branchTarget.proxyForScope)) {
+      if (!branchTarget.labeledStmtClassificationSet.test(
+              TargetStatementEnum::Branch) &&
+          !branchTarget.labeledStmtClassificationSet.test(
+              TargetStatementEnum::CompatibleBranch)) {
+        errorHandler.Say(branchTarget.parserCharBlock,
+            parser::MessageFormattedText{
+                "'%" PRIu64 "' not a branch target"_err_en_US, label});
```
klausler: Can this error message be improved, maybe to something like "statement at label %d cannot be used as the target of a branch"?

PR #173, opened 2018-09-04
More name resolution for derived types

tskeith:
Resolve procedure bindings, final procedures, and type parameters in derived types.

### DIFF
--- a/lib/common/fortran.h
+++ b/lib/common/fortran.h
@@ -28,5 +28,8 @@ ENUM_CLASS(TypeCategory, Integer, Real, Complex, Character, Logical, Derived)
 // Kinds of IMPORT statements. Default means IMPORT or IMPORT :: names.
 ENUM_CLASS(ImportKind, Default, Only, None, All)
 
+// Type parameters can be kind or len.
+ENUM_CLASS(TypeParamKindOrLen, Kind, Len)
+
 }  // namespace Fortran::common
 #endif  // FORTRAN_COMMON_FORTRAN_H_
--- a/lib/common/fortran.h
+++ b/lib/common/fortran.h
@@ -28,8 +28,8 @@ ENUM_CLASS(TypeCategory, Integer, Real, Complex, Character, Logical, Derived)
 // Kinds of IMPORT statements. Default means IMPORT or IMPORT :: names.
 ENUM_CLASS(ImportKind, Default, Only, None, All)
 
-// Type parameters can be kind or len.
-ENUM_CLASS(TypeParamKindOrLen, Kind, Len)
+// The attribute on a type parameter can be KIND or LEN.
+ENUM_CLASS(TypeParamAttr, Kind, Len)
 
 }  // namespace Fortran::common
 #endif  // FORTRAN_COMMON_FORTRAN_H_
--- a/lib/parser/parse-tree.h
+++ b/lib/parser/parse-tree.h
@@ -890,8 +890,7 @@ struct TypeParamDecl {
 // R734 type-param-attr-spec -> KIND | LEN
 struct TypeParamDefStmt {
   TUPLE_CLASS_BOILERPLATE(TypeParamDefStmt);
-  std::tuple<IntegerTypeSpec, common::TypeParamKindOrLen,
-      std::list<TypeParamDecl>>
+  std::tuple<IntegerTypeSpec, common::TypeParamAttr, std::list<TypeParamDecl>>
       t;
 };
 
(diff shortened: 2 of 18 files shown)


lib/parser/parse-tree.h
```diff
@@ -889,9 +889,10 @@ struct TypeParamDecl {
 //        integer-type-spec , type-param-attr-spec :: type-param-decl-list
 // R734 type-param-attr-spec -> KIND | LEN
 struct TypeParamDefStmt {
-  ENUM_CLASS(KindOrLen, Kind, Len)  // R734
   TUPLE_CLASS_BOILERPLATE(TypeParamDefStmt);
-  std::tuple<IntegerTypeSpec, KindOrLen, std::list<TypeParamDecl>> t;
+  std::tuple<IntegerTypeSpec, common::TypeParamKindOrLen,
```
sscalpone: Can Kind and Length ever be used interchangeably?  If not, shouldn't this be a variant of either a Kind or a Length instead of using the enum to discriminate between the two?

lib/semantics/resolve-names.cc
```diff
@@ -79,7 +79,7 @@ class ImplicitRules {
 // Provide Post methods to collect attributes into a member variable.
 class AttrsVisitor {
 public:
-  void BeginAttrs();
+  bool BeginAttrs();
```
sscalpone: Need a comment explaining when it returns true or false?

lib/common/fortran.h
```diff
@@ -28,5 +28,8 @@ ENUM_CLASS(TypeCategory, Integer, Real, Complex, Character, Logical, Derived)
 // Kinds of IMPORT statements. Default means IMPORT or IMPORT :: names.
 ENUM_CLASS(ImportKind, Default, Only, None, All)
 
+// Type parameters can be kind or len.
+ENUM_CLASS(TypeParamKindOrLen, Kind, Len)
```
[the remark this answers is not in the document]
tskeith: The standard uses "type-param" for type parameters. For example, KIND or LEN appears in a "type-param-def-stmt". In the context of a parameter statement or attribute they do not abbreviate it. We have been following the same convention.

lib/parser/parse-tree.h
tskeith: All the enum represents here is this reduction: `type-param-attr-spec -> KIND | LEN`
There is no data associated with it, so I don't think there is any value in using a variant.

lib/semantics/resolve-names.cc
tskeith: OK, I'll add one.

lib/semantics/resolve-names.cc
```diff
@@ -2138,17 +2214,76 @@ void DeclarationVisitor::SetType(
   symbol.SetType(type);
 }
 
-// ResolveNamesVisitor implementation
+// Find the Symbol for this derived type.
+const Symbol *DeclarationVisitor::ResolveDerivedType(const SourceName &name) {
```
schweitzpgi: might be a function that could return a `std::optional<T>`

lib/common/fortran.h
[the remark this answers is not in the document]
tskeith: But it's not a "type-param" so I don't think "TypeParam" is a good name.

Maybe "TypeParamAttr". It is "type-param-attr-spec" in the standard, but for other "attr-spec"s we use "Attr".

lib/semantics/mod-file.cc
```diff
@@ -165,23 +167,61 @@ ModFileWriter::symbolSet ModFileWriter::CollectSymbols(const Scope &scope) {
   return symbols;
 }
 
-void ModFileWriter::PutSymbol(const Symbol &symbol) {
+void ModFileWriter::PutSymbol(const Symbol &symbol, bool &didContains) {
```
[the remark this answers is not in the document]
tskeith: It's by-reference because the value is going in and out.

lib/semantics/mod-file.cc
```diff
@@ -165,23 +167,61 @@ ModFileWriter::symbolSet ModFileWriter::CollectSymbols(const Scope &scope) {
...
   std::visit(
       common::visitors{
           [&](const ModuleDetails &) { /* should be current module */ },
```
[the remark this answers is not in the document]
tskeith: I'll add an assertion if the information is available.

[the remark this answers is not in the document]
tskeith: Some of the lambdas don't need to capture anything so they have `[]` instead of `[&]`.

lib/semantics/resolve-names.cc
tskeith: It can't be `std::optional<Symbol>` because I don't want to copy Symbols, I need references to them. `std::optional<Symbol *>` isn't really an improvement.

PR #175, opened 2018-09-06
More name resolution

tskeith:
Procedure pointer components and PRIVATE and SEQUENCE statements in derived types.


lib/semantics/resolve-names.cc
```diff
@@ -36,6 +36,7 @@ using namespace parser::literals;
 class MessageHandler;
 
 static GenericSpec MapGenericSpec(const parser::GenericSpec &);
+static bool IsModule(const Scope &);
```
klausler: Would this be better as a member function `Scope::IsModule() const;"?

lib/semantics/resolve-names.cc
```diff
@@ -2100,24 +2122,36 @@ void DeclarationVisitor::Post(const parser::TypeParamDefStmt &x) {
...
-  // const auto *symbol{ResolveDerivedType(name)};
-  derivedTypeData_->extends = &name;
+  derivedTypeInfo_.extends = &x.v.source;
   return false;
 }
+
 bool DeclarationVisitor::Pre(const parser::PrivateStmt &x) {
-  derivedTypeData_->Private = true;
+  if (!IsModule(currScope().parent())) {
+    Say("PRIVATE is only allowed in a derived type that is"
```
klausler: This message may not need to be fatal if compilation can proceed.

lib/semantics/resolve-names.cc
```diff
@@ -2100,24 +2122,36 @@ void DeclarationVisitor::Post(const parser::TypeParamDefStmt &x) {
...
+        " in a module"_err_en_US);  // C766
+  } else if (derivedTypeInfo_.sawContains) {
+    derivedTypeInfo_.privateBindings = true;
+  } else if (!derivedTypeInfo_.privateComps) {
+    derivedTypeInfo_.privateComps = true;
+  } else {
+    Say("PRIVATE may not appear more than once in"
```
klausler: This message probably need not be fatal.

PR #177, opened 2018-09-10
Write private symbols to .mod files

tskeith:
... and some name resolution improvements.

### DIFF
--- a/test/semantics/resolve01.f90
+++ b/test/semantics/resolve01.f90
@@ -15,4 +15,9 @@
 integer :: x
 !ERROR: The type of 'x' has already been declared
 real :: x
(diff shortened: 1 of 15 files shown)


test/semantics/resolve01.f90
```diff
@@ -15,4 +15,9 @@
 integer :: x
 !ERROR: The type of 'x' has already been declared
 real :: x
+integer(8) :: i
+parameter(i=1,j=2,k=3)
+integer :: j
+!ERROR: The type of 'k' has already been implicitly declared
```
sscalpone: Does the symtab entry for k include the line number where k was first declared?

PR #182, opened 2018-09-12
Simplify representation of intrinsic types

tskeith:
Intrinsic types are now just a TypeCategory and a int kind. If no kind
is specified the default is used so that every type has an explicit
kind. This caused changes in the expected results of some of the tests.

Add support for "double precision" and "double complex".

Intrinsic types are now stored as values in DeclTypeSpec so none of the
KindedTypeHelper machinery is needed any more.

Eliminate DerivedTypeDef, DataComponentDef, ProcComponentDef,
TypeBoundProc. The components and bindings of a derived type are now
represented by the corresponding Scope.

### DIFF
--- a/lib/semantics/resolve-names.cc
+++ b/lib/semantics/resolve-names.cc
@@ -176,10 +176,9 @@ private:
   DerivedTypeSpec *derivedTypeSpec_{nullptr};
   std::unique_ptr<ParamValue> typeParamValue_;
 
-  void MakeIntrinsic(const IntrinsicTypeSpec &intrinsicTypeSpec);
+  void MakeIntrinsic(TypeCategory, int);
   void SetDeclTypeSpec(const DeclTypeSpec &declTypeSpec);
-  static KindParamValue GetKindParamValue(
-      const std::optional<parser::KindSelector> &kind);
+  static int GetKindParamValue(const std::optional<parser::KindSelector> &kind);
 };
 
 // Track statement source locations and save messages.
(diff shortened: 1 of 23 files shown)


test/semantics/modfile01.f90
```diff
@@ -27,11 +27,11 @@ module m
 
 !Expect: m.mod
 !module m
-!integer::i
-!integer,private::j
+!integer(4)::i
```
sscalpone: Any reason just to duplicate the test (to make sure default are set correctly with -i8/-r8?  But there will be other tests for that.

tskeith: There is no -i8/-r8 option yet.

lib/semantics/resolve-names.cc
```diff
@@ -176,10 +176,9 @@ class DeclTypeSpecVisitor : public AttrsVisitor {
   DerivedTypeSpec *derivedTypeSpec_{nullptr};
   std::unique_ptr<ParamValue> typeParamValue_;
 
-  void MakeIntrinsic(const IntrinsicTypeSpec &intrinsicTypeSpec);
+  void MakeIntrinsic(TypeCategory, int);
```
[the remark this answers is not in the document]
tskeith: Yes, I thought about that. My goal with this change is to make symbol table types more compatible with types in Fortran::evaluate and there the kind is just an int. We should consider a distinct type used both places.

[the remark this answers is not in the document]
klausler: Negative kind type parameter values are allowed for parameterized derived types.  The kind values that we have chosen for the intrinsic types are all greater than zero.

lib/semantics/type.cc
```diff
@@ -197,57 +79,43 @@ std::ostream &operator<<(std::ostream &o, const ShapeSpec &x) {
...
+IntrinsicTypeSpec::IntrinsicTypeSpec(TypeCategory category, int kind)
+  : category_{category}, kind_{kind ? kind : GetDefaultKind(category)} {
+  CHECK(category != TypeCategory::Derived);
+}
+
+int IntrinsicTypeSpec::GetDefaultKind(TypeCategory category) {
+  switch (category) {
+  case TypeCategory::Character: return evaluate::DefaultCharacter::kind;
+  //case TypeCategory::Complex: return evaluate::DefaultComplex::kind;
+  case TypeCategory::Complex: return 4;  // TEMP to work around bug
```
klausler: The default kind of COMPLEX will always be the same as the default kind of REAL.

PR #183, opened 2018-09-12
Semantic analysis of expressions (work in progress)

klausler:
Here's a pile of new code that makes progress on the semantic analysis of expressions and the representation of typed expressions produced by successful analysis.


lib/common/kind-variant.h
```diff
@@ -0,0 +1,121 @@
+// Copyright (c) 2018, NVIDIA CORPORATION.  All rights reserved.
```
klausler: Oops!  I wrote this code and then ended up not using it anywhere.  Please ignore it until it disappears from this review.

lib/common/template.h
```diff
@@ -0,0 +1,288 @@
...
+template<template<typename...> class T, typename... Ts>
+struct OverMembersHelper<T, std::variant<Ts...>> {
+  using type = T<Ts...>;
+};
+template<template<typename...> class T, typename... Ts>
+struct OverMembersHelper<T, std::tuple<Ts...>> {
+  using type = T<Ts...>;
+};
+
+template<template<typename...> class T, typename TorV>
```
tskeith: It took me a while to realize that "TorV" meant "tuple or variant". Maybe that should be mentioned in a comment.

lib/common/template.h
```diff
@@ -0,0 +1,288 @@
...
+  }
+  return std::nullopt;
+}
+
+// const std::variant<..., A, ...> -> std::optional<A>
+// i.e., when a variant holds a value of a particular type, return a copy
+// of that value in a std::optional<>.  The type A must be a valid
+// alternative for the variant.
+template<typename A, typename VARIANT>
+std::optional<A> GetIf(const VARIANT &u) {
```
tskeith: What's the benefit of this over just using std::get_if? They seem almost interchangeable, but we have a lot of uses std::get_if already.

lib/common/template.h
klausler: I'll replace `TorV` with a longer name.

lib/common/template.h
klausler: I'll examine the uses of GetIf that survive and see whether they can be rewritten in terms of get_if.

lib/evaluate/type.h
```diff
@@ -35,168 +38,245 @@ namespace Fortran::evaluate {
...
+
+  template<typename A> SomeScalar(const A &x) : u{x} {}
   template<typename A>
-  GenericScalar(std::enable_if_t<!std::is_reference_v<A>, A> &&x)
+  SomeScalar(std::enable_if_t<!std::is_reference_v<A>, A> &&x)
     : u{std::move(x)} {}
 
-  std::optional<std::int64_t> ToInt64() const;
-  std::optional<std::string> ToString() const;
+  auto ToInt64() const {
```
tskeith: I think these functions would be more readable with explicit return types.

lib/evaluate/type.h
```diff
@@ -35,168 +38,245 @@ namespace Fortran::evaluate {
...
+          if constexpr (TypeOf<decltype(x)>::category ==
+              TypeCategory::Logical) {
+            return {x.IsTrue()};
+          }
+          return std::nullopt;
+        },
+        u);
+  }
+
+  template<typename T> auto GetIf() const {
```
tskeith: Especially this one, where the return type isn't even in the function body.

lib/evaluate/type.h
klausler: OK.

lib/evaluate/type.h
klausler: Deleting this function anyway after earlier review comment.

klausler:
I've pushed an update that makes this branch current with the master branch and incorporates Tim's recent changes, so there's no longer any `dynamic_cast` usage.

PR #184, opened 2018-09-13
Add semantics checking for END xxx statements.

schweitzpgi:
Add a negative test for END xxx names.
Reverse the sequence of calls such that resolve-label is before
resolve-name.


lib/semantics/resolve-labels.cc
```diff
@@ -190,6 +190,15 @@ bool PresentAndEq(const std::optional<parser::Name> &name_a,
   }
 }
 
+bool PresentAndEq(
+    const std::optional<parser::Name> &name_a, const parser::Name &name_b) {
+  if (!name_a.has_value()) {
+    return true;
```
klausler: True?

lib/semantics/resolve-labels.cc
```diff
@@ -311,6 +320,156 @@ class ParseTreeAnalyzer {
...
+      if (endProgramStmt.v.has_value()) {
+        if (programStmt.v.ToString() != endProgramStmt.v->ToString()) {
+          errorHandler_.Say(currentPosition_,
+              parser::MessageFormattedText{
+                  "END PROGRAM name mismatch"_err_en_US});
+        }
+      } else {
+        errorHandler_.Say(currentPosition_,
+            parser::MessageFormattedText{
+                "END PROGRAM must have a program-name"_err_en_US});
```
klausler: The name is optional on END PROGRAM.

schweitzpgi: I'll fix this. The constraint reads "The program-name shall not be included in the end-program-stmt unless the optional program-stmt is used. _If included_, it shall be identical to the program-name specified in the program-stmt." and I confused myself about what optional object in the first sentence was implicitly referred to by "if included." Thanks.

lib/semantics/resolve-labels.cc
```diff
@@ -311,6 +320,156 @@ class ParseTreeAnalyzer {
...
 
+  // C1414
+  void Post(const parser::BlockData &blockData) {
+    if (!BothEqOrNone(
+            std::get<parser::Statement<parser::BlockDataStmt>>(blockData.t)
+                .statement.v,
+            std::get<parser::Statement<parser::EndBlockDataStmt>>(blockData.t)
+                .statement.v)) {
+      errorHandler_.Say(currentPosition_,
+          parser::MessageFormattedText{"BLOCK DATA name mismatch"_err_en_US});
```
tskeith: In these messages it would be friendly to include a pointer to the first name. It makes it easier for the user to spot the mismatch.

lib/semantics/resolve-labels.cc
```diff
@@ -190,6 +190,15 @@ bool PresentAndEq(const std::optional<parser::Name> &name_a,
   }
 }
 
+bool PresentAndEq(
+    const std::optional<parser::Name> &name_a, const parser::Name &name_b) {
```
tskeith: No underscores in argument names: nameA and nameB.

PR #185, opened 2018-09-14
Improved error messages:

schweitzpgi:
Update the tests to reflect new message
Improved error messages to show both source and destination of mismatches.

Use CharBlock::size.

### DIFF
--- a/lib/semantics/resolve-labels.cc
+++ b/lib/semantics/resolve-labels.cc
@@ -198,7 +198,9 @@ bool firstNameNoneOrBothEqual(
   }
 }
 
-unsigned sayLabel(parser::Label label) { return static_cast<unsigned>(label); }
+static unsigned SayLabel(parser::Label label) {
+  return static_cast<unsigned>(label);
+}
 
 struct UnitAnalysis {
   UnitAnalysis() { scopeModel.push_back(0); }
@@ -845,7 +847,7 @@ private:
     if (label < 1 || label > 99999) {
       errorHandler_.Say(currentPosition_,
           parser::MessageFormattedText{
-              "label '%u' is out of range"_err_en_US, sayLabel(label)});
+              "label '%u' is out of range"_err_en_US, SayLabel(label)});
     }
   }
 
@@ -859,7 +861,7 @@ private:
     if (!pair.second) {
       errorHandler_.Say(currentPosition_,
           parser::MessageFormattedText{
-              "label '%u' is not distinct"_err_en_US, sayLabel(label)});
+              "label '%u' is not distinct"_err_en_US, SayLabel(label)});
     }
   }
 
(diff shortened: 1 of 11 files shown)


lib/semantics/resolve-labels.cc
```diff
@@ -199,6 +199,8 @@ bool firstNameNoneOrBothEqual(
   }
 }
 
+unsigned sayLabel(parser::Label label) { return static_cast<unsigned>(label); }
```
klausler: Prefer `int` to `unsigned` apart from low-level bitwise code.

Name functions with capital letters.

This function should be `static`.

sscalpone: https://google.github.io/styleguide/cppguide.html#Integer_Types

lib/semantics/resolve-labels.cc
```diff
@@ -899,19 +994,23 @@ void CheckDoNesting(
...
       if (v2.first.begin() < v1.second.end() &&
           v1.second.begin() < v2.second.begin()) {
-        errorHandler.Say(v2.second,
-            parser::MessageFormattedText{"'%s' doesn't properly nest"_err_en_US,
-                v1.first.ToString().c_str()});
+        errorHandler
+            .Say(v1.first,
+                parser::MessageFormattedText{
+                    "DO loop doesn't properly nest"_err_en_US})
+            .Attach(v1.second, "which extends to here"_en_US)
```
sscalpone: Messages need to be collected into a single string or be very independent because when messages are localized the arguments might need to be put in different positions.  Using "and" and "which" are problematic in this case.  Another solution is to create three independent messages if it makes sense in context.

tskeith: I think these should be independent messages. It may also be too much information. I suggest just reporting the error and then pointing out the loop it conflicts with. Something like this:
````
test/semantics/label03.f90:28:3: error: DO loops are not properly nested
    do 10 i = 1, m
    ^^^^^^^^^^^^^^
test/semantics/label03.f90:29:6: DO 10 loop conflicts with this loop
       do 20 j = 1, n
       ^^^^^^^^^^^^^^
````

Comment (tskeith) [reply lib/semantics/resolve-labels.cc]: For semantic errors, I have been trying to use complete sentences, starting with a capital letter, no period. It would be good to have guidelines before we get too far. Also, when referring to names in the message I've been enclosing them in single quotes, mostly because that's easier than double quotes.

File: lib/semantics/resolve-labels.cc (hunk 1 above)
Comment (schweitzpgi) [reply lib/semantics/resolve-labels.cc]: Labels are already unsigned. This function is simply to force a type agreement between the C printf() format specifier and the argument.

File: lib/semantics/resolve-labels.cc (hunk 2 above)
Comment (schweitzpgi) [reply lib/semantics/resolve-labels.cc]: Our rule is/ought to be 1 error -> 1 error message, which is the point of Attach() as far as I understand it. Also, feedback from the original pull-request was to use the CharBlock information to emit source text and use Attach() rather than build up error messages in an ad hoc way.  This change corrects that. It was further suggested that some mechanisms for controlling the verbosity of error messages would be engineered at some point in the future, so we should go ahead and emit a message by default (at least for now). Finally, I agree that it makes sense to design the what/how/when/etc. of error messages and their reporting and not spend copious time polishing these placeholders.

Comment (klausler) [reply lib/semantics/resolve-labels.cc]: I think that a message attachment with a location is a helpful thing to have here.  The situation is analogous to errors about conflicting declarations.  The sorting of messages by location doesn't apply to attachments when the messages are presented to the user; i.e., attachments stay by their containing messages.

**Comment by schweitzpgi** [comment]
Merged to head and resolved conflicts. Reopening.

File: lib/semantics/resolve-labels.cc (hunk 1 above)
Comment (schweitzpgi) [reply lib/semantics/resolve-labels.cc]: Yes, will do.

File: lib/semantics/resolve-labels.cc [hunk 3]
```diff
@@ -359,10 +370,12 @@ class ParseTreeAnalyzer {
             if (const auto *otherPointer{
                     std::get_if<parser::Name>(&optionalGenericSpec->u)}) {
               if (namePointer->ToString() != otherPointer->ToString()) {
-                errorHandler_.Say(currentPosition_,
-                    parser::MessageFormattedText{
-                        "INTERFACE generic-name (%s) mismatch"_en_US,
-                        namePointer->ToString().c_str()});
+                errorHandler_
+                    .Say(currentPosition_,
+                        parser::MessageFormattedText{
```
tskeith: I don't think the explicit MessageFormattedText constructor is necessary. Message has a constructor that can take the position, fixed text, and arguments.

PR #186, opened 2018-09-14
Add top-level Semantics class

tskeith:
Refactor to create the Semantics class that is responsible for holding
state during semantics (the scope tree and messages) and managing the
logic of the various phases of semantic processing. Eliminate static
Scope::globalScope.

The messages generated during semantic processing are accumulated in a
Messages data member of Semantics so that individual phases don't need
to emit them to std::cerr. This is now done by the driver so that it has
control over where they go and eliminates other includes of iostream.
To do this, the messages object is passed in to the various semantics
operations.

Move DumpSymbols into semantics.cc: it doesn't belong in resolve-names.cc
and it depends on the global scope, so it's as good a place as any.
The call to RewriteParseTree is also moved to Semantics.

### DIFF
--- a/lib/semantics/resolve-labels.cc
+++ b/lib/semantics/resolve-labels.cc
@@ -216,6 +216,7 @@ struct UnitAnalysis {
 
 class ParseTreeAnalyzer {
 public:
+  ParseTreeAnalyzer(ParseTreeAnalyzer &&that) = default;
   ParseTreeAnalyzer(parser::Messages &errorHandler)
     : errorHandler_{errorHandler} {}
 
--- a/lib/semantics/resolve-labels.cc
+++ b/lib/semantics/resolve-labels.cc
@@ -18,7 +18,6 @@
 #include "../parser/parse-tree-visitor.h"
 #include <cctype>
 #include <cstdarg>
-#include <iostream>
 
 namespace Fortran::semantics {
 
@@ -217,12 +216,8 @@ struct UnitAnalysis {
 
 class ParseTreeAnalyzer {
 public:
-  ParseTreeAnalyzer() {}
-  ParseTreeAnalyzer(ParseTreeAnalyzer &&that)
-    : programUnits_{std::move(that.programUnits_)},
-      errorHandler_{std::move(that.errorHandler_)}, currentPosition_{std::move(
-                                                        that.currentPosition_)},
-      constructNames_{std::move(that.constructNames_)} {}
+  ParseTreeAnalyzer(parser::Messages &errorHandler)
+    : errorHandler_{errorHandler} {}
 
   template<typename A> constexpr bool Pre(const A &) { return true; }
   template<typename A> constexpr void Post(const A &) {}
@@ -827,7 +822,7 @@ private:
   }
 
   std::vector<UnitAnalysis> programUnits_;
-  parser::Messages errorHandler_;
+  parser::Messages &errorHandler_;
   parser::CharBlock currentPosition_{nullptr};
   ProxyForScope currentScope_{0};
   std::vector<std::string> constructNames_;
@@ -843,8 +838,9 @@ bool InInclusiveScope(const std::vector<ProxyForScope> &scopes,
   return true;
 }
 
-ParseTreeAnalyzer LabelAnalysis(const parser::Program &program) {
-  ParseTreeAnalyzer analysis;
+ParseTreeAnalyzer LabelAnalysis(
+    parser::Messages &errorHandler, const parser::Program &program) {
+  ParseTreeAnalyzer analysis{errorHandler};
   Walk(program, analysis);
   return analysis;
 }
@@ -1040,8 +1036,7 @@ void CheckDataTransferConstraints(const SourceStmtList &dataTransfers,
   CheckDataXferTargetConstraints(dataTransfers, labels, errorHandler);
 }
 
-bool CheckConstraints(ParseTreeAnalyzer &&parseTreeAnalysis,
-    const parser::CookedSource &cookedSource) {
+bool CheckConstraints(ParseTreeAnalyzer &&parseTreeAnalysis) {
   auto &errorHandler{parseTreeAnalysis.errorHandler()};
   for (const auto &programUnit : parseTreeAnalysis.programUnits()) {
     const auto &dos{programUnit.doStmtSources};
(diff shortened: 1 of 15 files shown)


lib/semantics/resolve-names.cc
```diff
@@ -218,14 +218,13 @@ class MessageHandler {
 
 private:
   // Where messages are emitted:
-  parser::Messages messages_;
+  parser::Messages *messages_;
```
klausler: Maybe this should become a reference that's established during construction.

lib/semantics/semantics.h
```diff
@@ -0,0 +1,47 @@
...
+  const parser::Messages &messages() const { return messages_; }
+  Semantics &set_searchDirectories(const std::vector<std::string> &);
+  Semantics &set_moduleDirectory(const std::string &);
+  bool AnyFatalError() const { return messages_.AnyFatalError(); }
+  bool Perform(parser::Program &);
+  void DumpSymbols(std::ostream &);
+
+private:
+  Scope globalScope_;
+  std::vector<std::string> directories_;
```
klausler: Consider putting the "." directory here in a default initializer rather than above in the default constructor.

tools/f18/f18.cc
```diff
@@ -211,20 +208,17 @@ std::string CompileFortran(
...
-    }
-    (void)Fortran::semantics::ValidateLabels(parseTree, parsing.cooked());
-    Fortran::semantics::ResolveNames(Fortran::semantics::Scope::globalScope,
-        parseTree, parsing.cooked(), directories);
-    Fortran::semantics::ModFileWriter writer;
-    writer.set_directory(driver.moduleDirectory);
-    writer.WriteAll();
-    writer.errors().Emit(std::cerr, parsing.cooked());
+    semantics.Perform(parseTree);
+    auto &messages{semantics.messages()};
```
klausler: Will warnings from semantics be sorted together with any parsing messages?

lib/semantics/resolve-names.cc
tskeith: I would have preferred that, but it requires passing it up through a series of superclass constructors and I decided it wasn't worth it. I'll take another look.

lib/semantics/semantics.h
tskeith: Good idea, thanks.

tools/f18/f18.cc
tskeith: Messages from parsing are emitted before doing any semantics processing. I didn't change that.

klausler: OK.  We'll have to fix that later.

lib/semantics/resolve-names.cc
schweitzpgi: In principle, it's just a fancy container. The container doesn't need to be threaded everywhere, if the contents in the container are ultimately forwarded into the "big container" owned by Semantics.

lib/semantics/resolve-names.cc
```diff
@@ -2940,37 +2948,4 @@ static GenericSpec MapGenericSpec(const parser::GenericSpec &genericSpec) {
...
-      }
-    }
-  }
-  for (const auto &child : scope.children()) {
-    DumpSymbols(os, child, indent);
-  }
-  --indent;
-}
-
-void DumpSymbols(std::ostream &os) { DumpSymbols(os, Scope::globalScope); }
```
schweitzpgi: I assume this is just debugging code?

lib/semantics/resolve-labels.cc
```diff
@@ -217,12 +216,8 @@ struct UnitAnalysis {
 
 class ParseTreeAnalyzer {
 public:
-  ParseTreeAnalyzer() {}
-  ParseTreeAnalyzer(ParseTreeAnalyzer &&that)
-    : programUnits_{std::move(that.programUnits_)},
-      errorHandler_{std::move(that.errorHandler_)}, currentPosition_{std::move(
-                                                        that.currentPosition_)},
-      constructNames_{std::move(that.constructNames_)} {}
+  ParseTreeAnalyzer(parser::Messages &errorHandler)
```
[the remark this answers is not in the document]
tskeith: I think it's the same as the implicitly-declared move constructor so it wouldn't change anything to take it out.
We could also leave it in with "= default".

lib/semantics/resolve-labels.cc
```diff
@@ -1053,15 +1048,12 @@ bool CheckConstraints(ParseTreeAnalyzer &&parseTreeAnalysis,
     const auto &dataTransfers{programUnit.formatStmtSources};
     CheckDataTransferConstraints(dataTransfers, labels, scopes, errorHandler);
   }
-  if (!errorHandler.empty()) {
```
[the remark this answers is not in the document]
tskeith: errorHandler is passed around to CheckLabelDoConstraints, CheckBranchConstraints, etc. So I thought is was more consistent to pass it into ParseTreeAnalyzer. Everyone has a reference to the same errorHandler and adds messages to it.

lib/semantics/resolve-labels.cc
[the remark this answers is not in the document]
klausler: +1 for `=default`.

lib/semantics/resolve-labels.h
```diff
@@ -16,18 +16,17 @@
...
 
 namespace Fortran::parser {
+class Messages;
 struct Program;
-class CookedSource;
 }  // namespace Fortran::parser
 
 namespace Fortran::semantics {
 
 /// \brief Validate the labels in the program
```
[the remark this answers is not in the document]
sscalpone: Keep the doygen :)

lib/semantics/resolve-names.cc
tskeith: DumpSymbols() is called when -fdebug-dump-symbols is specified.

lib/semantics/resolve-labels.h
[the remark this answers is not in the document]
schweitzpgi: 95% of it is already gone.

lib/semantics/semantics.cc
```diff
@@ -0,0 +1,99 @@
...
+  ValidateLabels(messages_, program);
+  if (AnyFatalError()) {
+    return false;
+  }
+  ResolveNames(messages_, globalScope_, program, directories_);
+  if (AnyFatalError()) {
+    return false;
+  }
+  RewriteParseTree(messages_, globalScope_, program);
+  if (AnyFatalError()) {
```
schweitzpgi: I believe AnyFatalError() is superlinear, no? Something like O(k*m_1 + (k-1)*m_2 + ... + 1*m_k) where k is the number of semantic passes and m_i are the number of error messages for pass i that are not fatal. ValidateLabels() originally built its own list of messages and threaded back a bool to avoid that.

lib/semantics/resolve-labels.cc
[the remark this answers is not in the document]
klausler: I don't recommend defining a move constructor *just* to disable a copy constructor, of course.  When going out of one's way to preclude copy construction (as I did in the parse tree data structures so that I would never have to worry about finding expensive accidental deep copies that should have been moves), there's `=delete`.  For data structures that constitute a more general-purpose library with unknown clients, it's best to support both copy and move semantics for construction and assignment.  I really prefer `=default` for these instead of explicit constructors so that we don't have to worry about missing the construction of a member when adding new ones to a class.  Another technique that is good is to support only move construction and assignment in the class, and supply a `Clone()` member function to create explicit copies for the rare cases when a copy is really needed (and to make them stand out).

lib/semantics/semantics.cc
klausler: Don't complicate designs to gain insignificant speedups.

PR #192, opened 2018-09-20
Fix problems with extended derived types

### LINKED ISSUE #187: Extended derived types can't access ancestor members
```
program main
  type ::t1
    integer :: t1mem
  end type t1
  type, extends(t1) :: t2
    integer :: t2mem
  end type t2
  type(t2) :: x2
  print *, x2%t1mem
end program main
```

elicits

```
../../test/component.f90:9:15: error: Component 't1mem' not found in derived type 't2'
    print *, x2%t1mem
                ^^^^^
../../test/component.f90:5:24: Declaration of 't2'
    type, extends(t1) :: t2
                         ^^
```

tskeith:
When looking for a component name in a derived type, also look in the
parent type. Before adding a component to a derived type, report an
error if it already has one with that name. Check that components are
accessible when they are accessed.

Add the "parent component" to derived types (i.e. a component with the
same name as the parent type). The symbol is marked with the
`ParentComp` flag so we can avoid writing it to `.mod` files.

Add calls to `add_occurrence()` so that those particular instances of
`parser::Name` get their symbol set.

Change `DeclareObjectEntity` and `DeclareProcEntity` to use `SourceName` as
the name passed in rather than `parser::Name`.

Fix some problems in `unparse-with-symbols.cc` on statements that both
define and reference names.

Fixes #187.

### DIFF
--- a/lib/semantics/symbol.h
+++ b/lib/semantics/symbol.h
@@ -272,7 +272,7 @@ public:
   ENUM_CLASS(Flag,
       Function,  // symbol is a function
       Subroutine,  // symbol is a subroutine
-      Implicit,  // symbol is implicity typed
+      Implicit,  // symbol is implicitly typed
       ModFile,  // symbol came from .mod file
       ParentComp  // symbol is the "parent component" of an extended type
   );
--- a/lib/semantics/symbol.h
+++ b/lib/semantics/symbol.h
@@ -269,11 +269,18 @@ std::string DetailsToString(const Details &);
 
 class Symbol {
 public:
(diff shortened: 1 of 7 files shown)


lib/semantics/symbol.h
```diff
@@ -269,11 +269,18 @@ std::string DetailsToString(const Details &);
 
 class Symbol {
 public:
-  ENUM_CLASS(Flag, Function, Subroutine, Implicit, ModFile);
+  ENUM_CLASS(Flag,
+      Function,  // symbol is a function
+      Subroutine,  // symbol is a subroutine
+      Implicit,  // symbol is implicity typed
```
klausler: "implicitly"

PR #193, opened 2018-09-20
canonicalization of DO constructs

schweitzpgi:
Add tests and finish up DO loop canonicalization
Rewrites the parse tree

### DIFF
--- a/lib/semantics/canonicalize-do.cc
+++ b/lib/semantics/canonicalize-do.cc
@@ -35,6 +35,12 @@ public:
     if (!labels_.empty() && statement.label.has_value() &&
         labels_.back() == *statement.label) {
       auto currentLabel{labels_.back()};
+      if constexpr (std::is_same_v<T, common::Indirection<parser::EndDoStmt>>) {
+        std::get<parser::ExecutableConstruct>(currentIter_->u).u =
+            parser::Statement<parser::ActionStmt>{
+                std::optional<parser::Label>{currentLabel},
+                parser::ContinueStmt{}};
+      }
       do {
         currentIter_ = MakeCanonicalForm(labelDoIters_.back(), currentIter_);
         labelDoIters_.pop_back();
--- a/lib/semantics/canonicalize-do.cc
+++ b/lib/semantics/canonicalize-do.cc
@@ -27,7 +27,7 @@ public:
     const auto &endIter{executionPart.v.end()};
     currentList_ = &executionPart.v;
     for (auto iter{executionPart.v.begin()}; iter != endIter; ++iter) {
-      iter = CheckStatement(iter);
+      iter = ConvertLabelDoToStructuredDo(iter);
     }
     return false;
   }
@@ -51,7 +51,7 @@ public:
   }
 
 private:
-  parser::Block SpliceBlock(
+  parser::Block ExtractBlock(
       parser::Block::iterator beginLoop, parser::Block::iterator endLoop) {
     parser::Block block;
     block.splice(block.begin(), *currentList_, ++beginLoop, ++endLoop);
@@ -82,12 +82,13 @@ private:
                 parser::NonLabelDoStmt{
                     std::make_tuple(std::optional<parser::Name>{},
                         ExtractLoopControl(startLoop))}},
-            SpliceBlock(startLoop, endLoop),
+            ExtractBlock(startLoop, endLoop),
             parser::Statement<parser::EndDoStmt>{std::optional<parser::Label>{},
                 parser::EndDoStmt{std::optional<parser::Name>{}}})};
     return startLoop;
   }
-  parser::Block::iterator CheckStatement(const parser::Block::iterator &iter) {
+  parser::Block::iterator ConvertLabelDoToStructuredDo(
+      const parser::Block::iterator &iter) {
     currentIter_ = iter;
     parser::ExecutionPartConstruct &executionPartConstruct{*iter};
     if (auto *executableConstruct = std::get_if<parser::ExecutableConstruct>(
--- a/lib/semantics/canonicalize-do.cc
+++ b/lib/semantics/canonicalize-do.cc
@@ -15,7 +15,7 @@
 #include "canonicalize-do.h"
 #include "../parser/parse-tree-visitor.h"
 
-namespace Fortran::semantics {
+namespace Fortran::parser {
 
 class CanonicalizationOfDoLoops {
 public:
@@ -23,7 +23,7 @@ public:
 
   template<typename T> bool Pre(T &) { return true; }
   template<typename T> void Post(T &) {}
-  bool Pre(parser::ExecutionPart &executionPart) {
+  bool Pre(ExecutionPart &executionPart) {
     const auto &endIter{executionPart.v.end()};
     currentList_ = &executionPart.v;
     for (auto iter{executionPart.v.begin()}; iter != endIter; ++iter) {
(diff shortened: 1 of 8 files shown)


lib/semantics/canonicalize-do.cc
```diff
@@ -0,0 +1,112 @@
...
+    }
+    return false;
+  }
+
+private:
+  parser::Block SpliceBlock(
+      parser::Block::iterator beginLoop, parser::Block::iterator endLoop) {
+    parser::Block block;
+    block.splice(block.begin(), *currentList_, ++beginLoop, ++endLoop);
+    return block;
```
tskeith: This function confuses me. It's moving elements from currentList_ to block, right? But even though block is returned, the return value isn't used. So if the goal is just to modify currentList_, isn't there a `remove` or `erase` function that does that? Even if `splice` is the best way to do this, I think the function could use a better name to make it clear what it is doing.

lib/semantics/canonicalize-do.cc
```diff
@@ -0,0 +1,118 @@
...
+        common::Indirection<parser::DoConstruct>{std::make_tuple(
+            parser::Statement<parser::NonLabelDoStmt>{
+                std::optional<parser::Label>{},
+                parser::NonLabelDoStmt{
+                    std::make_tuple(std::optional<parser::Name>{},
+                        ExtractLoopControl(startLoop))}},
+            SpliceBlock(startLoop, endLoop),
+            parser::Statement<parser::EndDoStmt>{std::optional<parser::Label>{},
+                parser::EndDoStmt{std::optional<parser::Name>{}}})};
+    return startLoop;
```
tskeith: I think `std::nullopt` is better than `std::optional<...>{}` in most cases, unless you think it's important to make the type explicit.

lib/semantics/canonicalize-do.cc
schweitzpgi: The return value is used. It is the body of the constructed DoConstruct.

lib/semantics/canonicalize-do.cc
schweitzpgi: My compiler was whining about that, so I made them explicit to make it stop.

lib/semantics/canonicalize-do.cc
tskeith: Thanks, I missed that. The statement started so many lines earlier that I didn't realize the call to SpliceBlock was an argument and not at statement of its own.

schweitzpgi: I can turn the entirety of MakeCanonicalForm() into a nested lambda expression. I don't think that would improve readability myself.

lib/semantics/canonicalize-do.cc
```diff
@@ -0,0 +1,118 @@
...
+public:
+  CanonicalizationOfDoLoops() = default;
+
+  template<typename T> bool Pre(T &) { return true; }
+  template<typename T> void Post(T &) {}
+  bool Pre(parser::ExecutionPart &executionPart) {
+    const auto &endIter{executionPart.v.end()};
+    currentList_ = &executionPart.v;
+    for (auto iter{executionPart.v.begin()}; iter != endIter; ++iter) {
+      iter = CheckStatement(iter);
```
klausler: Please be sure that you want to both update `iter` with `CheckStatement` *and* increment it in the `for` statement's increment expression.  If `CheckStatement` really just checks statements, I have no idea what its returned iterator might be pointing at.

lib/semantics/canonicalize-do.cc
```diff
@@ -0,0 +1,118 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "canonicalize-do.h"
+#include "../parser/parse-tree-visitor.h"
+
+namespace Fortran::semantics {
```
klausler: A lot of this code might be made more brief if it were in `namespace Fortran::parser`.  It doesn't seem to depend on much from semantics.

lib/semantics/canonicalize-do.cc
klausler: `Splice` isn't the best name here, I agree.  Maybe `Extract`.

lib/semantics/canonicalize-do.cc
```diff
@@ -0,0 +1,118 @@
...
+      if (auto *labelDoLoop = std::get_if<
+              parser::Statement<common::Indirection<parser::LabelDoStmt>>>(
+              &executableConstruct->u)) {
+        labelDoIters_.push_back(iter);
+        labels_.push_back(std::get<parser::Label>(labelDoLoop->statement->t));
+      } else if (!labels_.empty()) {
+        parser::Walk(executableConstruct->u, *this);
+      }
+    }
+    return currentIter_;
```
klausler: `currentIter_` is mysterious hidden shared state.  I don't know what it points at or which member functions can change it.

lib/semantics/canonicalize-do.cc
```diff
@@ -0,0 +1,118 @@
...
+                        ExtractLoopControl(startLoop))}},
+            SpliceBlock(startLoop, endLoop),
+            parser::Statement<parser::EndDoStmt>{std::optional<parser::Label>{},
+                parser::EndDoStmt{std::optional<parser::Name>{}}})};
+    return startLoop;
+  }
+  parser::Block::iterator CheckStatement(const parser::Block::iterator &iter) {
+    currentIter_ = iter;
+    parser::ExecutionPartConstruct &executionPartConstruct{*iter};
+    if (auto *executableConstruct = std::get_if<parser::ExecutableConstruct>(
```
klausler: C++17 fixed brace initialization with `auto` and we now use braces in these situations.

lib/semantics/canonicalize-do.h
```diff
@@ -0,0 +1,25 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef FORTRAN_SEMANTICS_CANONICALIZE_DO_H_
+#define FORTRAN_SEMANTICS_CANONICALIZE_DO_H_
+
```
klausler: A sentence or two describing what this module does would be informative.

lib/semantics/canonicalize-do.cc
tskeith: I don't think so either.

I think a different name for SpliceBlock would help. Maybe ExtractBlock.

lib/semantics/canonicalize-do.h
schweitzpgi: done

schweitzpgi: done

lib/semantics/canonicalize-do.cc
schweitzpgi: done

lib/semantics/canonicalize-do.cc
schweitzpgi: yes, it is necessary to keep the iterator pointing at a valid list member.

lib/semantics/canonicalize-do.cc
schweitzpgi: renamed to ExtractBlock, though the operation is std::list::splice

lib/semantics/canonicalize-do.cc
schweitzpgi: I'll add a comment.

lib/semantics/canonicalize-do.cc
schweitzpgi: ok

Issue #188, opened 2018-09-14
%RE and %IM complex part access


klausler:
These are parsed as component accesses and then failing in semantics as references to components of something that is not a derived type.  They should be silently passed through name resolution so that expression semantic analysis can take care of them, I think.

Issue #191, opened 2018-09-20
semantic error on correct usage


klausler:
```
complex :: ccopy, z
z = ccopy((1.0,2.0))
end
```
elicits "Use of 'ccopy' as a procedure conflicts with its declaration"

PR #194, opened 2018-09-24
Fix issues #188 and #191.

tskeith:


lib/semantics/resolve-names.cc
```diff
@@ -387,8 +387,10 @@ class ScopeHandler : public ImplicitRulesVisitor {
   std::optional<SubprogramKind> subpNamesOnly_;
 
   // Apply the implicit type rules to this symbol.
-  void ApplyImplicitRules(const SourceName &, Symbol &);
+  void ApplyImplicitRules(Symbol &);
   std::optional<const DeclTypeSpec> GetImplicitType(Symbol &);
+  bool ConvertToObjectEntity(Symbol &);
```
sscalpone: Convert what to ObjectEntity?

lib/semantics/resolve-names.cc
```diff
@@ -1228,16 +1232,15 @@ void ScopeHandler::EraseSymbol(const SourceName &name) {
...
-    symbol.set_details(ObjectEntityDetails{});
-  }
-  if (auto *details{symbol.detailsIf<ObjectEntityDetails>()}) {
-    if (!details->type()) {
-      if (const auto type{GetImplicitType(symbol)}) {
-        details->set_type(*type);
-      }
-    }
+void ScopeHandler::ApplyImplicitRules(Symbol &symbol) {
+  ConvertToObjectEntity(symbol);
```
sscalpone: Does GetType return a boolean?  Are we ok with non-boolean functions being used as booleans?

lib/semantics/resolve-names.cc
```diff
@@ -1252,6 +1255,36 @@ std::optional<const DeclTypeSpec> ScopeHandler::GetImplicitType(
   return type;
 }
 
+// Convert symbol to be a ObjectEntity or return false if it can't be.
+bool ScopeHandler::ConvertToObjectEntity(Symbol &symbol) {
+  if (symbol.has<ObjectEntityDetails>()) {
+    // nothing to do
+  } else if (symbol.has<UnknownDetails>()) {
+    symbol.set_details(ObjectEntityDetails{});
+  } else if (auto *details{symbol.detailsIf<EntityDetails>()}) {
```
sscalpone: Oh, this is the modern version of
if(a = b) {
...
}
What is the scope of details?

lib/semantics/resolve-names.cc
```diff
@@ -2245,20 +2264,27 @@ void DeclarationVisitor::Post(const parser::ProcInterface &x) {
...
   } else if (auto &type{GetDeclTypeSpec()}) {
     interface.set_type(*type);
+    isFunction = true;
   }
   auto attrs{GetAttrs()};
   if (currScope().kind() != Scope::Kind::DerivedType) {
     attrs.set(Attr::EXTERNAL);
   }
-  DeclareProcEntity(name, attrs, interface);
+  const auto &name{std::get<parser::Name>(x.t).source};
```
sscalpone: I'd feel more comfortable if there was an assertion indicating that only one of isF or isS can get true, if that's true.

lib/semantics/resolve-names.cc
```diff
@@ -2557,20 +2583,25 @@ const Symbol *ResolveNamesVisitor::ResolveDataRef(const parser::DataRef &x) {
...
     return nullptr;
   }
+  auto *type{base.GetType()};
   if (!type) {
     return nullptr;  // should have already reported error
   }
+  if (type->category() == DeclTypeSpec::Intrinsic &&
+      type->intrinsicTypeSpec().category() == TypeCategory::Complex) {
+    auto name{component.ToString()};
+    if (name == "re" || name == "im") {
```
sscalpone: If re & im are used an a component of a non complex, and are undefined, it might be nice to say that the prefix is not a complex (if you haven't ready done that).

lib/semantics/resolve-names.cc
tskeith: The symbol.

There is a comment on the function definition. I have been putting any comments on the definition rather than the declaration for private functions. In my view the declaration is boilerplate and ignored by human readers. Same with static non-member functions.

lib/semantics/resolve-names.cc
tskeith: It's a pointer. I'm okay with requiring `if (symbol.GetType() != nullptr)` if that's the style we want to follow.

lib/semantics/resolve-names.cc
sscalpone: Not clear to me whether llvm wants comments on the declaration or the definition -- Eric puts most comments on the declaration.  Google is clear that comments about the function interface belong on the declaration and comments about the inner workings of the function belong on the definition.   https://google.github.io/styleguide/cppguide.html#Function_Comments

lib/semantics/resolve-names.cc
tskeith: The error reported for `x%re` is either that `x` is not an object of derived type or, if it is, that `re` is not a component of the type.

In the first case we're guessing what the user meant: that `x` is complex or that `x` is a derived type. The former may be more likely, the latter was easier to implement.

lib/semantics/resolve-names.cc
tskeith: I'll look into adding consistency checking for the flags when they are set. There are probably other things that could be checked.

lib/semantics/resolve-names.cc
sscalpone: In this case, it's to help me as a reader of the code.

lib/semantics/resolve-names.cc
klausler: This is a *.cc file.  There is no public API needing documentation in this case.

lib/semantics/resolve-names.cc
klausler: If a function result value is being used only to determine a predicate, implicit conversion to `bool` is fine by me.

lib/semantics/resolve-names.cc
klausler: Few of the current set of `Symbol` flags can be simultaneously set and maybe it could become an `enum`.

lib/semantics/resolve-names.cc
tskeith: I don't understand this comment. What are you saying should happen?

sscalpone: I'm saying that my comment should have been on the bit about asserting function or subroutine.

lib/semantics/resolve-names.cc
tskeith: It turns out that setting Function or Subroutine on the symbol here is not necessary becuase DeclareProcEntity takes care of it.

PR #195, opened 2018-09-25
Categorize Fortran intrinsic procedures (and some more work on expressions)

klausler:
This pull request adds `documentation/intrinsics.md`, which categorizes and summarizes the many intrinsic procedures of Fortran.  We're nearing the point where we're going to want to somehow encode all of this information into some constexpr data structure in the front-end so that intrinsic calls (and other usage) can be properly checked and implemented.

(There's also some code changes that continue work on expression/variable semantics that I did before getting sidetracked into this study of intrinsics.)


documentation/intrinsics.md
```diff
@@ -0,0 +1,658 @@
...
+POPCNT(INTEGER(any) I) -> default INTEGER
+POPPAR(INTEGER(any) I) -> default INTEGER = IAND(POPCNT(I), Z'1')
+SHIFTA(INTEGER(k) I, INTEGER(any) SHIFT) -> INTEGER(k)
+SHIFTL(INTEGER(k) I, INTEGER(any) SHIFT) -> INTEGER(k)
+SHIFTR(INTEGER(k) I, INTEGER(any) SHIFT) -> INTEGER(k)
+TRAILZ(INTEGER(any) I) -> default INTEGER
+```
+
+Any typeless "BOZ" argument to `MERGE_BITS` is converted to `INTEGER(k)`.  Note that at most one
+of its first two arguments can be a typeless "BOZ" literal.
```
tskeith: I think this note would be better at the top of the section. I was puzzled by the `BOZ` arguments and they weren't mentioned in the General Rules.

documentation/intrinsics.md
```diff
@@ -0,0 +1,658 @@
...
+```
+ABS(REAL(k) A) -> REAL(k) = SIGN(A, 0.0)
+AIMAG(COMPLEX(k) Z) -> REAL(k) = Z%IM
+AINT(REAL(k) A [, KIND=k ]) -> REAL(KIND)
+ANINT(REAL(k) A [, KIND=k ] -> REAL(KIND)
+CONJG(COMPLEX(k) Z) -> COMPLEX(k) = CMPLX(Z%RE, -Z%IM)
+DIM(REAL(k) X, REAL(k) Y) -> REAL(k) = X-MIN(X,Y)
+DPROD(default REAL X, default REAL Y) -> DOUBLE PRECISION = DBLE(X)*DBLE(Y)
+EXP(REAL(k) X) -> REAL(k)
+INDEX(CHARACTER(k) STRING, CHARACTER(k) SUBSTRING [, LOGICAL(any) BACK, KIND=KIND(0) ]) -> INTEGER(KIND)
```
tskeith: I assume that any argument with a default value is optional. So even if you specify `BACK` here you don't have to specify `KIND`? Can we write this as:
`INDEX(CHARACTER(k) STRING, CHARACTER(k) SUBSTRING, LOGICAL(any) BACK=.false., KIND=KIND(0))`
and is that any clearer?

It seems like we could eliminate the square bracket syntax for arguments with defaults.

documentation/intrinsics.md
klausler: Yes.

documentation/intrinsics.md
klausler: Done where possible.  Since argument order matters when calls do not have keywords, some arguments with default values remain in brackets.

PR #196, opened 2018-10-01
DO CONCURRENT semantic constraints checking

schweitzpgi:
Add tests for DO CONCURRENT

### DIFF
--- a/lib/semantics/check-do-concurrent.cc
+++ b/lib/semantics/check-do-concurrent.cc
@@ -1,0 +1,205 @@
+// Copyright (c) 2018, NVIDIA CORPORATION.  All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "check-do-concurrent.h"
+#include "../parser/message.h"
+#include "../parser/parse-tree-visitor.h"
+
+namespace Fortran::semantics {
+
+using namespace parser::literals;
+
+// 11.1.7.5
+class DoConcurrentEnforcement {
+public:
+  DoConcurrentEnforcement(parser::Messages &messages) : messages_{messages} {}
+  template<typename T> bool Pre(const T &) { return true; }
+  template<typename T> void Post(const T &) {}
+  template<typename T> bool Pre(const parser::Statement<T> &statement) {
+    charBlock_ = statement.source;
+    return true;
+  }
+  // C1136
+  void Post(const parser::ReturnStmt &) {
+    messages_.Say(charBlock_,
+        parser::MessageFormattedText{
+            "RETURN not allowed in DO CONCURRENT"_err_en_US});
+  }
+  // C1137
+  void NoImageControl() {
+    messages_.Say(charBlock_,
+        parser::MessageFormattedText{
+            "image control statement not allowed in DO CONCURRENT"_err_en_US});
+  }
+  void Post(const parser::SyncAllStmt &) { NoImageControl(); }
+  void Post(const parser::SyncImagesStmt &) { NoImageControl(); }
+  void Post(const parser::SyncMemoryStmt &) { NoImageControl(); }
+  void Post(const parser::SyncTeamStmt &) { NoImageControl(); }
+  void Post(const parser::ChangeTeamConstruct &) { NoImageControl(); }
+  void Post(const parser::CriticalConstruct &) { NoImageControl(); }
+  void Post(const parser::EventPostStmt &) { NoImageControl(); }
+  void Post(const parser::EventWaitStmt &) { NoImageControl(); }
+  void Post(const parser::FormTeamStmt &) { NoImageControl(); }
+  void Post(const parser::LockStmt &) { NoImageControl(); }
+  void Post(const parser::UnlockStmt &) { NoImageControl(); }
+  void Post(const parser::StopStmt &) { NoImageControl(); }
+  void Post(const parser::EndProgramStmt &) { NoImageControl(); }
+
+  void Post(const parser::AllocateStmt &) {
+    if (ObjectIsCoarray()) {
+      messages_.Say(charBlock_,
+          parser::MessageFormattedText{
+              "ALLOCATE coarray not allowed in DO CONCURRENT"_err_en_US});
+    }
+  }
+  void Post(const parser::DeallocateStmt &) {
+    if (ObjectIsCoarray()) {
+      messages_.Say(charBlock_,
+          parser::MessageFormattedText{
+              "DEALLOCATE coarray not allowed in DO CONCURRENT"_err_en_US});
+    }
+    // C1140: deallocation of polymorphic objects
+  }
+  template<typename T> void Post(const parser::Statement<T> &) {
+    if (EndTDeallocatesCoarray()) {
+      messages_.Say(charBlock_,
+          parser::MessageFormattedText{
+              "implicit deallocation of coarray not allowed"
+              " in DO CONCURRENT"_err_en_US});
+    }
+  }
+  void Post(const parser::CallStmt &) {
+    // C1137: call move_alloc with coarray arguments
+    // C1139: call to impure procedure
+  }
+  // C1141
+  void Post(const parser::ProcedureDesignator &procedureDesignator) {
+    if (auto *name = std::get_if<parser::Name>(&procedureDesignator.u)) {
+      auto upperName{parser::ToUpperCaseLetters(name->ToString())};
+      if (upperName == "IEEE_GET_FLAG"s) {
+        messages_.Say(charBlock_,
+            parser::MessageFormattedText{
+                "IEEE_GET_FLAG not allowed in DO CONCURRENT"_err_en_US});
+      } else if (upperName == "IEEE_SET_HALTING_MODE"s) {
+        messages_.Say(charBlock_,
+            parser::MessageFormattedText{"IEEE_SET_HALTING_MODE not allowed"
+                                         " in DO CONCURRENT"_err_en_US});
+      } else if (upperName == "IEEE_GET_HALTING_MODE"s) {
+        messages_.Say(charBlock_,
+            parser::MessageFormattedText{"IEEE_GET_HALTING_MODE not allowed"
+                                         " in DO CONCURRENT"_err_en_US});
+      }
+    }
+  }
+
+  // C1138: extended ranges in DOs should be errors, not warnings
+
+  // 11.1.7.5
+  void Post(const parser::IoControlSpec &ioControlSpec) {
+    if (auto *charExpr =
+            std::get_if<parser::IoControlSpec::CharExpr>(&ioControlSpec.u)) {
+      if (std::get<parser::IoControlSpec::CharExpr::Kind>(charExpr->t) ==
+          parser::IoControlSpec::CharExpr::Kind::Advance) {
+        messages_.Say(charBlock_,
+            parser::MessageFormattedText{
+                "ADVANCE specifier not allowed in DO CONCURRENT"_err_en_US});
+      }
+    }
+  }
+
+private:
+  bool ObjectIsCoarray() { return false; } // placeholder
+  bool EndTDeallocatesCoarray() { return false; } // placeholder
+
+  parser::CharBlock charBlock_;
+  parser::Messages &messages_;
+};
+
+class DoConcurrentCollection {
+public:
+  DoConcurrentCollection(parser::Messages &messages) : messages_{messages} {}
+  template<typename T> constexpr bool Pre(const T &) { return true; }
+  template<typename T> constexpr void Post(const T &) {}
+  bool Pre(const parser::ExecutionPart &executionPart) {
+    const auto &cend{executionPart.v.cend()};
+    for (auto iter{executionPart.v.cbegin()}; iter != cend; ++iter) {
+      CheckDoConcurrent(iter);
+      currentIter_ = iter;
+    }
+    return true;
+  }
+  template<typename T> void Post(const parser::Statement<T> &statement) {
+    if (!labels_.empty() && statement.label.has_value() &&
+        labels_.back() == *statement.label) {
+      CheckConstraints(++labelDoIters_.back(), currentIter_);
+      labels_.pop_back();
+      labelDoIters_.pop_back();
+    }
+  }
+
+private:
+  void CheckConstraints(const parser::Block::const_iterator &begin,
+      const parser::Block::const_iterator &end) {
+    DoConcurrentEnforcement doConcurrentEnforcement{messages_};
+    for (auto iter = begin; iter != end; ++iter) {
+      Walk(*iter, doConcurrentEnforcement);
+    }
+  }
+  void CheckDoConcurrent(
+      const std::list<parser::ExecutionPartConstruct>::const_iterator &iter) {
+    const parser::ExecutionPartConstruct &executionPartConstruct{*iter};
+    if (auto *executableConstruct = std::get_if<parser::ExecutableConstruct>(
+            &executionPartConstruct.u)) {
+      if (auto *doConstruct =
+              std::get_if<common::Indirection<parser::DoConstruct>>(
+                  &executableConstruct->u)) {
+        if (std::get<std::optional<parser::LoopControl>>(
+                std::get<parser::Statement<parser::NonLabelDoStmt>>(
+                    (*doConstruct)->t)
+                    .statement.t)
+                .has_value()) {
+          CheckConstraints(std::get<parser::Block>((*doConstruct)->t).cbegin(),
+              std::get<parser::Block>((*doConstruct)->t).cend());
+        }
+      } else if (auto *labelDoLoop = std::get_if<parser::Statement<
+                     common::Indirection<parser::LabelDoStmt>>>(
+                     &executableConstruct->u)) {
+        if (std::get<std::optional<parser::LoopControl>>(
+                labelDoLoop->statement->t)
+                .has_value() &&
+            std::holds_alternative<parser::LoopControl::Concurrent>(
+                std::get<std::optional<parser::LoopControl>>(
+                    labelDoLoop->statement->t)
+                    ->u)) {
+          labelDoIters_.push_back(iter);
+          labels_.push_back(std::get<parser::Label>(labelDoLoop->statement->t));
+        }
+      }
+    }
+  }
+
+  parser::Messages &messages_;
+  std::vector<std::list<parser::ExecutionPartConstruct>::const_iterator>
+      labelDoIters_;
+  std::list<parser::ExecutionPartConstruct>::const_iterator currentIter_;
+  std::vector<parser::Label> labels_;
+};
+
+void CheckDoConcurrentConstraints(
+    parser::Messages &messages, const parser::Program &program) {
+  DoConcurrentCollection doConcurrentCollection{messages};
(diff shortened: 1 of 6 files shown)


lib/semantics/check-do-concurrent.cc
```diff
@@ -0,0 +1,205 @@
...
+private:
+  void CheckConstraints(const parser::Block::const_iterator &begin,
+      const parser::Block::const_iterator &end) {
+    DoConcurrentEnforcement doConcurrentEnforcement{messages_};
+    for (auto iter = begin; iter != end; ++iter) {
+      Walk(*iter, doConcurrentEnforcement);
+    }
+  }
+  void CheckDoConcurrent(
+      const std::list<parser::ExecutionPartConstruct>::const_iterator &iter) {
```
klausler: You can use the typedef `parser::Block` to save some typing.

lib/semantics/check-do-concurrent.cc
```diff
@@ -0,0 +1,205 @@
...
+                  &executableConstruct->u)) {
+        if (std::get<std::optional<parser::LoopControl>>(
+                std::get<parser::Statement<parser::NonLabelDoStmt>>(
+                    (*doConstruct)->t)
+                    .statement.t)
+                .has_value()) {
+          CheckConstraints(std::get<parser::Block>((*doConstruct)->t).cbegin(),
+              std::get<parser::Block>((*doConstruct)->t).cend());
+        }
+      } else if (auto *labelDoLoop = std::get_if<parser::Statement<
```
klausler: Why would there still be label DO statements in the program?  Aren't they canonicalized into block DO constructs by now?

lib/semantics/check-do-concurrent.cc
```diff
@@ -0,0 +1,205 @@
...
+
+private:
+  bool ObjectIsCoarray() { return false; } // placeholder
+  bool EndTDeallocatesCoarray() { return false; } // placeholder
+
+  parser::CharBlock charBlock_;
+  parser::Messages &messages_;
+};
+
+class DoConcurrentCollection {
```
klausler: Could you please use comments to document what this class is meant to do?  I can't tell from the name.

lib/semantics/check-do-concurrent.cc
```diff
@@ -0,0 +1,205 @@
...
+public:
+  DoConcurrentCollection(parser::Messages &messages) : messages_{messages} {}
+  template<typename T> constexpr bool Pre(const T &) { return true; }
+  template<typename T> constexpr void Post(const T &) {}
+  bool Pre(const parser::ExecutionPart &executionPart) {
```
klausler: This algorithm is similar to the one in labeled DO loop canonicalization and has similar issues with readability and correctness.  Let's resolve labeled DO loop canonicalization first and then it will be more clear what should be done here.

test/semantics/doconcurrent01.f90
```diff
@@ -0,0 +1,42 @@
...
+     return
+10 continue
+end subroutine do_concurrent_test1
+
+subroutine do_concurrent_test2(i,j,n,flag)
+  implicit none
+  integer :: i, j, n, flag, flag2
+  do concurrent (i = 1:n)
+    change team (j)
+      call ieee_get_flag(flag, flag2)
```
klausler: Without a `USE IEEE_EXCEPTIONS` this call should elicit a complaint about invoking a non-PURE procedure.

lib/semantics/check-do-concurrent.cc
```diff
@@ -0,0 +1,205 @@
...
+              " in DO CONCURRENT"_err_en_US});
+    }
+  }
+  void Post(const parser::CallStmt &) {
+    // C1137: call move_alloc with coarray arguments
+    // C1139: call to impure procedure
+  }
+  // C1141
+  void Post(const parser::ProcedureDesignator &procedureDesignator) {
+    if (auto *name = std::get_if<parser::Name>(&procedureDesignator.u)) {
```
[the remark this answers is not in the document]
schweitzpgi: ok

lib/semantics/check-do-concurrent.cc
```diff
@@ -0,0 +1,205 @@
...
+      auto upperName{parser::ToUpperCaseLetters(name->ToString())};
+      if (upperName == "IEEE_GET_FLAG"s) {
```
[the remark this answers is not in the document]
schweitzpgi: Understood. These checks were just a hand-waving sketch of what really needs to happen. I'll turn them into comments.

lib/semantics/check-do-concurrent.cc
```diff
@@ -0,0 +1,205 @@
...
+                                         " in DO CONCURRENT"_err_en_US});
+      }
+    }
+  }
+
+  // C1138: extended ranges in DOs should be errors, not warnings
+
+  // 11.1.7.5
+  void Post(const parser::IoControlSpec &ioControlSpec) {
+    if (auto *charExpr =
```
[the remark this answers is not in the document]
schweitzpgi: ok

lib/semantics/check-do-concurrent.cc
schweitzpgi: ok

lib/semantics/check-do-concurrent.cc
schweitzpgi: I implemented this pass first. No other reason.

lib/semantics/check-do-concurrent.cc
schweitzpgi: ok

lib/semantics/check-do-concurrent.cc
schweitzpgi: I'll revisit. Thanks.

lib/semantics/check-do-concurrent.cc
```diff
@@ -0,0 +1,270 @@
...
+    charBlock_ = statement.source;
+    if (statement.label.has_value()) {
+      labels_.insert(*statement.label);
+    }
+    return true;
+  }
+  // C1136
+  void Post(const parser::ReturnStmt &) {
+    messages_.Say(charBlock_,
+        parser::MessageFormattedText{
```
tskeith: Is the construction of a `parser::MessageFormattedText` necessary here? A `Message` can be constructed from a `MessageFixedText` so I think you can use just `messages_.Say(charBlock_, "..."_err_en_US);`.

lib/semantics/check-do-concurrent.cc
```diff
@@ -0,0 +1,270 @@
...
+      if (parser::ToUpperCaseLetters(
+              symbol.scope()->symbol()->name().ToString()) == moduleName) {
+        return true;
+      }
+    }
+    return false;
+  }
+
+  std::set<parser::Label> labels_;
+  parser::CharBlock charBlock_;
```
tskeith: I think it would be better for `charBlocks_` to have a name that indicates its purpose, e.g. `currStmt_`.

lib/semantics/check-do-concurrent.cc
```diff
@@ -0,0 +1,270 @@
...
+    }
+  }
+
+private:
+  bool anyObjectIsCoarray() { return false; }  // placeholder
+  bool anyObjectIsPolymorphic() { return false; }  // placeholder
+  bool EndTDeallocatesCoarray() { return false; }  // placeholder
+  bool isPure(Attrs &attrs) {
+    return attrs.HasAny({Attr::PURE}) ||
+        (attrs.HasAny({Attr::ELEMENTAL}) && !attrs.HasAny({Attr::IMPURE}));
```
tskeith: When you're testing a single attribute you can use `attrs.test(Attr::PURE)` etc.

lib/semantics/check-do-concurrent.cc
```diff
@@ -0,0 +1,270 @@
...
+          parser::MessageFormattedText{
+              "implicit deallocation of coarray not allowed"
+              " in DO CONCURRENT"_err_en_US});
+    }
+  }
+  // C1141: cannot call ieee_get_flag, ieee_[gs]et_halting_mode
+  void Post(const parser::ProcedureDesignator &procedureDesignator) {
+    if (auto *name{std::get_if<parser::Name>(&procedureDesignator.u)}) {
+      // C1137: call move_alloc with coarray arguments
+      if (parser::ToUpperCaseLetters(name->ToString()) == "MOVE_ALLOC"s) {
```
tskeith: The case of identifiers is normalized, so `name->ToString() == "move_alloc"s` should be good enough.

test/semantics/doconcurrent01.f90
schweitzpgi: done

lib/semantics/check-do-concurrent.cc
schweitzpgi: ok

lib/semantics/check-do-concurrent.cc
schweitzpgi: ok

lib/semantics/check-do-concurrent.cc
schweitzpgi: ok

lib/semantics/check-do-concurrent.cc
schweitzpgi: ok

lib/semantics/check-do-concurrent.cc
```diff
@@ -0,0 +1,254 @@
...
+  bool anyObjectIsCoarray() { return false; }  // placeholder
+  bool anyObjectIsPolymorphic() { return false; }  // placeholder
+  bool EndTDeallocatesCoarray() { return false; }  // placeholder
+  bool isPure(Attrs &attrs) {
+    return attrs.test(Attr::PURE) ||
+        (attrs.test(Attr::ELEMENTAL) && !attrs.test(Attr::IMPURE));
+  }
+  bool fromScope(const Symbol &symbol, const std::string &moduleName) {
+    if (symbol.scope() && symbol.scope()->IsModule()) {
+      if (symbol.scope()->symbol()->name().ToString() == moduleName) {
```
tskeith: `symbol.scope()` is the scope introduced by a symbol. `symbol.owner()` is the scope that it is in. Also, you need to call `GetUltimate()` to go from the local symbol to the one it is use-associated with. So the body of this function should be something like this:
```
    auto &owner{symbol.GetUltimate().owner()};
    return owner.IsModule() && owner.name().ToString() == moduleName;
```

lib/semantics/check-do-concurrent.cc
```diff
@@ -0,0 +1,171 @@
...
+  void Post(const parser::EventPostStmt &) { NoImageControl(); }
+  void Post(const parser::EventWaitStmt &) { NoImageControl(); }
+  void Post(const parser::FormTeamStmt &) { NoImageControl(); }
+  void Post(const parser::LockStmt &) { NoImageControl(); }
+  void Post(const parser::UnlockStmt &) { NoImageControl(); }
+  void Post(const parser::StopStmt &) { NoImageControl(); }
+  void Post(const parser::EndProgramStmt &) { NoImageControl(); }
+
+  void Post(const parser::AllocateStmt &) {
+    if (ObjectIsCoarray()) {
```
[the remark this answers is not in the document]
schweitzpgi: changed name of placeholder dummy function.

lib/semantics/check-do-concurrent.cc
```diff
@@ -0,0 +1,255 @@
...
+    if (std::get<std::optional<parser::LoopControl>>(
+            std::get<parser::Statement<parser::NonLabelDoStmt>>(doConstruct.t)
+                .statement.t)
+            .has_value() &&
+        std::holds_alternative<parser::LoopControl::Concurrent>(
+            std::get<std::optional<parser::LoopControl>>(
+                std::get<parser::Statement<parser::NonLabelDoStmt>>(
+                    doConstruct.t)
+                    .statement.t)
+                ->u)) {
```
tskeith: This would be easier to read if `std::get<std::optional<parser::LoopControl>>(std::get<parser::Statement<parser::NonLabelDoStmt>>(doConstruct.t).statement.t)` were extracted into a variable.

lib/semantics/check-do-concurrent.cc
schweitzpgi: Thanks. Ultimate meant nothing to me.

lib/semantics/check-do-concurrent.cc
```diff
@@ -0,0 +1,254 @@
...
+  }
+  // C1136
+  void Post(const parser::ReturnStmt &) {
+    messages_.Say(currentStatementSourcePosition_,
+        "RETURN not allowed in DO CONCURRENT"_err_en_US);
+  }
+  // C1137
+  void NoImageControl() {
+    messages_.Say(currentStatementSourcePosition_,
+        "image control statement not allowed in DO CONCURRENT"_err_en_US);
```
sscalpone: Is there a reason not to be specific here?  E.g. SYNC ALL statement is not allowed in DO CONCURRENT?

lib/semantics/check-do-concurrent.cc
```diff
@@ -0,0 +1,254 @@
...
+    }
+  }
+  void Post(const parser::DeallocateStmt &) {
+    if (anyObjectIsCoarray()) {
+      messages_.Say(currentStatementSourcePosition_,
+          "DEALLOCATE coarray not allowed in DO CONCURRENT"_err_en_US);
+    }
+    // C1140: deallocation of polymorphic objects
+    if (anyObjectIsPolymorphic()) {
+      messages_.Say(currentStatementSourcePosition_,
```
sscalpone: Can we point to the polymorphic objects in what I suppose is a list of objects?

lib/semantics/check-do-concurrent.cc
```diff
@@ -0,0 +1,254 @@
...
+          "DEALLOCATE polymorphic object(s) not allowed"
+          " in DO CONCURRENT"_err_en_US);
+    }
+  }
+  template<typename T> void Post(const parser::Statement<T> &) {
+    if (EndTDeallocatesCoarray()) {
```
sscalpone: I'm just a tourist, but I can't figure out what EndTD... stands for.

lib/semantics/check-do-concurrent.cc
```diff
@@ -0,0 +1,254 @@
...
+      messages_.Say(currentStatementSourcePosition_,
+          "implicit deallocation of coarray not allowed"
```
sscalpone: Will it be clear what coarray is being implicityly deallocated?

lib/semantics/check-do-concurrent.cc
```diff
@@ -0,0 +1,254 @@
...
+          " in DO CONCURRENT"_err_en_US);
+    }
+  }
+  // C1141: cannot call ieee_get_flag, ieee_[gs]et_halting_mode
+  void Post(const parser::ProcedureDesignator &procedureDesignator) {
+    if (auto *name{std::get_if<parser::Name>(&procedureDesignator.u)}) {
+      // C1137: call move_alloc with coarray arguments
+      if (name->ToString() == "move_alloc"s) {
+        if (anyObjectIsCoarray()) {
+          messages_.Say(currentStatementSourcePosition_,
```
sscalpone: Can we point to the corarray object(s)?  Perhaps have N messages so we don't have to worry about the (s)?

lib/semantics/check-do-concurrent.cc
```diff
@@ -0,0 +1,254 @@
...
+              " argument(s) not allowed"_err_en_US);
+        }
+      }
+      // C1139: call to impure procedure
+      if (name->symbol && !isPure(name->symbol->attrs())) {
+        messages_.Say(currentStatementSourcePosition_,
+            "call to impure subroutine in DO CONCURRENT not allowed"_err_en_US);
+      }
+      if (name->symbol && fromScope(*name->symbol, "ieee_exceptions"s)) {
+        if (name->ToString() == "ieee_get_flag"s) {
```
sscalpone: Is there a way we can not expose ToString() very often?  Perhaps name->Equals("") or something?  I don't have a good reason for hiding string other than just thinking my future self might not want to use String.

lib/semantics/check-do-concurrent.cc
schweitzpgi: None other than I just removed all the MessageFormattedText ctors.

lib/semantics/check-do-concurrent.cc
schweitzpgi: This is a placeholder only. It will necessarily change. Specifically, there is nothing to query at this point that says "at this END statement there are a bunch of objects to deallocate and some of them are polymorphic."

lib/semantics/check-do-concurrent.cc
schweitzpgi: I'm not in favor of introducing any variable names since it seems to always lead to review drag.

lib/semantics/check-do-concurrent.cc
schweitzpgi: All I see is ToString().

lib/semantics/check-do-concurrent.cc
klausler: I don't know what "review drag" is, but that expression really would be more readable if the reference to the extracted `std::optional` were made a local auto reference in an `if` predicate expression and its body, yes?  Otherwise, there's a multi-line common subexpression that leaves the reader wondering why it appears twice and whether the two instances are subtly distinct.

```
if (const auto &loopControl{std::get<std::optional<parser::LoopControl>>(...)}) {
  // use loopControl->
}
```

lib/semantics/check-do-concurrent.cc
```diff
@@ -215,32 +216,91 @@ class FindDoConcurrentLoops {
...
-      parser::Walk(
-          std::get<parser::Block>(doConstruct.t), doConcurrentEnforcement);
-      DoConcurrentLabelEnforce doConcurrentLabelEnforce{
-          messages_, doConcurrentEnforcement.labels()};
-      parser::Walk(
-          std::get<parser::Block>(doConstruct.t), doConcurrentLabelEnforce);
+  void Post(const parser::DoConstruct &doConstruct) {
+    auto &doStmt{
+        std::get<parser::Statement<parser::NonLabelDoStmt>>(doConstruct.t)};
+    if (std::get<std::optional<parser::LoopControl>>(doStmt.statement.t)
```
tskeith: The expression `std::get<std::optional<parser::LoopControl>>(doStmt.statement.t)` is repeated 4 times in this function. That makes it hard to read.

lib/semantics/check-do-concurrent.cc
```diff
@@ -215,32 +216,91 @@ class FindDoConcurrentLoops {
...
+  }
+  static void HasNoReferences(const std::vector<semantics::Symbol> &indexNames,
+      const parser::ScalarIntExpr &expression) {
+    // FIXME - implement
+  }
+  static bool IntegerVariable(const Symbol &variable) {
+    return variable.GetType()->category() !=
+        semantics::DeclTypeSpec::Intrinsic ||
+        variable.GetType()->intrinsicTypeSpec().category() !=
+        common::TypeCategory::Integer;
```
tskeith: It looks like this expression should be negated. Or maybe the function is meant to be `NotIntegerVariable`.

schweitzpgi:
Added prototype code for checking C1121 through C1130.

lib/semantics/check-do-concurrent.cc
klausler: This LGTM.

klausler:
Please resolve conflicts, then I'll approve.  Thanks!

schweitzpgi:
Conflicts are resolved now.

lib/semantics/check-do-concurrent.cc
klausler: You could cast the string to a CharBlock and then compare the CharBlocks, or extend CharBlock's comparison operators to accept string operands.

tskeith: `name->source == "ieee_get_flag"` should work now, right?

klausler: Oh, sure, right.

schweitzpgi: Thanks for the pointer. I'll change ToString(), etc. here.

PR #200, opened 2018-10-03
Add a test, a flag, a fix, and a parse tree visitor tweak.

klausler:
Add new test canondo04.f90 to exercise label DO loop canonicalization in more contexts.  The test doesn't pass; I probably won't merge this branch until it does.

Add new f18 flag "-fdebug-semantics" to cause semantic analysis to be run even if no specific semantic phase dumping is enabled.  Eventually this behavior should be the default and this flag can be reversed, but we'll want to first work through regression test suites with it and shake out any current bugs.

Fix the parsing of label DO loops nested within nonlabel DO constructs or containing nested nonlabel DO constructs; the fix is to capture the nonlabel DO loop nesting level when a label DO appears.  These situations came up while writing the new canondo04.f90 test above.

Tweak the parse tree visitation templates so that a `Block` is presented to the `Pre/Post` methods of a visitor/mutator rather than simply being traversed.  This is necessary for writing loop restructuring transformations that want to use the parse tree visitor pattern to find and handle nested constructs before dealing with outer constructs in a `Post` routine.  I put this template specialization into the central `parse-tree-visitor.h` header rather than specializing the behavior in a source file because it seems generally useful and didn't break any test.

### DIFF
--- a/lib/parser/parse-tree-visitor.h
+++ b/lib/parser/parse-tree-visitor.h
@@ -75,6 +75,9 @@ template<typename T, typename M> void Walk(std::optional<T> &x, M &mutator) {
     Walk(*x, mutator);
   }
 }
+// For most lists, just traverse the elements; but when a list constitutes
+// a Block (i.e., std::list<ExecutionPartConstruct>), also invoke the
+// visitor/mutator on the list itself.
 template<typename T, typename V> void Walk(const std::list<T> &x, V &visitor) {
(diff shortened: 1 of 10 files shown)


schweitzpgi:
Thanks for the test case.  I'm working on the problem.

klausler:
It would be best to do #201 first so that we never have a failing test.

PR #203, opened 2018-10-05
do a little polishing on the DO loop canonicalization code

schweitzpgi:
Build and pass the iteration state locally and more explicitly to eliminate the use of the Mutator object as a backdoor. Also directly use move semantics to build the DoConstruct.

### DIFF
--- a/lib/semantics/canonicalize-do.cc
+++ b/lib/semantics/canonicalize-do.cc
@@ -17,94 +17,72 @@
 
 namespace Fortran::parser {
 
-struct LabelInfo {
-  Block::iterator iter;
-  Label label;
-};
-
 class CanonicalizationOfDoLoops {
+  struct LabelInfo {
+    Block::iterator iter;
+    Label label;
+  };
+
 public:
-  CanonicalizationOfDoLoops(std::vector<LabelInfo> &labelInfos)
-    : labelInfos_{labelInfos} {}
   template<typename T> bool Pre(T &) { return true; }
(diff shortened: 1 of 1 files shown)


lib/semantics/canonicalize-do.cc
```diff
@@ -17,103 +17,72 @@
...
 
-struct LabelInfo {
-  Block::iterator iter;
-  Label label;
-};
-
 class CanonicalizationOfDoLoops {
+  struct LabelInfo {
+    Block::iterator iter;
+    Label label;
```
klausler: Note that the label can be computed easily by a function on the iterator.

lib/semantics/canonicalize-do.cc
```diff
@@ -17,103 +17,72 @@
...
-          labelInfos_.push_back(
-              LabelInfo{iter, std::get<Label>(labelDoLoop->statement->t)});
-        }
+  template<typename T>
+  void CanonicalizeIfMatch(Block &originalBlock, std::vector<LabelInfo> &stack,
+      Block::iterator &i, Statement<T> &statement) {
+    if (!stack.empty() && statement.label.has_value() &&
+        stack.back().label == *statement.label) {
+      auto currentLabel{stack.back().label};
+      if constexpr (std::is_same_v<T, common::Indirection<EndDoStmt>>) {
```
klausler: An END DO statement should never be encountered in a Block unless it bears a label used by an earlier label DO in the same Block.  That should be CHECKed.

Issue #202, opened 2018-10-05
Compilation failure on macOS


equal-l2:
- `Integer::HUGE()` collides with macro `HUGE` in `math.h`
- Requires casting in `UnparseVisitor::Before(const HollerithLiteralConstant &x)` (See #158)

ibara:
The `Integer::HUGE()` issue affects all *BSD systems.

equal-l2:
Is there any way to disable `__BSD_VISIBLE` while using POSIX.1-2008 (required for `strnlen`)?

According to FreeBSD headers,  `HUGE` macro is defined in `math.h` if `__BSD_VISIBLE` macro is defined.
`__BSD_VISIBLE` is defined in `sys/cdefs.h` unless either `_ANSI_SOURCE`, `_C99_SOURCE` or `_C11_SOURCE` is defined. 

On macOS, I can successfully disabled `HUGE` macro with `_POSIX_C_SOURCE`(defined as `200809L`) or `_XOPEN_SOURCE`(defined as `700`).

Or maybe can we solve the problem just with `#undef HUGE`?

klausler:
Does anybody have a clue how the macro HUGE in <math.h> gets exposed to the f18 source, which never includes that file?

equal-l2:
The full error log on macOS:
```
/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/c++   -I../include -Iinclude -I/usr/local/opt/llvm/include -std=c++17 -Wall -pedantic -g -DDEBUG -isysroot /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.14.sdk   -D__STDC_CONSTANT_MACROS -D__STDC_FORMAT_MACROS -D__STDC_LIMIT_MACROS -MD -MT lib/parser/CMakeFiles/FortranParser.dir/user-state.cc.o -MF lib/parser/CMakeFiles/FortranParser.dir/user-state.cc.o.d -o lib/parser/CMakeFiles/FortranParser.dir/user-state.cc.o -c ../lib/parser/user-state.cc
In file included from ../lib/parser/user-state.cc:17:
In file included from ../lib/parser/grammar.h:28:
In file included from ../lib/parser/stmt-parser.h:21:
In file included from ../lib/parser/token-parsers.h:28:
../lib/parser/../evaluate/integer.h:307:28: error: expected member name or ';' after declaration specifiers
[...]

Issue #158, opened 2018-08-02
Linking failure


pjh40:
The Hollerith unparse implementation needs some `static_cast<std::uint64_t>` in order to successfully link on Darwin.  Here is the diff vs revision 902054b
[diff-902054b.diff.txt](https://github.com/flang-compiler/f18/files/2254836/diff-902054b.diff.txt)

ibara:
This diff, applied to the current tip of the master branch, fixes the build on OpenBSD as well (using gcc-8.2.0).

equal-l2:
@pjh40 Will you make a PR for this?

pjh40:
Sorry, been away from the office.  I'll put one in today.

pjh40:
This issue was resolved by PR #205.  Thanks!

Issue #125, opened 2018-07-15
std::string::size_type causing problems on Mac gcc-8.1.0


pjh40:
Under gcc-8.1.0 (or libstdc++-8.1.0 to be more specific), the std::string::size_type is 'unsigned long'.  This causes compilation errors with the instantiation of 
```c++
template<int KIND>
typename CharacterExpr<KIND>::LengthExpr CharacterExpr<KIND>::LEN() const;
```
at expression.cc:211, because there is no Integer(unsigned long) ctor, and g++ is reluctant to convert UL to ULL.  This can be addressed with a `std::uint64_t(x.size())` in the base case of the visitor constructor, or, perhaps more robustly, adding `Integer(unsigned long)`.

klausler:
I will add a static_cast to avoid potential conflicts between the Integer constructors of fixed size integer types (e.g., std::uint64_t) and the implementation-defined types (e.g., "unsigned long"), which I've been trying to avoid using generally.

klausler:
Should be fixed as of commit 57e051eafa1088e83c7d1cef565add6903aed20a, please confirm when convenient.

pjh40:
works for me, thanks for the fix!

PR #204, opened 2018-10-09
Address differences between std::size_t and std::uint64_t on non-Linux systems

pjh40:
Added some static_cast<std::uint64_t> expressions to address differences
between std::size_t and std::uint64_t on Darwin and *BSD systems.  This
commit resolves issue #158 and partially resolves issue #202.

### DIFF
--- a/lib/parser/unparse.cc
+++ b/lib/parser/unparse.cc
@@ -191,9 +191,9 @@ public:
         encoding_ == Encoding::EUC_JP ? EUC_JPCharacterBytes
                                       : UTF8CharacterBytes)};
     if (chars.has_value()) {
-      Unparse(*chars);
+      Unparse(static_cast<std::uint64_t>(*chars));
     } else {
-      Unparse(x.v.size());
+      Unparse(static_cast<std::uint64_t>(x.v.size()));
     }
     Put('H');
   }


pjh40:
`Unparse` is specialized on `std::uint64_t`, but not `std::size_t`.  On Darwin and, apparently, *BSD, these are not the same types.  If you extract some `type_info` on these types, you end up with:
```
std::size_t
	type_info.name: m
	type_index.hash(): 14947210986724563927
	bytes: 8
std::string::size_type
	type_info.name: m
	type_index.hash(): 14947210986724563927
	bytes: 8
std::uint64_t
	type_info.name: y
	type_index.hash(): 15160494916537613830
	bytes: 8
long_hex_auto
	type_info.name: l
	type_index.hash(): 11540658564628934174
	bytes: 8
```
where `long_hex_auto` is a surrogate for a variable instance that appeared in one of the test cases
```
  auto long_hex_auto {0x123456789abcdef};
```
Without this modification, there is a linking error on Darwin because of an unresolved call to `Unparse`.  I've attached the ugly bit of code that I used to generate this information, if you want to compare the results on a Linux box.
[utype.cc.txt](https://github.com/flang-compiler/f18/files/2461498/utype.cc.txt)

pjh40:
For reference, `size_t` is `unsigned long`, and that hex constant is `long`:
```
unsigned long
	type_info.name: m
	type_index.hash(): 14947210986724563927
	bytes: 8
long
	type_info.name: l
	type_index.hash(): 11540658564628934174
	bytes: 8
```
so, it would be more accurate to say that `unsigned long` and `uint64_t` are not equivalent under Darwin.

klausler:
Would it make things more future-proof to add a conditional specialization for size_t?

pjh40:
I would think so, but you seemed hesitant to do so in our discussion on issue #125. If you elect to do so, I would suggest rejecting this PR and having you write the conditional version, as I imagine there is some modern C++ technique for doing so that I'm not proficient at.  Note that the same problem exists for `long` and `std::int64_t`, which is why the `long_hex_auto` case is covered above.

pjh40:
There is something else strange going on here, because compilation of this program:
```
#include <cinttypes>
#include <iostream>
#include <string>

std::string Unparse(int x) { return std::to_string(x); }
std::string Unparse(std::uint64_t x) { return std::to_string(x); }
std::string Unparse(std::int64_t x) { return std::to_string(x); }

int main() {
  std::uint64_t v_uint64_t {1};
  std::size_t v_size_t {1};
  
  std::cout << Unparse(1) << '\n';
  std::cout << Unparse(v_uint64_t) << '\n';
  std::cout << Unparse(v_size_t) << '\n';

  return 0;
}
```
fails with an 
```
convert.cc:15:32: error: call of overloaded 'Unparse(std::size_t&)' is ambiguous
   std::cout << Unparse(v_size_t) << '\n';
```
which indicates to me that the compiler *is* willing to convert, but just not in the f18 system.  I'll dig around some more.

klausler:
Tim and I discussed this and perhaps the thing to do is to define Unparse overloads for the four types int32_t, uint32_t, int64_t, and uint64_t, with the reasonable expectations that (1) they're distinct enough for overload resolution, and (2) size_t maps to an integer type to which one of them also maps.

pjh40:
I believe that will still fail.  The reason that the compiler was able to attempt a conversion in my example above is because I didn't have your "catch-all" template:
```
  template<typename T> double Unparse(const T &);  // not void, never used
```
When I add that in, linking of my program fails with an unresolved `double Unparse(long unsigned const&)`.  What we are seeing is that the *compiler* sees `unsigned long` as distinct from a `std::uint64_t`, so it turns to the generic template.  It knows that there is a difference between the types because the type_info is different, so you would always have this problem if you don't have an `Unparse(size_t)`.  Just for reference, here is the Darwin output for all the types in question:
```
std::uint64_t
	type_info.name: y
	type_index.hash(): 15160494916537613830
	bytes: 8
std::int64_t
	type_info.name: x
	type_index.hash(): 13272411544345499535
	bytes: 8
std::uint32_t
	type_info.name: j
	type_index.hash(): 10485857595211860659
	bytes: 4
std::int32_t
	type_info.name: i
	type_index.hash(): 6253375586064260614
	bytes: 4
unsigned long
	type_info.name: m
	type_index.hash(): 14947210986724563927
	bytes: 8
long
	type_info.name: l
	type_index.hash(): 11540658564628934174
	bytes: 8
unsigned int
	type_info.name: j
	type_index.hash(): 10485857595211860659
	bytes: 4
int
	type_info.name: i
	type_index.hash(): 6253375586064260614
	bytes: 4
```
So, the 4-byte integers will be convertible, but not the 8-byte (I don't know why).

pjh40:
Adding this to unparse.cc seems to work for me:
```
  template<
      typename = std::enable_if_t<!std::is_same_v<std::uint64_t, std::size_t>>>
  void Unparse(std::size_t x) {
    Put(std::to_string(x));
  }
```
If that doesn't look crazy, I'll:
1) go through and remove the static_casts from my fork and the ones that we put in on a previous commit
2) add this implementation and a similar one for longs.

pjh40:
I've got these changes made in another branch, but there are a couple of observations:
1) I didn't need to do anything for longs
2) I did need to make a conditional ctor for Integer
I'll merge them onto this PR if you're interested.

schweitzpgi:
unsigned int, unsigned long, and unsigned long long are distinct basic types (though the sizes can, of course, overlap).

tskeith:
I've created PR #205 as another solution to this problem.

Paul, for some reason I can't add you as a reviewer to the pull request. Please take a look and see if it fixes your problem. Peter suggested this approach and it seems nice and simple.

pjh40:
Okay, I'm trying branch tsk8 out now.  

Just FYI: gcc-8.2 produces a 407MB .s file for parsing.cc, which takes around 35 minutes for `as` to assemble into a 145MB .o file on my little laptop.  My turnaround time isn't great ;).  Be prepared for some package maintainers to be upset in the future.

pjh40:
The issue addressed by this PR was addressed through an alternative in PR #205

klausler:
Which build are you doing that yields such a large parsing.o file?

klausler:
Never mind, I see that it must have been the Debug build.

pjh40:
Yes, it was Debug.  Somewhat surprisingly, the Release build is *much* faster.

PR #205, opened 2018-10-10
Use fundamental types in overloadings of Unparse.

tskeith:
Different systems map std::size_t, string::size_type, etc. to different
fundamental types. To ensure they are all covered, make the overloadings
of Unparse for integer types use only fundamental types.

This fixes compilations problems on Darwin and *BSD systems.

### DIFF
--- a/lib/parser/unparse.cc
+++ b/lib/parser/unparse.cc
@@ -62,8 +62,11 @@ public:
   // Emit simple types as-is.
   void Unparse(const std::string &x) { Put(x); }
   void Unparse(int x) { Put(std::to_string(x)); }
-  void Unparse(std::uint64_t x) { Put(std::to_string(x)); }
-  void Unparse(std::int64_t x) { Put(std::to_string(x)); }
+  void Unparse(unsigned int x) { Put(std::to_string(x)); }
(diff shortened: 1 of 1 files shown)


tskeith:
This is a different solution to the problem addressed in PR #204.

pjh40:
This worked for me on Darwin + gcc-8.2.0 (MacPorts build).  Unable to test with LLVM due to issue #202.

Note that there is a similar problem with `size_t` being passed to the ctor of class `Integer` (in lib/evaluate/integer.h), although it could be potentially more dangerous because of the bit-twiddling going on there.  `Integer(size_t)` is needed in two places that I'm aware of:

* expression.cc, line 482
* variable.cc, line 123

It may make sense to refactor the `Integer` constructors:

1)  Make local functions that can decode `uint32_t`, `int32_t`, `uint64_t` and `int64_t` (or maybe, just the last two)
2) Make a simple ctor for each POD (as was done in Unparse) that converts to one of these `<cstdint>` types  and calls the appropriate decoder.

sscalpone:
@pjh40 We'll be relicensing soon to use the new llvm license.  Is it ok for us to relicense this change too to the new license?

http://llvm.org/foundation/relicensing/ for more information.

PR #206, opened 2018-10-10
Make conversions between C++ integral types and Integer fully generic.

klausler:
Intended to fix lingering and future portability problems by accepting all integer types in the constructor for the class template Integer.


tskeith:
Looks good. With this change we can remove the `static_cast<std::uint64_t>` added for Darwin in variable.cc and expression.cc.

pjh40:
Works for me, even with the Darwin-specific `static_cast<std::uint64_t>` statements removed.

PR #210, opened 2018-10-10
More name resolution

tskeith:


tskeith:
MiscDetails is the details used for a symbol where we don't need to save much information about the symbol.

lib/semantics/resolve-names.cc
```diff
@@ -2459,6 +2531,22 @@ bool DeclarationVisitor::OkToAddComponent(
   }
 }
 
+// ConstructNamesVisitor implementation
+
+bool ConstructNamesVisitor::CheckDef(const std::optional<parser::Name> &x) {
```
klausler: Is there a reason that this function has a `bool` result, given that it's always true?  Will it be extended in the future with code that could return false?

tskeith: No, I think it's just a historical artifact. I'm doing more work on constructs, so I'll clean this up in a future change.

PR #212, opened 2018-10-12
Intrinsic function tables

klausler:
Implement intrinsic procedure semantics via table-driven pattern matching.


lib/evaluate/intrinsics.cc
```diff
@@ -0,0 +1,1244 @@
...
+static constexpr CategorySet Int{TypeCategory::Integer};
+static constexpr CategorySet Real{TypeCategory::Real};
+static constexpr CategorySet Complex{TypeCategory::Complex};
+static constexpr CategorySet Char{TypeCategory::Character};
+static constexpr CategorySet Logical{TypeCategory::Logical};
+static constexpr CategorySet IntOrReal{Int | Real};
+static constexpr CategorySet Floating{Real | Complex};
+static constexpr CategorySet Numeric{Int | Real | Complex};
+static constexpr CategorySet Relatable{Int | Real | Char};
+static constexpr CategorySet IntrinsicType{
```
tskeith: Consider just `Intrinsic` for the name of this set. The ones above don't have `Type` in their names.

`AnyType` below seems okay because it is special (and just `Any` wouldn't be specific enough).

lib/evaluate/intrinsics.cc
```diff
@@ -0,0 +1,1244 @@
...
+  CategorySet categorySet;
+  KindCode kindCode{KindCode::none};
+  std::ostream &Dump(std::ostream &) const;
+};
+
+// Abbreviations for argument and result patterns in the intrinsic prototypes:
+
+// Match specific kinds of intrinsic types
+static constexpr TypePattern DftInt{Int, KindCode::defaultIntegerKind};
+static constexpr TypePattern DftReal{Real, KindCode::defaultRealKind};
```
tskeith: "Dft" is unfortunate as a prefix, but I guess we already have type named "DefaultReal".

lib/evaluate/intrinsics.h
```diff
@@ -15,12 +15,56 @@
...
-ENUM_CLASS(IntrinsicProcedure, IAND, IEOR, IOR, LEN, MAX, MIN)
+// Intrinsics are distinguished by their names and the characteristics
+// of their arguments, at least for now.
+using IntrinsicProcedure = const char *;  // not an owning pointer
 
+class Argument;
+
+struct CallCharacteristics {
+  parser::CharBlock name;
+  const Arguments &argument;
```
tskeith: Is `arguments` a better name for this field?

lib/parser/char-block.h
```diff
@@ -74,6 +74,13 @@ class CharBlock {
   bool operator>=(const CharBlock &that) const { return Compare(that) >= 0; }
   bool operator>(const CharBlock &that) const { return Compare(that) > 0; }
 
+  bool operator<(const char *that) const { return Compare(that) < 0; }
+  bool operator<=(const char *that) const { return Compare(that) <= 0; }
+  bool operator==(const char *that) const { return Compare(that) == 0; }
+  bool operator!=(const char *that) const { return Compare(that) != 0; }
+  bool operator>=(const char *that) const { return Compare(that) >= 0; }
+  bool operator>(const char *that) const { return Compare(that) > 0; }
```
tskeith: These are useful functions but using operators makes me uneasy. Especially equality, which should be symmetric.

test/evaluate/intrinsics.cc
```diff
@@ -0,0 +1,214 @@
...
+  using Real4 = Type<TypeCategory::Real, 4>;
+  using Real8 = Type<TypeCategory::Real, 8>;
+  using Complex4 = Type<TypeCategory::Complex, 4>;
+  using Complex8 = Type<TypeCategory::Complex, 8>;
+  using Char = Type<TypeCategory::Character, 1>;
+  using Log4 = Type<TypeCategory::Logical, 4>;
+
+  TestCall{table, "bad"}
+      .Push(Const(Scalar<Int4>{}))
+      .DoCall();  // bad intrinsic name
```
tskeith: How are the results verified? E.g. in this case that "bad" is not an intrinsic?

lib/evaluate/intrinsics.cc
klausler: I use `IntrinsicType` to avoid confusion in a module that's concerned with intrinsic functions.

It might be better to append `Type` to the names of the others.  Or rename `IntrinsicType` to `AnyIntrinsic` or `AnyIntrinsicType`.  Which sounds better to you?

lib/evaluate/intrinsics.cc
klausler: Yes.  But that `DefaultReal` type isn't really used and maybe I should remove or rename it.

lib/evaluate/intrinsics.cc
tskeith: I think I prefer adding `Type` to the others.

lib/evaluate/intrinsics.h
klausler: Ok.

lib/parser/char-block.h
klausler: What if I make them non-member friend overloads?

test/evaluate/intrinsics.cc
klausler: There's no valid return type specified as an argument to `DoCall`, so intrinsic table probing must fail and must generate a message.

lib/evaluate/intrinsics.cc
```diff
@@ -0,0 +1,1244 @@
...
+  TypePattern result;
+  Rank rank{Rank::elemental};
+  std::optional<SpecificIntrinsic> Match(const CallCharacteristics &,
+      const IntrinsicTypeDefaultKinds &,
+      parser::ContextualMessages &messages) const;
+  std::ostream &Dump(std::ostream &) const;
+};
+
+static const IntrinsicInterface genericIntrinsicFunction[]{
+    {"abs", {{"a", SameIntOrReal}}, SameIntOrReal},
```
schweitzpgi: It might help document to have some no-op annotation like "a"_arg. I don't think it's unreadable as is though.

lib/parser/char-block.h
tskeith: My concern is just that with these operators you can have `x == y` but you can't rewrite it as `y == x`. Similarly, `x > y` is not equivalent to `y < x`.

But the alternatives give you compilation errors (as opposed to getting the wrong answer), so maybe it isn't a big deal.

klausler: I should be more explicit.  Please see the current source.  The point of turning (some) of those member functions into friends was to enable the symmetric overloads.

lib/evaluate/intrinsics.cc
klausler: After I commit these changes, I will expunge and replace the types like DefaultReal with a better pervasive solution, and then rename these objects with better names once they're available.

lib/parser/char-block.h
tskeith: I get it now. Yes, that addresses my concerns.

PR #213, opened 2018-10-16
Fully parameterize default kinds of intrinsic types

klausler:
Eliminates old definitions of `DefaultInteger` in `type.h`.  Creates a new object to hold default kinds of intrinsic types and uses it everywhere.  Adds options to the development driver program to tweak those defaults.  Renames `dft`-like names in `intrinsics.cc` to better names that have been made available.  Changes intrinsic function result types to accommodate large arrays even when default INTEGER is small.

Also document that `T` and `F` are not supported as extensions.


tools/f18/f18.cc
```diff
@@ -439,6 +441,12 @@ int main(int argc, char *const argv[]) {
     } else if (arg.substr(0, 2) == "-U") {
       options.predefinitions.emplace_back(
           arg.substr(2), std::optional<std::string>{});
+    } else if (arg == "-r8" || arg == "-fdefault-real-8") {
+      defaultKinds.set_defaultRealKind(8);
+    } else if (arg == "-i8" || arg == "-fdefault-integer-8") {
+      defaultKinds.set_defaultIntegerKind(8);
```
tskeith: Does it make sense to allow the default real and integer kind to be set independently? Default integer and default real are both supposed to occupy "a single numeric storage unit" so changing one but not the other would violate that.

klausler: The way that these options kind of make sense, at least to me, in those circumstances, is by viewing the expanded default type as occupying two numeric storage units.

PR #215, opened 2018-10-22
Refactoring: add SemanticsContext class

tskeith:
The new SemanticsContext holds the state of semantics whose lifetime
spans all of the compilations. It contains the scope tree (and so all
symbols), the intrinsics table, messages, and the state of options that
affect semantics (default kinds, search directories, etc.)

### DIFF
--- a/lib/parser/message.h
+++ b/lib/parser/message.h
@@ -240,12 +240,23 @@ private:
 
 class ContextualMessages {
 public:
+  class SavedState {
+  public:
+    SavedState(ContextualMessages &, CharBlock);
+    ~SavedState();
+
+  private:
+    ContextualMessages &msgs_;
+    CharBlock at_;
+  };
+
   ContextualMessages(CharBlock at, Messages *m) : at_{at}, messages_{m} {}
-  ContextualMessages(CharBlock at, const ContextualMessages &context)
-    : at_{at}, messages_{context.messages_} {}
 
   CharBlock at() const { return at_; }
   Messages *messages() const { return messages_; }
+
+  // Set CharBlock for messages; restore when the returned value is deleted
+  SavedState PushLocation(const CharBlock &);
 
   template<typename... A> void Say(A &&... args) {
     if (messages_ != nullptr) {
--- a/lib/parser/message.h
+++ b/lib/parser/message.h
@@ -256,7 +256,7 @@ public:
   Messages *messages() const { return messages_; }
 
   // Set CharBlock for messages; restore when the returned value is deleted
-  SavedState PushLocation(const CharBlock &);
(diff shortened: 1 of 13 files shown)


lib/semantics/semantics.cc
```diff
@@ -20,51 +20,52 @@
...
+  writer.WriteAll();
+  if (AnyFatalError()) {
     return false;
   }
-  return true;
+  if (context_.debugExpressions()) {
+    parser::CharBlock whole{cooked_.data()};
+    parser::ContextualMessages contextualMessages{whole, &context_.messages()};
+    evaluate::FoldingContext foldingContext{contextualMessages};
+    AnalyzeExpressions(program_, foldingContext, context_);
```
tskeith: Should the FoldingContext be part of SemanticsContext. So far, nothing in SemanticsContext gets reset between source files, so FoldingContext would change that. If we include it, references to CookedSource and Program could also go in there.

klausler: The folding context wraps some contextualized messages and some floating-point flags.  The flags won't change.  I believe that the context of the contextualized messages is pretty meaningless at the top level; new higher-resolution contexts are created on demand during expression semantics so that error messages have precise locations.  Does that help?

lib/semantics/expression.cc
```diff
@@ -1226,17 +1227,17 @@ evaluate::MaybeExpr AnalyzeExpr(evaluate::FoldingContext &context,
...
+    : context_{context}, semanticsContext_{semanticsContext} {}
 
   template<typename A> bool Pre(A &) { return true /* visit children */; }
   template<typename A> void Post(A &) {}
 
   bool Pre(parser::Expr &expr) {
     if (expr.typedExpr.get() == nullptr) {
       if (MaybeExpr checked{
-              AnalyzeExpr(context_, defaults_, intrinsics_, expr)}) {
+              AnalyzeExpr(context_, semanticsContext_.defaultKinds(),
```
klausler: Or you could change AnalyzeExpr to just take a SemanticsContext reference and unpack it in there as necessary.  This would make it easier to add things to the semantics context later.

lib/parser/message.h
```diff
@@ -240,13 +240,24 @@ class Messages {
 
 class ContextualMessages {
 public:
+  class SavedState {
```
klausler: If I understand, you expect this code to be used like

```
{ auto namedoesntmatter{context.PushLocation(here)};
  ...
}
```

I was thinking along the lines of something like:

```
{ WithMessageLocation unusedname{context, here};
  ....
}
```

Modulo naming, I'm not sure which is more clear.  (Maybe "Push" shouldn't be in the name since there's no explicit "Pop".)

PR #218, opened 2018-10-26
More name resolution for submodules

tskeith:


lib/semantics/symbol.h
```diff
@@ -364,6 +364,7 @@ class Symbol {
 
   bool isSubprogram() const;
   bool HasExplicitInterface() const;
+  bool IsSeparateMp() const;
```
sscalpone: Is Mp a word from the standard or an abbreviation that could be spelled out?

tskeith: It is used in the grammar for _mp-subprogram-stmt_ but mostly it is spelled out.

I thought that `IsSeparateModuleProcedure` was too long, though it is not used much so maybe it's okay.

`Proc` is frequently used as an abbreviation. `Mod` is not except for in `ModFile`. So `IsSeparateModuleProc` would be okay.

sscalpone: +1 IsSeparateModuleProc

lib/semantics/resolve-names.cc
```diff
@@ -1839,23 +1861,39 @@ bool SubprogramVisitor::Pre(const parser::Suffix &suffix) {
   return true;
 }
 
+bool HasModulePrefix(const std::list<parser::PrefixSpec> &prefixes) {
+  for (auto &prefix : prefixes) {
```
klausler: `const`

lib/semantics/mod-file.cc
```diff
@@ -192,19 +192,19 @@ void ModFileWriter::PutDerivedType(const Symbol &typeSymbol) {
...
-  bool isExternal{attrs.test(Attr::EXTERNAL)};
-  std::ostream &os{isExternal ? decls_ : contains_};
-  if (isExternal) {
+  bool isInterface{details.isInterface()};
+  std::ostream &os{isInterface ? decls_ : contains_};
+  if (isInterface) {
     os << "interface\n";
   }
   PutAttrs(os, attrs, ""s, " "s);
-  auto &details{symbol.get<SubprogramDetails>()};
```
schweitzpgi: Is details used somewhere earlier? I don't see why it was moved up to the top of this member function.

lib/semantics/resolve-names.cc
```diff
@@ -262,6 +263,7 @@ class ImplicitRulesVisitor : public DeclTypeSpecVisitor, public MessageHandler {
 protected:
   void PushScope();
   void PopScope();
+  void ClearScopes() { implicitRules_.reset(nullptr); }
```
schweitzpgi: Isn't nullptr the default argument?

lib/semantics/resolve-names.cc
```diff
@@ -313,6 +315,9 @@ class ArraySpecVisitor {
 // Manage a stack of Scopes
 class ScopeHandler : public ImplicitRulesVisitor {
 public:
+  template<typename T> void Walk(const T &);
+  void set_this(ResolveNamesVisitor *x) { this_ = x; }
```
schweitzpgi: this and this_ look very similar.

lib/semantics/resolve-names.cc
```diff
@@ -1839,23 +1861,39 @@ bool SubprogramVisitor::Pre(const parser::Suffix &suffix) {
...
 void SubprogramVisitor::Post(const parser::SubroutineSubprogram &) {
   EndSubprogram();
 }
 
 bool SubprogramVisitor::Pre(const parser::FunctionSubprogram &x) {
-  const auto &name{std::get<parser::Name>(
-      std::get<parser::Statement<parser::FunctionStmt>>(x.t).statement.t)};
+  const auto &stmt{
+      std::get<parser::Statement<parser::FunctionStmt>>(x.t).statement};
+  bool hasModulePrefix{
```
schweitzpgi: could be const

lib/semantics/resolve-names.cc
```diff
@@ -1864,15 +1902,15 @@ void SubprogramVisitor::Post(const parser::FunctionSubprogram &) {
 bool SubprogramVisitor::Pre(const parser::InterfaceBody::Subroutine &x) {
   const auto &name{std::get<parser::Name>(
       std::get<parser::Statement<parser::SubroutineStmt>>(x.t).statement.t)};
-  return BeginSubprogram(name, Symbol::Flag::Subroutine, std::nullopt);
+  return BeginSubprogram(name, Symbol::Flag::Subroutine, false, std::nullopt);
```
schweitzpgi: could document "false" as "false /* no module prefix */" or possibly use default arguments here

lib/semantics/resolve-names.cc
```diff
@@ -3141,12 +3211,8 @@ bool ResolveNamesVisitor::Pre(const parser::MainProgram &x) {
   } else {
     PushScope(Scope::Kind::MainProgram, nullptr);
   }
-  if (auto &subpPart{
-          std::get<std::optional<parser::InternalSubprogramPart>>(x.t)}) {
-    subpNamesOnly_ = SubprogramKind::Internal;
-    Walk(*subpPart);
-    subpNamesOnly_ = std::nullopt;
-  }
+  auto &subpPart{std::get<std::optional<parser::InternalSubprogramPart>>(x.t)};
```
schweitzpgi: should subp be spelled out as subprogram?

lib/semantics/symbol.h
tskeith: OK

lib/semantics/mod-file.cc
tskeith: Yes, I added a call to `details.isInterface()` up a few lines.

lib/semantics/resolve-names.cc

lib/semantics/resolve-names.cc
tskeith: Yes, I'll change it

lib/semantics/resolve-names.cc
tskeith: I think that is often too verbose for something that is not part of an API. But if the consensus is we should always spell it out I'll change it.

lib/semantics/resolve-names.cc
tskeith: They are meant to. They have the same value but different types. This is to avoid having to cast `this` to `ResolveNamesVisitor *` when calling `Walk`.

lib/semantics/resolve-names.cc

PR #222, opened 2018-11-05
C interoperability: Fixes CFI_CDESC_T macro and CFI_establish function

jeanPerier:
This PR resolves the following issue:

**- `CFI_CDESC_T` (18.5.4)**

1. `CFI_CDESC_T` defined in 18.5.4 par 2 was defined as `CFI_DESC_T` (typo)

2. `CFI_CDESC_T` macro had a trailing `;` making it fail in C/C++

3. `CFI_CDESC_T(0)` triggered a compile error in C++and `C_CDESC_T(x)`, x>0 was over-allocating memory in C++.

Fix: The macro in ISO_Fortran_binding.h was renamed and different versions defined for C and C++.

**- `CFI_ establish` (18.5.5.5)**
1. `CFI_establish` was returning an error with arguments (`base_addr !=0`, `attribute = CFI_attribute_other`, ...) when this should succeed according to 18.5.5.5.

2. `CFI_establish` was not defining `dv->elem_len` to be equal to argument `elem_len` when the attribute argument was `CFI_attribute_char`. See `elem_len` Formal Parameters description in 18.5.5.5 par 3. Also see Note 2 from 18.5.3.

3. For arguments (`base_addr !=0`, `attribute = CFI_attribute_pointer`, `rank>0`,...)`CFI_establish` was defining the lower bounds of pointer descriptor to be one instead of zero. This is weird, but in 18.5.5.5 par 3, the standard defines they should be set to zero (note: I am using draft version N2162).

4. 'CFI_establish' was reading inside `extents` when `rank>0 && base_addr==nullptr`. The standard specifies `extents ` shall be ignored in such case.  See `elem_len` Formal Parameters description in 18.5.5.5 par 3. This caused segfaults on legal calls to `CFI_establish` with `extents=nullptr`.

Fix: For 1,3 and 4, `CFI_establish` implementation was modified. For 2, function `MinElemLen`, also in ISO_Fortran_binding.cc, was modified to return 0 for CFI_type_char. 

**- Unit tests were added for `CFI_CDESC_T macro` and `CFI_establish`**:
Tests added inside new file test/evaluate/ISO-Fortran-binding.cc. There is one bias I have to acknowledge which is that these are tests from C++ when the standard specifies interoperability from C. Nevertheless, I think C++ interoperability is just as important, and in the case of `CFI_establish `already representative. I might think about adding a pure C version later. The argument space visited in the tests might seem big, but as it takes <10ms to run it, I do not see why not.


include/flang/ISO_Fortran_binding.h
```diff
@@ -45,11 +45,19 @@ typedef unsigned char CFI_rank_t;
 // and should be used for all array indexing and loop bound calculations.
 typedef ptrdiff_t CFI_index_t;
 
-#define CFI_DESC_T(rank) \
+#ifdef __cplusplus
+// C++ does not support zero-sized array (rank=0)
+// Also, CFI_cdesc_t already contains 1 dim in cpp
+#define CFI_CDESC_T(rank) \
+  std::aligned_storage_t<sizeof(CFI_cdesc_t) + (rank - 1) * sizeof(CFI_dim_t), \
```
klausler: Why the `aligned_storage_t`?  Is the type `CFI_cdesc_t` not already aligned properly?

I am nervous about a potentially negative value being multiplied against an unsigned value.

klausler: Maybe one could avoid the C++ restriction on zero-length arrays by changing the type of the array in that case to a zero-size struct, i.e. `std::conditional_t<rank != 0, CFI_dim_t, MyLocalEmptyStructType> dim[rank | !rank];`.

jeanPerier: Well, my issue was with the size not with the alignment. Because of the `dim[1]` inside the definition of `CFI_cdesc_t` (independent of the CFI_CDESC_T macro definition), we would waste 24bytes in the case of a C++ user defined scalar descriptor if the macro uses a structure containing `CFI_cdesc_t`. So I wanted to replacing it with a structure containing only a `char[]` to allocate what is necessary (similarly to the `new char[bytes]` used by the runtime in descriptor.cc for storage allocation).  Then I was unsure that we would get the alignement required by `CFI_cdesc_t`, an therefore used `aligne_storage_t`.
 
Thanks for the suggestion of `std::conditional_t`.  To push the reasoning on the space saving, this means we would have to use it also inside the `CFI_cdesc_t` type definition inside ISO_Fortran_binding.h. I think it would work though it is unclear to me what could be the consequences of having such conditional type when the standard says the `CFI_cdesc_t `structure dim member is a `CFI_dim_t[]`. I cannot imagine any user code that it would break.
Was your idea also to use such conditional type in `CFI_cdesc_t `definition or only in `CFI_CDESC_T `macro?

klausler: If the rank is zero the code should never care about the type of the `dim[]` data member, although it might affect the behavior of the code in descriptor.h that locates the addendum.

The conditional type would not be used in the definition of `CFI_cdesc_t` because the predicate has to be testable at compilation time.

The other thing that you could do is just waste the 24 bytes in the case of a descriptor for a scalar in C++.

include/flang/ISO_Fortran_binding.h
```diff
@@ -139,6 +125,28 @@ typedef struct CFI_cdesc_t {
   CFI_dim_t dim[CFI_ISO_FORTRAN_BINDING_FLEXIBLE_ARRAY]; /* must appear last */
 } CFI_cdesc_t;
 
+/* 18.5.4 */
+#ifdef __cplusplus
+// C++ does not support zero-sized array (rank=0)
+// Also, CFI_cdesc_t already contains 1 dim in cpp
+template<bool B, int rank> struct Cond_CFI_cdesc_t {
```
klausler: Please make sure that these declarations are hidden in a namespace.

schweitzpgi: If you really need a flexible sized object (with size >= 0) that's co-allocated in juxtaposition in C++, I'd suggest perhaps a class-specific member placement new and some accessor members.

jeanPerier: Thanks placement new is an interesting idea. I am not sure I can use it for what I want here though. I just want to create a type (not object) that is the correct size and alignment to hold a `CFI_cdesc_t` for an element of rank `rank` . The standard does not really care of what this type is made of. From what I understand of placement new, it would more fit to the next step to be done by the user:
`CFI_CDESC_T(rank) storage;
CFI_cdesc_t * dv = new (&storage) CFI_cdesc_t;`
Maybe it is possible to optimize the space by using member placement new inside the CFI_cdesc_t class in C++ but the standard is really precise on what CFI_cdesc_t structure should be so I do not want to tweak it too much in C++.
For now, I think CFI_CDESC_T is good enough in C++.

PR #223, opened 2018-11-07
Resolve expressions in symbols

tskeith:
The "Resolve expressions in symbols" commit contains the main part of these changes. The others are refactorings.

### DIFF
--- a/lib/evaluate/variable.cc
+++ b/lib/evaluate/variable.cc
@@ -27,10 +27,10 @@ using namespace Fortran::parser::literals;
 
 namespace Fortran::evaluate {
 
-int GetSymbolRank(const Symbol *symbol) { return symbol->Rank(); }
+int GetSymbolRank(const Symbol &symbol) { return symbol.Rank(); }
 
-const parser::CharBlock &GetSymbolName(const Symbol *symbol) {
-  return symbol->name();
+const parser::CharBlock &GetSymbolName(const Symbol &symbol) {
+  return symbol.name();
 }
 
 // Constructors, accessors, mutators
--- a/lib/evaluate/variable.cc
+++ b/lib/evaluate/variable.cc
@@ -27,6 +27,12 @@ using namespace Fortran::parser::literals;
 
 namespace Fortran::evaluate {
 
+int GetSymbolRank(const Symbol *symbol) { return symbol->Rank(); }
+
+const parser::CharBlock &GetSymbolName(const Symbol *symbol) {
+  return symbol->name();
+}
+
 // Constructors, accessors, mutators
 
 Triplet::Triplet(std::optional<Expr<SubscriptInteger>> &&l,
--- a/lib/evaluate/variable.h
+++ b/lib/evaluate/variable.h
@@ -48,8 +48,8 @@ template<typename A> struct Variable;
 using IndirectSubscriptIntegerExpr =
     CopyableIndirection<Expr<SubscriptInteger>>;
 
-int GetSymbolRank(const Symbol *);
-const parser::CharBlock &GetSymbolName(const Symbol *);
+int GetSymbolRank(const Symbol &);
+const parser::CharBlock &GetSymbolName(const Symbol &);
 
 // R913 structure-component & C920: Defined to be a multi-part
 // data-ref whose last part has no subscripts (or image-selector, although
@@ -267,7 +267,7 @@ public:
 
   int Rank() const {
     return std::visit(
-        common::visitors{[](const Symbol *sym) { return GetSymbolRank(sym); },
+        common::visitors{[](const Symbol *sym) { return GetSymbolRank(*sym); },
             [](const auto &x) { return x.Rank(); }},
         u);
   }
@@ -282,7 +282,7 @@ public:
 
   std::ostream &Dump(std::ostream &o) const {
     std::visit(common::visitors{[&](const Symbol *sym) {
-                                  o << GetSymbolName(sym).ToString();
+                                  o << GetSymbolName(*sym).ToString();
                                 },
                    [&](const auto &x) { x.Dump(o); }},
         u);
--- a/lib/evaluate/variable.h
+++ b/lib/evaluate/variable.h
@@ -26,11 +26,14 @@
 #include "type.h"
 #include "../common/idioms.h"
 #include "../common/template.h"
-#include "../semantics/symbol.h"
 #include <optional>
 #include <ostream>
 #include <variant>
 #include <vector>
+
+namespace Fortran::semantics {
+class Symbol;
+}
 
 namespace Fortran::evaluate {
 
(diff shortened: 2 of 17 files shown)


lib/semantics/mod-file.cc
```diff
@@ -324,13 +360,37 @@ void PutProcEntity(std::ostream &os, const Symbol &symbol) {
...
+    os << ':';
+  } else {
+    PutExpr(os, x.GetExplicit());
+  }
+}
+
+void PutExpr(std::ostream &os, const LazyExpr &expr) {
+  if (expr.Get()) {
+    // TODO: Dump does not necessarily produce Fortran code
+    expr.Get()->Dump(os);
```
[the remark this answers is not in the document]
klausler: Coming soon.  `Dump` will become `AsFortran`.

test/evaluate/CMakeLists.txt
```diff
@@ -40,9 +40,11 @@ add_executable(expression-test
 )
 
 target_link_libraries(expression-test
-  FortranEvaluate
   FortranEvaluateTesting
+  FortranEvaluate
   FortranParser
+  FortranSemantics
+  FortranEvaluate
```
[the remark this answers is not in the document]
klausler: Can cmake merge binary libraries together?  That would be a good solution if possible.

lib/evaluate/variable.cc
```diff
@@ -27,6 +27,12 @@ using namespace Fortran::parser::literals;
 
 namespace Fortran::evaluate {
 
+int GetSymbolRank(const Symbol *symbol) { return symbol->Rank(); }
```
schweitzpgi: Random thought: would it help at all to have a type SymbolRef that encapsulates the whole pointer-to-immutable-object syntactic fluff?

schweitzpgi: Should the rank concept be encapsulated to enforce correctness? As a bare int, one could theoretically get a rank of -45288, for instance.

lib/evaluate/variable.h
```diff
@@ -276,7 +282,7 @@ template<typename A> class Designator {
 
   std::ostream &Dump(std::ostream &o) const {
     std::visit(common::visitors{[&](const Symbol *sym) {
-                                  o << sym->name().ToString();
+                                  o << GetSymbolName(sym).ToString();
```
schweitzpgi: I'm starting to wonder if having a conversion operator std::string for CharBlock would be a nice thing.

lib/evaluate/variable.cc
klausler: I would prefer a reference to a pointer here, but I wouldn't want to see a typedef for either.  I can tell what `const Symbol &` means without having to look at any other definition.

lib/evaluate/variable.h
klausler: Go for it, see whether it would be a net improvement to readability and maintainability.

lib/semantics/resolve-names.cc
```diff
@@ -1547,8 +1558,8 @@ bool ModuleVisitor::Pre(const parser::Module &x) {
   const auto &name{
       std::get<parser::Statement<parser::ModuleStmt>>(x.t).statement.v.source};
   auto &subpPart{std::get<std::optional<parser::ModuleSubprogramPart>>(x.t)};
-  auto &symbol{BeginModule(name, false, subpPart)};
-  MakeSymbol(name, symbol.details());
+  BeginModule(name, false, subpPart);
```
schweitzpgi: Should BeginModule() still be returning a Symbol& ?

lib/semantics/mod-file.cc
```diff
@@ -339,7 +399,7 @@ void PutEntity(
     std::ostream &os, const Symbol &symbol, std::function<void()> writeType) {
   writeType();
   PutAttrs(os, symbol.attrs());
-  PutLower(os << "::", symbol) << '\n';
+  PutLower(os << "::", symbol);
```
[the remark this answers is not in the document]
tskeith: I had to move around where the newline was emitted in order to handle writing out of initializations.

lib/semantics/resolve-names.cc
tskeith: We're still using the return value in the submodule case, but it's possible it could follow the pattern of the module case.

tskeith: Regarding symbol ownership: Yes, Symbols are all owned by Scopes, which are in a tree rooted at the global scope, which is in the SemanticsContext. That's why we can freely use pointers and references to Symbols and Scopes without worrying about lifetimes.

lib/semantics/resolve-names.cc
```diff
@@ -2408,7 +2433,14 @@ void DeclarationVisitor::Post(const parser::ComponentDecl &x) {
     attrs.set(Attr::PRIVATE);
   }
   if (OkToAddComponent(name)) {
-    DeclareObjectEntity(name, attrs);
+    auto &symbol{DeclareObjectEntity(name, attrs)};
+    if (auto *details{symbol.detailsIf<ObjectEntityDetails>()}) {
+      if (auto &init{std::get<std::optional<parser::Initialization>>(x.t)}) {
+        if (auto *initExpr{std::get_if<parser::ConstantExpr>(&init->u)}) {
+          details->set_init(GetExpr(*initExpr));
```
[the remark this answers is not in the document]
tskeith: The ComponentDecl case is similar but not identical.

lib/semantics/type.h
```diff
@@ -202,23 +225,47 @@ class GenericSpec {
   friend std::ostream &operator<<(std::ostream &, const GenericSpec &);
 };
 
-// The value of a len type parameter
-using LenParamValue = Bound;
+// A type parameter value: integer expression or assumed or deferred.
+class ParamValue {
+public:
+  static const ParamValue Assumed() { return ParamValue(Category::Assumed); }
+  static const ParamValue Deferred() { return ParamValue(Category::Deferred); }
```
schweitzpgi: Should this be ParamValue { } ?

test/evaluate/CMakeLists.txt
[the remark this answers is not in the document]
schweitzpgi: I vote for combining these somehow. Circular lib dependences may not be considered friendly.

lib/semantics/type.h
klausler: Or just `{Category::...}`, which would be okay here from a readability standpoint since the type of the function result is right there on the same line.

lib/evaluate/variable.cc
tskeith: I'll change it to a reference.

lib/semantics/type.h
tskeith: OK, I'll change this.

PR #225, opened 2018-11-14
Roll up a bunch of changes

klausler:
This is a large pull request that really should have been implemented
as multiple smaller sets of changes.

- Documentation updates
- Renaming Dump() member functions to AsFortran() for use in module file
  output
- Implementing conversions of Real to scaled decimals for use in
  formatting for use in AsFortran()
- Fixing Real::Read() for decimal->binary conversion
- Initial placeholder declarations for static compiler-created data
- Use of static data for literal strings used as anonymous parents of
  substring references, in place of special-purpose representations
  of character literal substrings (e.g. 'FORTRAN'(2:4))
- Changed ActualArgument::isAssumedRank data flag member into a predicate
  function of the argument expression
- Representation of array constructors in expressions
- Cleaned up warning messages from recent merges into master
- Implemented Real::AINT() because I needed it for an early take on
  real -> scaled decimal conversion; turned out not to be necessary for that
- Split up GetSymbol(bool first/last) member functions into two separate
  functions (per class), GetBaseObject() and GetLastSymbol(), for readability
  reasons and so that GetBaseObject() could return a reference to static data
  as well as to a symbol.  GetFirstSymbol() is provided for those classes
  that can always return a symbol.  (E.g., for A(J,K)%F, the last symbol
  is F and the first symbol is A.)
- Renamed the lower and upper indices of a substring reference to
  "lower" and "upper", in place of "first" and "last", to avoid confusion
  with those names in the STL.

### DIFF
--- a/documentation/C++style.md
+++ b/documentation/C++style.md
@@ -190,6 +190,7 @@ the definition of an external destructor.
 optional deep copy semantics; reassignable.
 Often better than a reference (due to ownership) or `std::unique_ptr<>`
 (due to non-nullability and copyability).
+Can be wrapped in `std::optional<>` when nullability is required.
 * `CountedReference<>`: A nullable pointer with shared ownership via
 reference counting, null by default, shallowly copyable, reassignable.
 Safe to use *only* when the data are private to just one
@@ -199,19 +200,15 @@ of that standard feature is prohibitive.
 
 A feature matrix:
 
-| pointer | nullable | default null | owning | reassignable | copyable |
-| ------- | -------- | ------------ | ------ | ------------ | -------- |
-| `*p` | yes | no | no | yes | shallowly |
-| `&r` | no | n/a | no | no | shallowly |
-| `unique_ptr<>` | yes | yes | yes | yes | no |
-| `shared_ptr<>` | yes | yes | yes | yes | shallowly |
-| `OwningPointer<>` | yes | yes | yes | yes | no |
-| `Indirection<>` | no | n/a | yes | yes | optionally deeply |
-| `CountedReference<>` | yes | yes | yes | yes | shallowly |
-
-Note that `std::optional<Indirection<>>` checks all the boxes, if you
-need a nullable `Indirection<>`.
-
+| pointer | nullable | default null | owning | reassignable | copyable | undefined type ok? |
+| ------- | -------- | ------------ | ------ | ------------ | -------- | ------------------ |
+| `*p` | yes | no | no | yes | shallowly | yes |
+| `&r` | no | n/a | no | no | shallowly | no |
+| `unique_ptr<>` | yes | yes | yes | yes | no | no |
+| `shared_ptr<>` | yes | yes | yes | yes | shallowly | no |
+| `OwningPointer<>` | yes | yes | yes | yes | no | yes |
+| `Indirection<>` | no | n/a | yes | yes | optionally deeply | no |
+| `CountedReference<>` | yes | yes | yes | yes | shallowly | no |
 
 ### Overall design preferences
 Don't use dynamic solutions to solve problems that can be solved at
--- a/documentation/C++style.md
+++ b/documentation/C++style.md
@@ -196,6 +196,23 @@ Safe to use *only* when the data are private to just one
 thread of execution.
 Used sparingly in place of `std::shared_ptr<>` only when the overhead
 of that standard feature is prohibitive.
+
+A feature matrix:
+
+| pointer | nullable | default null | owning | reassignable | copyable |
+| ------- | -------- | ------------ | ------ | ------------ | -------- |
+| `*p` | yes | no | no | yes | shallowly |
+| `&r` | no | n/a | no | no | shallowly |
+| `unique_ptr<>` | yes | yes | yes | yes | no |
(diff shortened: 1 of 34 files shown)


documentation/C++style.md
```diff
@@ -164,35 +164,52 @@ There are many -- perhaps too many -- means of indirect addressing
...
-Used sparely in place of `std::shared_ptr<>` only when the overhead
+Used sparingly in place of `std::shared_ptr<>` only when the overhead
 of that standard feature is prohibitive.
+
+A feature matrix:
+
+| pointer | nullable | default null | owning | reassignable | copyable | undefined type ok? |
+| ------- | -------- | ------------ | ------ | ------------ | -------- | ------------------ |
+| `*p` | yes | no | no | yes | shallowly | yes |
+| `&r` | no | n/a | no | no | shallowly | no |
```
tskeith: Does "undefined type ok?" mean that all you need is a forward declaration of the type? That should be "yes" for references.

documentation/C++style.md
```diff
@@ -164,35 +164,52 @@ There are many -- perhaps too many -- means of indirect addressing
...
+| `unique_ptr<>` | yes | yes | yes | yes | no | no |
+| `shared_ptr<>` | yes | yes | yes | yes | shallowly | no |
+| `OwningPointer<>` | yes | yes | yes | yes | no | yes |
+| `Indirection<>` | no | n/a | yes | yes | optionally deeply | no |
+| `CountedReference<>` | yes | yes | yes | yes | shallowly | no |
+
```
tskeith: I think this table is really helpful. If the columns were lined up in the markdown it would be easier to read without a markdown viewer. Like this:
```
| pointer              | nullable | default null | owning | reassignable | copyable          | undefined type ok? |
| -------------------- | -------- | ------------ | ------ | ------------ | ----------------- | ------------------ |
| `*p`                 | yes      | no           | no     | yes          | shallowly         | yes                |
| `&r`                 | no       | n/a          | no     | no           | shallowly         | no                 |
| `unique_ptr<>`       | yes      | yes          | yes    | yes          | no                | no                 |
| `shared_ptr<>`       | yes      | yes          | yes    | yes          | shallowly         | no                 |
| `OwningPointer<>`    | yes      | yes          | yes    | yes          | no                | yes                |
| `Indirection<>`      | no       | n/a          | yes    | yes          | optionally deeply | no                 |
| `CountedReference<>` | yes      | yes          | yes    | yes          | shallowly         | no                 |
```

lib/semantics/type.cc
```diff
@@ -270,7 +270,7 @@ void ExprResolver::Resolve(Symbol &symbol) {
   if (auto *type{symbol.GetType()}) {
     if (type->category() == DeclTypeSpec::TypeDerived) {
       DerivedTypeSpec &dts{type->derivedTypeSpec()};
-      for (auto & [ name, value ] : dts.paramValues()) {
+      for (auto &[name, value] : dts.paramValues()) {
```
tskeith: I formatted this originally with clang-format version 5.0.1. I've switched to 6.0.1 and now it matches yours.

lib/evaluate/expression.cc
```diff
@@ -74,19 +74,15 @@ std::ostream &LogicalOperation<KIND>::Infix(std::ostream &o) const {
...
 }
 
-template<typename T> std::ostream &Constant<T>::Dump(std::ostream &o) const {
+template<typename T>
+std::ostream &Constant<T>::AsFortran(std::ostream &o) const {
   if constexpr (T::category == TypeCategory::Integer) {
     return o << value.SignedDecimal() << '_' << T::kind;
   } else if constexpr (T::category == TypeCategory::Real ||
       T::category == TypeCategory::Complex) {
     return o << value.DumpHexadecimal() << '_' << T::kind;
```
tskeith: Doesn't this mean that some real values will still come out in hex?

tskeith:
A couple of small points on formatting expressions:
1. Module files are currently written out in lower case. Expressions use upper case (for example, `1.E10_4` and `.AND.`). Do we want them to be consistent?
2. Some expressions get extra parentheses added around them. E.g. `x+y` -> `(x+y)` -> `((x+y))`. It would be nice if the parsing+formatting process were idempotent.

klausler:
> A couple of small points on formatting expressions:
> 
> 1. Module files are currently written out in lower case. Expressions use upper case (for example, `1.E10_4` and `.AND.`). Do we want them to be consistent?

I think that I can emit everything in lower case, will do.

> 2. Some expressions get extra parentheses added around them. E.g. `x+y` -> `(x+y)` -> `((x+y))`. It would be nice if the parsing+formatting process were idempotent.

Agree.  This turns out to be a bit harder than I thought it would be.  The compiler needs to preserve original parentheses and also needs to add them to the output to ensure correct precedence of operators (e.g., `PARAMETER(X=A+B, Y=C*X)` might become `Y=C*(A+B)`).  Both of these kinds of parentheses look like mandatory original source parentheses to the next round of parsing.  Since `((x))` becomes `(x)` in expression rewriting, there shouldn't be anything in the output that's triply nested -- i.e., things should become idempotent after a couple of rounds.

PR #227, opened 2018-11-19
 Implement CFI_set_pointer, CFI_section, CFI_select_part. Minor changes in other CFI funcs

jeanPerier:
This commit contains:

- Implementation of 18.5.5 CFI_setpointer, CFI_select_part and CFI_section as well as related unit tests.
Note that CFI_select_part and CFI_section also set lower bounds to zero. For nonallocatable nonpointer, this is to be in line with 18.5.3 par 3. For pointers, I have not found a clear indication, but based on CFI_establish behavior, it seems reasonable to set them to zero also.

- Fix of lower bound value in C descriptor:
CFI_establish now also sets lower bounds to zero for nonallocatable nonpointer data object descriptors according to 18.5.3 par 3. This is not done if no base_addr is given because it will be done in CFI_select_part and CFI_section in this case.

- Fix for character type in CFI_allocate:
It seems that CFI_type_cptr was used to refer to Fortran character type instead of CFI_type_char. The elem_len argument has to be considered for Fortran character type data objects for which the type encoding in the descriptor is CFI_type_char from what I understand.

### DIFF
--- a/runtime/ISO_Fortran_binding.cc
+++ b/runtime/ISO_Fortran_binding.cc
@@ -252,7 +252,7 @@ int CFI_section(CFI_cdesc_t *result, const CFI_cdesc_t *source,
     return CFI_INVALID_DESCRIPTOR;
   }
 
-  // For zero-sized arrays, base_addr is proc-dep (see 18.5.3).
+  // For zero-sized arrays, base_addr is processor-dependent (see 18.5.3).
   // We keep it on the source base_addr
   result->base_addr = isZeroSized ? source->base_addr
                                   : reinterpret_cast<void *>(shiftedBaseAddr);
--- a/runtime/ISO_Fortran_binding.cc
+++ b/runtime/ISO_Fortran_binding.cc
@@ -225,12 +225,12 @@ int CFI_section(CFI_cdesc_t *result, const CFI_cdesc_t *source,
 
   bool isZeroSized{false};
   for (int j{0}; j < source->rank; ++j) {
-    const CFI_dim_t &dim = source->dim[j];
-    const CFI_index_t srcLB = dim.lower_bound;
-    const CFI_index_t srcUB = srcLB + dim.extent - 1;
-    const CFI_index_t lb = (lower_bounds != nullptr) ? lower_bounds[j] : srcLB;
-    const CFI_index_t ub = (upper_bounds != nullptr) ? upper_bounds[j] : srcUB;
-    const CFI_index_t stride = (strides != nullptr) ? strides[j] : 1;
+    const CFI_dim_t &dim{source->dim[j]};
+    const CFI_index_t srcLB{dim.lower_bound};
+    const CFI_index_t srcUB{srcLB + dim.extent - 1};
+    const CFI_index_t lb{lower_bounds != nullptr ? lower_bounds[j] : srcLB};
+    const CFI_index_t ub{upper_bounds != nullptr ? upper_bounds[j] : srcUB};
+    const CFI_index_t stride{strides != nullptr ? strides[j] : 1};
 
     if (stride == 0 && lb != ub) {
       return CFI_ERROR_OUT_OF_BOUNDS;
@@ -240,7 +240,7 @@ int CFI_section(CFI_cdesc_t *result, const CFI_cdesc_t *source,
         return CFI_ERROR_OUT_OF_BOUNDS;
       }
       shiftedBaseAddr += (lb - srcLB) * dim.sm;
-      extent[j] = (stride != 0) ? 1 + (ub - lb) / stride : 1;
+      extent[j] = stride != 0 ? 1 + (ub - lb) / stride : 1;
     } else {
       isZeroSized = true;
       extent[j] = 0;
--- a/runtime/ISO_Fortran_binding.cc
+++ b/runtime/ISO_Fortran_binding.cc
@@ -33,6 +33,9 @@ void *CFI_address(
 
 int CFI_allocate(CFI_cdesc_t *descriptor, const CFI_index_t lower_bounds[],
     const CFI_index_t upper_bounds[], std::size_t elem_len) {
+  if (descriptor == nullptr) {
+    return CFI_INVALID_DESCRIPTOR;
+  }
   if (descriptor->version != CFI_VERSION) {
     return CFI_INVALID_DESCRIPTOR;
   }
@@ -80,6 +83,9 @@ int CFI_allocate(CFI_cdesc_t *descriptor, const CFI_index_t lower_bounds[],
 }
 
 int CFI_deallocate(CFI_cdesc_t *descriptor) {
+  if (descriptor == nullptr) {
+    return CFI_INVALID_DESCRIPTOR;
+  }
   if (descriptor->version != CFI_VERSION) {
     return CFI_INVALID_DESCRIPTOR;
   }
@@ -157,6 +163,9 @@ int CFI_establish(CFI_cdesc_t *descriptor, void *base_addr,
   if (type < CFI_type_signed_char || type > CFI_type_struct) {
     return CFI_INVALID_TYPE;
   }
+  if (descriptor == nullptr) {
+    return CFI_INVALID_DESCRIPTOR;
+  }
   std::size_t minElemLen{MinElemLen(type)};
   if (minElemLen > 0) {
     elem_len = minElemLen;
@@ -204,8 +213,10 @@ int CFI_section(CFI_cdesc_t *result, const CFI_cdesc_t *source,
   CFI_index_t extent[CFI_MAX_RANK];
   CFI_index_t actualStride[CFI_MAX_RANK];
   CFI_rank_t resRank{0};
-  char *shiftedBaseAddr{static_cast<char *>(source->base_addr)};
 
+  if (result == nullptr || source == nullptr) {
+    return CFI_INVALID_DESCRIPTOR;
+  }
   if (source->rank == 0) {
     return CFI_INVALID_RANK;
   }
@@ -223,6 +234,7 @@ int CFI_section(CFI_cdesc_t *result, const CFI_cdesc_t *source,
     return CFI_ERROR_BASE_ADDR_NULL;
   }
 
+  char *shiftedBaseAddr{static_cast<char *>(source->base_addr)};
   bool isZeroSized{false};
   for (int j{0}; j < source->rank; ++j) {
     const CFI_dim_t &dim{source->dim[j]};
@@ -269,6 +281,9 @@ int CFI_section(CFI_cdesc_t *result, const CFI_cdesc_t *source,
 
 int CFI_select_part(CFI_cdesc_t *result, const CFI_cdesc_t *source,
     std::size_t displacement, std::size_t elem_len) {
+  if (result == nullptr || source == nullptr) {
+    return CFI_INVALID_DESCRIPTOR;
+  }
   if (result->rank != source->rank) {
     return CFI_INVALID_RANK;
   }
@@ -302,6 +317,9 @@ int CFI_select_part(CFI_cdesc_t *result, const CFI_cdesc_t *source,
 
 int CFI_setpointer(CFI_cdesc_t *result, const CFI_cdesc_t *source,
     const CFI_index_t lower_bounds[]) {
+  if (result == nullptr) {
+    return CFI_INVALID_DESCRIPTOR;
+  }
   if (result->attribute != CFI_attribute_pointer) {
     return CFI_INVALID_ATTRIBUTE;
   }
--- a/runtime/ISO_Fortran_binding.cc
+++ b/runtime/ISO_Fortran_binding.cc
@@ -20,6 +20,13 @@
 namespace Fortran::ISO {
 extern "C" {
 
+static inline constexpr bool IsCharacterType(CFI_type_t ty) {
+  return ty == CFI_type_char;
+}
+static inline constexpr bool IsAssumedSize(const CFI_cdesc_t *dv) {
+  return dv->rank > 0 && dv->dim[dv->rank - 1].extent == -1;
+}
+
 void *CFI_address(
     const CFI_cdesc_t *descriptor, const CFI_index_t subscripts[]) {
   auto p = reinterpret_cast<char *>(descriptor->base_addr);
@@ -55,7 +62,7 @@ int CFI_allocate(CFI_cdesc_t *descriptor, const CFI_index_t lower_bounds[],
       descriptor->type > CFI_type_struct) {
     return CFI_INVALID_TYPE;
   }
-  if (descriptor->type != CFI_type_char) {
+  if (!IsCharacterType(descriptor->type)) {
     elem_len = descriptor->elem_len;
     if (elem_len <= 0) {
       return CFI_INVALID_ELEM_LEN;
@@ -203,10 +210,6 @@ int CFI_is_contiguous(const CFI_cdesc_t *descriptor) {
   return 1;
 }
 
-static inline bool IsAssumedSize(const CFI_cdesc_t *dv) {
-  return dv->rank > 0 && dv->dim[dv->rank - 1].extent == -1;
-}
-
 int CFI_section(CFI_cdesc_t *result, const CFI_cdesc_t *source,
     const CFI_index_t lower_bounds[], const CFI_index_t upper_bounds[],
     const CFI_index_t strides[]) {
@@ -297,7 +300,7 @@ int CFI_select_part(CFI_cdesc_t *result, const CFI_cdesc_t *source,
     return CFI_INVALID_DESCRIPTOR;
   }
 
-  if (result->type != CFI_type_char) {
+  if (!IsCharacterType(result->type)) {
     elem_len = result->elem_len;
   }
   if (displacement + elem_len > source->elem_len) {
--- a/runtime/ISO_Fortran_binding.cc
+++ b/runtime/ISO_Fortran_binding.cc
@@ -39,7 +39,7 @@ int CFI_allocate(CFI_cdesc_t *descriptor, const CFI_index_t lower_bounds[],
   if (descriptor->attribute != CFI_attribute_allocatable &&
       descriptor->attribute != CFI_attribute_pointer) {
     // Non-interoperable object
-    return CFI_INVALID_DESCRIPTOR;
+    return CFI_INVALID_ATTRIBUTE;
   }
   if (descriptor->attribute == CFI_attribute_allocatable &&
       descriptor->base_addr != nullptr) {
@@ -52,7 +52,7 @@ int CFI_allocate(CFI_cdesc_t *descriptor, const CFI_index_t lower_bounds[],
       descriptor->type > CFI_type_struct) {
     return CFI_INVALID_TYPE;
   }
-  if (descriptor->type != CFI_type_cptr) {
+  if (descriptor->type != CFI_type_char) {
     elem_len = descriptor->elem_len;
     if (elem_len <= 0) {
       return CFI_INVALID_ELEM_LEN;
@@ -133,6 +133,7 @@ static constexpr std::size_t MinElemLen(CFI_type_t type) {
     minElemLen = 2 * sizeof(long double);
     break;
   case CFI_type_Bool: minElemLen = 1; break;
+  case CFI_type_cptr: minElemLen = sizeof(void *); break;
   }
   return minElemLen;
 }
--- a/runtime/ISO_Fortran_binding.cc
+++ b/runtime/ISO_Fortran_binding.cc
@@ -195,7 +195,7 @@ int CFI_is_contiguous(const CFI_cdesc_t *descriptor) {
 }
 
 static inline bool IsAssumedSize(const CFI_cdesc_t *dv) {
-  return (dv->rank > 0 && dv->dim[dv->rank - 1].extent == -1);
+  return dv->rank > 0 && dv->dim[dv->rank - 1].extent == -1;
 }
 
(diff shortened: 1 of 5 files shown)


runtime/ISO_Fortran_binding.cc
```diff
@@ -193,20 +199,150 @@ int CFI_is_contiguous(const CFI_cdesc_t *descriptor) {
...
+}
+
 int CFI_section(CFI_cdesc_t *result, const CFI_cdesc_t *source,
     const CFI_index_t lower_bounds[], const CFI_index_t upper_bounds[],
     const CFI_index_t strides[]) {
-  return CFI_INVALID_DESCRIPTOR;  // TODO
+  CFI_index_t extent[CFI_MAX_RANK];
+  CFI_index_t actualStride[CFI_MAX_RANK];
+  CFI_rank_t resRank{0};
+  char *shiftedBaseAddr{reinterpret_cast<char *>(source->base_addr)};
```
tskeith: I think reinterpret_cast should only be used when absolutely necessary. In this case static_cast is fine.

runtime/ISO_Fortran_binding.cc
```diff
@@ -193,20 +199,150 @@ int CFI_is_contiguous(const CFI_cdesc_t *descriptor) {
...
+    resRank += (stride != 0);
+  }
+  if (resRank != result->rank) {
+    return CFI_INVALID_DESCRIPTOR;
+  }
+
+  // For zero-sized arrays, base_addr is proc-dep (see 18.5.3).
+  // We keep it on the source base_addr
+  result->base_addr = isZeroSized ? source->base_addr
+                                  : reinterpret_cast<void *>(shiftedBaseAddr);
```
tskeith: Here the cast isn't needed at all. But if you think it makes it clearer, use static_cast.

test/evaluate/ISO-Fortran-binding.cc
```diff
@@ -210,14 +210,469 @@ int main() {
...
+      CFI_MAX_RANK, extents);
+  for (CFI_index_t *lowerBounds : lowerBoundCases) {
+    EstablishLowerBounds(dv, lowerBounds);
+    for (CFI_type_t type : validTypeCases) {
+      for (bool contiguous : {true, false}) {
+        std::size_t size = ByteSize(type, 12);
+        dv->elem_len = size;
+        for (int i{0}; i < dv->rank; ++i) {
+          // contiguous
+          dv->dim[i].sm = size;
```
tskeith: Is this line meant to still be here?

test/evaluate/ISO-Fortran-binding.cc
```diff
@@ -358,23 +358,17 @@ static void check_CFI_allocate(CFI_cdesc_t *dv,
...
-    expectedRetCode = CFI_INVALID_DESCRIPTOR;  // current implem
+    expectedRetCode = CFI_INVALID_ATTRIBUTE;
   }
   if (rank > 0 && (lower_bounds == nullptr || upper_bounds == nullptr)) {
     ++numErr;
     expectedRetCode = CFI_INVALID_EXTENT;
   }
 
-  // memory allocation is the only remaining unpredicatble failure
+  // Memory allocation failures are unpredicatble in this test.
```
tskeith: Typo "unpredictable".

test/evaluate/ISO-Fortran-binding.cc
```diff
@@ -210,14 +210,463 @@ int main() {
...
+static void check_CFI_address(
+    const CFI_cdesc_t *dv, const CFI_index_t subscripts[]) {
+  // 18.5.5.2
+  void *addr = CFI_address(dv, subscripts);
+  const Descriptor *desc = reinterpret_cast<const Descriptor *>(dv);
+  void *addrCheck = desc->Element<void>(subscripts);
+  MATCH(true, addr == addrCheck);
+}
+
+// Helper function to set lower bound of decsriptor
```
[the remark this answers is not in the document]
jeanPerier: Thanks, I have turned on spell checks on vim.

runtime/ISO_Fortran_binding.cc
jeanPerier: No, I do not think it adds a lot, I have removed it that makes it more readable.  Thanks

test/evaluate/ISO-Fortran-binding.cc
jeanPerier: No, I have removed it.

test/evaluate/ISO-Fortran-binding.cc
jeanPerier: corrected

test/evaluate/ISO-Fortran-binding.cc
```diff
@@ -210,14 +210,463 @@ int main() {
...
+    MATCH(source->dim[j].sm, result->dim[j].sm);
+    MATCH(lower_bounds[j], result->dim[j].lower_bound);
+  }
+}
+
+int main() {
+  TestCdescMacroForAllRanksSmallerThan<CFI_MAX_RANK>();
+  run_CFI_establish_tests();
+  run_CFI_address_tests();
+  // TODO: calrify CFI_allocate -> CFI_type_cptr/CFI_type_char
```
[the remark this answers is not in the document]
jeanPerier: corrected

runtime/ISO_Fortran_binding.cc
jeanPerier: yes indeed, thanks.

test/evaluate/ISO-Fortran-binding.cc
```diff
@@ -210,14 +210,459 @@ int main() {
...
+      CFI_index_t subscripts[]{
+          result->dim[0].lower_bound + i, result->dim[1].lower_bound + j};
+      MATCH(i * 2 + j * 64,
+          *reinterpret_cast<int *>(CFI_address(result, subscripts)));
+    }
+  }
+
+  // Test for Fortran character type
+  retCode = CFI_establish(
+      result, nullptr, CFI_attribute_other, CFI_type_char, 2, rank, nullptr);
```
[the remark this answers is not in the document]
jeanPerier: Ok, I may have missed something here. Let me explain you why I believe CFI_type_char is Fortran characters for the following reasons:

- Table 18.4 (in section 18.5.4):
The table is split in four parts: The first regarding integers contains CFI_type_signed_char for C `signed char`. Then there is a part on real types and one on complex types. Last, there is a part with what I understand are non numerical types. This parts defines CFI_type_char for C `char`. These lead me to understand that CFI_type_char refers to Fortran CHARACTER when CFI_type_signed_char is INTEGER*1.
- CFI_type_cptr is associated in this table to void* that I understand as the C pointer types rather than Fortran CHARACTER. I understood it as linked to the C_PTR type described in sections 18.3.2, though I am not sure they directly relate.
- In table 18.2 (in section 18.3.1):
Fortran CHARACTER is associated to C_CHAR from ISO_C_BINDING module and to the C type `char`. Note 2 of 18.5.3 links C_CHAR, and hence `char` to Fortran CHARACTER type in the context of descriptors.
In table 18.2, the standard also refers to C `unsigned char` and `signed char` for the numerical types INTEGER*1.

Again, I may just have skipped an important paragraph. I will give it a deeper look then.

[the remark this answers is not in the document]
schweitzpgi: In 18.3.1 paragraph 1 3rd sentence, the standard mentions that the CHARACTER to C_CHAR/char mapping in table 18.2 applies if and only if the LEN type parameter is 1. Note that CHARACTER values with LEN > 1 cannot be represented in C's char type.

runtime/ISO_Fortran_binding.cc
```diff
@@ -193,20 +194,149 @@ int CFI_is_contiguous(const CFI_cdesc_t *descriptor) {
...
-  return CFI_INVALID_DESCRIPTOR;  // TODO
+  CFI_index_t extent[CFI_MAX_RANK];
+  CFI_index_t actualStride[CFI_MAX_RANK];
+  CFI_rank_t resRank{0};
+  char *shiftedBaseAddr{static_cast<char *>(source->base_addr)};
+
+  if (source->rank == 0) {
+    return CFI_INVALID_RANK;
+  }
+  if (IsAssumedSize(source) && upper_bounds == nullptr) {
```
schweitzpgi: Since the arguments are all pointers, do any of the rest of them need nullptr checks?

runtime/ISO_Fortran_binding.cc
```diff
@@ -193,20 +194,149 @@ int CFI_is_contiguous(const CFI_cdesc_t *descriptor) {
...
+  }
+
+  bool isZeroSized{false};
+  for (int j{0}; j < source->rank; ++j) {
+    const CFI_dim_t &dim = source->dim[j];
+    const CFI_index_t srcLB = dim.lower_bound;
+    const CFI_index_t srcUB = srcLB + dim.extent - 1;
+    const CFI_index_t lb = (lower_bounds != nullptr) ? lower_bounds[j] : srcLB;
+    const CFI_index_t ub = (upper_bounds != nullptr) ? upper_bounds[j] : srcUB;
+    const CFI_index_t stride = (strides != nullptr) ? strides[j] : 1;
```
schweitzpgi: I believe "var{expr}" initializer syntax is preferred to "var = expr"

test/evaluate/ISO-Fortran-binding.cc
```diff
@@ -79,8 +80,9 @@ static void AddNoiseToCdesc(CFI_cdesc_t *dv, CFI_rank_t rank) {
 }
 
 #ifdef VERBOSE
-static void DumpTestWorld(void *bAddr, CFI_attribute_t attr, CFI_type_t ty,
-    std::size_t eLen, CFI_rank_t rank, CFI_index_t *eAddr) {
+static void DumpTestWorld(const void *bAddr, CFI_attribute_t attr,
```
schweitzpgi: A void * object cannot be dereferenced, so it's not clear that adding "const" really adds much.

test/evaluate/ISO-Fortran-binding.cc
```diff
@@ -210,14 +210,459 @@ int main() {
...
+static void EstablishLowerBounds(CFI_cdesc_t *dv, CFI_index_t *sub) {
+  for (int i{0}; i < dv->rank; ++i) {
+    dv->dim[i].lower_bound = sub[i];
+  }
+}
+
+// Helper to get size without making internal compiler functions accessible
+static std::size_t ByteSize(CFI_type_t ty, std::size_t size) {
+  CFI_CDESC_T(0) storage;
+  CFI_cdesc_t *dv = reinterpret_cast<CFI_cdesc_t *>(&storage);
```
schweitzpgi: reinterpret_cast<T> should be used as a last resort.

test/evaluate/ISO-Fortran-binding.cc
```diff
@@ -210,14 +210,459 @@ int main() {
...
+  return (retCode == CFI_SUCCESS) ? dv->elem_len : 0;
+}
+
+static void run_CFI_address_tests() {
+  // Test CFI_address defined in 18.5.5.2
+  // Create test world
+  CFI_index_t extents[CFI_MAX_RANK];
+  CFI_CDESC_T(CFI_MAX_RANK) dv_storage;
+  CFI_cdesc_t *dv{reinterpret_cast<CFI_cdesc_t *>(&dv_storage)};
+  static char base;
```
schweitzpgi: Why is this static?  Can a stack address not be used?

test/evaluate/ISO-Fortran-binding.cc
```diff
@@ -210,14 +210,459 @@ int main() {
...
+      }
+    }
+  }
+}
+
+static void run_CFI_section_tests() {
+  // simple tests
+  bool testPreConditions{true};
+  constexpr CFI_index_t m = 5, n = 6, o = 7;
+  constexpr CFI_rank_t rank = 3;
```
schweitzpgi: More curly brace opportunities

test/evaluate/ISO-Fortran-binding.cc
```diff
@@ -210,14 +210,459 @@ int main() {
...
+    char name[name_len];
+  } Galaxy;
+
+  const CFI_rank_t rank{2};
+  constexpr CFI_index_t universeSize[] = {2, 3};
+  Galaxy universe[universeSize[1]][universeSize[0]];
+
+  for (int j{0}; j < universeSize[1]; ++j) {
+    for (int i{0}; i < universeSize[0]; ++i) {
+      universe[j][i].distance = i + j * 32;
```
schweitzpgi: This looks like a test, but can't these loops be interchanged, etc. to initialize the buffers in C's row-major order?

test/evaluate/ISO-Fortran-binding.cc
```diff
@@ -210,14 +210,459 @@ int main() {
...
+  }
+
+  displacement = offsetof(Galaxy, name) + 2;
+  elem_len = 2;  // not ignored this time
+  retCode = CFI_select_part(result, source, displacement, elem_len);
+  MATCH(CFI_SUCCESS, retCode);
+
+  baseAddrShiftedOk =
+      (reinterpret_cast<char *>(source->base_addr) + displacement ==
+          result->base_addr);
```
schweitzpgi: Does our f18 coding standard want bool typed expressions enclosed in outer parenthesis?

test/evaluate/ISO-Fortran-binding.cc
```diff
@@ -210,14 +210,459 @@ int main() {
...
+          reinterpret_cast<char *>(CFI_address(result, subscripts))[1]);
+    }
+  }
+}
+
+static void run_CFI_setpointer_tests() {
+  constexpr CFI_rank_t rank = 3;
+  CFI_CDESC_T(rank) resStorage, srcStorage;
+  CFI_cdesc_t *result = reinterpret_cast<CFI_cdesc_t *>(&resStorage);
+  CFI_cdesc_t *source = reinterpret_cast<CFI_cdesc_t *>(&srcStorage);
```
schweitzpgi: more curly brace syntax opportunities

test/evaluate/ISO-Fortran-binding.cc
tskeith: I prefer to minimize unnecessary parentheses.

klausler: Not that I know of.  Did `clang-format` add them?  Could it be configured to remove them?

test/evaluate/ISO-Fortran-binding.cc
klausler: Is it even necessary here?  If so, could the CFI_CDESC_T macro be tweaked to avoid it?

Also: braces, please.

test/evaluate/ISO-Fortran-binding.cc
jeanPerier: No it does not. [Google C++ style guide](https://google.github.io/styleguide/cppguide.html#Boolean_Expressions) says "_Feel free to insert extra parentheses judiciously because they can be very helpful in increasing readability when used appropriately_". I have to admit that this is probably not a judicious use. It is a personal habit, I will stop doing this.
Parenthesis removed.

runtime/ISO_Fortran_binding.cc
jeanPerier: Thanks, changed as recommended

test/evaluate/ISO-Fortran-binding.cc
jeanPerier: Currently, it is needed.
I have changed CFI_CDESC_T macro in C++ to avoid this. It now inherits from CFI_cdesc_t so that there is an implicit conversion to the base class and no cast is needed anymore.

test/evaluate/ISO-Fortran-binding.cc
jeanPerier: It can, I have removed the static keyword.

schweitzpgi: Okay. This pattern is repeated frequently, so you may want to look for more of these.

test/evaluate/ISO-Fortran-binding.cc
jeanPerier: Thanks, I went through the whole file again to fixed all similar cases.

test/evaluate/ISO-Fortran-binding.cc
jeanPerier: fixed

test/evaluate/ISO-Fortran-binding.cc
jeanPerier: Thanks, I removed them everywhere but in a function where the address is returned through the arguments. I do not like giving a stack address in this situation even if it is not supposed to be used in this case.

klausler: If you have to return the address of a static variable, please make sure that it's `const`-qualified.  That should cause the static variable to reside in a read-only page and help detect cases where user code tries to write to it.

runtime/ISO_Fortran_binding.cc
jeanPerier: No other reason than readability, but it this to be a personal opinion. I will remove them everywhere to be consistent.

klausler: The needless parentheses are not a big problem.  I think it is best to avoid adding unnecessary details like those parentheses because they can be distracting to the reader of the code, who is going to wonder why they are there since the expression is sufficiently clear without them.

But I do use parentheses that aren't strictly necessary for a correct parse of an expression when there are operations being used that are commonly misunderstood by many C programmers.  For example, I will write `x = y << (z+1);` even though addition binds more tightly than shifts do, simply to avoid a case that has led to bugs in the past.

GCC will emit warnings when parentheses are not used on `(a && b) || c` though, and I really don't like having to use them there, and do so just to get a clean build.

test/evaluate/ISO-Fortran-binding.cc
jeanPerier: It is because I wanted it to be clear that this code is initializing data related to a Fortran variable. I changed it to the usual C's row major order and I have added a comment regarding what Fortran indexes it initializes.

runtime/ISO_Fortran_binding.cc
jeanPerier: After discussing with @klausler , I have added nullptr checks for descriptor pointers in all CFI functions that return error codes. For the one that do not, I first need to make a function to crash nicely. This function would not belong only to the ISO runtime part so I need to make it somewhere else first.

runtime/ISO_Fortran_binding.cc
```diff
@@ -170,7 +187,7 @@ int CFI_establish(CFI_cdesc_t *descriptor, void *base_addr,
   descriptor->attribute = attribute;
   descriptor->f18Addendum = 0;
   std::size_t byteSize{elem_len};
-  std::size_t lower_bound{attribute != CFI_attribute_pointer};
+  const std::size_t lower_bound{0};
```
schweitzpgi: Might want to consider "constexpr"

runtime/ISO_Fortran_binding.cc
```diff
@@ -20,6 +20,13 @@
...
+static inline constexpr bool IsCharacterType(CFI_type_t ty) {
+  return ty == CFI_type_char;
+}
+static inline constexpr bool IsAssumedSize(const CFI_cdesc_t *dv) {
+  return dv->rank > 0 && dv->dim[dv->rank - 1].extent == -1;
+}
+
 void *CFI_address(
     const CFI_cdesc_t *descriptor, const CFI_index_t subscripts[]) {
   auto p = reinterpret_cast<char *>(descriptor->base_addr);
```
schweitzpgi: prefer static_cast<>

runtime/ISO_Fortran_binding.cc
```diff
@@ -196,17 +213,151 @@ int CFI_is_contiguous(const CFI_cdesc_t *descriptor) {
...
+
+  if (!IsCharacterType(result->type)) {
+    elem_len = result->elem_len;
+  }
+  if (displacement + elem_len > source->elem_len) {
+    return CFI_INVALID_ELEM_LEN;
+  }
+
+  result->base_addr = reinterpret_cast<void *>(
+      displacement + reinterpret_cast<char *>(source->base_addr));
```
schweitzpgi: Another static_cast<> opportunity.  Since the base_addr member is declared void*, it might make sense to consider adding some functions to freely cast it to char* (and back to void* for assignments).  The idea being to put functions as a firewall between using the pointer and the pointer's (required but not very amenable) declared type.

runtime/ISO_Fortran_binding.cc
jeanPerier: that fits better thanks

runtime/ISO_Fortran_binding.cc
jeanPerier: Thanks. I took the opportunity to also change the size_t types in this function because I do not see good reasons to use size_t for the rank.

runtime/ISO_Fortran_binding.cc
jeanPerier: That is a good idea, there is actually already something doing a firewall over this pointer runtime/descriptor.h. When the pointer is seen as a Descriptor object it is possible to use the member function Element<T>(offset) to get a pointer of type T.
I do not really want to use it here because I would rather stay away from the Descriptor class when implementing ISO required functions and work only with the ISO descriptor (CFI_cdesc_t). In the rest of the rest runtime, I will follow your suggestion and avoid repeating the cast.

jeanPerier:
Successfully tested to merge with master branch and passed regression tests.

PR #234, opened 2018-12-03
Refactoring of expr analysis API and implementation

klausler:
After a discussion last week with Tim, here's a refactoring and simplification of the API into expression analysis as well as the implementation of that analysis.  The old `ExprAnalyzer` class is gone, as is the `AnalyzeHelper` internal template function.  Now there is just an `AnalyzeExpr` template function with a contextual argument.  Some specializations of `AnalyzeExpr` are exposed in the header file, but most appear in the implementation code.

Further, the constraint checking necessary for expressions required by the Standard to be scalar, constant, &/or integer values is now performed in a deferred manner so that the error reporting context is more accurate.  The constraint check is part of the expression analysis context and they can be nested.

The set of changed source lines is large due to the many little tweaks made throughout the code to move to the new single template function, and exacerbated by the need to rearrange some of the utility subroutines in `expression.cc` to get their definitions ahead of their uses (which wasn't a concern before this change when they were all member functions).

### DIFF
--- a/lib/semantics/assignment.cc
+++ b/lib/semantics/assignment.cc
@@ -1,0 +1,80 @@
+// Copyright (c) 2018, NVIDIA CORPORATION.  All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "assignment.h"
+#include "expression.h"
+#include "semantics.h"
+#include "symbol.h"
+#include "../common/idioms.h"
+#include "../evaluate/expression.h"
+#include "../parser/parse-tree-visitor.h"
+#include "../parser/parse-tree.h"
+
+using namespace Fortran::parser::literals;
+
+namespace Fortran::semantics {
+
+template<typename A>
+void AnalyzeExecutableStmt(SemanticsContext &, const parser::Statement<A> &) {}
+template<>
+void AnalyzeExecutableStmt(SemanticsContext &context,
+    const parser::Statement<parser::AssignmentStmt> &stmt) {}
+template<>
+void AnalyzeExecutableStmt(SemanticsContext &context,
+    const parser::Statement<parser::PointerAssignmentStmt> &stmt) {}
+template<>
+void AnalyzeExecutableStmt(SemanticsContext &context,
+    const parser::Statement<parser::WhereStmt> &stmt) {}
+template<>
+void AnalyzeExecutableStmt(SemanticsContext &context,
+    const parser::Statement<parser::ForallStmt> &stmt) {}
+
+void AnalyzeAssignment(SemanticsContext &context,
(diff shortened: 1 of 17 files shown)


lib/semantics/expression.h
```diff
@@ -27,30 +29,100 @@ template<typename> struct Integer;
...
+  semantics::SemanticsContext &context() const { return context_; }
+
+  template<typename... A> void Say(A... args) {
+    context_.foldingContext().messages.Say(std::forward<A>(args)...);
+  }
+
+  void CheckConstraints(std::optional<Expr<SomeType>> &);
+  bool ScalarConstraint(Expr<SomeType> &);
+  bool ConstantConstraint(Expr<SomeType> &);
+  bool IntegerConstraint(Expr<SomeType> &);
```
sscalpone: Need logical?

klausler: Haven't yet (nor CHARACTER) but I'll deal with it now.

lib/semantics/assignment.cc
```diff
@@ -0,0 +1,80 @@
...
+  bool Pre(parser::Statement<parser::AssignmentStmt> &stmt) {
+    AnalyzeAssignment(context_, stmt);
+    return false;
+  }
+
+private:
+  SemanticsContext &context_;
+};
+
+void AnalyzeAssignments(parser::Program &program, SemanticsContext &context) {
```
tskeith: This is never called, as far as I can see. Is it just for testing?

lib/semantics/expression.h
```diff
@@ -27,30 +29,100 @@ template<typename> struct Integer;
...
+  return AnalyzeExpr(withCheck, expr.thing);
+}
+
+template<typename A>
+std::optional<Expr<SomeType>> AnalyzeExpr(
+    ExpressionAnalysisContext &context, const parser::Integer<A> &expr) {
+  ExpressionAnalysisContext withCheck{
+      context, &ExpressionAnalysisContext::ConstantConstraint};
+  return AnalyzeExpr(withCheck, expr.thing);
+}
```
tskeith: I think we also need `AnalyzeExpr(context, const common::Indirection<A> &expr)` to handle types like `parser::ScalarIntConstantExpr` that have `Indirection` in them.

lib/semantics/expression.cc
```diff
@@ -29,25 +29,53 @@
...
 
 using common::TypeCategory;
 
-using MaybeExpr = std::optional<Expr<SomeType>>;
+// Constraint checking
+void ExpressionAnalysisContext::CheckConstraints(MaybeExpr &expr) {
+  if (inner_ != nullptr) {
+    inner_->CheckConstraints(expr);
+  }
+  if (constraint_ != nullptr && expr.has_value()) {
```
tskeith: This function might be clearer if it had `if (expr.has_value()) {...}` around the whole body. At first glance it seems like the first `if` might do something on a `nullopt` expr, but I don't think it does.

lib/semantics/assignment.cc
klausler: It's the beginning of the assignment statement semantic analysis code I was writing before embarking on this refactoring job.

lib/semantics/expression.h
klausler: Ok, will hoist that one out to the header file as well.

lib/semantics/expression.cc
klausler: Ok.

tskeith: This was a bad suggestion on my part. The call to `inner_->CheckConstraints(expr)` can cause expr to be reset, so we **do** need to check `expr.has_value()` before trying to use the contained value.

PR #237, opened 2018-12-05
Use `std::optional<>::value()` where appropriate in place of `operator*`.

klausler:
Some uses of `*x` for `std::optional<> x` are better (more safely) expressed as `x.value()` when there is not a clear nearby presence test.

No performance cost was measurable on f90_correct.

Style guide was updated to encourage use of `value()` whenever presence of a value might not be obvious.

### DIFF
--- a/lib/parser/parse-tree-visitor.h
+++ b/lib/parser/parse-tree-visitor.h
@@ -66,12 +66,12 @@ template<typename M> void Walk(format::IntrinsicTypeDataEditDesc &, M &);
 // Traversal of needed STL template classes (optional, list, tuple, variant)
 template<typename T, typename V>
 void Walk(const std::optional<T> &x, V &visitor) {
-  if (x) {
+  if (x.has_value()) {
     Walk(*x, visitor);
   }
 }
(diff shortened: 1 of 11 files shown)


lib/parser/parse-tree-visitor.h
```diff
@@ -66,12 +66,12 @@ template<typename M> void Walk(format::IntrinsicTypeDataEditDesc &, M &);
 // Traversal of needed STL template classes (optional, list, tuple, variant)
 template<typename T, typename V>
 void Walk(const std::optional<T> &x, V &visitor) {
-  if (x) {
+  if (x.has_value()) {
```
tskeith: I have realized that the reason I don't usually use `.has_value()` in `if` conditions is because you can't use it when declaring a variable in the condition (like with `escape` above in characters.h). So it seems more consistent never to use it.

Of course you could consistently use `.has_value()` as below, but it's not an improvement.
```
if (std::optional<char> escape{BackslashEscapeChar(ch)}; escape.has_value()) {
```

PR #238, opened 2018-12-06
Rework expression evaluation in symbol table

tskeith:
Rather than fill inserting symbols into the symbol table and evaluating expressions after name resolution, now do it at the same time. When we create or find a symbol for a Name, enter it immediately. When we encounter an expression that goes in the symbol table (bound, kind parameter, etc.), evaluate it immediately.


sscalpone:
Typo "Rather than *fill* inserting symbols...."

PR #239, opened 2018-12-07
Preserve order of type parameters

tskeith:
Type parameters were sorted by the order of the type-param-def-stmts.
But we need to preserve the order of the type-param-name-list.
The is the order of positional parameters in a derived-type-spec.

So add `paramNames` to `DerivedTypeDetails` to preserve the original
order. Using this allows us to write module files with both the
type-param-name-list and type-param-def-stmts in the original order.

Also fix a bug where a duplicate type-param-def caused a spurious
extra error. If `MakeTypeSymbol()` reports an error we should not
call `SetType()` because it will just report another error.

### DIFF
--- a/lib/semantics/mod-file.cc
+++ b/lib/semantics/mod-file.cc
@@ -176,14 +176,12 @@ void ModFileWriter::PutDerivedType(const Symbol &typeSymbol) {
   }
   PutLower(decls_ << "::", typeSymbol);
   auto &typeScope{*typeSymbol.scope()};
-  if (details.hasTypeParams()) {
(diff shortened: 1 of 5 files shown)


sscalpone:
Does the design doc say why the order is important for type-param-def-stmts (etc)?

tskeith:
We have to preserve the order of the type-param-list to correctly represent the type because if you have `type(t(1, 2, 3))` you have to know which parameter is associated with which value. They have to be written to the .mod file for the same reason. I'm not sure where that would go in a design doc.

PR #241, opened 2018-12-08
More precise locations for construct name errors; some refactoring

klausler:
While beginning work on FORALL and WHERE constructs, I noticed that locations on construct name errors could be made more precise (highlighting the names themselves rather than entire statements).  So I went and made them so, doing some refactoring along the way to abstract some common code patterns in resolve-labels.cc and reduce code.

### DIFF
--- a/lib/semantics/resolve-labels.cc
+++ b/lib/semantics/resolve-labels.cc
@@ -14,10 +14,12 @@
 
 #include "resolve-labels.h"
 #include "../common/enum-set.h"
+#include "../common/template.h"
 #include "../parser/message.h"
 #include "../parser/parse-tree-visitor.h"
 #include <cctype>
 #include <cstdarg>
+#include <type_traits>
 
 namespace Fortran::semantics {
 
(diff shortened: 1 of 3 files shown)


lib/semantics/resolve-labels.cc
```diff
@@ -379,118 +381,62 @@ class ParseTreeAnalyzer {
...
-      }
-    } else {
-      if (std::get<parser::Statement<parser::EndProgramStmt>>(mainProgram.t)
-              .statement.v.has_value()) {
-        errorHandler_.Say(currentPosition_,
-            parser::MessageFormattedText{
-                "END PROGRAM cannot have a program-name"_err_en_US});
+    auto &endProgramStmt{
+        std::get<parser::Statement<parser::EndProgramStmt>>(mainProgram.t)
+            .statement};
```
tskeith: It seems like this could be simplified further by use GetStmtName() here too. Maybe:
```
const parser::CharBlock *endName{GetStmtName(
    std::get<parser::Statement<parser::EndProgramStmt>>(mainProgram.t))};
if (const auto &program{
    std::get<std::optional<parser::Statement<parser::ProgramStmt>>>(mainProgram.t)}) {
  const parser::CharBlock *name{GetStmtName(*program)};
...
```

klausler: Good idea, and I think that it can be even further simplified.

klausler:
Last call for comments

PR #247, opened 2018-12-17
Support character types

tskeith:
The real functional change is in the 2nd commit. The other two are mostly reorganizing and refactoring.

### DIFF
--- a/lib/semantics/resolve-names.cc
+++ b/lib/semantics/resolve-names.cc
@@ -2348,8 +2348,7 @@ Symbol &DeclarationVisitor::DeclareObjectEntity(
 
 void DeclarationVisitor::Post(const parser::IntrinsicTypeSpec::Character &x) {
   if (!charInfo_.length) {
-    charInfo_.length = ParamValue{SomeExpr{
-        evaluate::AsExpr(evaluate::Constant<evaluate::SubscriptInteger>{1})}};
+    charInfo_.length = ParamValue{1};
   }
   SetDeclTypeSpec(currScope().MakeCharacterType(
       std::move(*charInfo_.length), charInfo_.kind));
@@ -2370,8 +2369,7 @@ void DeclarationVisitor::Post(const parser::TypeParamValue &x) {
 }
 void DeclarationVisitor::Post(const parser::CharLength &x) {
   if (const auto *length{std::get_if<std::int64_t>(&x.u)}) {
-    charInfo_.length = ParamValue{SomeExpr{evaluate::AsExpr(
-        evaluate::Constant<evaluate::SubscriptInteger>{*length})}};
+    charInfo_.length = ParamValue{*length};
   }
 }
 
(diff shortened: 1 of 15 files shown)


lib/semantics/resolve-names.cc
```diff
@@ -65,7 +66,7 @@ class ImplicitRules {
   void set_isImplicitNoneExternal(bool x) { isImplicitNoneExternal_ = x; }
   void set_inheritFromParent(bool x) { inheritFromParent_ = x; }
   // Get the implicit type for identifiers starting with ch. May be null.
-  std::optional<const DeclTypeSpec> GetType(char ch) const;
+  const DeclTypeSpec *GetType(char ch) const;
```
klausler: If the function's result cannot be a null pointer, please make its return type a reference.

lib/semantics/resolve-names.cc
```diff
@@ -2363,23 +2346,55 @@ Symbol &DeclarationVisitor::DeclareObjectEntity(
   return symbol;
 }
 
-void DeclarationVisitor::Post(const parser::DeclarationTypeSpec::Type &x) {
-  if (const auto *symbol{ResolveDerivedType(DeclTypeSpec::TypeDerived)}) {
-    GetDeclTypeSpec()->derivedTypeSpec().set_scope(*symbol->scope());
+void DeclarationVisitor::Post(const parser::IntrinsicTypeSpec::Character &x) {
+  if (!charInfo_.length) {
+    charInfo_.length = ParamValue{SomeExpr{
+        evaluate::AsExpr(evaluate::Constant<evaluate::SubscriptInteger>{1})}};
```
klausler: Could also be written as `evaluate::Expr<SubscriptInteger>{1}` if you think that would be more clear.

lib/semantics/type.cc
```diff
@@ -111,7 +111,7 @@ std::ostream &operator<<(std::ostream &o, const ParamValue &x) {
 IntrinsicTypeSpec::IntrinsicTypeSpec(TypeCategory category, int kind)
   : category_{category}, kind_{kind} {
   CHECK(category != TypeCategory::Derived);
-  CHECK(kind > 0);
+  CHECK(kind >= 0);
```
klausler: Is the significance of kind==0 described somewhere?  Does it mean "default kind for this category"?  If so, the default kinds for each category should be available to you now as part of the semantics context.

lib/semantics/resolve-names.cc
tskeith: It can be null. The change was to return the DeclTypeSpec by reference rather than by value, but it's still optional.

lib/semantics/resolve-names.cc
tskeith: Yes, I've changed it, and added a way to construct ParamValue directly from an int.

lib/semantics/type.cc
tskeith: I've changed it so the kind is always explicit in IntrinsicTypeSpec and in what is written to the .mod file.

Issue #242, opened 2018-12-10
intrinsics-test build failure gcc-8.2/macOS


pjh40:
Trying to link test/evaluate/intrinsics-test on a gcc-8.2/macOS build, I receive a bunch of duplicate symbol errors of the form:
```
duplicate symbol Fortran::evaluate::ExpressionBase<Fortran::evaluate::Type<(Fortran::common::TypeCategory)0, 1> >::derived() in:
    lib/evaluate/libFortranEvaluate.a(expression.cc.o)
    lib/evaluate/libFortranEvaluate.a(fold.cc.o)
```

klausler:
Is this "issue" still a problem?

PR #248, opened 2018-12-20
Define non-inline versions of ExpressionBase::derived()

pjh40:
Addresses issue #242 by making non-inline definitions for the two versions of `ExpressionBase::derived()` declared in `lib/evaluate/expression.h`.  GCC-8.2 on macOS was choosing to create non-inline instances of derived() during the explicit instantiations of `ExpressionBase` found in `expression.cc` and `fold.cc`.  During the linking phase of any executable, the linker failed when it found these duplicate definitions.

While this solution works, it removes the opportunity to inline the trivial derived() functions.  Another solution would be to make all of the templates related to `ExpressionBase` in `expression.cc` and `fold.cc` available in a single .cc file, where the explicit instantiation `FOR_EACH_TYPE_AND_KIND(template class ExpressionBase)` is done once. This approach would allow inlining, but would require something like template implementation headers that could be included into the instantiation .cc file.

### DIFF
--- a/lib/evaluate/expression.cc
+++ b/lib/evaluate/expression.cc
@@ -189,6 +189,7 @@ template<typename T> DynamicType ArrayConstructor<T>::GetType() const {
   return result.GetType();
 }
 
+#if defined(__APPLE__) && defined(__GNUC__)
 template<typename A>
 typename ExpressionBase<A>::Derived &ExpressionBase<A>::derived() {
   return *static_cast<Derived *>(this);
@@ -198,6 +199,7 @@ template<typename A>
 const typename ExpressionBase<A>::Derived &ExpressionBase<A>::derived() const {
   return *static_cast<const Derived *>(this);
 }
+#endif
 
 template<typename A>
 std::optional<DynamicType> ExpressionBase<A>::GetType() const {
--- a/lib/evaluate/expression.cc
+++ b/lib/evaluate/expression.cc
@@ -190,6 +190,16 @@ template<typename T> DynamicType ArrayConstructor<T>::GetType() const {
 }
 
 template<typename A>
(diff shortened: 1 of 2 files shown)


pjh40:
I did try inline (and constexpr) to no avail.  I'm not entirely convinced that it is a compiler bug:

1. `inline` is always up to the compiler, so it is free to generate an explicit instance;
1. `private` is an attribute only within the compiler: private class members still have global visibility in the object file unless you take some explicit steps to change that;
1. multiple explicit instantiations seems rather suspect to me.  I think that is the real issue.

I'm going to play around with a test case in clang and report back...

klausler:
The bug is that an (implicitly) `inline` member function has been made externally visible.  This is independent of whether its calls are expanded inline or not.  Member functions whose bodies appear in a class definition are implicitly `inline` in C++ and should never be externalized.

Class member accessibility has nothing to do with it, I think that's a red herring.

tskeith:
It builds successfully for me on MacOS with clang 7.0.2 without this change, so that also suggests it might be a compiler bug.

I would like to get this change in surrounded by `#if defined(__APPLE__) && defined(__GNUC__)` as @klausler suggests. @pjh40, is that okay with you? Would you like me to make the change?

pjh40:
I tend to believe that it is some problem with the MacPorts (a package manager) gcc install and how it interacts with the linker.  I'll put the `#if` guards in as suggested by @tskeith, just so that I can test it out on the broken system.  

BTW, Tim, where are you getting a working macOS clang from?  I've had problems with the runtime library that prevent me from compiling f18 with it.  Honestly, if I had a working clang, I may just withdraw this patch, as I don't want to introduce cruft that may just be from a broken developer environment.

pjh40:
Okay, good to get confirmation that you also saw the gcc problem.  The current PR branch should resolve it for you.

PR #250, opened 2018-12-29
Change expression constraint checking

klausler:
Back out the recent change for deferred expression constraint checking, revert to more straightforward model with improved error locations.

### DIFF
--- a/lib/semantics/expression.h
+++ b/lib/semantics/expression.h
@@ -66,9 +66,7 @@ template<typename A> CharBlock FindSourceLocation(const A &x) {
 namespace Fortran::evaluate {
 class ExpressionAnalysisContext {
 public:
-  ExpressionAnalysisContext(semantics::SemanticsContext &sc) : context_{sc} {}
-  ExpressionAnalysisContext(ExpressionAnalysisContext &i)
-    : context_{i.context_}, inner_{&i} {}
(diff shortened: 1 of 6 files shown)


lib/semantics/expression.h
```diff
@@ -18,18 +18,34 @@
 #include "semantics.h"
 #include "../common/indirection.h"
 #include "../evaluate/expression.h"
+#include "../evaluate/tools.h"
 #include "../evaluate/type.h"
+#include "../parser/parse-tree-visitor.h"
+#include "../parser/parse-tree.h"
```
tskeith: It would be nice if we could avoid including parse-tree.h (and parse-tree-visitor.h) in other .h files because it is large and has an impact on compile times. In this case the only source file that newly depends on it is type.cc, so it may not be a big deal, but that may change over time.

klausler: Agreed.  The options as I saw them were to (1) instantiate the `AnalyzeExpr` template in `expression.h` using `extern template` on every concrete parse tree type that might get passed in, (2) avoid the use of the parse tree visitor in favor of a more specialized suite of mutually recursive template functions that dig out source locations, or (3) just use the visitor.  But maybe I can push the visitor down into `expression.cc`; will try.

PR #253, opened 2019-01-07
Symbol changes for PDT instantiation

klausler:
These are all of the changes to `lib/semantics/symbol.{h,cc}` needed in order to implement instantiation of parameterized derived types.  Changes to other files in `lib/semantics` necessary to build & pass tests have been included as necessary but are incomplete.  Changes to folding that were withheld from the previous PR because they depended on these symbol table changes are also included.

Some places in the symbol table that admitted generic expressions have been restricted to INTEGER expressions (of any kind).

A new predicate, `Symbol::IsDescriptor()`, is included.  It is intended to be true when a symbol denotes an object that at runtime will have (i.e., will *be*) a descriptor.

Other changes will follow to complete the implementation of PDTs in types, scopes, and name resolution.  When complete, these changes will make `DerivedTypeSpec` represent each (potentially) distinct instantiation of a derived type in a scope.  The scope associated with each of those `DerivedTypeSpec` instances will (in the case of a PDT) be a clone of the original abstract derived type's scope in which each appearance of a type parameter will have been replaced with the actual parameter expression's value.  This change includes the implementation of `Symbol::Instantiate` which is the member function that clones a symbol in a new scope.

The expected results of some tests have been tweaked.

### DIFF
--- a/lib/evaluate/fold.cc
+++ b/lib/evaluate/fold.cc
@@ -187,6 +187,21 @@ Expr<Type<TypeCategory::Integer, KIND>> FoldOperation(
     return Expr<IntKIND>{TypeParamInquiry<KIND>{
         FoldOperation(context, std::move(*component)), inquiry.parameter}};
   }
+  if (context.pdtInstance != nullptr &&
+      std::get<const Symbol *>(inquiry.u) == nullptr) {
+    // "bare" type parameter: replace with actual value
+    const semantics::Scope *scope{context.pdtInstance->scope()};
+    CHECK(scope != nullptr);
+    const semantics::Symbol *symbol{scope->FindSymbol(inquiry.parameter)};
+    CHECK(symbol != nullptr);
+    const auto *details{symbol->detailsIf<semantics::TypeParamDetails>()};
+    CHECK(details != nullptr);
+    CHECK(details->init().has_value());
+    Expr<SomeInteger> expr{*details->init()};
+    return Fold(context,
+        Expr<IntKIND>{
+            Convert<IntKIND, TypeCategory::Integer>(std::move(expr))});
+  }
   return Expr<IntKIND>{std::move(inquiry)};
 }
 
--- a/lib/evaluate/fold.cc
+++ b/lib/evaluate/fold.cc
@@ -1,4 +1,4 @@
-// Copyright (c) 2018-2019-2019, NVIDIA CORPORATION.  All rights reserved.
+// Copyright (c) 2018-2019, NVIDIA CORPORATION.  All rights reserved.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
--- a/lib/evaluate/fold.cc
+++ b/lib/evaluate/fold.cc
@@ -1,4 +1,4 @@
-// Copyright (c) 2018-2019, NVIDIA CORPORATION.  All rights reserved.
+// Copyright (c) 2018-2019-2019, NVIDIA CORPORATION.  All rights reserved.
 //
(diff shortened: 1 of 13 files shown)


lib/evaluate/fold.cc
```diff
@@ -1,4 +1,4 @@
-// Copyright (c) 2018-2019, NVIDIA CORPORATION.  All rights reserved.
+// Copyright (c) 2018-2019-2019, NVIDIA CORPORATION.  All rights reserved.
```
tskeith: Something went wrong here.

klausler: Sure did!  Thanks

test/semantics/modfile12.f90
```diff
@@ -48,8 +48,8 @@ subroutine bar(x)
 !  integer(4),parameter::n=3_4
 !  integer(4),parameter::l=-3_4
 !  real(4),parameter::e=3.333333432674407958984375e-1_4
-!  real(4)::x(1_4:13_8)
-!  real(4)::y(1_8:8_4)
+!  real(4)::x(int(1_4,kind=8):13_8)
```
tskeith: Can `int(1_4,kind=8)` be folded to `1_8`?

test/semantics/symbol06.f90
```diff
@@ -97,7 +97,7 @@ subroutine s1
  i = x%t1
  !REF: /s1/i
  !REF: /s1/x
- !DEF: /s1/t3/t2 ObjectEntity TYPE(t2)
+ !DEF: /s1/t3/t2 ObjectEntity TYPE(t1)
```
tskeith: I don't understand this change. t3 has an implicit member named t2 whose type is type(t2).

lib/semantics/symbol.cc
```diff
@@ -56,19 +57,66 @@ void ObjectEntityDetails::set_type(const DeclTypeSpec &type) {
...
+              if (const auto *details{param->detailsIf<TypeParamDetails>()}) {
+                if (details->attr() == common::TypeParamAttr::Len) {
+                  return true;
+                }
+              }
+            }
+          }
+        }
+      }
+    } else if (type_->category() == DeclTypeSpec::Category::ClassDerived ||
```
tskeith: ClassDerived will go into the previous branch of the if. AsDerived returns non-null in that case.

klausler: Right, I didn't remove this case when I extended it.

test/semantics/symbol06.f90
klausler: t2 is the local alias for the use-associated type name t1, and the code that adds the parent component to t3 ends up working with the fully resolved type rather than the local name.

test/semantics/modfile12.f90
klausler: Will do.

klausler: Done.

lib/semantics/resolve-names.cc
```diff
@@ -2834,14 +2852,14 @@ Symbol *DeclarationVisitor::FindExplicitInterface(const parser::Name &name) {
...
       return symbol;
     }
-    const auto *extends{scope->symbol()->get<DerivedTypeDetails>().extends()};
-    if (!extends) {
+    if (const Symbol * parent{scope->symbol()->GetParent()}) {
+      scope = parent->scope();
+    } else {
       return nullptr;
     }
-    scope = extends->scope();
```
tskeith: I think this was clearer with the "increment" expression as the last statement of the body of the loop.

klausler: OK, will revert.

PR #256, opened 2019-01-08
More name resolution

tskeith:
Resolve names in BIND and PASS attributes.
Resolve %kind and %len on intrinsic types.


tskeith:
My goal is to have every `parser::Name` have a resolved symbol, assuming there are no errors in the program. This way we can detect failure to resolve some. So I want to create symbols for %kind, %len, %re, %im. When evaluating expressions it shouldn't be necessary to do string comparisons on these again, just check the `MiscDetails::Kind` in the symbol.

klausler:
Understood.  I'll work through the merge with my work in progress.

PR #257, opened 2019-01-09
FIR draft

schweitzpgi:


lib/fir/afforestation.h
```diff
@@ -0,0 +1,33 @@
+// Copyright (c) 2018, NVIDIA CORPORATION.  All rights reserved.
```
klausler: 2019

lib/fir/fir-basicblock.h
```diff
@@ -0,0 +1,71 @@
...
+  void insertBefore(Statement *stmt, Statement *before = nullptr);
+  static BasicBlock *Create(
+      Region *parentRegion, BasicBlock *insertBefore = nullptr);
+  const Statement *getTerminator() const;
+  Statement *getTerminator();
+  void SetRegion(Region *region);
+  void addPred(BasicBlock *bb);
+
+private:
+  StatementListType statementList;
```
klausler: statementList_;

lib/fir/fir-builder.h
```diff
@@ -0,0 +1,68 @@
...
+#include <initializer_list>
+
+namespace Fortran::IR {
+
+struct FIRBuilder {
+  explicit FIRBuilder(BasicBlock &block)
+    : cursorRegion{block.getParent()}, cursorBlock{&block} {}
+  template<typename A> Statement &Insert(A *s);
+  template<typename A> Statement &InsertTerminator(A *s);
+  void SetInsertionPoint(BasicBlock *bb);
```
klausler: Does this set the "insertion point" to the end of the block?  Would it precede any terminating statement?  Can one set the insertion point before/after any other statement?

lib/fir/fir-builder.h
```diff
@@ -0,0 +1,68 @@
...
+  void ClearInsertionPoint();
+  BasicBlock *GetInsertionPoint() const;
+  Statement &createDummyStmt(const parser::CharBlock &pos);
```
klausler: These names should be capitalized.

lib/fir/fir-builder.h
```diff
@@ -0,0 +1,68 @@
...
+  Statement &createSwitch(const A &condition, BasicBlock *defaultCase,
+      const SwitchStmt::ValueSuccPairListType &rest);
+  void PushBlock(BasicBlock *block);
+  BasicBlock *PopBlock();
+  void dump() const;
+  void SetCurrentRegion(Region *region);
+  Region *GetCurrentRegion() const;
+
+private:
+  Region *cursorRegion;
```
klausler: _

lib/fir/fir-basicblock.h
```diff
@@ -0,0 +1,71 @@
...
+#ifndef FORTRAN_IR_BASICBLOCK_H_
+#define FORTRAN_IR_BASICBLOCK_H_
+
+namespace Fortran::IR {
+class BasicBlock;
+class Statement;
+struct Stmt_impl;
+}
+
+#include "fir-region.h"
```
klausler: Why `"fir-region.h"` and not `<fir-region.h>`?  Are the quoted header names more in the LLVM style?

lib/fir/fir-basicblock.h
```diff
@@ -0,0 +1,71 @@
...
+#include "mixin.h"
+#include "statements.h"
+#include "llvm/ADT/ilist.h"
+#include <iostream>
```
klausler: We can't include iostream, at least not permanently.

lib/fir/fir-basicblock.h
```diff
@@ -0,0 +1,71 @@
+// Copyright (c) 2018, NVIDIA CORPORATION.  All rights reserved.
```
klausler: You probably don't need to prefix the names of the files in `fir/` with `fir-`.

lib/fir/statements.h
```diff
@@ -0,0 +1,231 @@
...
+  static constexpr unsigned TrueIndex{0u};
+  static constexpr unsigned FalseIndex{1u};
+  Expression *condition;
+  BasicBlock *succs[2];
+};
+
+struct SwitchStmt : public TerminatorStmt_impl {
+  using ValueSuccPairType = std::pair<Expression *, BasicBlock *>;
+  using ValueSuccPairListType = std::vector<ValueSuccPairType>;
+  ~SwitchStmt() = default;
```
klausler: Why the explicitly defaulted destructors?  Aren't the destructors defaulted anyway?

lib/fir/statements.h
```diff
@@ -0,0 +1,231 @@
...
+  ~UnreachableStmt() = default;
+
+private:
+  explicit UnreachableStmt() = default;
+};
+
+struct ActionStmt_impl : public Stmt_impl {
+  using ActionTrait = std::true_type;
+  // TODO: DynamicType is a placeholder for now
+  std::optional<evaluate::DynamicType> type_{std::nullopt};
```
klausler: `std::optional<>` items are null by default.

lib/fir/statements.h
```diff
@@ -0,0 +1,231 @@
...
+};
+
+struct AssignmentStmt
```
klausler: It would be great if assignments could be templated on their types like expressions are.

lib/fir/statements.h
```diff
@@ -0,0 +1,231 @@
...
+  friend Statement;
+  ~DeallocateStmt() = default;
+  static DeallocateStmt *Create(const parser::DeallocateStmt *s);
+
+private:
+  explicit DeallocateStmt(const parser::DeallocateStmt *s) : dealloc{s} {}
+  const parser::DeallocateStmt *dealloc;
+};
+
+struct CastingStmt_impl : public ActionStmt_impl {};
```
klausler: Data conversions will be explicit operations in RHS expressions.

lib/fir/statements.h
```diff
@@ -0,0 +1,231 @@
...
+
+struct ScopeStmt_impl : public ActionStmt_impl {};
+struct ScopeEnterStmt : public ScopeStmt_impl {};
+struct ScopeExitStmt : public ScopeStmt_impl {};
+struct PHIStmt : public ActionStmt_impl {};
+struct EvalStmt : public ActionStmt_impl {};
+struct CallStmt_impl : public ActionStmt_impl {};
+
+struct CallStmt : public CallStmt_impl {
+  friend Statement;
```
klausler: Could you please add a comment or two to this file explaining the relationships between the various classes?  I don't know why `Statement` is made a `friend` of all the specific statement classes, or why `CallStmt` inherits from an empty `CallStmt_impl` class.

lib/semantics/scope.h
```diff
@@ -25,11 +25,16 @@
...
+namespace Fortran::fir {
+class ControlFlowAnalyzer;
+}
+
 namespace Fortran::semantics {
 
 using namespace parser::literals;
 
 class Scope {
+  friend fir::ControlFlowAnalyzer;
```
klausler: Maybe `Scope`'s API can be extended to serve the needs of `ControlFlowAnalyzer` without needing to expose all of its content.

lib/fir/afforestation.h
```diff
@@ -0,0 +1,33 @@
...
+
+namespace Fortran::parser {
+class Program;
+}
+
+namespace Fortran::semantics {
+class SemanticsContext;
+}
+
+namespace Fortran::IR {
```
tskeith: I think the namespace should be "fir" if the directory is.

lib/fir/afforestation.h
```diff
@@ -0,0 +1,33 @@
...
+class Program;
+
+Program *CreateFortranIR(const parser::Program &program,
```
tskeith: `CreateFIR` would be more consistent with the other names.

lib/fir/fir-procedure.h
```diff
@@ -0,0 +1,83 @@
...
+#define FORTRAN_IR_PROCEDURE_H_
+
+namespace Fortran::IR {
+class Procedure;
+struct GraphWriter;
+}
+
+#include "fir-program.h"
+#include "fir-region.h"
+#include "mixin.h"
```
tskeith: `mixin.h` isn't in this PR.

lib/fir/fir-program.h
```diff
@@ -0,0 +1,58 @@
...
+  friend GraphWriter;
+  using ProcedureListType = llvm::iplist<Procedure>;
+  using ProcedureMapType = llvm::StringMap<Procedure *>;
+
+  explicit Program(llvm::StringRef id);
+  ~Program();
+  void insertBefore(Procedure *subprog, Procedure *before = nullptr);
+  ProcedureListType &getSublist(Procedure *);
+  bool containsProcedure(llvm::StringRef name);
+  std::string getName() const;
```
tskeith: We've been using `parser::CharBlock` for names. Is there any reason not to continue with that?

lib/fir/statements.h
```diff
@@ -0,0 +1,231 @@
...
+struct Evaluation
+  : public SumTypeCopyMixin<
+        std::variant<Expression *, Variable *, PathVariable *>> {
+  SUM_TYPE_COPY_MIXIN(Evaluation)
+  Evaluation(PathVariable *pv) : SumTypeCopyMixin{pv};
+  template<typename A> Evaluation(A *a) : SumTypeCopyMixin{a} {}
+  std::string dump() const;
+};
+
+struct Stmt_impl {
```
tskeith: Does `_impl` mean the struct is not meant to be part of the API? Maybe add an explanation to `C++style.md`.

lib/fir/fir-basicblock.h
```diff
@@ -0,0 +1,71 @@
...
+  using reverse_iterator = StatementListType::reverse_iterator;
+  using const_reverse_iterator = StatementListType::const_reverse_iterator;
+
+  BasicBlock(const BasicBlock &) = delete;
+  BasicBlock &operator=(const BasicBlock &) = delete;
+  ~BasicBlock();
+  StatementListType &getSublist(Statement *);
+  void insertBefore(Statement *stmt, Statement *before = nullptr);
+  static BasicBlock *Create(
+      Region *parentRegion, BasicBlock *insertBefore = nullptr);
```
tskeith: How is memory managed for pointers like the one returned here?

lib/fir/fir-program.h
klausler: CharBlock works great for names that are in the source stream but the bridge to LLVM will have to use LLVM's string representation and will most likely also have to deal with names that aren't in the original source.  But we should try to retain CharBlock names as far along as we can for names that are in the source stream so that optimization info messages can have provenance.

lib/fir/fir-basicblock.h
schweitzpgi: Follows the standard coding convention.

https://gcc.gnu.org/onlinedocs/cpp/Search-Path.html

lib/fir/fir-basicblock.h
schweitzpgi: Why not?

lib/fir/afforestation.h
schweitzpgi: OK

lib/fir/afforestation.h
schweitzpgi: I disagree. It reads as doubly redundant.

lib/fir/fir-basicblock.h
schweitzpgi: True. I borrowed from Swift.

lib/fir/fir-basicblock.h
schweitzpgi: Ownership like in LLVM.

lib/fir/fir-basicblock.h
schweitzpgi: I really dislike the useless trailing underscore.

lib/fir/fir-builder.h
schweitzpgi: Yes. Conventionally, the constructing visitation will build the IR in a lexical order. However, yes, the LLVM builder also supports repositioning the cursor.

lib/fir/fir-procedure.h
schweitzpgi: Yeah. I'm thinking it's throw-away code.

lib/fir/fir-program.h
schweitzpgi: Have to use some LLVM data structures to use other LLVM data structures.

lib/fir/statements.h
schweitzpgi: These are for abstract classes for ad hoc polymorphism and a place for shared code.

lib/fir/statements.h
schweitzpgi: I honestly don't recall now, but I might have picked them up from the SIL code? True, they don't appear to be useful here.

lib/fir/statements.h
schweitzpgi: Thanks. There is plenty of cleanup work.

lib/fir/statements.h
schweitzpgi: More vestigal stuff from LLVM.

lib/fir/statements.h
schweitzpgi: Sorry. I didn't scrub this code at all. I think this is just more left-over scraps from some on-the-fly hacking.

lib/semantics/scope.h
schweitzpgi: Yet another TODO and rubbish from prototype hacking...

lib/fir/fir-basicblock.h
klausler: [](https://llvm.org/docs/CodingStandards.html#include-iostream-is-forbidden)

It's a special case of the LLVM rule against statically allocated objects with constructors being present in libraries.

lib/fir/fir-basicblock.h
klausler: Then propose that we change the conventions being used throughout f18 for naming private data members.  If the convention is stupid, we'll remove or change it.  But please don't just ignore it.

schweitzpgi:
>  I am not clear how to drive them when (e.g.) expanding an array constructor into a sequence of assignments and loops during expression analysis.  Maybe you could walk us through the protocol.

I'm not sure I follow this question. Am I correctly understanding the implication that the semantic analysis of an expression will want to create some discrete piece of CFG?

The assumption I made was of a transformational compiler pass, where there will be a pass that traverses/visits the incoming post-semantic analysis representation and converts that to FIR.

To the specific question, the working (and entirely fluid at this point) idea is:

1. Semantic analysis will correctly identify and annotate every array assignment
2. Inject assignments into the FIR structure as generic assignment statements
3. As one of the successive lowering passes over FIR, recognize each array assignment (say, by its rank) and perform a local rewrite to modify the CFG to add the loop(s) and an element assignment statement

> Is the "fir-statement.def" file missing?

Yes. I'll see if I can add it.

klausler:
The semantic analysis of expressions generates a representation of each expression as part of its results.  This is necessary because those expressions are required by the symbol table.  But yes, semantic analysis of assignments need not emit any IR so long as the LHS & RHS expression representations are saved in their slots in the parse tree (which they are).

lib/fir/afforestation.h
tskeith: Then the directory and namespace should both be "ir". We shouldn't have two different names for the same thing.

lib/fir/fir-basicblock.h
tskeith: What does that mean?

schweitzpgi: Say you want to build a compilation unit. That's an object. You create the object and reference it with a pointer. The compilation unit has three functions and six global variables.  The program owns those nine objects, and they are referenced by pointers.  Function A has 12 basic blocks.  Those are objects that are owned by the function and referenced by pointers. Basic block "entry" has 57 instructions.  The basic block owns those instructions, which are referenced by pointers.

tskeith: In that case wouldn't it be better to use `std::unique_ptr<BasicBlock>` instead of `BasicBlock *`?

tskeith: There is a section in `C++style.md` about pointers. Another option for an owning pointer is `common::Indirection<>`, if you want it to be non-nullable.

PR #258, opened 2019-01-09
cherrypick expression changes

klausler:
These are all of the remaining changes to expression analysis and representation that have been made mostly as part of supporting parameterized derived type instantiation.  They build and pass tests independently and other work in lib/sematics that remains to be reviewed depends upon them.

(Unrelated, but included: many former template functions in lib/semantics/expression.cc that had no template parameters have been converted into regular overloaded functions.)

### DIFF
--- a/lib/semantics/expression.h
+++ b/lib/semantics/expression.h
@@ -16,6 +16,7 @@
 #define FORTRAN_SEMANTICS_EXPRESSION_H_
 
 #include "semantics.h"
+#include "../common/fortran.h"
 #include "../common/indirection.h"
 #include "../evaluate/expression.h"
 #include "../evaluate/tools.h"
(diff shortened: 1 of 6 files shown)


lib/semantics/expression.h
```diff
@@ -189,6 +204,25 @@ std::optional<Expr<SomeType>> AnalyzeExpr(
   }
   return result;
 }
+
+template<typename L, typename R>
+bool AreConformable(const L &left, const R &right) {
+  int leftRank{left.Rank()};
+  if (left.Rank() == 0) {
```
tskeith: This could be `if (leftRank == 0)`

klausler: Right.

lib/semantics/expression.h
```diff
@@ -71,15 +74,28 @@ class ExpressionAnalysisContext {
 
   semantics::SemanticsContext &context() const { return context_; }
 
+  FoldingContext &GetFoldingContext() const {
+    return context_.foldingContext();
```
tskeith: I think there is a place in resolve-names.cc that can use FoldingContext() too.

klausler: There's a couple, will do.

PR #259, opened 2019-01-10
Eas2

schweitzpgi:
Add document for Fortran IR

### DIFF
--- a/documentation/fortran-ir.md
+++ b/documentation/fortran-ir.md
@@ -101,7 +101,7 @@ While each control-flow source statement is explicit in the traversal, it can be
 
 Above, various Fortran executable constructs were discussed with respect to how they (may) give rise to control flow.  These Fortran statements are mapped to a small number of FIR statements: ReturnStmt, BranchStmt, SwitchStmt, IndirectBrStmt, and UnreachableStmt.
 
-_ReturnStmt_: execution leaves the enclosing Procedure. A ReturnStmt can return and optional value. This would appear for RETURN statements or at END SUBROUTINE.
+_ReturnStmt_: execution leaves the enclosing Procedure. A ReturnStmt can return an optional value. This would appear for RETURN statements or at END SUBROUTINE.
 
 _BranchStmt_: execution of the current basic block ends. If the branch is unconditional then control transfers to exactly one successor basic block. If the branch is conditional then control transfers to exactly one of two successor blocks depending on the true/false value of the condition. All successors must be in the current Procedure. Unconditional branches would appear for GOTO statements. Conditional branches would appear for IF constructs, IF statements, etc.
 
@@ -110,6 +110,8 @@ _SwitchStmt_: Exactly one of multiple successors is selected based on the contro
 _IndirectBrStmt_: A variable is loaded with the address of a basic block in the containing Procedure. Control is transferred to the contents of this variable. An IndirectBrStmt also requires a complete list of potential basic blocks that may be loaded into the variable. This would appear for ASSIGNED GOTO.
 
 Supporting ASSIGNED GOTO offers a little extra challenge as the ASSIGN GOTO statement's list of target labels is optional.  If that list is not present, then the procedure must be analyzed to find ASSIGN statements. The implementation proactively looks for ASSIGN statements and keeps a dictionary mapping an assigned Symbol to its set of targets. When constructing the CFG, ASSIGNED GOTOs can be processed as to potential targets either from the list provided in the ASSIGNED GOTO or from the analysis pass.
+
+Alternatively, ASSIGNED GOTO could be implemented as a _SwitchStmt_ that tests on a compiler-defined value and fully elaborates all potential target basic blocks.
 
 _UnreachableStmt_: If control reaches an unreachable statement, then an error has occurred. Calls to library routines that do not return should be followed by an UnreachableStmt.  An example would be the STOP statement.
 
--- a/documentation/fortran-ir.md
+++ b/documentation/fortran-ir.md
@@ -29,9 +29,9 @@ In Fortran, there are a number of statements that result in control flow to stat
 
 #### Structured Control Flow
 
-Fortran has executable constructs that imply three basic control flow forms.  The first form is a structured loop (DO construct)<sup>2</sup>. The second form is a structured cascade of conditional branches (IF construct, IF statement,<sup3</sup> WHERE construct).  The third form is a structured multiway branch (SELECT CASE, SELECT RANK, and SELECT TYPE constructs).  The FORALL construct, while it implies a semantic model of interleaved iterations, can be modeled as a special single-entry single-exit region in FIR perhaps with backstage marker statements.<sup>4</sup>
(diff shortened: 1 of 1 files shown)


documentation/fortran-ir.md
```diff
@@ -0,0 +1,194 @@
...
+
+### Operations
+
+The compiler's to be determined optimization passes will inform us as to the exact composition of FIR at the operations level.  This details here will necessarily change, so please read them with a grain of salt.
+
+The plan (see CFG-WP) is that statements (actions) will be a veneer model of Fortran syntactical executable constructs. Fortran statements will correspond one to one with actions. Actions will be composed of and own objects of Fortran::evaluate::GenericExprWrapper. Values of type GenericExprWrapper will have Fortran types. This implies that actions will not be in an explicit data flow representation and have optional type information.<sup>6</sup> Initially, values will bind to symbols in a context and have an implicit use-def relation. An action statement may entail a "big step" operation with many side-effects. No semantics has been defined at this time.  Actions may reference other non-executable statements from the parse tree in some to be determined manner.
```
klausler: To clarify: all types will have been resolved by the time the CFG is created, and so expressions that appear in actions will use the least general `Expr<>` specialization.  For example, predicates in `IF` statements will all be expressions that have a concrete LOGICAL type with kind=1, not a more general expression type.  Assignment actions will be templatized on type, I think, since RHS expressions will have had explicit conversion operations applied to force them to the type of the LHS.  So instances of the most generic expression type (`Expr<SomeType>`) should be pretty rare in the CFG and probably isolated to places like the operands of `STOP` and `PAUSE` statements.

documentation/fortran-ir.md
```diff
@@ -0,0 +1,194 @@
...
+* Pro: Constructing the linearized form can be done with a simple traversal of the parse tree.
+* Pro: Once composed the linearized form can be traversed and a CFG directly constructed.  This greatly reduces bookkeeping of contextual state.
+
+### Details
+
+#### Grappling with Control Flow
+
+Above, various Fortran executable constructs were discussed with respect to how they (may) give rise to control flow.  These Fortran statements are mapped to a small number of FIR statements: ReturnStmt, BranchStmt, SwitchStmt, IndirectBrStmt, and UnreachableStmt.
+
+_ReturnStmt_: execution leaves the enclosing Procedure. A ReturnStmt can return and optional value. This would appear for RETURN statements or at END SUBROUTINE.
```
klausler: s/and/an/

documentation/fortran-ir.md
```diff
@@ -0,0 +1,194 @@
...
+
+_SwitchStmt_: Exactly one of multiple successors is selected based on the control expression. Successors are pairs of case expressions and basic blocks.  If the control expression compares to the case expression and returns true, then that control transfers to that block. There may be one special block, the default block, that is selected if none of the case expressions compares true. This would appear for SELECT CASE, SELECT TYPE, SELECT RANK, COMPUTED GOTO, WRITE with exceptional condition label specificers, alternate return specifiers, etc.
+
+_IndirectBrStmt_: A variable is loaded with the address of a basic block in the containing Procedure. Control is transferred to the contents of this variable. An IndirectBrStmt also requires a complete list of potential basic blocks that may be loaded into the variable. This would appear for ASSIGNED GOTO.
```
klausler: `ASSIGNED GOTO` can be implemented as a switch, too.

documentation/fortran-ir.md
```diff
@@ -0,0 +1,194 @@
...
+
+_UnreachableStmt_: If control reaches an unreachable statement, then an error has occurred. Calls to library routines that do not return should be followed by an UnreachableStmt.  An example would be the STOP statement.
+
+#### Scope
+
+In the CFG-WP, scopes are meant to be captured by a pair of backstage statements for entering and exiting a particular scope. In structured code, these pairs would not be problematic; however, control flow in Fortran is ad hoc, particularly in legacy Fortran. In short, Fortran does not have a clean sense of structure with respect to scope.
+
+To separate concerns, FIR will construct the ad hoc CFG and impose bounding boxes over regions of that graph to demarcate and superimpose scope structures on that CFG. Any GOTO-like statements that are side-entries and side-exits to the region will be explicit.
```
klausler: Are there actually cases in Fortran where one can jump into any part of a scope besides its starting point, apart from the special weird case of `ENTRY` statements?  You can probably simplify the problem by assuming that scopes are single-entry multi-exit regions.

documentation/fortran-ir.md
```diff
@@ -0,0 +1,194 @@
...
+
+#### Structure
+
+_Program_: A program instance is the top-level object that contains the representation of all the code being compiled, the compilation unit. It contains a list of procedures and a reference to the global symbol table.
+
+_Procedure_: This is a named Fortran procedure (subroutine or function). It contains a (hierarchical) list of regions. It also owns the master list of all basic blocks for the procedure.
+
+_Region_: A region is owned by a procedure or by another region. A region owns a reference to a scope in the symbol table tree. The list of delineated basic blocks can also be requested from a region.
+
+_Basic block_: A basic block is owned by a procedure. A basic block owns a list of statements. The last statement in the list must be a terminator, and no other statement in the list can be a terminator. A basic block owns a list of its predecessors, which are also basic blocks. (Precisely, it is this level of FIR that is the CFG.)
```
klausler: Are terminators in a 1-1 correspondence with their blocks?

documentation/fortran-ir.md
schweitzpgi: Fixed, thanks.

documentation/fortran-ir.md
schweitzpgi: Yes.

documentation/fortran-ir.md
schweitzpgi: We aren't disallowing "ugly" GOTOs that jump in or out of regions to support dusty deck Fortran.

documentation/fortran-ir.md
schweitzpgi: True. I'll make a note of that.

documentation/fortran-ir.md
schweitzpgi: I believe the refined type is encoded in the type of the 'v' member of the GenericExprWrapper, right? An assignment statement's type could then simply project the type of the LHS expression.

documentation/fortran-ir.md
klausler: There's extended-range DO loops, but DO loops shouldn't constitute regions unless they're OpenMP or OpenACC constructs, and those don't support extended ranges.  Other constructs with regions (`BLOCK`, `SELECT TYPE/RANK`, `ASSOCIATE`)  come with control flow constraints that we enforce (or must, if we don't).  Can you think of a specific instance of a scope where we'd need to represent an incoming flow arc that doesn't arrive at its head?

documentation/fortran-ir.md
klausler: What I propose is an general `Assignment` action that wraps a variant whose alternatives are all (LHS, RHS) pairs of identically typed expressions.

PR #263, opened 2019-01-18
Make the trailing underscore optional.

schweitzpgi:


documentation/C++style.md
```diff
@@ -40,7 +40,7 @@ remove the inclusion before committing.
 (e.g., `clear()` and `size()` member functions in a class that implements
 a container).
 1. Non-public data members should be named with leading miniscule (lower-case)
-letters, internal camelCase capitalization, and a trailing underscore,
+letters, internal camelCase capitalization, and an optional trailing underscore,
```
klausler: If it's optional, then some classes will use it, and others will not, and the reader loses the benefit of consistency that makes a naming convention worthwhile.  We should either use the convention uniformly or discard it or replace it.

In this specific case of references to class private data members, it's helpful to the reader that they stand out as such by never appearing as a bare name whose declaration may be in another file.  Some people use `this->foo` by convention and I read `foo_` as shorthand for that usage.  If you don't want to use `foo_`, we should discuss replacing it with `this->foo`.

Issue #211, opened 2018-10-11
real-test failures on Power9


pjh40:
I suspect that this is similar to the problems reported in issue #148, but, for completeness, real-test has 480 failures on an IBM Power9 Linux system:
```
uname: Linux cn2024 4.14.0-49.10.1.el7a.ppc64le #1 SMP Thu Jul 5 17:54:38 UTC 2018 ppc64le ppc64le ppc64le GNU/Linux
GCC 8.2.0: Target: powerpc64le-unknown-linux-gnu
```

[real-fail.txt](https://github.com/flang-compiler/f18/files/2469157/real-fail.txt)

klausler:
Fixed by PR#266

pjh40:
All tests pass on Power9 release build, thanks!

Issue #148, opened 2018-07-26
[testing] real-test fails 480 test cases on AArch64 (all of them pass on x86_64)


pawosm-arm:
I've compiled f18 with system-provided gcc/g++ 7.3.0 on similarly configured Gentoo Linux systems, one x86_64, one AArch64. On x86_64, all `make test` tests are succeeding. On AArch64, one of those tests is failing, which is real-test.
```
      Start  6: Real
 6/47 Test  #6: Real .............................***Failed    3.49 sec
```
I've tried to run real-test binary manually. On x86_64 it reports success:
```
$ ./real-test
all 16805389 tests PASS
```
On AArch64 it reports 480 failed tests:
```
$ ./real-test
real.cc:345: FAIL: FlagsToBits(prod.flags) == 0x10, not 0x18
        0 0x800001 * 0xbf7ffffe
real.cc:345: FAIL: FlagsToBits(prod.flags) == 0x10, not 0x18
        0 0x800001 * 0x3f7ffffe
real.cc:345: FAIL: FlagsToBits(prod.flags) == 0x10, not 0x18
        0 0x80800001 * 0xbf7ffffe
real.cc:345: FAIL: FlagsToBits(prod.flags) == 0x10, not 0x18
        0 0x80800001 * 0x3f7ffffe
real.cc:345: FAIL: FlagsToBits(prod.flags) == 0x10, not 0x18
        0 0xbf7ffffe * 0x800001
real.cc:345: FAIL: FlagsToBits(prod.flags) == 0x10, not 0x18
        0 0xbf7ffffe * 0x80800001
real.cc:345: FAIL: FlagsToBits(prod.flags) == 0x10, not 0x18
        0 0x3f7ffffe * 0x800001
real.cc:345: FAIL: FlagsToBits(prod.flags) == 0x10, not 0x18
        0 0x3f7ffffe * 0x80800001
real.cc:345: FAIL: FlagsToBits(prod.flags) == 0x10, not 0x18
        0 0x1000001 * 0xbefffffe
real.cc:345: FAIL: FlagsToBits(prod.flags) == 0x10, not 0x18
        0 0x1000001 * 0x3efffffe
real.cc:345: FAIL: FlagsToBits(prod.flags) == 0x10, not 0x18
        0 0x81000001 * 0xbefffffe
real.cc:345: FAIL: FlagsToBits(prod.flags) == 0x10, not 0x18
        0 0x81000001 * 0x3efffffe
[...]

PR #266, opened 2019-01-22
Make x86 fp flag edge case behavior emulation conditional.

klausler:
Fixes issue #211 and probably #148 (but that's unconfirmed).

Some code in the floating-point evaluation library unconditionally emulates an x86 edge case behavior in which the Underflow flag isn't always set when an inexact product of a multiply rounds an subnormal up to a normal number.  This PR makes that behavior conditional on a new "x86 behavior emulation" flag in an expanded `Rounding` struct.  All tests now pass on OpenPower and maybe this fix is sufficient for ARM.

### DIFF
--- a/lib/evaluate/common.h
+++ b/lib/evaluate/common.h
@@ -118,9 +118,22 @@ template<typename A> struct ValueWithRealFlags {
   RealFlags flags;
 };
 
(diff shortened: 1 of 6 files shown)


sscalpone:
The PGI buildbot for OpenPOWER is reporting green now.  (It had been yellow because of this issue.)

PR #271, opened 2019-01-29
Semantic analysis & folding of array constructors

klausler:
Works for intrinsic types except CHARACTER, should be enough to let Jean get started on intrinsic folding of arrays.  Includes fixes to semantic analysis of statement entities (implied DO loop indices), distinguishing them from construct entities, and cleaning up those (distinguishing concurrent indices from locality specs).

Some bugs & missing features encountered during development and testing were addressed.  One was correcting ambiguous parses of substrings that look like array sections or elements (e.g., `A(1:2)`) during expression semantics.


lib/evaluate/fold.cc
```diff
@@ -209,6 +212,102 @@ Expr<T> FoldOperation(FoldingContext &context, Designator<T> &&designator) {
...
+    } else {
+      return Expr<T>{std::move(array)};
+    }
+  }
+
+private:
+  bool FoldArray(const CopyableIndirection<Expr<T>> &expr) {
+    Expr<T> folded{Fold(context_, Expr<T>{*expr})};
+    if (auto *c{UnwrapExpr<Constant<T>>(folded)}) {
+      // Copy elements in Fortran array element order
```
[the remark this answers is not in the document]
jeanPerier: Edit: From our discussion, I understand it is to avoid exposing the iterator and tempting people to calculate offsets on their own rather than using At which could lead to many bugs.
That is fine for me. I will try to centralize as much as possible this access pattern while folding elemental intrinsic to avoid duplicating it.

lib/evaluate/fold.h
```diff
@@ -51,7 +52,11 @@ std::optional<Expr<T>> Fold(
 template<typename T>
 const Scalar<T> *GetScalarConstantValue(const Expr<T> &expr) {
   if (const auto *c{UnwrapExpr<Constant<T>>(expr)}) {
-    return &c->value;
+    if (c->size() == 1) {
```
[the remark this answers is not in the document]
klausler: Yes, it is.  The result is unambiguous and there didn't seem to me to be any debugging benefit to checking the rank and crashing if nonzero.

lib/semantics/expression.cc
```diff
@@ -1015,15 +1110,227 @@ static MaybeExpr AnalyzeExpr(ExpressionAnalysisContext &context,
...
+      // be well-defined.
+      // TODO: Possible language extension: use the most general type of
+      // the values as the type of a numeric constructed array, convert all
+      // of the other values to that type.  Alternative: let the first value
+      // determine the type, and convert the others to that type.
+      type = std::move(xType);
+      values.Push(std::move(*x));
+    } else if (typesMustMatch) {
+      if (static_cast<const DynamicType &>(*type) ==
+          static_cast<const DynamicType &>(xType)) {
```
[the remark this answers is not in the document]
klausler: The lengths may not be known at compilation time, so that's why the Fortran standard uses "shall" instead of imposing a constraint here.

lib/semantics/expression.cc
```diff
@@ -1015,15 +1110,227 @@ static MaybeExpr AnalyzeExpr(ExpressionAnalysisContext &context,
...
+                type = DynamicTypeWithLength{IntType::GetType()};
+              }
+              ArrayConstructorContext nested{exprContext, type, typesMustMatch};
+              parser::CharBlock name;
+              nested.Push(Expr<SomeType>{
+                  Expr<SomeInteger>{Expr<IntType>{ImpliedDoIndex{name}}}});
+              values.Push(ImpliedDo<SomeType>{name, std::move(*lower),
+                  std::move(*upper), std::move(*stride),
+                  std::move(nested.values)});
+            }
```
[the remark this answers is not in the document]
klausler: A bad parse should preclude running semantic analysis.  However, if those expressions failed semantic checking, they'll be vacant optional values here, and we're in an error recovery situation.

test/semantics/modfile18.f90
```diff
@@ -0,0 +1,38 @@
...
+
+module m
+  real, parameter :: a0 = 1.0_8
+  real, parameter :: a1(2) = [real::2.0, 3.0]
+  real, parameter :: a2(2) = [4.0, 5.0]
+  real, parameter :: a3(0) = [real::]
+  real, parameter :: a4(55) = [real::((1.0*k,k=1,j),j=1,10)]
+  real, parameter :: a5(:) = [6.0, 7.0, 8.0]
+  real, parameter :: a6(2) = [9, 10]
+  real, parameter :: a7(6) = [([(1.0*k,k=1,j)],j=1,3)]
```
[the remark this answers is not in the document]
klausler: Good idea, will do.

lib/semantics/expression.cc
```diff
@@ -1015,15 +1110,227 @@ static MaybeExpr AnalyzeExpr(ExpressionAnalysisContext &context,
...
+ArrayConstructorValues<T> MakeSpecific(
+    ArrayConstructorValues<SomeType> &&from) {
+  ArrayConstructorValues<T> to;
+  for (ArrayConstructorValue<SomeType> &x : from.values) {
+    std::visit(
+        common::visitors{
+            [&](CopyableIndirection<Expr<SomeType>> &&expr) {
+              auto *typed{UnwrapExpr<Expr<T>>(*expr)};
+              CHECK(typed != nullptr);
+              to.Push(std::move(*typed));
```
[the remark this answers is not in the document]
klausler: The comment is obsolete and I will update it.  The conversions are done when the elements are appended to the intermediate array constructor.

lib/semantics/type.h
```diff
@@ -280,7 +280,10 @@ class DeclTypeSpec {
   bool IsNumeric(TypeCategory) const;
   const NumericTypeSpec &numericTypeSpec() const;
   const LogicalTypeSpec &logicalTypeSpec() const;
-  const CharacterTypeSpec &characterTypeSpec() const;
+  const CharacterTypeSpec &characterTypeSpec() const {
```
tskeith: Why make this function different from the others in that it is in line? (I don't see a reason for any of them to be.)

test/semantics/resolve30.f90
```diff
@@ -28,16 +28,3 @@ subroutine s2
...
-  block
-    import, only: j
-    type t
-      !ERROR: 'i' from host scoping unit is not accessible due to IMPORT
-      real :: x(10) = [(i, &
-        !ERROR: 'i' from host scoping unit is not accessible due to IMPORT
-        i=1,10)]
-    end type
-  end block
-end subroutine
```
tskeith: Was this a bad test?

lib/semantics/scope.cc
```diff
@@ -265,27 +267,32 @@ const DeclTypeSpec &Scope::InstantiateIntrinsicType(
...
-            .data());
+  int kind{
+      semanticsContext.defaultKinds().GetDefaultKind(intrinsic->category())};
+  if (auto value{evaluate::ToInt64(copy)}) {
+    if (evaluate::IsValidKindOfIntrinsicType(intrinsic->category(), *value)) {
+      kind = *value;
+    } else {
+      foldingContext.messages().Say(
+          "KIND parameter value (%jd) of intrinsic type %s "
+          "did not resolve to a supported value"_err_en_US,
```
tskeith: When do we get this error as opposed to "... is not a supported type" from `semantics/expression.cc`?

lib/semantics/resolve-names.cc
```diff
@@ -2960,29 +2998,64 @@ bool DeclarationVisitor::Pre(const parser::StructureConstructor &x) {
...
+  bool implicit{false};
+  if (prev == nullptr) {
+    // Declare the name as an object in the enclosing scope so that
+    // the name can't be repurposed there later as something else.
+    prev = &MakeSymbol(InclusiveScope(), name.source, Attrs{});
+    ApplyImplicitRules(*prev);
+    implicit = true;
+  }
+  if (!ConvertToObjectEntity(*prev)) {
+    Say2(name, "Locality attribute not allowed on '%s'"_err_en_US, *prev,
```
tskeith: Is there a test for this case?

lib/semantics/resolve-names.cc
```diff
@@ -3212,10 +3302,37 @@ void ConstructVisitor::Post(const parser::DoConstruct &x) {
...
+    Say2(name, "Index name '%s' conflicts with existing identifier"_err_en_US,
+        *prev, "Previous declaration of '%s'"_en_US);
+    return;
+  } else {
+    if (auto *type{prev->GetType()}) {
+      symbol.SetType(*type);
+    }
+    if (const auto *details{prev->detailsIf<ObjectEntityDetails>()}) {
+      if (details->IsArray()) {
+        Say2(name, "Index variable '%s' is not scalar"_err_en_US, *prev,
```
tskeith: Will this be caught by the call to `CheckScalarIntegerType` below? If the check is needed here there should probably be a return after the error is reported.

lib/semantics/expression.cc
```diff
@@ -320,12 +416,12 @@ static MaybeExpr ApplySubscripts(ExpressionAnalysisContext &context,
...
-            using Ty = std::decay_t<decltype(base)>;
-            if constexpr (common::HasMember<Ty, decltype(ArrayRef::u)>) {
-              return CompleteSubscripts(
-                  context, ArrayRef{std::move(base), std::move(subscripts)});
-            }
+          [&](Component &&c) {
+            return CompleteSubscripts(
+                context, ArrayRef{std::move(c), std::move(subscripts)});
+          },
+          [&](auto &&) -> MaybeExpr {
```
tskeith: Can't `dataRef.u` contain an `ArrayRef` here?

lib/semantics/type.h
klausler: It makes the implementation of the function available to code in the evaluation library without introducing a cycle of dependency between the two shared binary libraries.

test/semantics/resolve30.f90
klausler: I believe so.  The scope of a statement entity is simply the statement in which it appears, and the rules about conflicts with the containing scope's use of the same name shouldn't apply here since `i` doesn't appear in the outer scope.

lib/semantics/scope.cc
klausler: The error in `scope.cc` will be emitted only for a parameterized derived type component with an intrinsic type whose kind is an expression that depends on the kind type parameters of the PDT.   Resolution of those errors is necessarily deferred until the PDT is instantiated. The error in `expression.cc` will come out for all other bad cases of `KindSelector` as they are encountered in declarations, array constructors, and (soon) structure constructors and `ALLOCATE` statements.

lib/semantics/resolve-names.cc
klausler: Not that I can find.  I'll add one to this PR.

klausler:
Thanks for the review!  Will merge presently.

lib/semantics/expression.cc
tskeith: I hit this assert on the following program:
```
real, pointer :: x(:), y(:)
allocate(x(3))
y => x(1:3)
end
```

klausler: Will investigate and fix before merging.

klausler: Fixed in branch.

Issue #269, opened 2019-01-29
Conversion of Real from lower to higher precision gives unexpected results


jeanPerier:
Calling `lib/evaluate/real.h` function `Convert `with an argument from a less precise type gives wrong result.
  
```
template<typename A> static ValueWithRealFlags<Real> Convert(
const A &x, Rounding rounding = defaultRounding) 
```

**Suspected Root Cause:**
line 284: `Fraction fraction{Fraction::ConvertUnsigned(xFraction).value};` should be right shifted by `-bitsLost` so that the MSB of the old value remains the MSB of the new value.

**Code reproducing the issue**
```
  #include "../../lib/evaluate/type.h"
  #include <iostream>
  using namespace Fortran::evaluate::value;
  float x{0.5f};
  auto rx{*reinterpret_cast<Real<Integer<32>, 24>*>(&x)};
  auto ry{Real<Integer<64>, 53>::Convert(rx).value}; //issue here
  auto y{*reinterpret_cast<double*>(&ry)};
  std::cout<<"expected: "<<x<<" result: "<<y<<std::endl; //should be identical
```
Currently, the output for me is `"expected: 0.5 result: 9.31323e-10"`

klausler:
Thanks for the bug report.  If the problem is resolved, please close this bug.

jeanPerier:
Tested ok after PR #272 merge. Thanks

Issue #270, opened 2019-01-29
Conversion of Denormal floating-point values gives unexpected results


jeanPerier:
The issue occurs both when going from and to denormal values in `lib/evaluate/real.h`  `Convert` function.

Going from a value that is not denormal in a type towards a less precise type where the value is denormal gives unexpected result. The conversion should right shift the significant.

**Suspected Root Cause:**
It seems The `Normalize` function is not handling negative exponents. They are going through case `else if (lshift == 0)` line 283 and the new exponent is set to 1 without any shifts. However, maybe it should be up to the caller of Normalize to handle negative exponent up-front, not sure.

**Code reproducing the issue:**
  ```
#include "../../lib/evaluate/type.h"
  #include <iostream>
  using FROM = Real<Integer<32>, 24>;
  using TO = Real<Integer<64>, 53>;
  FROM x32{typename FROM::Word{std::uint64_t{8}}};
  ValueWithRealFlags<TO> x64{TO::Convert(x32)};
  ValueWithRealFlags<FROM> xx32{FROM::Convert(x64.value)}; //converting non denormal toward denormal
  auto hx{*reinterpret_cast<float*>(&x32)};
  auto hxx{*reinterpret_cast<float*>(&xx32)};
  std::cout<<"expected: "<<hx<<" result: "<<hxx<<std::endl; //should be identical
```
Current output for me is "`expected: 1.12104e-44 result: 1.17549e-38`".

klausler:
If this bug is resolved to your satisfaction, please close this report when convenient.  Thanks!

PR #272, opened 2019-01-30
Fix bugs #269 & #270

klausler:


klausler:
All of your cases now pass.

PR #277, opened 2019-02-07
Name resolution for namelist and data statements

tskeith:


lib/semantics/mod-file.cc
```diff
@@ -280,6 +289,7 @@ void ModFileWriter::PutUseExtraAttr(
 }
 
 // Collect the symbols of this scope sorted by their original order, not name.
+// Namelists are an exception: they are sorted to the end.
```
vdonaldson: What does "sorted to the end" mean?  Something about the fact that a later statement can add group objects to a group from a previous statement?  (8.9p3)

tskeith: It means the symbols for namelists appear at the end of the sorted vector of symbols, not in order of first occurrence of the name as with other symbols. This is so that the declarations of the objects in the namelist are sure to occur before the namelist statement.

I'll try to make the comment clearer.

lib/semantics/mod-file.cc
```diff
@@ -293,7 +303,13 @@ std::vector<const Symbol *> CollectSymbols(const Scope &scope) {
     }
   }
   std::sort(sorted.begin(), sorted.end(), [](const Symbol *x, const Symbol *y) {
-    return x->name().begin() < y->name().begin();
+    bool xIsNml{x->has<NamelistDetails>()};
+    bool yIsNml{y->has<NamelistDetails>()};
+    if (xIsNml != yIsNml) {
+      return xIsNml < yIsNml;
```
klausler: One doesn't often see comparisons of Booleans.

Consider `return yIsNml;`.  It'll do the same thing and might be more clear.  (Hard to tell, it's subjective.)

lib/semantics/resolve-names.cc
```diff
@@ -1421,6 +1425,10 @@ void ScopeHandler::SayAlreadyDeclared(
   Say2(name, "'%s' is already declared in this scoping unit"_err_en_US, prev,
       "Previous declaration of '%s'"_en_US);
 }
+void ScopeHandler::SayWithDecl(
```
klausler: I'm going to have to remember to use this one in work in progress.  Thanks.

lib/semantics/symbol.cc
```diff
@@ -213,6 +214,11 @@ void Symbol::SetType(const DeclTypeSpec &type) {
       details_);
 }
 
+bool Symbol::IsArray() const {
+  const auto *details{std::get_if<ObjectEntityDetails>(&details_)};
```
klausler: Functions can return arrays.  Do you want `IsArray()` to be true for them too?  If not, maybe `IsObjectArray()` would be better.

lib/semantics/mod-file.cc
tskeith: What I really want is a template that combines two Compare functions into one. I.e. compare the isNml property and then the name location. But I couldn't find anything like that in the C++ standard library.

That way of thinking about it is what led me to this implementation. I see that `return yIsNml;` is equivalent, but I had to stop and think about it.

Does anyone else have an opinion on which is clearer?

lib/semantics/symbol.cc
tskeith: Thanks, I'll change it to `IsObjectArray`.

lib/semantics/mod-file.cc
klausler: Maybe wrap the sequence in a function with naming that conveys that primary and secondary keys are being used.

```
template<typename PRIMARY, typename SECONDARY>
bool IsLess(PRIMARY x1, SECONDARY x2, PRIMARY y1, SECONDARY y2) {
  return x1 != y1 ? x1 < y1 : x2 < y2;
}
```

PR #280, opened 2019-02-08
Fix for both issues 274 and 278

schweitzpgi:
Makes each subprogram/procedure a distinct scope and fixes the bug in stack initialization.

### DIFF
--- a/lib/semantics/resolve-labels.cc
+++ b/lib/semantics/resolve-labels.cc
@@ -228,10 +228,13 @@ public:
   }
 
   // see 11.1.1
-  bool Pre(const parser::ProgramUnit &) { return PushNewScope(); }
-  bool Pre(const parser::FunctionSubprogram &) { return PushSubscope(); }
-  bool Pre(const parser::SubroutineSubprogram &) { return PushSubscope(); }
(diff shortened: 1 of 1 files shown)


lib/semantics/resolve-labels.cc
```diff
@@ -495,8 +494,9 @@ class ParseTreeAnalyzer {
     currentScope_ = programUnits_.back().scopeModel.size() - 1;
     return true;
   }
-  bool PushNewScope() {
+  bool InitializeNewScopeContext() {
```
klausler: These pushed subscopes still get popped, right?

schweitzpgi:
> LGTM if all tests still pass and bugs stayed fixed. Thanks.

I changed the algorithm a bit. Rather than popping back to the root at the END statement, the scope tree is left as created.  The next scope (if any) just forcibly resets the (pristine) tree state.

klausler:
Is the scope tree still needed for anything?

schweitzpgi:
> Is the scope tree still needed for anything?

Oh, yes. Very much so. All subsequent constraint checks may use it.

PR #281, opened 2019-02-11
More name resolution improvements

tskeith:


lib/semantics/expression.cc
```diff
@@ -775,15 +771,13 @@ static MaybeExpr AnalyzeExpr(
     return AsMaybeExpr(ConvertToKind<TypeCategory::Integer>(
         *kind, AsExpr(ImpliedDoIndex{n.source})));
   } else if (n.symbol == nullptr) {
-    context.Say(
-        n.source, "TODO INTERNAL: name was not resolved to a symbol"_err_en_US);
+    // error should have been reported in name resolution
```
schweitzpgi: I've been using `CHECK(false);` macros in these "something previous should've caught this already" cases.

lib/semantics/expression.cc
```diff
@@ -1779,20 +1770,10 @@ Expr<SubscriptInteger> ExpressionAnalysisContext::Analyze(TypeCategory category,
...
-                size = defaultKind;
-              }
-            } else if (!evaluate::IsValidKindOfIntrinsicType(category, size)) {
-              Say("%s*%jd is not a supported type"_err_en_US,
-                  parser::ToUpperCaseLetters(EnumToString(category)).data(),
-                  size);
+            if (!CheckIntrinsicSize(category, size)) {
               size = defaultKind;
+            } else if (category == TypeCategory::Complex) {
+              size /= 2;
```
schweitzpgi: Maybe add a comment as to where in the standard this adjustment comes from?

lib/semantics/expression.cc
klausler: `CHECK(!"internal error message");` is better.

lib/semantics/expression.cc
klausler: It doesn't come from the standard.  Types with byte sizes like `COMPLEX*16` are universally implemented extensions that predate the standard's concept of kinds.  While kind codes are processor-dependent and arbitrary, the standard does require that the same kind codes for `REAL` have to work with `COMPLEX` (7.4.3.3).  So `COMPLEX*16` means `COMPLEX(KIND=8)`.

lib/semantics/rewrite-parse-tree.cc
```diff
@@ -88,7 +85,7 @@ class RewriteMutator {
 // Check that name has been resolved to a symbol
 void RewriteMutator::Post(parser::Name &name) {
   if (name.symbol == nullptr && errorOnUnresolvedName_) {
-    messages_.Say(name.source, "Internal: no symbol found for '%s'"_en_US,
+    messages_.Say(name.source, "Internal: no symbol found for '%s'"_err_en_US,
```
klausler: This becomes a `CHECK` in the future, right?

lib/semantics/expression.cc
tskeith: In this case it's not an internal error. The comment is meant to indicate that there is no need to report an error here because it was (hopefully) reported in name resolution.

Maybe it would be clearer just to say "error was reported in name resolution". Which is will be true once name resolution is complete and bug-free.

lib/semantics/rewrite-parse-tree.cc
tskeith: Yes, I think after I implement COMMON and EQUIVALENCE I'll change it to a CHECK.

PR #286, opened 2019-02-14
Add support for common blocks

tskeith:
A symbol for a common block has `CommonBlockDetails` which contains
a list of the symbols that are in the common block.

The name of the symbol for the blank common block is the empty string.
That preserves the property that every symbol name is a substring of
the cooked source. We use the 0-length substring starting at the COMMON
statement so that when symbols are sorted by the location of the start
of the name it ends up in the right place.

Some of the checks on members of common blocks don't happen until the
end of the scope. They can't happen earlier because we don't necessarily
know the type and attributes.

Enhance `test_errors.sh` so that multiple errors can be expected for
a single line.

### DIFF
--- a/lib/semantics/mod-file.cc
+++ b/lib/semantics/mod-file.cc
@@ -173,6 +173,15 @@ void ModFileWriter::PutSymbol(
             }
             decls_ << '\n';
           },
+          [&](const CommonBlockDetails &x) {
+            PutLower(decls_ << "common/", symbol);
+            char sep = '/';
+            for (const auto *object : x.objects()) {
+              PutLower(decls_ << sep, *object);
+              sep = ',';
+            }
+            decls_ << '\n';
+          },
           [&](const FinalProcDetails &) {
             PutLower(typeBindings << "final::", symbol) << '\n';
           },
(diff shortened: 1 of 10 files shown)


test/semantics/test_errors.sh
```diff
@@ -47,7 +47,11 @@ fi
 # $actual has errors from the compiler; $expect has them from !ERROR comments in source
 # Format both as "<line>: <text>" so they can be diffed.
 sed -n 's=^[^:]*:\([^:]*\):[^:]*: error: =\1: =p' $log > $actual
-{ echo; cat $src; } | cat -n | sed -n 's=^ *\([0-9]*\). *\!ERROR: *=\1: =p' > $expect
+awk '
```
klausler: Thanks for this feature, but I might keep using continuation lines in tests.

tskeith:
I didn't mark objects that appear in common blocks but I will do so in a future PR. Is a flag good enough or do you think it would be useful to have a pointer back to the common block the symbol is a member of?

klausler:
I suggested a flag because I couldn't think of a use case for a link, but now that I think about it more, you might need/want one for EQUIVALENCE processing.

lib/semantics/mod-file.cc
```diff
@@ -173,6 +173,15 @@ void ModFileWriter::PutSymbol(
             }
             decls_ << '\n';
           },
+          [&](const CommonBlockDetails &x) {
+            PutLower(decls_ << "common/", symbol);
+            char sep = '/';
```
jeanPerier: Not the most useful feedback regarding your PR, but just because I saw it: `char sep{'/'};`

PR #287, opened 2019-02-15
Structure constructor semantics

klausler:
Semantic analysis, representation, and folding of structure constructors.  The semantic analysis is performed during expression semantics, so that the code will be in the right place after generic function call resolution is implemented.

Constants got a lot of working over, and constant Character objects are now in their own template specialization.  Constant character array elements are no longer distinct std::strings, but are instead all packed into one string in Fortran element order, anticipating their use in folding of transformational intrinsics.  (Constant derived type values are not yet packed in storage.)

Some further work on array constructors was done after taking a few passes with this code over some regression test suites.  A language extension implemented by PGI and another compiler was discovered and now supported regarding character array constructors with distinct lengths on their character literals.

Some TODOs remain with structure and array constructors with regard to type conformance checking; that'll come next week.


lib/semantics/mod-file.cc
```diff
@@ -195,8 +195,9 @@ void ModFileWriter::PutSymbol(
 void ModFileWriter::PutDerivedType(const Symbol &typeSymbol) {
   auto &details{typeSymbol.get<DerivedTypeDetails>()};
   PutAttrs(decls_ << "type", typeSymbol.attrs(), ","s, ""s);
-  if (!details.extends().empty()) {
-    PutLower(decls_ << ",extends(", details.extends().ToString()) << ')';
+  if (const DerivedTypeSpec * extends{typeSymbol.GetParentTypeSpec()}) {
+    PutLower(decls_ << ",extends(", extends->typeSymbol().name().ToString())
+        << ')';
```
tskeith: There is an overloading of `PutLower` that takes a symbol instead of a name:
`PutLower(decls_ << ",extends(", extends->typeSymbol()) << ')';`

klausler: Thank you.

lib/semantics/symbol.h
```diff
@@ -211,20 +213,28 @@ class DerivedTypeDetails {
...
   // declarations appear in the derived type definition(s).
   std::list<SourceName> paramNames_;
-  std::list<const Symbol *> paramDecls_;
-  // These are the declarations of the derived type's components in component
+  SymbolList paramDecls_;
+  // These are the names of the derived type's components in component
   // order.  A parent component, if any, appears first in this list.
-  std::list<const Symbol *> components_;
-  SourceName extends_;
+  std::list<SourceName> components_;
```
tskeith: `componentNames_` might be a better name for this field. In the functions `OrderComponents()` and `GetParentComponent()`, "component" means component symbol, not name.

klausler: Sure.

lib/semantics/symbol.cc
```diff
@@ -616,20 +618,53 @@ std::list<SourceName> DerivedTypeDetails::OrderParameterNames(
...
+SymbolList DerivedTypeDetails::OrderComponents(const Scope &scope) const {
+  SymbolList result;
+  for (SourceName name : components_) {
+    auto iter{scope.find(name)};
+    if (iter != scope.cend()) {
+      const Symbol &symbol{*iter->second};
+      if (symbol.test(Symbol::Flag::ParentComp)) {
+        CHECK(result.empty());
+        const DerivedTypeSpec &spec{
+            *symbol.get<ObjectEntityDetails>().type()->AsDerived()};
```
tskeith: Since you're assuming the type is a derived type, you could use `symbol.get<ObjectEntityDetails>().type()->derivedTypeSpec()`. It asserts it is a derived type and returns a reference.

Maybe we shouldn't have both `AsDerived()` and `derivedTypeSpec()`.

klausler: Thanks; fixed it here and in one other place that I found.

Issue #283, opened 2019-02-12
Recursive PDTs don't work yet


klausler:
The implementation of PDTs needs to be rearranged a little so that the type name and type parameters are visible for use in the definitions of components within the derived type definition.

klausler:
Fix coded and tested, PR to follow shortly.

klausler:
Fixed.

PR #291, opened 2019-02-18
Fix recursive use of parameterized derived types (issue #283)

klausler:


lib/semantics/resolve-names.cc
```diff
@@ -2809,13 +2811,18 @@ void DeclarationVisitor::Post(const parser::DerivedTypeSpec &x) {
...
+// in this Pre() routine so that recursive use of the derived type can be
+// supported in the components.
+bool DeclarationVisitor::Pre(const parser::DerivedTypeDef &x) {
+  Walk(std::get<parser::Statement<parser::DerivedTypeStmt>>(x.t));
+  Walk(std::get<std::list<parser::Statement<parser::TypeParamDefStmt>>>(x.t));
   auto &scope{currScope()};
   CHECK(scope.symbol() != nullptr);
   CHECK(scope.symbol()->scope() == &scope);
   auto &details{scope.symbol()->get<DerivedTypeDetails>()};
   auto &stmt{std::get<parser::Statement<parser::DerivedTypeStmt>>(x.t)};
```
tskeith: If this declaration were moved up it could be used in line 2818.

lib/semantics/resolve-names.cc
```diff
@@ -2848,39 +2856,45 @@ void DeclarationVisitor::Post(const parser::DerivedTypeDef &x) {
...
 }
 bool DeclarationVisitor::Pre(const parser::DerivedTypeStmt &x) {
   return BeginAttrs();
 }
 void DeclarationVisitor::Post(const parser::DerivedTypeStmt &x) {
+  // Resolve the EXTENDS() clause before creating the derived
+  // type's symbol to foil attempts to recursively extend a type.
+  auto *extendsName{derivedTypeInfo_.extends};
+  const Symbol *extendsType{nullptr};
+  if (extendsName != nullptr) {
```
tskeith: Can we check here for extendsName being the same as the derived type name and provide a better error, like "derived type cannot extend itself"?

lib/semantics/resolve-names.cc
klausler: Thanks.

lib/semantics/resolve-names.cc
klausler: Sure.

PR #292, opened 2019-02-20
Indirection to host rte for folding and draft of rte interface description structure

jeanPerier:
This PR replaces #275 from which it is derived and intends to answer to the comments.

It adds an indirection layer to host runtime in order to avoid direct dependencies on a particular library while folding intrinsic functions with the host runtime.

It defines a structure `RteProcedureSymbol` that can describe runtime procedures regardless of whether they are host or target runtime procedures.
This structure mainly holds a compiler internal name as well as the types of the arguments and results in an host independent way. This allow to query for runtime functions based on these internal names and F18 argument types.
`HostRteProcedureSymbol `and `TargetRteProcedureSymbol `inherit from this data structure with different purposes. `TargetRteProcedureSymbol` holds the actual symbol name for a certain library.
`HostRteProcedureSymbol` holds a function pointer that can be used for folding purposes.
A big part of this PR deals with providing a safe interaction with the host runtime. The following features are implemented:
- It is impossible to directly call the "raw" function pointers inside a `HostRteProcedureSymbol`. Instead, the code in the folding has to make a statically typed query to the `HostRte` structures containing the `HostRteProcedureSymbols`.
If a `HostRteProcedureSymbol` is found with matching name and types, `HostRtewill` return a statically typed callable taking F18 constants as arguments.
- This callable is dealing with all the casts to host types
- This callable is dealing with the floating point errors/traps/status flags that may be used by the host runtime function.

So far, this PR is focused on elemental functions, but the idea is that it should be extendable for transformational runtime function taking descriptor arguments. Ideally, the TargetRteProcedureSymbol should be able to describe any kind of runtime function, not only intrinsics (e.g. omp runtime). But this is not an absolute goal so far.

Another point to consider, is that it will be possible to dynamically generates `HostRteProcedureSymbols `from `TargetRteProcedureSymbols `if the library is available on the host (POC to be implemented in future PR).
For now, `HostRteProcedureSymbols` are created by the other possibility to construct them: from functions defined in headers included in the compiler.

The new files have small descriptions. You should start by looking at rte-interface.h and fold.cc/rte.cc to see how it is used.
rte.h and host.h are template implementations that one should not have to look to use the interface.

For simplicity, some feature of #275 were dropped for now:

- implicit cast to "bigger" type mainly to allow fp16 folding
- Mapping intrinsic functions to the folding code with a hash of the name

Update 22-feb-2019 (commits 8161f77 and 1c2093f):

- Answer some comments.
- Adding libpgmath support for folding:
Some of the scalar functions of libpgmath are now described in rte.cc (162 symbols are described) and the logic to dynamically load these function is implemented for systems supporting dlopen.
Then, C++ standard library function are added to the HostRte table if an equivalent was not already defined.
If dlopen is not supported or libpgmath.so not found, all functions come from the C++ standard library.


lib/evaluate/fold.cc
```diff
@@ -156,35 +162,298 @@ ComplexPart FoldOperation(FoldingContext &context, ComplexPart &&complexPart) {
...
+  using namespace intrinsicHelper;
+  using T = Type<TypeCategory::Real, KIND>;
+  for (std::optional<ActualArgument> &arg : funcRef.arguments()) {
+    if (arg.has_value()) {
+      *arg->value = FoldOperation(context, std::move(*arg->value));
+    }
+  }
+  if (auto *intrinsic{std::get_if<SpecificIntrinsic>(&funcRef.proc().u)}) {
+    const std::string name{intrinsic->name};
+    if (name == "acos" || name == "acosh") {
```
tskeith: Is there eventually going to be a branch in this if-statement for every real intrinsic? It would be nice to have a way to break up this function so it doesn't end up being thousands of lines long.

lib/evaluate/fold.cc
```diff
@@ -156,35 +162,298 @@ ComplexPart FoldOperation(FoldingContext &context, ComplexPart &&complexPart) {
...
+      if (auto callable{
+              context.hostRte().GetHostProcedureWrapper<Scalar, T, T>(name)}) {
+        return FoldElementalIntrinsic<T, T>(
+            context, std::move(funcRef), *callable);
+      } else {
+        context.messages().Say(
+            "%s(real(kind=%d)) cannot be folded on host"_en_US, name.c_str(),
```
tskeith: Is this a user error or internal error? What does the user do when they get this error?

lib/evaluate/rte.cc
```diff
@@ -0,0 +1,166 @@
...
+  return std::cyl_bessel_j(static_cast<HostT>(n), x);
+}
+template<typename HostT> static HostT Bessel_yn(std::int64_t n, HostT x) {
+  return std::cyl_neumann(static_cast<HostT>(n), x);
+}
+
+template<typename HostT> void AddLibmRealHostProcedure(HostRte &hostRte) {
+  using F = FuncPointer<HostT, HostT>;
+  using F2 = FuncPointer<HostT, HostT, HostT>;
+  HostRteProcedureSymbol libmSymbols[]{{"acos", F{std::acos}, true},
```
tskeith: Should there be an enum of intrinsics so we don't have to do string comparisons on the names?

lib/evaluate/rte.cc
```diff
@@ -0,0 +1,166 @@
...
+#if __x86_64__
+  if (context.flushSubnormalsToZero()) {
+    currentFenv_.__mxcsr |= 0x8000;  // result
+    currentFenv_.__mxcsr |= 0x0040;  // operands
+  } else {
+    currentFenv_.__mxcsr &= ~0x8000;  // result
+    currentFenv_.__mxcsr &= ~0x0040;  // operands
+  }
+#else
+  // TODO other architecture
```
tskeith: Maybe we should have a #warning here.

lib/evaluate/rte.cc
```diff
@@ -0,0 +1,166 @@
...
+  AddLibmComplexHostProcedure<long double>(*this);
+}
+
+void HostFloatingPointEnvironment::SetUpHostFloatingPointEnvironment(
+    FoldingContext &context) {
+  errno = 0;
+  if (feholdexcept(&originalFenv_) != 0) {
+    context.messages().Say(
+        "Folding with host runtime: feholdexcept() failed: %s"_en_US,
+        std::strerror(errno));
```
tskeith: What can the user do when they get this error?

lib/evaluate/rte.h
```diff
@@ -0,0 +1,176 @@
...
+// function of the structures defined in rte-interface.h. It should only be
+// included if these member functions are used, else rte-interface.h is
+// sufficient. This is to avoid circular dependencies. The below implementation
+// cannot be defined in .cc file because it would be too cumbersome to decide
+// which version should be instantiated in a generic way.
+
+#include "host.h"
+#include "rte-interface.h"
+#include "type.h"
+#include "../../lib/common/template.h"
```
tskeith: You can remove "../lib/" from the path.

test/evaluate/folding.cc
```diff
@@ -39,7 +47,6 @@ struct TestGetScalarConstantValue {
 };
 
 int main() {
-  using TestTypes = AllIntrinsicTypes;
-  RunOnTypes<TestGetScalarConstantValue, TestTypes>::Run();
+  RunOnTypes<TestGetScalarConstantValue, AllIntrinsicTypes>::Run();
```
tskeith: Are there tests for the new functionality?

lib/evaluate/fold.cc
jeanPerier: The user code is correct, this was check when resolving intrinsics.  It is reporting to the user an F18 limitation. I think it is hard to guarantee that it will be possible to fold all intrinsics with all possible types on all host architectures.
Pgfortran gives an error in this case (e.g. PGF90-S-0155-Intrinsic not supported in initialization: acosh)
Depending on whether the value is really needed in the front-end or not (e.g. in a kind argument), the compiler may still be able to compile the program or not. That is why it is not yet a hard error.

I am not sure what a user can do. Probably try to adapt his code (I see very little use cases of using an acos with 128bits floating point to get a kind type argument) or add to F18 a host runtime function that can fold this on his host if it is really needed.
Would you rephrase the message to make it clear it is a compiler limitation?

lib/evaluate/fold.cc
jeanPerier: A way could be to handle all the cases that do not use host runtime (this will still be around 30/40 func for integers) and then to have a single call to the host runtime to try to fold it regardless the intrinsics.  The HostRte interface could be simplified into something like:
`hostRte.TryFolding(context, std::move(funcRef));`
This requires some work in HostRte and maybe a few duplicates with fold.cc, but it is probably cleaner. I think it is also possible to make this part independent of the actual intrinsics. This would make it easier to extend. I will describe my idea better before implementing it.

lib/evaluate/rte.cc
jeanPerier: If we reduce the number of hand written comparisons on the names by better delegating it to a map lookup when there is no need to do something particular, then I am not sure it is worth having to define such an enum somewhere and having to map it with the string names and maintain it. I will try to go with the hostRte.TryFolding(funcRef) proposal that should greatly reduce the number of hand written string comparison first.

lib/evaluate/rte.cc
jeanPerier: To be honest, I did not find why this function could failed exactly apart that from its definition it can. So I am not entirely sure what are the consequences of a failure and what should happen. I have no idea what the user should do appart from compiling again and reporting if the issue persists.
Would you make it a fatal error to be safe?

lib/evaluate/rte.h
jeanPerier: thx

test/evaluate/folding.cc
jeanPerier: No, they were actually added in a PR fixing GetScalarConstantValue. The changes appearing here are unfortunate left over modifications from PR #275. I should have copied the original file after the cherry-pick from PR #275.
In PR #275, I had added tests for some intrinsic functions folding. For now, I have removed them until the interface is stable.

lib/evaluate/CMakeLists.txt
```diff
@@ -28,6 +28,7 @@ add_library(FortranEvaluate
   tools.cc
   type.cc
   variable.cc
+  rte.cc
```
klausler: What does "rte" stand for?  Maybe "run-time environment"?  Perhaps "library" would be more clear.

lib/evaluate/fold.cc
```diff
@@ -156,35 +162,298 @@ ComplexPart FoldOperation(FoldingContext &context, ComplexPart &&complexPart) {
       FoldOperation(context, std::move(complex)), complexPart.part()};
 }
 
-template<typename T>
-Expr<T> FoldOperation(FoldingContext &context, FunctionRef<T> &&funcRef) {
-  ActualArguments args{std::move(funcRef.arguments())};
-  for (std::optional<ActualArgument> &arg : args) {
+namespace intrinsicHelper {
```
klausler: I don't think that you need a namespace like this in a *.cc file around some types and explicitly `static` functions.

lib/evaluate/fold.cc
```diff
@@ -156,35 +162,298 @@ ComplexPart FoldOperation(FoldingContext &context, ComplexPart &&complexPart) {
...
+  if ((... && std::get<I>(scalars).has_value())) {
+    if constexpr (std::is_same_v<WrapperType<TR, TA...>,
+                      ScalarFuncWithContext<TR, TA...>>) {
+      return Expr<TR>{Constant<TR>{func(context, *std::get<I>(scalars)...)}};
+    } else if constexpr (std::is_same_v<WrapperType<TR, TA...>,
+                             ScalarFunc<TR, TA...>>) {
+      return Expr<TR>{Constant<TR>{func(*std::get<I>(scalars)...)}};
+    }
+  }
+  // TODO: handle arrays when Constant<T> can represent them
```
klausler: `Constant<>` now represents constant arrays.

lib/evaluate/fold.cc
```diff
@@ -156,35 +162,298 @@ ComplexPart FoldOperation(FoldingContext &context, ComplexPart &&complexPart) {
...
+  using namespace intrinsicHelper;
+  using T = Type<TypeCategory::Integer, KIND>;
+  for (std::optional<ActualArgument> &arg : funcRef.arguments()) {
     if (arg.has_value()) {
       *arg->value = FoldOperation(context, std::move(*arg->value));
     }
   }
   if (auto *intrinsic{std::get_if<SpecificIntrinsic>(&funcRef.proc().u)}) {
-    std::string name{intrinsic->name};
+    const std::string name{intrinsic->name};
```
klausler: Why add the `const` qualifier to a local variable?

lib/evaluate/fold.cc
```diff
@@ -156,35 +162,298 @@ ComplexPart FoldOperation(FoldingContext &context, ComplexPart &&complexPart) {
...
+                *funcRef.arguments()[0]->value)}) {
           return std::visit([](auto &kx) { return kx.LEN(); }, charExpr->u);
         }
       } else {
         common::die("len() result not SubscriptInteger");
       }
-    } else {
-      // TODO: many more intrinsic functions
+    } else if (name == "iand") {
+      if (auto *x{std::get_if<BOZLiteralConstant>(
```
klausler: First, you should loop here rather than copying and pasting code.  Second, since these conversions of typeless BOZ literals to the type of the function result are a common pattern, maybe the intrinsic function table lookup code in `intrinsics.cc` should be performing them.

lib/evaluate/host.h
```diff
@@ -0,0 +1,229 @@
...
+
+template<> struct HostTypeHelper<Type<TypeCategory::Integer, 4>> {
+  using Type = std::int32_t;
+};
+
+template<> struct HostTypeHelper<Type<TypeCategory::Integer, 8>> {
+  using Type = std::int64_t;
+};
+
+// no int 128bit
```
klausler: Why not?

lib/evaluate/host.h
```diff
@@ -0,0 +1,229 @@
...
+
+// no float 16bits
```
klausler: Why not?

lib/evaluate/host.h
```diff
@@ -0,0 +1,229 @@
...
+};
+
+template<> struct HostTypeHelper<Type<TypeCategory::Character, 1>> {
+  using Type = std::string;
+};
+
+template<> struct HostTypeHelper<Type<TypeCategory::Character, 2>> {
+  using Type = std::u16string;
+};
+
```
klausler: Missing `CHARACTER(KIND=4)`.

You could use the typedefs in `type.h` for these types of string constants rather than duplicating them.

lib/evaluate/host.h
```diff
@@ -0,0 +1,229 @@
...
+}
+
+// Utility to find "bigger" types that exist on host. By bigger, it is meant
+// that the bigger type can represent all the values of the smaller types
+// without information loss.
+template<TypeCategory cat, int KIND> struct NextBiggerReal {
+  using Type = void;
+};
+template<TypeCategory cat> struct NextBiggerReal<cat, 2> {
+  using Type = Fortran::evaluate::Type<cat, 3>;
```
klausler: Conversions from IEEE 16-bit real to the truncated 32-bit real will lose precision.

lib/evaluate/host.h
```diff
@@ -0,0 +1,229 @@
...
+};
+template<TypeCategory cat> struct NextBiggerReal<cat, 3> {
+  using Type = Fortran::evaluate::Type<cat, 4>;
+};
+template<TypeCategory cat> struct NextBiggerReal<cat, 4> {
+  using Type = Fortran::evaluate::Type<cat, 8>;
+};
+
```
klausler: 64-bit real can convert to 80 or 128-bit real.

lib/evaluate/rte-interface.h
```diff
@@ -0,0 +1,130 @@
...
+
+namespace Fortran::evaluate {
+class FoldingContext;  // forward declare
+namespace rte {
+
+using TypeCode = unsigned char;
+
+template<typename TR, typename... TA> using FuncPointer = TR (*)(TA...);
+
+enum PassBy { Ref, Val };
```
klausler: Should probably be `enum class`.

lib/evaluate/rte-interface.h
```diff
@@ -0,0 +1,130 @@
...
+#include <map>
+#include <optional>
+#include <string>
+#include <vector>
+
+namespace Fortran::evaluate {
+class FoldingContext;  // forward declare
+namespace rte {
+
+using TypeCode = unsigned char;
```
klausler: Are you using single character encodings of types rather than an enumeration or (category,kind) ordered pair?

Are there mapping functions to encode/decode these single character type codes?

lib/evaluate/CMakeLists.txt
jeanPerier: Yes it stands "run-time environment".  I agree "environment" is not appropriate and rte may be unclear. I find library also unclear on its own. I will go for "runtime-library". What about the following naming scheme:
`rte.cc -> runtime-library.cc`
`rte.h -> runtime-library-templates.h` (make it anoying to include because it should not be included if rte-interface.h is sufficient)
`rte-interface.h -> runtime-library.h`
For the namespace "rte", I don't know. Fortran::runtime is taken for the actual runtime implementation. Should I just remove the rte namespace for now?
Also, when they are many libraries described in runtime-library.cc, maybe they should go in different files of another folder, for now let's keep it simple.

lib/evaluate/fold.cc
jeanPerier: yes, I will remove it.

lib/evaluate/fold.cc
jeanPerier: Changed the TODO to reflect that. That is the next one on my list once when this PR is OK.

lib/evaluate/fold.cc
jeanPerier: I am afraid the code below will grow a lot and I want to give the guarantee to the reader when he sees "name" somewhere below that it was not changed in the code in-between. Plus, our style guideline says to use it generously.

lib/evaluate/fold.cc
jeanPerier: OK, I added a TODO for the second point and applied the first point for now.

lib/evaluate/host.h
jeanPerier: I do not know what I could try mapping here. I think std::int128_t does no exit. I can maybe try to add a conditional type on "long long int" if it is 128bit. Do you know of any architecture where this is the case?

lib/evaluate/host.h
jeanPerier: It is implicitly mapped to UnsupportedType. There is no "half" in the standard yet. Are they any architecture where float is 16bits ? Then it may be worth adding a conditional that tries mapping it to float.

lib/evaluate/host.h
jeanPerier: Thanks, I missed it. that will make maintenance easier.

lib/evaluate/host.h
jeanPerier: Now mapped to IEEE 32bits

lib/evaluate/host.h
jeanPerier: Added the last two possible conversions

lib/evaluate/rte-interface.h
jeanPerier: changed accordingly

lib/evaluate/rte-interface.h
jeanPerier: The actual mapping is defined in rte.h. see typeCodeOf<T> and TypeOf<TypeCode n>.
This is both to avoid dependencies and also because I do not really want to make it possible for the user to interact with the type encoding directly.
TypeCode is simply the index of a type inside the tuple `RteTypes` containing all the types that are allowed to describe a runtime function symbol. For now RteTypes are all the intrinsic types.
I do not want to define an enum for now because it will be annoying define and maintain.
I did not consider using pairs (category, kind) because I had the void and descriptor types in mind that do not really fit inside this encoding and are special. It is easy to just add an empty class DescriptorType and VoidType to reflect that.
I tried to centralized everything interacting with this encoding so that it can "mature" when trying to describe more complex runtime functions.

lib/evaluate/CMakeLists.txt
schweitzpgi: Regarding "make it annoying": In other projects, public/exported header files are placed in a directory (tree) and the `-I` option is used to select that directory.  This makes it easy to replace entire interfaces with no code changes. Meanwhile, header files that are privately shared within the implementation live in the same place as the source files.  Since module/library A only has access to its own directory and whatever lives at the `-I` directory (set), it can access module B's public interface but doesn't have backdoor access to module B's private parts.

lib/evaluate/fold.cc
klausler: Yes, it does recommend generous use of `const`.  It makes sense to `const`-qualify pointers and references and class members as much as we can.  Const qualification on local variables is really rare, which is why I asked about it.

lib/evaluate/host.h
klausler: Both GCC and clang have `__int128_t`.

lib/evaluate/host.h
klausler: The IEEE-754 2008 standard has `binary16`, and some environments have the alternative `bfloat16` which is just 32-bit floating-point without the least-order 16 bits.  f18 currently supports both, and the runtime library might have to support one or both.  I don't think that either form is supported by gcc or clang or by standard C math libraries.

lib/evaluate/fold.cc
tskeith: I think it's reasonable to use it for mutable objects too (as opposed const on intrinsic types). You're providing a guarantee similar to const references that you won't call non-const methods on it.

lib/evaluate/CMakeLists.txt
jeanPerier: Why not, that would make things clear. I believe though that it is a repository wide decision and it should not be taken inside this PR. Let's discuss this with the others.

lib/evaluate/host.h
jeanPerier: OK, I will add `__int128_t` with some ifdef, it seems that it is not defined on every architecture  ([src](https://gcc.gnu.org/onlinedocs/gcc/_005f_005fint128.html)) so I will also check that `__SIZEOF_INT128__` is defined ([src](https://gcc.gnu.org/ml/gcc-help/2015-08/msg00185.html)) and protect the definition with:
`#if (defined(__GNUC__) || defined(__clang__)) && defined(__SIZEOF_INT128__)`
Also, I am confused regarding what is the difference between  `__int128` and `__int128_t` but I get a warning from g++ when using `__int128` that I do not get with `__int128_t` so I will go for this last one.

> ISO C++ does not support ‘__int128’ for ‘type name’ [-Wpedantic]

lib/evaluate/host.h
jeanPerier: If gcc and clang are not supporting it right now, I propose transforming the comment regarding the host/f18 types mapping into a TODO for when gcc and clang supports it.
For folding purposes, we can convert to "bigger" supported floating-point. I have to find a good place to make the conversions.

lib/evaluate/fold.cc
jeanPerier: TODO is now implemented in commit c2fec22.

lib/evaluate/rte.cc
tskeith: I think I would treat it as an internal error until we find a case where it can actually happen.

lib/evaluate/fold.cc
tskeith: If it's a case where the language requires that it be folded to a value, I think it should be a fatal error (like PGI). If we can still compile without folding I'm not sure we need a diagnostic at all. i don't know if you can distinguish those cases.

lib/evaluate/fold.cc
```diff
@@ -156,35 +162,354 @@ ComplexPart FoldOperation(FoldingContext &context, ComplexPart &&complexPart) {
...
+  if ((... && (std::get<I>(args) != nullptr))) {
+    // Compute the shape of the result based on shapes of arguments
+    std::vector<std::int64_t> shape;
+    int rank;
+    const std::vector<std::int64_t> *shapes[sizeof...(TA)]{
+        &std::get<I>(args)->shape()...};
+    const int ranks[sizeof...(TA)]{std::get<I>(args)->Rank()...};
+    for (unsigned int i{0}; i < sizeof...(TA); ++i) {
+      if (ranks[i] > 0) {
+        if (rank == 0) {
```
tskeith: Was `rank` supposed to be initialized to 0?

jeanPerier: Yes. Thanks for catching this highway to undefined behaviors...

lib/evaluate/rte.cc
jeanPerier: Done in commit f776c8d. I changed all errors to internal error.
I only left a warning when the rounding mode cannot be set as expected because it does not exists. What happens then is well defined, the user just needs to know that the results might not be bit to bit comparable to what he expects.

lib/evaluate/rte.cc
jeanPerier: I went for a runtime TODO warning in commit f776c8d. I do not want to annoy people for a feature they may not use.

lib/evaluate/CMakeLists.txt
jeanPerier: Taking into consideration that everything that is done here is about intrinsic procedures, I have committed the following renaming in commit 7cfd33b:
 
rte-interface.h -> intrinsics-library.h
rte.h -> intrinsics-library-templates.h
rte.cc -> intrinsics-library.cc

`RteProcedureSymbol` becomes `IntrinsicProcedureRuntimeDescription`
`HostRteProcedureSymbol` becomes `HostRuntimeIntrinsicProcedure`
`TargetRteProcedureSymbol` becomes `TargetRuntimeIntrinsicProcedure`
`HostRte` becomes `HostIntrinsicProceduresLibrary`.
`TargetRteLibrary` becomes `TargetIntrinsicProceduresLibrary`.

the rte namespace is removed.

lib/evaluate/rte.cc
pawosm-arm: For AArch64 architecture, please either look into merged flang's pull requests https://github.com/flang-compiler/flang/pull/488 and https://github.com/flang-compiler/flang/pull/489 and also into two of my comments under f18's bug report https://github.com/flang-compiler/f18/issues/148 (https://github.com/flang-compiler/f18/issues/148#issuecomment-408269774 and https://github.com/flang-compiler/f18/issues/148#issuecomment-409990207) or discuss and synchronize addition of missing implementation during our regular calls.

lib/evaluate/rte.cc
```diff
@@ -0,0 +1,166 @@
...
+    return;
+  }
+  if (fegetenv(&currentFenv_) != 0) {
+    context.messages().Say(
+        "Folding with host runtime: fegetenv() failed: %s"_en_US,
+        std::strerror(errno));
+    return;
+  }
+#if __x86_64__
```
pawosm-arm: Do we really need to use those architecture specific fields of fenv_t directly? _FPU_GETCW()/_FPU_SETCW() pair of macros are defined in <fpu_control.h> which should be available to all glibc-based systems, this could help us avoid use of preprocessor-based checking for architecture (#if __x86_64__) - each place like that makes it less possible to use f18 as a cross-compiler and can make it harder to be accepted by LLVM people.

jeanPerier: I think the `#ifdef` is unavoidable here because `<fpu_control.h>` is unfortunately only about the X87 FPU on X86_64 and not SSE. `_FPU_GETCW`/`_FPU_SETCW` resolve to assembly instructions `fnldcw`/`fnstcw `on x86_64 which are related to the x87 fpu control word (FPCR).
X86_64 FPCR does not give control over denormal value flushing. What matters for X86_64 is the `mxcsr` register which is related to SSE. It needs to be addressed directly for to control denormal value flushing.

Anyway, thanks for bringing the topic and for pointing to these macros for Aarch64 hosts. I will try to move the `#ifdef` in `host.h` file from this PR that should centralize such host dependent behavior (So far, I think having `#ifdef` in `host.h` is preferable compared to having different files for different hosts.). Once this PR is merged, Aarch64 implementation can be added . Will you want to do it?

lib/evaluate/intrinsics-library.cc
```diff
@@ -50,29 +50,53 @@ bool HostIntrinsicProceduresLibrary::HasEquivalentProcedure(
...
+
+// C++ Bessel functions take a floating point as first arguments.
 // Fortran Bessel functions take an integer.
-template<typename HostT> static HostT Bessel_jn(std::int64_t n, HostT x) {
+template<typename HostT> static HostT Bessel_j0(HostT x) {
+  return std::cyl_bessel_j(0., x);
+}
+template<typename HostT> static HostT Bessel_y0(HostT x) {
+  return std::cyl_neumann(0., x);
+}
```
tskeith: When I compile with clang, I get these errors:
```
../../lib/evaluate/intrinsics-library.cc:58:15: error: no member named 'cyl_bessel_j' in namespace 'std'
  return std::cyl_bessel_j(0., x);
         ~~~~~^
../../lib/evaluate/intrinsics-library.cc:61:15: error: no member named 'cyl_neumann' in namespace 'std'
  return std::cyl_neumann(0., x);
         ~~~~~^
```

jeanPerier: Thanks for testing and catching this. I had only checked clang with libstd++... It seems Libc++ has no support yet of C++17 new math functions.
I changed the code to use C Bessel functions instead. It compiles and pass tests with both GCC, clang + GNU libstd++ and clang + libc++ without warnings.

lib/evaluate/rte.cc
pawosm-arm: What is the easiest way to test this piece of code?

jeanPerier: Good point. I did not add a test for that. It cannot be tested yet with a Fortran program because there is not options to control SubnormalFlushing in f18 tool yet.
It can be tested with a C++ program that tries to fold an expression that requires host runtime and set the context with different subnormal flushing options.
I made a small test that use `log(subnormal-value)` to detect subnormal flushing. If subnormal values are flushed log(0) returns -infinity. If subnormal values are not flushed, it returns a negative finite value.
See commit 35ebaf3.
It passes on X86_64 and once the logic is here for Aarch64 in host.cc, it should pass there also.

PR #294, opened 2019-02-20
Fortran IR: part 2, lowest layer of the IR hierarchy. Specifically, Basic Blocks

schweitzpgi:
contain Statements.  The statement layer is still evolving as the IR takes better shape and other code in the compiler is written.


lib/IntermediateRepresentation/statement.def
```diff
@@ -0,0 +1,171 @@
...
+HANDLE_TERM_STMT(3, Switch,      SwitchStmt)
+HANDLE_TERM_STMT(4, SwitchCase,  SwitchCaseStmt)
+HANDLE_TERM_STMT(5, SwitchType,  SwitchTypeStmt)
+HANDLE_TERM_STMT(6, SwitchRank,  SwitchRankStmt)
+HANDLE_TERM_STMT(7, IndirectBr,  IndirectBrStmt)
+HANDLE_TERM_STMT(8, Unreachable, UnreachableStmt)
+LAST_TERM_STMT(8)
+
+// Standard actions - These instructions capture computations as
+// evalute::expressions
```
klausler: `evaluate`

lib/IntermediateRepresentation/statement.def
```diff
@@ -0,0 +1,171 @@
...
+FIRST_BINARY_STMT(9)
+HANDLE_BINARY_STMT(9,  Assign,    AssignmentStmt)
+HANDLE_BINARY_STMT(10, PtrAssign, PointerAssignStmt)
+HANDLE_BINARY_STMT(11, LblAssign, LabelAssignStmt)
```
klausler: Fortran's `ASSIGN` statement is so restricted that I doubt we should use the general representations for expressions and variables to represent it.

lib/IntermediateRepresentation/statement.def
```diff
@@ -0,0 +1,171 @@
...
+HANDLE_BINARY_STMT(12, Expr,      ExprStmt)
```
klausler: Fortran, unlike C, doesn't have expression statements.  Or is this for capturing values into temporaries?

lib/IntermediateRepresentation/stmt.h
```diff
@@ -0,0 +1,50 @@
...
+#define FORTRAN_INTERMEDIATEREPRESENTATION_STMT_H_
+
+#include "basicblock.h"
+#include "mixin.h"
+#include "statements.h"
+
+namespace Fortran::IntermediateRepresentation {
+
+/// Sum type over all statement classes
+struct Statement : public SumTypeMixin<std::variant<
```
klausler: There's some facilities in `lib/common/template.h` for incrementally constructing and combining variants that might make for less macro-y code here.  They're used in `lib/evaluate/expression.h` to compose the alternatives allowed for expressions of each type category.

lib/IntermediateRepresentation/statement.def
schweitzpgi: That's why I made a distinct statement.

lib/IntermediateRepresentation/statement.def
schweitzpgi: The latter.  I'll come up with a new name.

lib/IntermediateRepresentation/stmt.h
schweitzpgi: Yes, I saw that later.

lib/IntermediateRepresentation/statement.def
schweitzpgi: Thanks.

lib/FIR/statement.def
```diff
@@ -0,0 +1,171 @@
+// Copyright (c) 2019, NVIDIA CORPORATION.  All rights reserved.
```
klausler: As we discussed last week, there are other ways of implementing the definitions of the various statement data structures besides including a source file with various definitions of preprocessor macros that are going to be more readable and maintainable in the very long term.  Some combination of templates, inheritance, and singly-defined macros for boilerplate will likely replace the mechanism that you have here.  I am okay with approving this PR with this file in it, but one of us is going to have to replace it soon, I think.

lib/FIR/statements.h
```diff
@@ -0,0 +1,470 @@
...
+#define HANDLE_STMT(num, opcode, name) struct name;
+#include "statement.def"
+
+struct Statement;
+
+CLASS_TRAIT(StatementTrait)
+CLASS_TRAIT(TerminatorTrait)
+CLASS_TRAIT(ActionTrait)
+
+struct Evaluation
```
klausler: These top-level `struct`s have inheritance and private members, and I think that they should be `class`es.

PR #296, opened 2019-02-20
Fortran IR: part 4. Dotty graph visualization and other miscellaneous

schweitzpgi:
headers.

### DIFF
--- a/lib/IntermediateRepresentation/common.h
+++ b/lib/IntermediateRepresentation/common.h
@@ -1,0 +1,107 @@
+// Copyright (c) 2019, NVIDIA CORPORATION.  All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef FORTRAN_INTERMEDIATEREPRESENTATION_COMMON_H_
+#define FORTRAN_INTERMEDIATEREPRESENTATION_COMMON_H_
+
+#include "../common/idioms.h"
+#include "../common/indirection.h"
+#include "../evaluate/expression.h"
+#include "../evaluate/type.h"
+#include "../evaluate/variable.h"
+#include "../parser/parse-tree.h"
+#include "../semantics/symbol.h"
+
+// Some useful, self-documenting macros for failure modes
+#define STRINGIFY(X) #X
+#define LINE2STRING(X) STRINGIFY(X)
+#define AT_HERE " at " __FILE__ "(" LINE2STRING(__LINE__) ")"
+#define DIE Fortran::common::die
+#define SEMANTICS_FAILED(STRING) DIE("semantics bug: " STRING AT_HERE)
+#define SEMANTICS_CHECK(CONDITION, STRING) \
+  if (CONDITION) { \
+  } else { \
+    DIE("semantics bug: " STRING AT_HERE); \
+  }
+#define WRONG_PATH() DIE("control should not reach here" AT_HERE)
+
+namespace Fortran::IntermediateRepresentation {
+class Statement;
+class BasicBlock;
+struct Program;
+struct GraphWriter;
+
+struct Attribute {
+  enum { IntentIn, IntentOut, IntentInOut } attribute;
+  unsigned short position;
+};
+using FunctionType = evaluate::SomeType;  // TODO: what should this be?
+using AttributeList = std::vector<Attribute>;
+enum struct LinkageTypes { Public, Hidden, External };
+using Expression = evaluate::GenericExprWrapper;
+#if 0
+struct Variable {
+  // TODO: should semantics::Symbol be removed?
+  template<typename... Ts> struct GVT {
+    using type =
+        std::variant<const semantics::Symbol *, evaluate::Variable<Ts>...>;
+  };
+  Variable(const semantics::Symbol *symbol) : u{symbol} {}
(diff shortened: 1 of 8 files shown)


lib/IntermediateRepresentation/common.h
```diff
@@ -0,0 +1,107 @@
...
+#define WRONG_PATH() DIE("control should not reach here" AT_HERE)
+
+namespace Fortran::IntermediateRepresentation {
+class Statement;
+class BasicBlock;
+struct Program;
+struct GraphWriter;
+
+struct Attribute {
+  enum { IntentIn, IntentOut, IntentInOut } attribute;
```
klausler: This `enum` might be a useful thing to have in `lib/common/fortran.h` as an `enum class Intent`.

Should `Value` also be in that list?

lib/IntermediateRepresentation/mixin.h
```diff
@@ -0,0 +1,114 @@
...
+#include "llvm/ADT/ilist.h"
+#include <optional>
+#include <tuple>
+#include <type_traits>
+#include <variant>
+
+namespace Fortran::IntermediateRepresentation {
+
+template<typename T, typename E = void> struct SumTypeMixin {};
+template<typename T>  // T must be std::optional<...>
```
klausler: You could enforce that constraint by defining the `struct` as a specialization `SumTypeMixin<std::optional<T>, ...>`.

lib/IntermediateRepresentation/mixin.h
```diff
@@ -0,0 +1,114 @@
...
+struct SumTypeMixin<T, std::enable_if_t<std::variant_size_v<T>>> {
```
klausler: Be advised, I don't think that zero-sized variants are legal in C++.  (They do seem to work with g++ and clang, though.)

lib/IntermediateRepresentation/mixin.h
```diff
@@ -0,0 +1,114 @@
...
+  std::transform(first, last, std::back_inserter(out.first),
+      [](auto &&a) -> decltype(a.first) { return a.first; });
+  std::transform(first, last, std::back_inserter(out.second),
+      [](auto &&a) -> decltype(a.second) { return a.second; });
+  return out;
+}
+
+}
+
+#endif
```
klausler: Please put a comment on the terminal `#endif` in header files that matches the guard at the top of the file.

lib/IntermediateRepresentation/common.h
```diff
@@ -0,0 +1,107 @@
...
+  RuntimeCallEventPost,
+  RuntimeCallEventWait,
+  RuntimeCallSyncAll,
+  RuntimeCallSyncImages,
+  RuntimeCallSyncMemory,
+  RuntimeCallSyncTeam,
+  RuntimeCallLock,
+  RuntimeCallUnlock,
+  RuntimeCallSIZE = RuntimeCallUnlock - RuntimeCallFailImage + 1
+};
```
jeanPerier: The RuntimeCallType seems to be only related with CoArrays. What is your expectation regarding intrinsics functions? Do you expect that they will be simple CallStmt?

schweitzpgi: At the moment, this is correct. Statements that clearly need some sort of runtime support code are in this list and nothing more. (Similarly for IO runtime, but I made that a different bucket.) The idea here is that these "runtime calls" might be changing the state of the environment in ways that an optimization pass on FIR might have to be concerned about.

Regarding intrinsics, the idea was to have generic intrinsic procedures (even if the name is a family of functions) be lowered to an instance of a CallStmt since they have the same semantics, effectively, as an invocation of a user-defined procedure.  But that's still squishy and function calls are currently embedded in evaluate::Expression trees.

lib/IntermediateRepresentation/common.h
```diff
@@ -0,0 +1,107 @@
...
+#include "../parser/parse-tree.h"
+#include "../semantics/symbol.h"
+
+// Some useful, self-documenting macros for failure modes
+#define STRINGIFY(X) #X
+#define LINE2STRING(X) STRINGIFY(X)
+#define AT_HERE " at " __FILE__ "(" LINE2STRING(__LINE__) ")"
+#define DIE Fortran::common::die
+#define SEMANTICS_FAILED(STRING) DIE("semantics bug: " STRING AT_HERE)
+#define SEMANTICS_CHECK(CONDITION, STRING) \
```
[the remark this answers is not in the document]
schweitzpgi: Right. I just extended that idea a little.

lib/IntermediateRepresentation/common.h
schweitzpgi: Yes, maybe they should go there.

Would `Value` mean pass-by-value here?  Relatedly, FIR will have to be able to represent and distinguish between references/pointers and values (that are not references).  Also with BIND(C), it makes some sense that `Call` is an abstraction over calling conventions.

lib/IntermediateRepresentation/mixin.h
schweitzpgi: Fixed buggy comments

lib/IntermediateRepresentation/common.h
klausler: `Value` would mean "dummy argument has the VALUE attribute", and unfortunately that's not the same thing as pass-by-value.  In Fortran it basically means "pass by value, if it fits in a register, otherwise pass by address and immediately make a local copy on entry".  The FIR probably needs to be capable of describing argument-passing methods in terms of implementation as well or instead of their Fortran dummy argument attributes so that arguments to C & assembly code in runtime libraries can be characterized.

schweitzpgi: Thanks for the explanation. I agree that it probably makes a lot of sense to add VALUE here with its strong interactions with INTENT as well.

lib/IntermediateRepresentation/mixin.h
schweitzpgi: I'm going to have to change this code a bit for clang++.

lib/IntermediateRepresentation/mixin.h
schweitzpgi: Done

PR #297, opened 2019-02-20
Fotran IR: part 5. Afforestation is the lowering of concrete parse trees

schweitzpgi:
and semantics information into the Fortran IR.

### DIFF
--- a/lib/IntermediateRepresentation/afforestation.cc
+++ b/lib/IntermediateRepresentation/afforestation.cc
@@ -1,0 +1,1963 @@
+// Copyright (c) 2019, NVIDIA CORPORATION.  All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "afforestation.h"
+#include "builder.h"
+#include "mixin.h"
+#include "../evaluate/fold.h"
+#include "../evaluate/tools.h"
+#include "../parser/parse-tree-visitor.h"
+#include "llvm/ADT/SmallSet.h"
+#include "llvm/Support/FileSystem.h"
+#include "llvm/Support/raw_ostream.h"
+
+namespace Fortran::IntermediateRepresentation {
+
+static llvm::raw_ostream *debugChannel;
+static llvm::raw_ostream &DebugChannel() {
+  return debugChannel ? *debugChannel : llvm::errs();
+}
+static void SetDebugChannel(llvm::raw_ostream *output) {
+  debugChannel = output;
+}
+void SetDebugChannel(const std::string &filename) {
+  std::error_code ec;
+  SetDebugChannel(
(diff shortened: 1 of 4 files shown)


lib/IntermediateRepresentation/afforestation.h
```diff
@@ -0,0 +1,37 @@
...
+}
+
+namespace Fortran::semantics {
+class SemanticsContext;
+}
+
+namespace Fortran::IntermediateRepresentation {
+class Program;
+
+void SetDebugChannel(const std::string &filename);
```
klausler: Maybe SemanticsContext should contain a `std::ostream *` for optional debugging output.

lib/IntermediateRepresentation/afforestation.h
```diff
@@ -0,0 +1,37 @@
...
+
+Program *CreateFortranIR(const parser::Program &program,
+    semantics::SemanticsContext &semanticsContext, bool debugLinearIR);
```
klausler: That debug parameter could default to `false`.

lib/IntermediateRepresentation/afforestation.cc
```diff
@@ -0,0 +1,1963 @@
...
+  LinearLabelRef getNext() {
+    LinearLabelRef next{counter++};
+    auto cap{referenced.capacity()};
+    if (cap < counter) referenced.reserve(2 * cap);
+    referenced[next] = false;
+    return next;
+  }
+  void setReferenced(LinearLabelRef label) { referenced[label] = true; }
+  bool isReferenced(LinearLabelRef label) const { return referenced[label]; }
+  std::vector<bool> referenced;
```
klausler: `std::set<int>` might be safer with respect to expansion and more readable (since this is actually meant to represent a set).  The extra overhead won't matter.

lib/IntermediateRepresentation/afforestation.cc
```diff
@@ -0,0 +1,1963 @@
...
+  auto iter{ad.labelMap.find(label)};
+  if (iter == ad.labelMap.end()) {
+    LinearLabel ll{ad.labelBuilder};
+    ll.setReferenced();
+    ad.labelMap.insert({label, ll});
+    return ll;
+  }
+  return iter->second;
+}
+std::tuple<const parser::Name *, LinearLabelRef, LinearLabelRef> FindStack(
```
klausler: Perhaps a `using` typedef here for the tuple, to make it clear that the result type and stack element types are identical?

lib/IntermediateRepresentation/afforestation.cc
```diff
@@ -0,0 +1,1963 @@
...
+    const SwitchTypeStmt::ValueSuccPairListType &rest) {
+  builder->CreateSwitchType(condition, defaultCase, rest);
+}
+
+struct FortranIRLowering {
+  using LabelMapType = std::map<LinearLabelRef, BasicBlock *>;
+  using Closure = std::function<void(const LabelMapType &)>;
+
+  FortranIRLowering(semantics::SemanticsContext &sc, bool debugLinearIR)
+    : fir_{new Program("program_name")}, semanticsContext_{sc},
```
klausler: Is `program_name` a placeholder for something that should be passed in as an argument or a constituent of the SemanticsContext?

lib/IntermediateRepresentation/afforestation.h
schweitzpgi: Good idea

lib/IntermediateRepresentation/afforestation.cc
schweitzpgi: Yes, it is just a placeholder for now. It could be taken from a PROGRAM statement (or like) if available.

PR #302, opened 2019-02-25
RFC: Design proposal for ERROR STOP statement semantic analysis.

pawosm-arm:
It's RFC document only. Any comments welcomed.

### DIFF
--- a/documentation/ErrorStop.md
+++ b/documentation/ErrorStop.md
@@ -1,0 +1,103 @@
+<!--
+Copyright (c) 2019, Arm Ltd.  All rights reserved.
+-->
+
+# Design proposal for ERROR STOP statement semantic analysis
+
+## Current state
+
+The overview of semantic analysis process is briefly described in the "Overview of Compiler Phases" document held in the F18 source code repository ([https://github.com/flang-compiler/f18/blob/master/documentation/Overview.md](https://github.com/flang-compiler/f18/blob/master/documentation/Overview.md)).
+
+The process is orchestrated by the `Perform()` method in `Semantics` class as defined in `lib/semantics/semantics.cc` file:
+
+~~~
+bool Semantics::Perform() {
+  return ValidateLabels(context_.messages(), program_) &&
+      parser::CanonicalizeDo(program_) &&  // force line break
+      ResolveNames(context_, program_) &&
+      RewriteParseTree(context_, program_) &&
+      StatementSemanticsPass1{context_}.Walk(program_) &&
+      StatementSemanticsPass2{context_}.Walk(program_) &&
+      ModFileWriter{context_}.WriteAll();
+}
+
+~~~
+
+Following steps are performed:
+
+1. Validate labels (see [Semantics: Resolving Labels and Construct Names](https://github.com/flang-compiler/f18/blob/master/documentation/LabelResolution.md))
+
+2. Canonicalize DO statements
+
+3. Resolve names (produce *tree of scopes* populated with symbols and types)
+
+4. Rewrite Parse Tree
+
+5. Statement semantics analysis pass 1 (see below)
+
+6. Statement semantics analysis pass 2 (see below)
+
+7. Run Mod File Writer
+
+Pass 1 of the Statements Semantic Analysis currently analyzes experssions (semantic checks performed on all expressions):
+
+~~~
+using StatementSemanticsPass1 = SemanticsVisitor<ExprChecker>;
+~~~
+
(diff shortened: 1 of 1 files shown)


sscalpone:
The NVIDIA team has been discussing how to implement semantics for general statements for a while, albeit casually around the water cooler instead of in mailing lists.

We appreciate your idea to have a separate entry point for each statement; it's clean & easy to extend. 
Our worry is that having multiple traversals of the parse tree may be too slow and the independence of each traversal might complicate sharing infrastructure.  Our thought is to have a single visitor & a semantic static that's available to each type of statement being parsed.  Please expect a counter proposal to be published soon.  Let me know if I've misinterpreted what you'd proposed.

In either case, the parts of semantic checking for DO CONCURRENT that don't involve labels would be folded into the new structure.

pawosm-arm:
Hi Steve,
I think you've got my idea right, I can see your point too. For what I can read from the rest of the comment, I'd need to wait and see how things are going to change soon before moving any further...

tskeith:
@pawosm-arm, please take a look at PR #315. It is a different approach that doesn't require walking the parse tree for each separate check to be performed.

sscalpone:
A couple things -- please restructure the code to follow the model that was started with check-do-current.  Check the check-* routines to see how they handle displaying error messages.  They use a messages object; I think you are using fprintf.

test/semantics/CMakeLists.txt
```diff
@@ -152,6 +152,10 @@ set(IF_TESTS
   if_stmt01.f90
 )
 
+set(ANY_TESTS
```
sscalpone: I don't think we're ready for another generic category (ANY == generic).  You can add your test under ERROR_TESTS; it's fine if it doesn't get an error.

pawosm-arm:
Hi Steve,
This (second) commit is only an RFC companion and only discloses the subjects that need to be covered. Separate PR will be issued when the work is done (seems like all pieces are known by now). Your comments are useful regardless.

sscalpone:
@pawosm-arm Do you want to rework this PR to update and include just the documentation?

sscalpone:
I am getting a compilation error:

In file included from /home/sjs/work/pgi/f18/test/lib/semantics/tools.cc:15:
/home/sjs/work/pgi/f18/test/lib/semantics/tools.h:102:53: error: ‘SemanticsContext’ does not name a type; did you mean ‘semantics’?
     const evaluate::GenericExprWrapper &expr, const SemanticsContext &context);

on this branch.

pawosm-arm:
> @pawosm-arm Do you want to rework this PR to update and include just the documentation?

Done.

pawosm-arm:
Updated tools function name. Force pushed, as there was no review since last force push and the change was one-worder.

documentation/ErrorStop.md
```diff
@@ -0,0 +1,103 @@
...
+private:
+  SemanticsContext &context_;
+};
+~~~
+
+As we can see, whenever the visitor's `Pre()` method is called for any given node of the parse tree, the `Enter()` method of a given checker is executed for that node. Similarly, whenever the visitor's `Post()` method is called for any given node of the parse tree, the `Leave()` method of a given checker is executed for that node.
+
+In order to prevent code duplications, useful functions (to be called from within the checkers) were extracted into `tools.cc` file and exposed to te `semantics` namespace. E.g. `ExprChecker` makes good use of `semantics::FindExternallyVisibleObject()` function.
+
+## Proposed checker of ERROR STOP statements
```
sscalpone: I think everything before line 91 is fairly generic.  I think it can be deleted.  If you think some of it should be preserved, perhaps you could move it to the documentation for semantics.

sscalpone: This is no long the proposed checked.  It is the checker.

documentation/ErrorStop.md
```diff
@@ -0,0 +1,103 @@
...
+
+`ERROR STOP` and also `STOP` are very simple statements for which the Standard specifies small number of constraints, most of them are functional rather than semantic.
+
+1. The `StopChecker` class should be defined to inherit from `BaseChecker`. It should contain one private field, a reference to the current semantic context.
```
sscalpone: Reflect that the state is finished.  "...clans is defined to inherit..."   Etc.

pawosm-arm:
I don't see much use of this document now. I guess this RFC PR can be closed.

sscalpone:
Up to you.

PR #304, opened 2019-02-27
Procedure(id) semantics

klausler:
Semantic analysis for identifiers being used as procedure interfaces on pointer declarations.  These are changes needed for the larger task of validating structure constructor component values.  "Unrestricted" specific intrinsic names are accepted (and given the INTRINSIC attribute).  Pointer assignment statements to these intrinsics are also checked.

The procedure "characteristics" defined in 15.3 now have a representation, which is more than was strictly needed to represent recognition of the unrestricted specific intrinsics, but the characteristics will very soon be needed for procedure call argument compatibility checking and generic resolution.

Also cleaned up usage of `DynamicType`, adding a pointer field to the character type description in the symbol table.


test/semantics/resolve45.f90
```diff
@@ -1,76 +0,0 @@
-! Copyright (c) 2019, NVIDIA CORPORATION.  All rights reserved.
```
tskeith: This looks like a merge glitch. This file shouldn't be removed.

lib/semantics/resolve-names.cc
```diff
@@ -3019,6 +3044,51 @@ bool DeclarationVisitor::Pre(const parser::ProcComponentDefStmt &) {
...
+bool DeclarationVisitor::Pre(const parser::ProcPointerInit &x) {
+  if (auto *name{std::get_if<parser::Name>(&x.u)}) {
+    return !NameIsKnownOrIntrinsic(*name);
+  }
+  return true;
+}
+bool DeclarationVisitor::Pre(const parser::ProcInterface &x) {
+  if (auto *name{std::get_if<parser::Name>(&x.u)}) {
+    if (!FindSymbol(*name) &&
+        !HandleUnrestrictedSpecificIntrinsicFunction(*name)) {
```
tskeith: Can you use `NameIsKnownOrIntrinsic` here?

test/semantics/resolve46.f90
```diff
@@ -0,0 +1,28 @@
...
+  intrinsic :: null ! a weird special case
+  !ERROR: 'haltandcatchfire' is not a known intrinsic procedure
+  intrinsic :: haltandcatchfire
+  procedure(sin), pointer :: p
+  p => alog ! valid use of an unrestricted specific intrinsic
+  p => alog10 ! ditto, but already declared intrinsic
+  p => cos ! ditto, but also generic
+  p => tan ! a generic & an unrestricted specific, not already declared
+  !TODO ERROR: a restricted specific, to be caught in ass't semantics
+  p => amin1
```
jeanPerier: What about intrinsics that are only generic and not specific? For instance:
```
intrinsic :: bessel_j0 ! a generic intrinsic name only
!TODO ERROR: a generic intrinsic name only
p => bessel_j0
```

lib/evaluate/intrinsics.cc
```diff
@@ -613,10 +622,13 @@ static const IntrinsicInterface genericIntrinsicFunction[]{
...
 
+// The following table contains the intrinsic functions listed in
+// Tables 16.2 and 16.3 in Fortran 2018.  The "unrestricted" functions
+// in Table 16.2 can be used as actual arguments, PROCEDURE() interfaces,
+// and procedure pointer targets.
 struct SpecificIntrinsicInterface : public IntrinsicInterface {
   const char *generic{nullptr};
-  bool isRestrictedSpecific{
-      false};  // when true, can only be called, not passed
+  bool isRestrictedSpecific{false};
```
jeanPerier: There was no functional change here, but the rewriting made me realize that both `SpecificIntrinsicInterface` and `SpecificIntrinsic` have `false` has a default value for `isRestrictedSpecific`.
This means that when `IntrinsicProcTable::Probe` finds an intrinsic in the generic table, the `specificIntrinsic` member of the returned `SpecificCall` is currently default initialized to isRestrictedSpecific{false}.
This is probably wrong if generic intrinsics that are not specific intrinsics shall not be used in a procedure pointer assignment.
I think SpecificIntrinsic.isRestrictedSpecific is not used anywhere so far, but maybe Probe should set isRestrictedSpecific to false for intrinsics that were found in the generic table just to be safe.

test/semantics/resolve45.f90
klausler: Right.

lib/semantics/resolve-names.cc
klausler: Yes.  And that leaves `NameIsKnownOrIntrinsic` as the only remaining caller of `HandleUnrestrictedSpecificIntrinsic`, but I think that I'll let it stay as its own function rather than manually folding it into its only caller.

test/semantics/resolve46.f90
klausler: Thanks

klausler:
> It looks like the copyrights for f20180-grammar.txt, enum-set.h, and indirection.h need to be updated.

Done.

PR #308, opened 2019-02-28
Rename to Fortran.h, update style guide, run clang-format on everything

klausler:
Address my TODOs from last week's meeting on style and naming.


lib/FIR/builder.h
```diff
@@ -133,7 +133,6 @@ struct FIRBuilder {
   BasicBlock *cursorBlock_;
   std::vector<BasicBlock *> blockStack_;
 };
-
```
schweitzpgi: I wonder if this noise is because we have different versions of clang-format?

klausler: Maybe.  I'm using 6.0.   I don't know what you're using.  If it's newer, then maybe you can tell me, and we can upgrade to it.

schweitzpgi: I'm using tip (which is a moving 9.0).
I'll use 6.0 from now on.

klausler: 6.0 is getting pretty old.  The newest clang-format binary that I can find here at PGI is a 7.0.  I will try it out on f18 and see how big the resulting patch would be.  If it's not too bad, we should upgrade.

klausler: clang-format 7.0 didn't change anything from the 6.0 results.  I'll continue to use it unless somebody can point me to a more recent stable release.

PR #311, opened 2019-03-02
More structure constructor semantics checking

klausler:
Structure constructor semantics remain incomplete, but here's another step, and I'm sending it out for review before continuing.  I needed a bunch of utility routines to implement some complicated rules and I packaged them into their own new module in lib/semantics/tools.{h,cc} because they looked likely to be useful in other situations.

### DIFF
--- a/lib/semantics/expression.cc
+++ b/lib/semantics/expression.cc
@@ -16,6 +16,7 @@
 #include "scope.h"
 #include "semantics.h"
 #include "symbol.h"
+#include "tools.h"
 #include "../common/idioms.h"
 #include "../evaluate/common.h"
 #include "../evaluate/fold.h"
@@ -1484,12 +1485,23 @@ static MaybeExpr AnalyzeExpr(ExpressionAnalysisContext &context,
           }
         } else {
           CHECK(symbol->has<semantics::ObjectEntityDetails>());
+          // C1594(4)
+          if (!isNULL) {
+            const auto &innermost{context.context().FindScope(expr.source)};
+            if (const auto *pureFunc{
+                    semantics::FindPureFunctionContaining(&innermost)}) {
+              if (semantics::IsOrHasPointerComponent(*symbol) &&
+                  semantics::IsExternallyVisibleObject(*value, *pureFunc)) {
+                context.Say(expr.source,
+                    "Externally visible object must not be associated with a pointer in a PURE function"_err_en_US);
+              }
+            }
+          }
           if (symbol->attrs().test(semantics::Attr::POINTER)) {
             if (!isNULL) {
               // TODO C7104: check that object pointer components are
               // being initialized with compatible object designators
-              context.Say(expr.source,
-                  "TODO: non-null object pointer component value not implemented yet"_err_en_US);
+              // TODO pmk WIP this is next
             }
           } else if (MaybeExpr converted{
                          ConvertToType(*symbol, std::move(*value))}) {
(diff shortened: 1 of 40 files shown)


lib/semantics/expression.cc
```diff
@@ -1470,9 +1474,46 @@ static MaybeExpr AnalyzeExpr(ExpressionAnalysisContext &context,
       }
       unavailable.insert(symbol->name());
       if (MaybeExpr value{AnalyzeExpr(context, expr)}) {
-        // TODO pmk: C7104, C7105 check that pointer components are
-        // being initialized with data/procedure designators appropriately
-        result.Add(*symbol, std::move(*value));
+        bool isNULL{std::holds_alternative<NullPointer>(value->u)};
```
tskeith: I think `isNull` is a better spelling, to be consistent with `NullPointer`.

klausler: The capitalization was meant to suggest the NULL() intrinsic function, as I've used all-caps for other intrinsic functions in the evaluation library.

lib/semantics/expression.cc
```diff
@@ -1470,9 +1474,46 @@ static MaybeExpr AnalyzeExpr(ExpressionAnalysisContext &context,
...
+          CHECK(symbol->has<semantics::ObjectEntityDetails>());
+          // C1594(4)
+          if (!isNULL) {
+            const auto &innermost{context.context().FindScope(expr.source)};
+            if (const auto *pureFunc{
+                    semantics::FindPureFunctionContaining(&innermost)}) {
+              if (semantics::IsOrHasPointerComponent(*symbol) &&
+                  semantics::IsExternallyVisibleObject(*value, *pureFunc)) {
+                context.Say(expr.source,
+                    "Externally visible object must not be associated with a pointer in a PURE function"_err_en_US);
```
tskeith: Would this message be more helpful if it referred to the symbol?

lib/semantics/expression.cc
tskeith: OK, makes sense.

lib/semantics/expression.cc
klausler: Yes, when the component is a pointer; maybe not when the component contains a nested pointer component several levels down.  From the expression, it's not always possible to dig out a clear symbol for the globally visible object -- it can be USE associated with renaming.  But I can try to do better on both, maybe by extending those predicates to return symbol pointers instead of simple truth values.

klausler: Done.

klausler:
Added an improvement to the `Indirection` class template family API suggested by Tim: instead of working like a pointer with `operator*` and `operator->`, these should work like `optional` with `has_value()` and `value()` interfaces.

klausler:
If any approving reviewer would like to take another look at the PR before submission, now's your chance.  A fair number of additional changes have crept in.

lib/semantics/tools.h
```diff
@@ -0,0 +1,74 @@
...
+
+namespace Fortran::semantics {
+
+const Symbol *FindCommonBlockContaining(const Symbol &object);
+const Scope *FindProgramUnitContaining(const Scope &);
+const Scope *FindProgramUnitContaining(const Symbol &);
+const Scope *FindPureFunctionContaining(const Scope *);
+
+bool IsCommonBlockContaining(const Symbol &block, const Symbol &object);
+bool IsAncestor(const Scope *maybeAncestor, const Scope &maybeDescendent);
```
tskeith: `IsAncestor` could be referring to submodules and modules. So I think this needs a different name (or at least a comment). Maybe `IsAncestorScope` or `IsNestedIn` with the arguments reversed.

klausler: Ok, done.

PR #315, opened 2019-03-06
Create framework for checking statement semantics

tskeith:
Add `SemanticsVisitor` as the visitor class to perform statement
semantics checks. Its template parameters are "checker" classes
that perform the checks. They have `Enter` and `Leave` functions
that are called for the corresponding parse tree nodes (`Enter`
before the children, `Leave` after). Unlike `Pre` and `Post` in
visitors they cannot prevent the parse tree walker from visiting
child nodes.

Existing checks have been incorporated into this framework:
- `ExprChecker` replaces `AnalyzeExpressions()`
- `AssignmentChecker` replaces `AnalyzeAssignments()`
- `DoConcurrentChecker` replaces `CheckDoConcurrentConstraints()`

Adding a new checker requires:
- defining the checker class:
  - with BaseChecker as virtual base class
  - constructible from `SemanticsContext`
  - with Enter/Leave functions for nodes of interest
- add the checker class to the template parameters of `StatementSemantics`

Because these checkers and also `ResolveNamesVisitor` require tracking
the current statement source location, that has been moved into
`SemanticsContext`. `ResolveNamesVisitor` and `SemanticsVisitor`
update the location when `Statement` nodes are encountered, making it
available for error messages.

`AnalyzeKindSelector()` now has access to the current statement through
the context and so no longer needs to have it passed in.

Test `assign01.f90` was added to verify that `AssignmentChecker` is
actually doing something.

### DIFF
--- a/lib/semantics/semantics.cc
+++ b/lib/semantics/semantics.cc
@@ -48,10 +48,12 @@ template<typename... C> struct SemanticsVisitor : public virtual C... {
     return true;
   }
   template<typename N> void Post(const N &node) { Leave(node); }
+  void Walk(const parser::Program &program) { parser::Walk(program, *this); }
 };
 
-using StatementSemantics =
-    SemanticsVisitor<ExprChecker, AssignmentChecker, DoConcurrentChecker>;
+using StatementSemanticsPass1 = SemanticsVisitor<ExprChecker>;
+using StatementSemanticsPass2 =
+    SemanticsVisitor<AssignmentChecker, DoConcurrentChecker>;
 
 SemanticsContext::SemanticsContext(
     const common::IntrinsicTypeDefaultKinds &defaultKinds)
@@ -102,8 +104,11 @@ bool Semantics::Perform() {
   if (AnyFatalError()) {
     return false;
   }
-  StatementSemantics visitor{context_};
-  parser::Walk(program_, visitor);
+  StatementSemanticsPass1{context_}.Walk(program_);
+  if (AnyFatalError()) {
+    return false;
+  }
+  StatementSemanticsPass2{context_}.Walk(program_);
   if (AnyFatalError()) {
     return false;
   }
--- a/lib/semantics/semantics.cc
+++ b/lib/semantics/semantics.cc
@@ -24,12 +24,34 @@
 #include "scope.h"
 #include "symbol.h"
 #include "../common/default-kinds.h"
+#include "../parser/parse-tree-visitor.h"
 #include <ostream>
 
 namespace Fortran::semantics {
 
 static void DoDumpSymbols(std::ostream &, const Scope &, int indent = 0);
 static void PutIndent(std::ostream &, int indent);
+
+// A parse tree visitor that calls Enter/Leave functions from each checker
+// class C supplied as template parameters. Enter is called before the node's
+// children are visited, Leave is called after. No two checkers may have the
+// same Enter or Leave function. Each checker must be constructible from
+// SemanticsContext and have BaseChecker as a virtual base class.
+template<typename... C> struct SemanticsVisitor : public virtual C... {
+  using C::Enter...;
+  using C::Leave...;
+  using BaseChecker::Enter;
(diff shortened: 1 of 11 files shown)


lib/semantics/semantics.h
```diff
@@ -113,6 +119,12 @@ class Semantics {
   parser::Program &program_;
   const parser::CookedSource &cooked_;
 };
-}
 
+// Base class for semantics checkers.
+struct BaseChecker {
+  template<typename C> void Enter(const C &x) {}
```
klausler: No argument name is necessary.

lib/semantics/semantics.cc
```diff
@@ -24,13 +24,35 @@
...
+// same Enter or Leave function. Each checker must be constructible from
+// SemanticsContext and have BaseChecker as a virtual base class.
+template<typename... C> struct SemanticsVisitor : public virtual C... {
+  using C::Enter...;
+  using C::Leave...;
+  using BaseChecker::Enter;
+  using BaseChecker::Leave;
+  SemanticsVisitor(SemanticsContext &context) : C{context}... {}
+  template<typename N> bool Pre(const N &node) {
+    Enter(node);
```
jeanPerier: Does the "No two checkers may have the same Enter or Leave function" limitation implies that two checkers that belongs to the same visitor cannot perform their checks on a same node?
Is it only a technical limitations or is it also a wish to avoid potential conflicts between the checkers?
If it is only a technical limitations, you could with a few changes call `(...,C::Enter(node));` instead of `Enter(node);` so that you can call multiple checkers on the same node.

lib/semantics/semantics.h
tskeith: OK

lib/semantics/semantics.cc
tskeith: The limitation is intentional, with the hope that it will help avoid conflicts. We will have to reconsider when we encounter the first case where two checkers want to do something on the same node.

We can easily add multiple passes of checkers when that happens, or relax the constraint as you suggest.

Another thing that might lead to more than one pass is the fact that expression checking is (at least for now) included in the first pass. If there were to be a checker that needed an expression to be analyzed before the `Pre` call on the `Expr` node, it would need to be in a later pass.

jeanPerier: OK thanks for detailing, I am fine with this conservative approach.

tskeith:
> I have some observation on using virtual inheritance in this commit (as well as in resolve-names.cc in semantics directory). I do realize that all methods of BaseChecker class are templates, hence they can't be virtual (not to mention pure virtual), in effect BaseChecker class can't be abstract. Still, adding 'virtual' keyword to public inheritance is not helping here as it has quite different purpose:
> 
> > Virtual base classes, used in virtual inheritance, is a way of preventing multiple "instances" of a givenclass appearing in an inheritance hierarchy when using multiple inheritance.

I'm not clear on what the concern is here. Each checker extends `BaseChecker` and `SemanticsVisitor` extends all of the checkers. Virtual inheritance is used to ensure there is only one copy of `BaseChecker` in the inheritance tree.

pawosm-arm:
> > I have some observation on using virtual inheritance in this commit (as well as in resolve-names.cc in semantics directory). I do realize that all methods of BaseChecker class are templates, hence they can't be virtual (not to mention pure virtual), in effect BaseChecker class can't be abstract. Still, adding 'virtual' keyword to public inheritance is not helping here as it has quite different purpose:
> > > Virtual base classes, used in virtual inheritance, is a way of preventing multiple "instances" of a givenclass appearing in an inheritance hierarchy when using multiple inheritance.
> 
> I'm not clear on what the concern is here. Each checker extends `BaseChecker` and `SemanticsVisitor` extends all of the checkers. Virtual inheritance is used to ensure there is only one copy of `BaseChecker` in the inheritance tree.

Indeed, SemanticsVisitor is a variadic template and that's how the magic (that requires virtual inheritance) is done.

PR #316, opened 2019-03-06
Add Expression traversal framework, use it to reimplement IsConstantE…

klausler:
…xpr()

This is a proof-of-concept for a template-based data structure traversal method that I discussed a bit with Tim this week.  The basic idea is that multiple visitation classes could be mixed together to perform a pass over a tree structure (`Expr<>` in this case).  This particular example just uses one visitation class to demonstrate a replacement for the suite of recursive functions that used to be used to implement `IsConstantExpr()` in fold.cc.

No rush on this review; I'm pushing it out there so that Tim and others can see it and comment on the approach, and think about whether it would work for parse tree traversal for executable statement semantics.


lib/evaluate/traversal.h
```diff
@@ -0,0 +1,214 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef FORTRAN_EVALUATE_TRAVERSAL_H_
+#define FORTRAN_EVALUATE_TRAVERSAL_H_
+
+#include "expression.h"
+#include <type_traits>
+
+// Implements an expression traversal utility framework.
```
tskeith: I think we need a comment explaining how to use this framework. That your class should extend `TraversalBase`. What it means to implement `Handle` vs. `Pre` and `Post` (and what happens if you implement all three). What happens after you call `Return`.

klausler: Yes, of course.

tskeith:
In `check-do-concurrent.cc` we have `GatherReferencesFromExpression()` which walks a `parser::Expr` and find the symbols referenced. Would this framework be a better way to accomplish that, assuming the expression has been resolved?

klausler:
> In `check-do-concurrent.cc` we have `GatherReferencesFromExpression()` which walks a `parser::Expr` and find the symbols referenced. Would this framework be a better way to accomplish that, assuming the expression has been resolved?

Maybe.  I'll give it a try.

lib/evaluate/traversal.h
```diff
@@ -19,22 +19,54 @@
...
+  // from any void-valued client callback.
   template<typename A> std::nullptr_t Handle(const A &) { return nullptr; }
-  template<typename A> void Pre(const A &) {}
-  template<typename A> void Post(const A &) {}
-  template<typename... A> void Return(A &&... x) {
-    result_.emplace(std::move(x)...);
+  template<typename A> std::nullptr_t Pre(const A &) { return nullptr; }
+  template<typename A> std::nullptr_t Post(const A &) { return nullptr; }
+
+  void Stop() { done_ = true; }
```
tskeith: I think `Return()` would be a better name for this function, to make it clear it does the same thing as the other return, except for setting the result.

klausler:
> > In `check-do-concurrent.cc` we have `GatherReferencesFromExpression()` which walks a `parser::Expr` and find the symbols referenced. Would this framework be a better way to accomplish that, assuming the expression has been resolved?
> 
> Maybe. I'll give it a try.

Done and pushed; please take a look when convenient.

lib/semantics/check-do-concurrent.cc
```diff
@@ -299,10 +300,21 @@ static CS GatherLocalVariableNames(
   }
   return names;
 }
+
 static CS GatherReferencesFromExpression(const parser::Expr &expression) {
-  GatherSymbols gatherSymbols;
-  parser::Walk(expression, gatherSymbols);
-  return gatherSymbols.symbols;
+  // Use the new expression traversal framework if possible, for testing.
+  if (expression.typedExpr.has_value()) {
```
tskeith: It looks we would want expressions to be fully resolved before doing this. So perhaps expression resolution should be in a separate pass than these checks in PR #315.

klausler: Sure.  Once that's assured, this code can be simplified.

PR #317, opened 2019-03-07
Simplify Semantics::Perform

tskeith:
`Semantics::Perform` is mostly a series of calls followed by a check
for fatal errors. There is more error checking logic than real code.

To make it clearer, change each of the phases it calls to return true
on success so that `Perform` can just call them one after the other.

### DIFF
--- a/lib/semantics/semantics.cc
+++ b/lib/semantics/semantics.cc
@@ -37,18 +37,26 @@ static void PutIndent(std::ostream &, int indent);
 // children are visited, Leave is called after. No two checkers may have the
 // same Enter or Leave function. Each checker must be constructible from
 // SemanticsContext and have BaseChecker as a virtual base class.
-template<typename... C> struct SemanticsVisitor : public virtual C... {
+template<typename... C> class SemanticsVisitor : public virtual C... {
+public:
   using C::Enter...;
   using C::Leave...;
   using BaseChecker::Enter;
   using BaseChecker::Leave;
-  SemanticsVisitor(SemanticsContext &context) : C{context}... {}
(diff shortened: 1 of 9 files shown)


lib/semantics/semantics.cc
```diff
@@ -91,30 +99,13 @@ const Scope &SemanticsContext::FindScope(
...
-  }
-  StatementSemanticsPass2{context_}.Walk(program_);
-  if (AnyFatalError()) {
-    return false;
-  }
-  ModFileWriter writer{context_};
-  writer.WriteAll();
-  return !AnyFatalError();
+  return ValidateLabels(context_.messages(), program_) &&
+      parser::CanonicalizeDo(program_) &&  // force line break
```
tskeith: The comment is there to prevent clang-format from joining this line with the next.
Is there a better way to achieve that?

jeanPerier: You can also use `// clang-format off` just before and re-activate formatting with `// clang-format on`. But I think your comment is better because formatting is still applied inside your line.

schweitzpgi: `clang-format off` is a good option for a block of code, but—I agree with Jean—this comment is a better way to split a line and still get some formatting.

klausler: It's not the end of the world if the calls don't each occupy their own line.  But you might also try giving `ResolveNames` a longer name so it doesn't fit.

Or use template magic to build a tuple of lambdas from an initializer-list and call its members in order until one fails; the constructor call's items could have a trailing comma, like common::visitors, and clang-format might keep it prettier.  But that cure seems worse that the original problem.

PR #319, opened 2019-03-07
FIR update

schweitzpgi:
There are a number of changes.

  - makes FIR a first-class library in the build
  - reactions to some previous review comment fixes
  - start decomposing some of the "big step" statements
  - various bug fixes
  - port everything over to the new pointer wrapper interfaces


lib/FIR/statement.def
```diff
@@ -26,18 +26,18 @@
 #define LAST_TERM_STMT(num)
```
schweitzpgi: This file is obsolete. I'll delete it.

lib/FIR/stmt.h
```diff
@@ -22,13 +22,14 @@
 namespace Fortran::FIR {
```
schweitzpgi: This file is obsolete. I'll delete it.

lib/FIR/afforestation.cc
```diff
@@ -25,6 +25,10 @@
 
 namespace Fortran::FIR {
 namespace {
+template<typename A> Expression *ExprRef(const A &a) {
```
tskeith: It looks like `A` has to be `parser::Expr`. Nothing else in the parse tree has `typedExpr`. So this doesn't need to be a template.

tskeith: You could consider adding an overload that takes `const common::Indirection<parser::Expr>`. That's pretty common the the parse tree, so it would simplify some of the places you call `ExprRef`.

schweitzpgi: ok

schweitzpgi: good idea

lib/FIR/afforestation.cc
```diff
@@ -1317,84 +1337,150 @@ struct FortranIRLowering {
...
+    return nullptr;
   }
   CallArguments CreateCallArguments(
       const std::list<parser::ActualArgSpec> &arguments) {
     return CallArguments{};
   }
 
+  Expression VariableToExpression(const parser::Variable &var) {
+    auto maybe{semantics::AnalyzeVariable(semanticsContext_, var)};
+    return {std::move(*maybe)};
```
tskeith: Can't `maybe` be `std::nullopt`?

schweitzpgi: It shouldn't be (I don't think), but I can add a `CHECK` to get a nicer looking crater message.

lib/FIR/builder.h
```diff
@@ -15,106 +15,118 @@
...
 
 struct FIRBuilder {
   explicit FIRBuilder(BasicBlock &block)
     : cursorRegion_{block.getParent()}, cursorBlock_{&block} {}
-  template<typename A> Statement &Insert(A &&s) {
+  template<typename A> Statement *Insert(A &&s) {
     CHECK(GetInsertionPoint());
     auto *statement{new Statement(GetInsertionPoint(), s)};
-    return *statement;
+    return statement;
```
tskeith: This should probably be `common::Indirection` instead of a raw pointer so that ownership is clear.

lib/FIR/common.h
```diff
@@ -50,24 +50,13 @@ using FunctionType = evaluate::SomeType;  // TODO: what should this be?
 using AttributeList = std::vector<Attribute>;
 enum struct LinkageTypes { Public, Hidden, External };
 using Expression = evaluate::GenericExprWrapper;
```
tskeith: Why use `GenericExprWrapper` rather than `Expr<SomeType>`?

lib/FIR/mixin.h
```diff
@@ -23,9 +23,11 @@
 
 namespace Fortran::FIR {
 
+inline constexpr bool has_size(std::size_t size) { return size > 0; }
+
 template<typename T, typename E = void> struct SumTypeMixin {};
 template<typename T>  // T must be std::variant<...>
-struct SumTypeMixin<T, std::enable_if_t<std::variant_size_v<T>>> {
+struct SumTypeMixin<T, std::enable_if_t<has_size(std::variant_size_v<T>)>> {
```
tskeith: Can you add comments saying what these mixin classes do?

lib/FIR/afforestation.cc
klausler: Or use `maybe.value()`.

tools/f18/f18.cc
```diff
@@ -405,6 +416,10 @@ int main(int argc, char *const argv[]) {
       driver.dumpUnparse = true;
     } else if (arg == "-funparse-with-symbols") {
       driver.dumpUnparseWithSymbols = true;
+    } else if (arg == "-fdotty") {
+      driver.dumpGraph = true;
+    } else if (arg == "-fdebug-dump-linear-ir") {
+      driver.debugLinearFIR = true;
```
tskeith: The help message should be updated with these new options too.

lib/semantics/expression.cc
```diff
@@ -422,21 +422,20 @@ static MaybeExpr CompleteSubscripts(
...
+                        },
+                        [&](Component &&c) {
+                          return CompleteSubscripts(context,
+                              ArrayRef{std::move(c), std::move(subscripts)});
+                        },
+                        [&](auto &&) -> MaybeExpr {
+                          CHECK(!"bad base for ArrayRef");
+                          return std::nullopt;
+                        },
+                    },
```
tskeith: This might have been formatted with the wrong version of clang-format.

lib/FIR/afforestation.cc
```diff
@@ -1317,84 +1337,150 @@ struct FortranIRLowering {
...
+    } else {
+      handleDefinedAssignmentStmt(stmt);
+    }
+  }
+
+  struct AllocOpts {
+    std::optional<Expression *> mold;
+    std::optional<Expression *> source;
+    std::optional<Expression> stat;
+    std::optional<Expression> errmsg;
```
tskeith: Couldn't all four of these be the same type? Either `Expression *` or `std::optional<Expression>`?

lib/FIR/builder.h
klausler: And then `Indirection::Make()` can be used to create & construct the value.

schweitzpgi:
> Do I need to change something in my environment to build this new code? When I try to build it, I get error messages from make:
> 

Thanks, I'll take a look.

lib/FIR/afforestation.cc
```diff
@@ -1057,59 +1085,21 @@ const parser::Format *FindReadWriteFormat(
...
-              return AlwaysTrueExpression();  // FIXME
-            },
-        },
-        loopControl->u);
-  }
-  return AlwaysTrueExpression();
+
+// create an integer constant as an expression
+static Expression CreateConstant(int value) {
+  auto result{common::SearchTypes(
```
klausler: I think this code will create an expression of type INTEGER(kind=value) with a constant value of 1, which is probably not what you want to do.

lib/FIR/afforestation.cc
schweitzpgi: done

lib/FIR/afforestation.cc
schweitzpgi: Right. No, that's not correct.

lib/semantics/expression.cc
schweitzpgi: ok

lib/FIR/builder.h
schweitzpgi: Ownership is not being passed to the caller. The idiom here is that a factory constructs FIR statements. Those statements will often need to be referenced to construct more FIR, and all the FIR will have a lifetime longer than the factory.

lib/FIR/common.h
schweitzpgi: No special reason. `GenericExprWrapper` was the class in the parse tree, so it was just going with the flow. On the other hand, my intention is to use `Expression` everywhere in FIR so that replacing these with the optimal class can be done seamlessly.

lib/FIR/mixin.h
schweitzpgi: Yes

tools/f18/f18.cc
schweitzpgi: ok

lib/FIR/afforestation.cc
schweitzpgi: I did try that but ran into some problems that I can't quite recall now. I'll make a note to revisit.

tskeith:
> > Do I need to change something in my environment to build this new code? When I try to build it, I get error messages from make:
> 
> Thanks, I'll take a look.
> 
> I'm able to build the whole enchilada with clang 7. (There are warnings, but no errors.)

I'm seeing the same thing: builds with clang 7 but undefined symbols at link time with GCC.

Is there a specific version of LLVM we need to build with?

lib/FIR/afforestation.cc
```diff
@@ -941,19 +942,18 @@ static std::vector<SwitchRankStmt::ValueType> populateSwitchValues(
...
+                     const auto &e{exp.thing.thing.thing.value()};
+                     result.emplace_back(SwitchRankStmt::Exactly{ExprRef(e)});
+                   },
+                   [&](const parser::Star &) {
+                     result.emplace_back(SwitchRankStmt::AssumedSize{});
+                   },
+                   [&](const parser::Default &) {
+                     result.emplace_back(SwitchRankStmt::Default{});
+                   },
+               },
```
tskeith: It looks like this was formatted with the wrong clang-format.

lib/FIR/mixin.h
```diff
@@ -25,6 +30,7 @@ namespace Fortran::FIR {
 
 inline constexpr bool has_size(std::size_t size) { return size > 0; }
 
+// implementation of a (moveable) sum type (variant)
 template<typename T, typename E = void> struct SumTypeMixin {};
```
tskeith: How does this compare to `UNION_CLASS_BOILERPLATE` in `parse-tree.h`. I.e. how would I choose which to use if I wanted to create a sum type?

lib/FIR/builder.h
klausler: Can you please clarify the ownership of the FIR nodes?  If it's not being granted to the caller of the factory, but the FIR will outlive the factory, who owns it?

klausler:
> > > Do I need to change something in my environment to build this new code? When I try to build it, I get error messages from make:
> > 
> > 
> > Thanks, I'll take a look.
> > I'm able to build the whole enchilada with clang 7. (There are warnings, but no errors.)
> 
> I'm seeing the same thing: builds with clang 7 but undefined symbols at link time with GCC.
> 
> Is there a specific version of LLVM we need to build with?

I'd really like to clean up any warnings from clang 7 in existing code and add that compiler to my build & test scripts.  Is there a designated clang 7 binary here in house to which you can point me?

schweitzpgi:
Thanks.  I was able to reproduce the linker issue finally.

`/home/sw/thirdparty/llvm/linux86-64/latest/lib/libLLVMSupport.a(ManagedStatic.cpp.o): In function `void std::call_once<void (&)()>(std::once_flag&, void (&)())':
ManagedStatic.cpp:(.text._ZSt9call_onceIRFvvEJEEvRSt9once_flagOT_DpOT0_[_ZSt9call_onceIRFvvEJEEvRSt9once_flagOT_DpOT0_]+0xf): undefined reference to `std::__get_once_mutex()'
ManagedStatic.cpp:(.text._ZSt9call_onceIRFvvEJEEvRSt9once_flagOT_DpOT0_[_ZSt9call_onceIRFvvEJEEvRSt9once_flagOT_DpOT0_]+0x45): undefined reference to `std::__once_functor'
ManagedStatic.cpp:(.text._ZSt9call_onceIRFvvEJEEvRSt9once_flagOT_DpOT0_[_ZSt9call_onceIRFvvEJEEvRSt9once_flagOT_DpOT0_]+0xbb): undefined reference to `std::__set_once_functor_lock_ptr(std::unique_lock<std::mutex>*)'
ManagedStatic.cpp:(.text._ZSt9call_onceIRFvvEJEEvRSt9once_flagOT_DpOT0_[_ZSt9call_onceIRFvvEJEEvRSt9once_flagOT_DpOT0_]+0x11f): undefined reference to `std::__set_once_functor_lock_ptr(std::unique_lock<std::mutex>*)'
collect2: error: ld returned 1 exit status
`

All of these missing symbols are STL.  So it appears, at least for me, that it's not clang, but a C++ library mismatch.

lib/FIR/mixin.h
```diff
@@ -25,6 +30,7 @@ namespace Fortran::FIR {
...
 template<typename T>  // T must be std::variant<...>
```
klausler: If T must be std::variant<>, it would be better to templatize on `typename... T` and then use `std::variant<T...>` below in place of bare T.

lib/FIR/builder.h
schweitzpgi: Ownership is managed exactly like LLVM, since the code reuses LLVM containers.

> Say you want to build a compilation unit. That's an object. You create the object and reference it with a pointer. The compilation unit has three functions and six global variables. The program owns those nine objects, and they are referenced by pointers. Function A has 12 basic blocks. Those are objects that are owned by the function and referenced by pointers. Basic block "entry" has 57 instructions. The basic block owns those instructions, which are referenced by pointers.

klausler: That is a clear explanation.  Does this text or something similar appear in the FIR headers?  That would be great.

schweitzpgi: I'll add it.

lib/FIR/mixin.h
schweitzpgi: I've been considering replacing this with the code from `parse`, but not urgently.

lib/FIR/CMakeLists.txt
```diff
@@ -0,0 +1,30 @@
...
+  graph-writer.cc
+  procedure.cc
+  program.cc
+  region.cc
+  statements.cc
+)
+
+target_link_libraries(FortranFIR
+  FortranCommon
+  clangBasic
```
tskeith: After merging PR #321 you'll need to remove clangBasic here.

PR #320, opened 2019-03-07
Rearrange/simplify expr analysis code

klausler:
The organization of the expression semantics analysis code has an ExpressionAnalysisContext class defined in its header file to hold some state and an AnalyzeExpr function with many overloads and template instantiations to implement analysis of all the various parse tree nodes that can appear in an expression.  The intent was to hide most of the AnalyzeExpr overloads in the implementation file `expression.cc` and only expose the interfaces in the header file API that need to be used from the outside world.

I suspected that this organization was more trouble than it was worth, and it led to frequent confusion on my part, and I got tired of always passing around references to the contextual object as arguments to AnalyzeExpr.  So I tried dropping back to a single class with a lot of private member functions.  This puts some more noise into the header file to distract the reader, but I think it's a net win and it saves nearly 200 lines of code.  Also, I renamed ExpressionAnalysisContext to ExpressionAnalyzer.

This PR carves this code rearrangement out of some other work in progress so that this large messy non-behavioral change doesn't obscure actual work that's coming along later.

Reviewers: Rather than stepping through a large number of changed lines, you might be better off just looking at the files before/after.

### DIFF
--- a/lib/semantics/expression.cc
+++ b/lib/semantics/expression.cc
@@ -111,11 +111,6 @@ std::optional<DataRef> ExtractDataRef(std::optional<A> &&x) {
   return std::nullopt;
 }
 
-struct CallAndArguments {
-  ProcedureDesignator procedureDesignator;
-  ActualArguments arguments;
-};
-
 struct DynamicTypeWithLength : public DynamicType {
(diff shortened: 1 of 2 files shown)


klausler:
Ping.

klausler:
Thanks!

PR #322, opened 2019-03-09
Fix two ambiguous parsing cases by patching parse tree in expression semantics

klausler:
The ambiguous `A(1,2)` might be a function call, array element, or structure constructor.  The parser will parse it as a function call.  This patch adds code to expression semantics to detect when it should be replaced (prior to semantic analysis).  The replacement takes place in the parse tree so that any tools can benefit from a corrected parse.

Prior to this change, the only way I could test structure constructors was to force their parses through the use of no-op type parameters (e.g., `T(0)(1, 2)`).  Now that "bare" structure constructors can be parsed as such, they're tested by a new test that copies an earlier one and removes the no-op type parameters.

While I'm at it, replace the code that fixed a misparsed `A(1:2)` array section when it's meant to be a substring; similarly, the corrected parse is left in the parse tree.

### DIFF
--- a/lib/semantics/expression.cc
+++ b/lib/semantics/expression.cc
@@ -322,16 +322,13 @@ static void FixMisparsedSubstring(const parser::Designator &d) {
                     },
                     arrElement.base.u)}) {
               const Symbol &ultimate{symbol->GetUltimate()};
-              if (const auto *details{
-                      ultimate.detailsIf<semantics::ObjectEntityDetails>()}) {
-                if (const semantics::DeclTypeSpec * type{details->type()}) {
-                  if (!details->IsArray() &&
-                      type->category() == semantics::DeclTypeSpec::Character) {
-                    // The ambiguous S(j:k) was parsed as an array section
-                    // reference, but it's now clear that it's a substring.
-                    // Fix the parse tree in situ.
-                    mutate.u = arrElement.ConvertToSubstring();
-                  }
+              if (const semantics::DeclTypeSpec *type{ultimate.GetType()}) {
+                if (!ultimate.IsObjectArray() &&
+                    type->category() == semantics::DeclTypeSpec::Character) {
+                  // The ambiguous S(j:k) was parsed as an array section
+                  // reference, but it's now clear that it's a substring.
+                  // Fix the parse tree in situ.
+                  mutate.u = arrElement.ConvertToSubstring();
                 }
               }
             }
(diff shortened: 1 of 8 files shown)


tskeith:
Maybe the fact that expression analysis has to mutate the parse tree means it should be outside of the `SemanticsVisitor` framework. It could be in a mutating visitor; that would make it clear it's changing the parse tree and avoid the need for `const_cast`.

I assume that none of the other semantics passes will need to do that. Do you think that is correct?

lib/semantics/expression.cc
```diff
@@ -333,7 +298,52 @@ MaybeExpr ExpressionAnalyzer::TopLevelChecks(DataRef &&dataRef) {
...
+                        [](common::Indirection<parser::StructureComponent>
+                                &sc) { return sc.value().component.symbol; },
+                        [](auto &) -> Symbol * { return nullptr; },
+                    },
+                    arrElement.base.u)}) {
+              const Symbol &ultimate{symbol->GetUltimate()};
+              if (const auto *details{
+                      ultimate.detailsIf<semantics::ObjectEntityDetails>()}) {
+                if (const semantics::DeclTypeSpec * type{details->type()}) {
+                  if (!details->IsArray() &&
```
tskeith: You could use `ultimate.GetType()` and `ultimate.IsObjectArray()` here and that would avoid the need for `details`.

klausler:
> Maybe the fact that expression analysis has to mutate the parse tree means it should be outside of the `SemanticsVisitor` framework. It could be in a mutating visitor; that would make it clear it's changing the parse tree and avoid the need for `const_cast`.
> 
> I assume that none of the other semantics passes will need to do that. Do you think that is correct?

I think that we can list the cases where Fortran is ambiguous without recourse to symbolic information: function call vs array element reference vs unparameterized structure constructor; array section vs substring; statement function definition vs array assignment; DEC-style structure component references vs user-defined binary operator use.  Are there others?  Anyway, the first two really have to be handled on demand during name resolution and its necessary expression analysis.  The statement function one can be handled between declaration processing and executable statement semantics.  The fourth one seems best handled in the parser.

Since those first two are the onlyparse tree repairs that expression semantics will have to make, AFAICT, using const_cast in the midst of otherwise read-only access to the parse tree seems like the right approach.  I don't want the bulk of expression analysis to be at risk of inadvertently changing the parse tree.

lib/semantics/expression.cc
klausler: Yes, but then there would only be nine lines in a row afterwards with nothing but right braces on them.

PR #329, opened 2019-03-13
Add cmake option to link with FIR and LLVM

tskeith:
The cmake option -DLINK_WITH_FIR=ON causes the f18 driver to be linked
against FIR and LLVM. When it is set to OFF (the default), the call into
FIR is #ifdef'd out.

The source in lib/FIR is always built; this option only affects linking.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -14,6 +14,8 @@
 
 cmake_minimum_required(VERSION 3.9.0)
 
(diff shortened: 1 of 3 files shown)


schweitzpgi:
Building by omitting the "middle end" seems like a bad idea.

We should make f18 work as a project or tool inside the LLVM environment/ecosystem.

sscalpone:
We should flip the sense to default with FIR.

tskeith:
OK. PR #331

PR #330, opened 2019-03-13
f18-parse-demo

klausler:
Add a simple f18 parser demonstration tool (stripped down from the f18 scaffolding driver).

### DIFF
--- a/tools/f18/CMakeLists.txt
+++ b/tools/f18/CMakeLists.txt
@@ -36,4 +36,5 @@ add_executable(f18-parse-demo
 
 target_link_libraries(f18-parse-demo
   FortranParser
+  FortranEvaluate
 )
--- a/tools/f18/CMakeLists.txt
+++ b/tools/f18/CMakeLists.txt
@@ -36,5 +36,4 @@ add_executable(f18-parse-demo
 
 target_link_libraries(f18-parse-demo
   FortranParser
-  FortranSemantics
 )
--- a/tools/f18/CMakeLists.txt
+++ b/tools/f18/CMakeLists.txt
@@ -29,3 +29,12 @@ target_link_libraries(f18
   ${FORTRAN_FIR_LIB}
   ${LLVM_COMMON_LIBS}
 )
+
+add_executable(f18-parse-demo
+  f18-parse-demo.cc
+)
+
+target_link_libraries(f18-parse-demo
+  FortranParser
+  FortranSemantics
+)
--- a/tools/f18/f18-parse-demo.cc
+++ b/tools/f18/f18-parse-demo.cc
@@ -1,0 +1,476 @@
+// Copyright (c) 2018-2019, NVIDIA CORPORATION.  All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// F18 parsing demonstration.
+//   f18-parse-demo [ -E | -fdump-parse-tree | -funparse-only ]
+//     foo.{f,F,f77,F77,f90,F90,&c.}
+//
+// By default, runs the supplied source files through the F18 preprocessing and
+// parsing phases, reconstitutes a Fortran program from the parse tree, and
+// passes that Fortran program to a Fortran compiler identified by the $F18_FC
+// environment variable (defaulting to gfortran).  The Fortran preprocessor is
+// always run, whatever the case of the source file extension.  Unrecognized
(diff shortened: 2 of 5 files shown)


tools/f18/f18-parse-demo.cc
```diff
@@ -0,0 +1,476 @@
...
+  return tspec.tv_nsec * 1.0e-9 + tspec.tv_sec;
+}
+
+struct DriverOptions {
+  DriverOptions() {}
+  bool verbose{false};  // -v
+  bool compileOnly{false};  // -c
+  std::string outputPath;  // -o path
+  std::vector<std::string> searchDirectories{"."s};  // -I dir
+  std::string moduleDirectory{"."s};  // -module dir
```
tskeith: I don't think this is needed.

tools/f18/f18-parse-demo.cc
```diff
@@ -0,0 +1,476 @@
...
+  bool forcedForm{false};  // -Mfixed or -Mfree appeared
+  bool warnOnNonstandardUsage{false};  // -Mstandard
+  bool warningsAreErrors{false};  // -Werror
+  Fortran::parser::Encoding encoding{Fortran::parser::Encoding::UTF8};
+  bool parseOnly{false};
+  bool dumpProvenance{false};
+  bool dumpCookedChars{false};
+  bool dumpUnparse{false};
+  bool dumpUnparseWithSymbols{false};
```
tskeith: Not needed.

tools/f18/f18-parse-demo.cc
```diff
@@ -0,0 +1,476 @@
...
+#include "../../lib/common/default-kinds.h"
+#include "../../lib/parser/characters.h"
+#include "../../lib/parser/features.h"
+#include "../../lib/parser/message.h"
+#include "../../lib/parser/parse-tree-visitor.h"
+#include "../../lib/parser/parse-tree.h"
+#include "../../lib/parser/parsing.h"
+#include "../../lib/parser/provenance.h"
+#include "../../lib/parser/unparse.h"
+#include "../../lib/semantics/dump-parse-tree.h"
```
tskeith: It would make sense for this tool to only depend on parser and not semantics.

I guess ParseTreeDumper is in semantics because if a Name has a Symbol, that gets dumped. We could decouple ParseTreeDumper from semantics by passing in a Name dumping function. The version of that function for this tool would just print the name.

tools/f18/f18-parse-demo.cc
klausler: Right.

tools/f18/f18-parse-demo.cc
klausler: Right.

tools/f18/f18-parse-demo.cc
klausler: Right.  Does anybody use the parse tree dumper?  Maybe it just doesn't need the symbol dump.

tskeith: I use the parse tree dumper and I think it's useful for understanding the parse tree.
I would be okay with removing the symbol dump part. Mostly I use `-fdebug-dump-symbols` anyway.

klausler: I'll try moving the parse tree dumper into the parser proper.

klausler: Done.

klausler:
Adding Alexis Perry as a reviewer as I've heard she may be interested in an f18 parser demonstration tool.

tools/f18/CMakeLists.txt
```diff
@@ -36,4 +36,5 @@ add_executable(f18-parse-demo
 
 target_link_libraries(f18-parse-demo
   FortranParser
+  FortranEvaluate
```
tskeith: I think we could avoid this dependency by adding a dummy `GenericExprWrapper` to `f18-parse-demo.cc`.

Depending on evaluate doesn't seem right, but this cure may be worse than the disease.

klausler: I tried that.  Will try again.

tskeith: Adding this to `f18-parse-demo.cc` before the other includes seems to work:
```
#include "../../lib/common/indirection.h"
namespace Fortran::evaluate {
struct GenericExprWrapper {
  bool operator==(const GenericExprWrapper &) const { return false; }
};
}
DEFINE_OWNING_DESTRUCTOR(OwningPointer, Fortran::evaluate::GenericExprWrapper)
```

klausler: That's pretty much what I've discovered as well.  I'll take care of it.

tskeith: Maybe a cleaner approach would be to put the above in a .cc file in parser. Then the dummy version would hopefully be linked in if there wasn't a real version. So f18 would get the real one from evaluate and f18-parse-demo would get the dummy.

This would help other tools that need to depend only the parser, assuming there will be some. Otherwise they would each have to add their own dummy GenericExprWrapper.

klausler: We don't want duplicate entry points in the f18 libraries.  I've defined a stub file for the tools that f18-parse-demo can use.  This is a temporary situation and it needs a better fix.  Please stand by, I'll push new code shortly.

PR #334, opened 2019-03-15
FIR sync: replace the stubs for translating terminal symbols to expressions; st…

schweitzpgi:
…art pushing Value type into FIR; cleanup terminator succ_blocks; remove parser element from disassociate; etc.

### DIFF
--- a/lib/FIR/afforestation.cc
+++ b/lib/FIR/afforestation.cc
@@ -23,26 +23,6 @@
 #include "llvm/Support/FileSystem.h"
 #include "llvm/Support/raw_ostream.h"
 
-namespace Fortran::howdowedothis {
-
-// need to be able to convert variable-like things to Expressions, or forego
-// using Expressions entirely. The .typedExpr data member is only on
-// parser::Expr nodes, which is not sufficient.
-
-semantics::MaybeExpr AnalyzeVariable(
-    semantics::SemanticsContext &context, const parser::Variable &var) {
-  return {};
-}
-semantics::MaybeExpr AnalyzeName(
-    semantics::SemanticsContext &context, const parser::Name &name) {
-  return {};
-}
-semantics::MaybeExpr AnalyzeDataRef(
-    semantics::SemanticsContext &context, const parser::DataRef &dataRef) {
-  return {};
-}
-}
-
 namespace Fortran::FIR {
 namespace {
 Expression *ExprRef(const parser::Expr &a) { return &a.typedExpr.value(); }
@@ -66,7 +46,9 @@ struct LinearLabelBuilder {
   LinearLabelRef getNext() {
     LinearLabelRef next{counter++};
     auto cap{referenced.capacity()};
-    if (cap < counter) referenced.reserve(2 * cap);
+    if (cap < counter) {
+      referenced.reserve(2 * cap);
+    }
     referenced[next] = false;
     return next;
   }
@@ -108,8 +90,8 @@ struct LinearGoto {
 };
 
 struct LinearReturn
-  : public SumTypeCopyMixin<std::variant<const parser::FailImageStmt *,
-        const parser::ReturnStmt *, const parser::StopStmt *>> {
+  : public SumTypeCopyMixin<const parser::FailImageStmt *,
+        const parser::ReturnStmt *, const parser::StopStmt *> {
   SUM_TYPE_COPY_MIXIN(LinearReturn)
   template<typename T> LinearReturn(const T &stmt) : SumTypeCopyMixin{&stmt} {}
 };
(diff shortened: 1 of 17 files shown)


lib/FIR/value.h
```diff
@@ -0,0 +1,50 @@
...
+  Value() : SumTypeCopyMixin{NOTHING} {}
+};
+
+inline bool IsNothing(Value value) {
+  return std::visit(
+      common::visitors{
+          [](Nothing &) { return true; },
+          [](auto *) { return false; },
+      },
+      value.u);
```
tskeith: Could use `std::holds_alternative<Nothing>(value.u)` instead.

lib/FIR/value.h
```diff
@@ -0,0 +1,50 @@
...
+class Procedure;
+class DataObject;
+
+class Value : public SumTypeCopyMixin<Nothing, DataObject *, Statement *,
+                  BasicBlock *, Procedure *> {
+public:
+  SUM_TYPE_COPY_MIXIN(Value)
+  template<typename A> Value(A *a) : SumTypeCopyMixin{a} {}
+  Value(const Nothing &n) : SumTypeCopyMixin{n} {}
+  Value(Nothing &&n) : SumTypeCopyMixin{std::move(n)} {}
```
tskeith: Why would you want to move a Nothing? There is nothing in it to move.

lib/FIR/statements.h
```diff
@@ -283,30 +266,29 @@ class ActionStmt_impl : public Stmt_impl {
...
-  Statement *leftValue() const { return value_[0]; }
-  Statement *rightValue() const { return value_[1]; }
+  Value leftValue() const { return value_[0]; }
+  Value rightValue() const { return value_[1]; }
 
 private:
-  explicit IncrementStmt(Statement *v1, Statement *v2);
-  Statement *value_[2];
+  explicit IncrementStmt(Value v1, Value v2);
+  Value value_[2];
```
tskeith: Why not just have data members named `leftValue_` and `rightValue_`?

lib/FIR/statements.h
```diff
@@ -283,30 +266,29 @@ class ActionStmt_impl : public Stmt_impl {
...
-  Statement *rightValue() const { return value_[2]; }
+  Value direction() const { return value_[0]; }
+  Value leftValue() const { return value_[1]; }
+  Value rightValue() const { return value_[2]; }
 
 private:
-  explicit DoConditionStmt(Statement *dir, Statement *left, Statement *right);
-  Statement *value_[3];
+  explicit DoConditionStmt(Value dir, Value left, Value right);
+  Value value_[3];
```
tskeith: Same here.

lib/FIR/statements.h
```diff
@@ -592,6 +580,14 @@ class Statement : public SumTypeMixin<std::variant<ReturnStmt,  //
     parent->insertBefore(this);
   }
   std::string dump() const;
+
+  static std::size_t offsetof_impl() {
+    return reinterpret_cast<std::size_t>(&static_cast<Statement *>(nullptr)->u);
```
tskeith: I think this is undefined behavior, dereferencing a nullptr. Why do you need this?

lib/FIR/statements.h
```diff
@@ -592,6 +580,14 @@ class Statement : public SumTypeMixin<std::variant<ReturnStmt,  //
...
+  }
+  static Statement *From(Stmt_impl *stmt) {
+    return reinterpret_cast<Statement *>(
+        reinterpret_cast<char *>(stmt) - Statement::offsetof_impl());
```
tskeith: This has me baffled. Can you add a comment about what's going on and why it's safe?

lib/FIR/statements.h
```diff
@@ -602,6 +598,21 @@ inline std::list<BasicBlock *> succ_list(BasicBlock &block) {
...
+  return Statement::From(returnValue_);
+}
+
+inline ApplyExprStmt *GetApplyExpr(Statement *stmt) {
+  return std::visit(
+      common::visitors{
+          [](ApplyExprStmt &s) { return &s; },
+          [](auto &) -> ApplyExprStmt * { return nullptr; },
+      },
+      stmt->u);
```
tskeith: `return std::get_if<ApplyExprStmt>(&stmt->u);`

lib/FIR/mixin.h
```diff
@@ -66,33 +60,28 @@ struct SumTypeCopyMixin<T, std::enable_if_t<has_size(std::variant_size_v<T>)>> {
...
   ProductTypeMixin(const ProductTypeMixin &) = delete;
   ProductTypeMixin &operator=(const ProductTypeMixin &) = delete;
   ProductTypeMixin() = delete;
-  T t;
+  std::tuple<Ts...> t;
 };
 
 // implementation of a (moveable) maybe type
-template<typename T, typename E = void> struct MaybeMixin {};
 template<typename T>  // T must be std::optional<...>
```
tskeith: With this change the comment is not longer true.

lib/FIR/statements.h
schweitzpgi: This may not be the way to go.  I'll make an analogy to explain this particular implementation however.
Say we have a world of envelopes and contents.  For some purposes, it is great to deal with envelopes without regard to the letter inside them.  For other purposes, it is better to know what is inside the envelope. (Say maybe knowing which are bills, which are checks, etc.) All this was intended to be was a quick way to map from "this electric bill" back to the envelope it is packaged in. This should work fine in a world where we know all content is wrapped in its own envelope.
Unfortunately, C++ doesn't have a robust and cheap way of encoding this relationship apparently.
For your amusement, here's a link to a more elaborate "envelope finder" scheme in LLVM. http://llvm.org/docs/ProgrammersManual.html#the-waymarking-algorithm

lib/FIR/mixin.h
```diff
@@ -125,6 +114,12 @@ template<typename A, typename B> B &Unzip(B &out, A first, A last) {
       [](auto &&a) -> decltype(a.second) { return a.second; });
   return out;
 }
+
+template<typename A, typename B> B &UnzipSnd(B &out, A first, A last) {
```
tskeith: Can you add a comment like the ones on Zip and Unzip?

Also, I think the name should be spelled out, `UnzipSecond`.

lib/FIR/value.h
schweitzpgi: done

lib/FIR/value.h
schweitzpgi: removed

lib/FIR/mixin.h
schweitzpgi: removed

lib/FIR/afforestation.cc
```diff
@@ -865,30 +816,16 @@ struct ControlFlowAnalyzer {
...
-struct SwitchTypeArguments {
-  Evaluation exp;
+template<typename T> struct SwitchArgs {
+  Value exp;
   LinearLabelRef defLab;
-  std::vector<SwitchTypeStmt::ValueType> types;
+  std::vector<T> values;
   std::vector<LinearLabelRef> labels;
 };
+struct SwitchArguments : public SwitchArgs<SwitchStmt::ValueType> {};
```
tskeith: Why not `using SwitchArguments = SwitchArgs<SwitchStmt::ValueType>`?

lib/FIR/common.h
```diff
@@ -37,13 +37,30 @@
...
+CLASS_TRAIT(ValueTrait)
+
+class Value_impl {
+public:
+  using ValueTrait = std::true_type;
+
+  std::string dump() const { return {}; }
+};
+
+struct Nothing {};
```
tskeith: For what it's worth, there is a standard unit type: `std::monostate`.

klausler:
This PR was not approved before being merged.

schweitzpgi:
> This PR was not approved before being merged.

I can't find a button that will re-open this one.

PR #336, opened 2019-03-16
Fix parse of .TRUE._8

klausler:
Since we allow leading underscores on names as an extension, the parser was misparsing `.TRUE._8` as the application of a user-defined operator named `.TRUE.` to an object named `_8`.  User-defined operators must be distinct from logical literal constants and intrinsic operators.  Resolve the problem by attempting parses of primaries before parses of unary user-defined operators in the grammar.

### DIFF
--- a/lib/parser/token-parsers.h
+++ b/lib/parser/token-parsers.h
@@ -704,10 +704,22 @@ constexpr auto rawName{nonDigitIdChar >> many(nonDigitIdChar || digit)};
 TYPE_PARSER(space >> sourced(rawName >> construct<Name>()))
 constexpr auto keyword{construct<Keyword>(name)};
 
+constexpr auto logicalTRUE{
+    (".TRUE."_tok ||
+        extension<LanguageFeature::LogicalAbbreviations>(".T."_tok)) >>
+    pure(true)};
+constexpr auto logicalFALSE{
+    (".FALSE."_tok ||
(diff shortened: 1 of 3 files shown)


lib/parser/token-parsers.h
```diff
@@ -704,10 +704,22 @@ constexpr auto rawName{nonDigitIdChar >> many(nonDigitIdChar || digit)};
...
 constexpr auto keyword{construct<Keyword>(name)};
 
+constexpr auto logicalTRUE{
+    (".TRUE."_tok ||
+        extension<LanguageFeature::LogicalAbbreviations>(".T."_tok)) >>
+    pure(true)};
+constexpr auto logicalFALSE{
+    (".FALSE."_tok ||
+        extension<LanguageFeature::LogicalAbbreviations>(".F."_tok)) >>
+    pure(false)};
```
tskeith: If the LogicalAbbreviations language feature is enabled, we should also report an error if the user tries to define an operator named ".T." or ".F.", right? I happen to be working on that right now.

I'm happy to see this feature is off by default.

klausler: That is correct (although I did not test it).  This is one of the few extensions that we cannot enable by default without putting conforming programs at risk (unless we start tracking defined `OPERATOR(.foo.)` names as state in the parser).

PR #337, opened 2019-03-16
Revert "FIR sync: replace the stubs for translating terminal symbols …

schweitzpgi:
…to expressions; st…"


klausler:
It's probably more messy to revert that PR at this point than to leave it in, and I can still add comments to the original review.

schweitzpgi:
> It's probably more messy to revert that PR at this point than to leave it in, and I can still add comments to the original review.

I was trying to work through some of these various build issues on different machines and was having issues with git not getting synched on top of that.  Sorry about the mistaken premature merge.

PR #339, opened 2019-03-18
Squash free-form white space in prescanner around parentheses.

klausler:
Before and after `(`, and before `)`, in free-form source outside a preprocessing directive, discard any white space rather than squashing it into a single space character while prescanning.  This should give Tim canonical contiguous names in the cooked character stream for generic interfaces like `OPERATOR ( / )`.

### DIFF
--- a/lib/parser/prescan.cc
+++ b/lib/parser/prescan.cc
@@ -398,7 +398,7 @@ bool Prescanner::NextToken(TokenSequence &tokens) {
       if (*at_ == '\n') {
         // Discard white space at the end of a line.
       } else if (!inPreprocessorDirective_ &&
-          (previous == '(' || *at_ == ')')) {
+          (previous == '(' || *at_ == '(' || *at_ == ')')) {
         // Discard white space before/after parentheses, unless in a
         // preprocessor directive.  This helps yield space-free contiguous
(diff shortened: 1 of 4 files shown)


lib/parser/prescan.cc
```diff
@@ -391,11 +391,24 @@ bool Prescanner::NextToken(TokenSequence &tokens) {
...
-      // Discard white space at the end of a line.
       const auto theSpace{at_};
+      char previous{at_ <= start_ ? ' ' : at_[-1]};
       NextChar();
       SkipSpaces();
-      if (*at_ != '\n') {
+      if (*at_ == '\n') {
+        // Discard white space at the end of a line.
+      } else if (!inPreprocessorDirective_ &&
+          (previous == '(' || *at_ == '(' || *at_ == ')')) {
```
tskeith: What about when `previous` is ')'? As in: `use m, only: operator(+) , operator(-)`

We can skip the space after the ')' either here or when parsing. But skipping it here matches the description of "Discard whitespace before/after parentheses".

klausler: Any space after `)` is skipped by the parser and doesn't need to be squashed by the prescanner.  I improved the comment.

PR #340, opened 2019-03-19
Avoid capturing trailing space with parentheses

tskeith:
When we have cooked source like `generic :: operator(+) => ...`
the name of the symbol we want to create is "operator(+)" but the
parser includes the following space in the `CharBlock` it captures
for that generic spec.

The fix has two parts:
1. Define another user-defined string token matcherm `""_nosptok`.
   It works like `""_tok` except it does not skip trailing spaces.
2. Make use of it in closing parenthesis of the `parenthesized()`
   token matcher and the various `GenericSpec` matchers.

### DIFF
--- a/documentation/ParserCombinators.md
+++ b/documentation/ParserCombinators.md
@@ -143,7 +143,9 @@ is built.  All of the following parsers consume characters acquired from
   the combinator `/`.
 * `"..."_sptok` is a string match in which the spaces are required in
    free form source.
-* `parenthesized(p)` is shorthand for `"(" >> p / ")"`.
+* `"..."_nosptok` is like `"..."_tok` except that trailing spaces are
+  not skipped.
+* `parenthesized(p)` is shorthand for `"(" >> p / ")"_nosptok`.
 * `bracketed(p)` is shorthand for `"[" >> p / "]"`.
 * `nonEmptyList(p)` matches a comma-separated list of one or more
   instances of p.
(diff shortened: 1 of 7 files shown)


tskeith:
This potentially affects everywhere we use `parenthesized()` but I couldn't find any problems with that. One noticeable side effect was how source is associated with sub-expressions. For example,
```
dt31.f90:16:29: error: subscripts may be applied only to an object or component
  integer, parameter ::  m2 = qs(2) / qs(1)
                              ^^^^^
```
Without this change the compiler includes the following space in the source range (i.e. it prints one more caret).

tskeith:
> I do not follow when you need to add "_sptok" after parentheses and when "_tok" is fine.

It only makes a difference if we are saving the source location of a parse, i.e. it is nested inside a `SourcedParser`. In the bug I'm fixing it makes a difference because the source location of "operator(+)" is going to be used as the name of the symbol.

In the examples you give, they are not part of a name or expression, so we probably aren't saving the source location.

jeanPerier:
> It only makes a difference if we are saving the source location of a parse, i.e. it is nested inside a `SourcedParser`. In the bug I'm fixing it makes a difference because the source location of "operator(+)" is going to be used as the name of the symbol.

Ok, thanks. After looking at #338, the difference is clear.

tskeith:
> I had comments but must have failed to hit the right button to share them.
> 
> In general, the parsers that constitute the f18 parser consume trailing spaces so that parsers can generally assume that they don't have to consume leading spaces. Having a few token parsers that don't eat their trailing spaces might break parsers that run afterwards.
> 
> I would suggest instead trying to modify the source-capturing parser combinator to trim exterior white space. That would solve your problems and not affect parsing behavior.
> 
> If you keep this change, please fix the spelling of "training" to "trailing".

OK, PR #343 tries to follow your suggestion.

PR #343, opened 2019-03-19
Strip whitespace from parse tree source locations

tskeith:
This is instead of PR #340.

Parse tree nodes have `source` fields that map the node back to
a range in the cooked input stream. These are used for names and
to locate error messages.

This change strips spaces from the beginning and end of those
source ranges before they are saved. This fixes two problems:
1. For a statement like this: `generic :: operator(+) => ...`,
   the name of the symbol we want to create is "operator(+)" but
   the parser was including the following space in the CharBlock
   it captured.
2. Error messages referring to parts of expressions sometimes included
   extra spaces in the source locations. For example:
   ```
   t.f90:9:29: error: subscripts may be applied only to an object or component
    integer, parameter :: b = a(1) + 1
                              ^^^^^
   t.f90:4:8: error: operands to LOGICAL operation must be LOGICAL
    i = j .or. k
         ^^^^^^^
   ```

### DIFF
--- a/lib/parser/basic-parsers.h
+++ b/lib/parser/basic-parsers.h
@@ -1344,11 +1344,9 @@ public:
     auto result{parser_.Parse(state)};
     if (result.has_value()) {
       const char *end{state.GetLocation()};
-      if (start < end && start[0] == ' ') {
-        ++start;
+      for (; start < end && start[0] == ' '; ++start) {
       }
-      if (start < end && end[-1] == ' ') {
-        --end;
+      for (; start < end && end[-1] == ' '; --end) {
       }
       result->source = CharBlock{start, end};
     }
--- a/lib/parser/basic-parsers.h
+++ b/lib/parser/basic-parsers.h
@@ -1343,7 +1343,14 @@ public:
     const char *start{state.GetLocation()};
     auto result{parser_.Parse(state)};
     if (result.has_value()) {
-      result->source = CharBlock{start, state.GetLocation()};
+      const char *end{state.GetLocation()};
+      if (start < end && start[0] == ' ') {
+        ++start;
+      }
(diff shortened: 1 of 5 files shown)


lib/parser/basic-parsers.h
```diff
@@ -1343,7 +1343,14 @@ template<typename PA> class SourcedParser {
     const char *start{state.GetLocation()};
     auto result{parser_.Parse(state)};
     if (result.has_value()) {
-      result->source = CharBlock{start, state.GetLocation()};
+      const char *end{state.GetLocation()};
+      if (start < end && start[0] == ' ') {
```
klausler: Maybe these should be `while` loops out of an overabundance of caution?

tskeith: OK. I actually had it that way initially.

tskeith:
> LGTM. If this works out, we can simplify the code in the prescanner that deletes spacesbefore `(`.

I think we still need the prescanner change for spaces in the middle of the source range, as in `operator ( + )`.

klausler:
> > LGTM. If this works out, we can simplify the code in the prescanner that deletes spacesbefore `(`.
> 
> I think we still need the prescanner change for spaces in the middle of the source range, as in `operator ( + )`.

Yes, internal spaces have to be removed by the prescanner, but not external ones before the `(`.

klausler:
> > > LGTM. If this works out, we can simplify the code in the prescanner that deletes spacesbefore `(`.
> > 
> > 
> > I think we still need the prescanner change for spaces in the middle of the source range, as in `operator ( + )`.
> 
> Yes, internal spaces have to be removed by the prescanner, but not external ones before the `(`.

Oh, wait -- you want to capture the source including the `OPERATOR` keyword before the `(`, right?  So never mind.

PR #344, opened 2019-03-20
Update the build instructions in README.md.

sscalpone:
Update README.md with a list of compilers that are known to compile f18. Explain how to configure LLVM on the cmake cmd line. Add more detail about how to compile with clang++.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,47 +1,92 @@
 <!--
-Copyright (c) 2018, NVIDIA CORPORATION.  All rights reserved.
+Copyright (c) 2018-2019, NVIDIA CORPORATION.  All rights reserved.
 -->
 
-# f18
+# F18
 
-F18 is a front-end for Fortran.
-It is intended to replace the existing front-end in the Flang compiler.
+F18 is a ground-up implementation of a Fortran frontend written in modern C++.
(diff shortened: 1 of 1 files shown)


psteinfeld:
Thanks for doing this update.

I would use "front end" as two words rather than "frontend".

Also, I'm personally guilty of this, but I recommend that we get into the habit of capitalizing "f18" to "F18" everywhere.

Lastly, I recommend that we include some actual commands that work for building F18 rather than instructions like "cmake <your custom options> where/you/put/the/source/f18".

README.md
```diff
@@ -1,85 +1,142 @@
...
 
-To get the correct LLVM and clang libraries included in your f18 build,
-set the environment variable
-`Clang_DIR`
-to the `lib/cmake/clang` directory in the clang install directory.
+Or GCC can be defined on the f18 cmake command line
+where `<GCC_DIRECTORY>` is the path to a GCC installation with bin, lib, etc:
+```
+cmake -DGCC=<GCC_DIRECTORY> ...
+```
```
tskeith: It looks like this paragraph isn't supposed to be here. It's not an alternative to the previous sentence.

sscalpone: Reordered so it makes sense.

Issue #288, opened 2019-02-15
Restrictions for exit statement not handled in do concurrent semantic checks


kiranchandramohan:
Restrictions for exit statement is not handled in do concurrent semantic checks.
C1167 states that an exit statement shall not appear in a "do concurrent" construct if it belongs to that construct or an outer construct.

For e.g : the exit from do concurrent via exit mydoc in the following example should not be allowed. While Flang catches this, F18 does not.
```
subroutine run_par(n)
  mydoc: do concurrent(j=1:n)
  mydo:    do k=1,n
             If (k==5) exit mydoc
             If (j==10) exit mydo
           end do mydo
         end do mydoc
end subroutine
```
I have the following questions,
-> Is the plan to cover this in the semantics for exit statement?
-> Are you currently working on this?
-> Would this be a well-defined work for me or my colleagues to get started with F18?

klausler:
f18 has a semantics pass that enforces the control flow restrictions on DO CONCURRENT already, so this missed case is probably a bug.  I'll assign this to @schweitzpgi initially, and he can answer your questions better than I can.

schweitzpgi:
Hi Kiran,

Yes, it looks like an oversight and C1167 isn't in the `do concurrent` checks yet. Thanks for reporting it.

The ultimate plan is to correctly enforce all of the constraints in the standard.  No one is presently working on this semantics check, as far as I know.  It should be straightforward to add this constraint to the `DoConcurrentLabelEnforce` class.  Please let me know if you'd like to do it or have questions.

kiranchandramohan:
Thanks Peter and Eric for your response. Yes, I would like to do this.

kiranchandramohan:
Fixed by the following commit. Thanks @sscalpone.
https://github.com/flang-compiler/f18/commit/f5f6ed8e983bbaffe078e4b206efe95a975a6c85

PR #345, opened 2019-03-20
C1167 : Check for exit statments in do-concurrent

kiranchandramohan:
Addresses https://github.com/flang-compiler/f18/issues/288

### DIFF
--- a/lib/semantics/check-do-concurrent.cc
+++ b/lib/semantics/check-do-concurrent.cc
@@ -40,6 +40,7 @@ class DoConcurrentEnforcement {
 public:
   DoConcurrentEnforcement(parser::Messages &messages) : messages_{messages} {}
   std::set<parser::Label> labels() { return labels_; }
+  std::set<parser::CharBlock> names() { return names_; }
   template<typename T> bool Pre(const T &) { return true; }
   template<typename T> void Post(const T &) {}
   template<typename T> bool Pre(const parser::Statement<T> &statement) {
@@ -47,6 +48,47 @@ public:
     if (statement.label.has_value()) {
       labels_.insert(*statement.label);
     }
+    return true;
+  }
+  // C1167
+  bool Pre(const parser::WhereConstructStmt &s) {
+    addName(std::get<std::optional<parser::Name>>(s.t));
+    return true;
+  }
+  bool Pre(const parser::ForallConstructStmt &s) {
+    addName(std::get<std::optional<parser::Name>>(s.t));
+    return true;
+  }
+  bool Pre(const parser::ChangeTeamStmt &s) {
+    addName(std::get<std::optional<parser::Name>>(s.t));
+    return true;
+  }
+  bool Pre(const parser::CriticalStmt &s) {
+    addName(std::get<std::optional<parser::Name>>(s.t));
+    return true;
+  }
+  bool Pre(const parser::LabelDoStmt &s) {
+    addName(std::get<std::optional<parser::Name>>(s.t));
+    return true;
+  }
+  bool Pre(const parser::NonLabelDoStmt &s) {
+    addName(std::get<std::optional<parser::Name>>(s.t));
+    return true;
+  }
+  bool Pre(const parser::IfThenStmt &s) {
+    addName(std::get<std::optional<parser::Name>>(s.t));
+    return true;
+  }
+  bool Pre(const parser::SelectCaseStmt &s) {
+    addName(std::get<std::optional<parser::Name>>(s.t));
+    return true;
+  }
+  bool Pre(const parser::SelectRankStmt &s) {
+    addName(std::get<0>(s.t));
+    return true;
+  }
+  bool Pre(const parser::SelectTypeStmt &s) {
+    addName(std::get<0>(s.t));
     return true;
   }
   // C1136
@@ -160,7 +202,13 @@ private:
     }
     return false;
   }
+  void addName(const std::optional<parser::Name> &nm) {
+    if (nm.has_value()) {
+      names_.insert(nm.value().source);
+    }
+  }
 
+  std::set<parser::CharBlock> names_;
   std::set<parser::Label> labels_;
   parser::CharBlock currentStatementSourcePosition_;
   parser::Messages &messages_;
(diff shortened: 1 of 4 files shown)


psteinfeld:
Note that the copyrights on all of the test files are out of date.

kiranchandramohan:
> Note that the copyrights on all of the test files are out of date.

Thanks @psteinfeld. Updated.

lib/semantics/check-do-concurrent.cc
```diff
@@ -160,25 +192,38 @@ class DoConcurrentEnforcement {
     }
     return false;
   }
+  bool addName(const std::optional<parser::Name> &nm) {
+    if (nm.has_value())
```
klausler: Please use braces on all structured statements for consistency with the rest of F18.  `clang-format` might be able to add them for you.

It would be more clear if this function returned `void` and its callers returned explicit `true`, rather than forwarding a `true` result back.  As written, these `Pre` routines look as if their return values that guide the parse tree traversal might be determined by the prior presence of the names in the set.

lib/semantics/check-do-concurrent.cc
```diff
@@ -160,25 +192,38 @@ class DoConcurrentEnforcement {
...
+      names_.insert(nm.value().ToString());
+    return true;
+  }
 
+  bool inDoConcurrent{false};
```
klausler: We conventionally append an underscore `_` to the names of private data members in classes.

lib/semantics/check-do-concurrent.cc
```diff
@@ -160,25 +192,38 @@ class DoConcurrentEnforcement {
...
+                           std::set<parser::Label> &&labels,
+                           std::set<std::string> &&names)
+    : messages_{messages}, labels_{labels}, names_{names} {}
   template<typename T> bool Pre(const T &) { return true; }
   template<typename T> bool Pre(const parser::Statement<T> &statement) {
     currentStatementSourcePosition_ = statement.source;
     return true;
   }
+  bool Pre(const parser::DoConstruct &d) {
+    do_depth_++;
```
klausler: It's a very minor point, but it's somewhat more idiomatic to use pre-increment (`++do_depth_`) in modern C++.  (And pre-decrement below.)

lib/semantics/check-do-concurrent.cc
```diff
@@ -204,6 +249,20 @@ class DoConcurrentLabelEnforce {
...
   void Post(const parser::EndLabel &endLabel) { checkLabelUse(endLabel.v); }
   void Post(const parser::EorLabel &eorLabel) { checkLabelUse(eorLabel.v); }
+  void Post(const parser::DoConstruct &d) { do_depth_--; }
+  void checkName(const std::optional<parser::Name> &nm) {
+    if (!nm.has_value()) {
+      if (do_depth_ == 0) {
+        messages_.Say(currentStatementSourcePosition_,
+            "exit from DO CONCURRENT construct"_err_en_US);
+      }
+      return;
```
klausler: Might be more clear if written as an `else` rather than a `return`.

lib/semantics/check-do-concurrent.cc
```diff
@@ -204,6 +249,20 @@ class DoConcurrentLabelEnforce {
...
+    }
+    if (names_.find(nm.value().ToString()) == names_.end()) {
+      messages_.Say(currentStatementSourcePosition_,
+          "exit from DO CONCURRENT construct"_err_en_US);
```
schweitzpgi: You might consider adding the name of the DO construct to the error message.

lib/semantics/check-do-concurrent.cc
```diff
@@ -204,6 +249,20 @@ class DoConcurrentLabelEnforce {
...
+  void Post(const parser::DoConstruct &d) { do_depth_--; }
+  void checkName(const std::optional<parser::Name> &nm) {
+    if (!nm.has_value()) {
+      if (do_depth_ == 0) {
+        messages_.Say(currentStatementSourcePosition_,
+            "exit from DO CONCURRENT construct"_err_en_US);
+      }
+      return;
+    }
+    if (names_.find(nm.value().ToString()) == names_.end()) {
```
schweitzpgi: I wonder if a comment that the nesting of named constructs is assumed to have been (and should be!) previously checked by the label resolution pass.

test/semantics/doconcurrent06.f90
```diff
@@ -0,0 +1,44 @@
...
+  integer :: j,k,n
+  mytest: if (n>0) then
+  mydoc:    do concurrent(j=1:n)
+              do concurrent(k=1:n)
+                if (k==5) exit
+                if (j==10) exit mytest
+              end do
+            end do mydoc
+          end if mytest
+end subroutine do_concurrent_test2
```
schweitzpgi: Should we have a test that interleaves a nest of DO and DO CONCURRENT loops, say, up to 4 or 6 total?

lib/semantics/check-do-concurrent.cc
```diff
@@ -160,25 +192,38 @@ class DoConcurrentEnforcement {
     }
     return false;
   }
+  bool addName(const std::optional<parser::Name> &nm) {
+    if (nm.has_value())
+      names_.insert(nm.value().ToString());
```
tskeith: There is no need to create a string here. Instead you could make names_ a set of parser::CharBlock. There here insert `nm.value().source`.

sscalpone:
Hi Kiran, thanks for the pull request.  I'll merge this commit after the issues raised in the comments are resolved.  (The NVIDIA convention has been to approve a pull request with the understanding that issues raised in comments will be addressed or further discussed.)

My thought is that the person who opened the issue should resolve the issue, so please respond with more comments and commits.  Thanks!

lib/semantics/check-do-concurrent.cc
kiranchandramohan: Thanks @klausler. I tried using clang-format, but that showed some differences with the existing code. Just wanted to check whether you use clang-format with any settings or is it a plain run of clang-format?

kiranchandramohan: Please ignore my comment. I now find that running without any style options shows no differences.

klausler:
We use clang-format using the settings in the “.clang-format” file in the top-level directory.

lib/semantics/check-do-concurrent.cc
klausler: There's a `.clang-format` file in the top-level directory that contains the settings that we use.

lib/semantics/check-do-concurrent.cc
kiranchandramohan: Thanks, I have made the suggested change.

test/semantics/doconcurrent06.f90
kiranchandramohan: I have added some tests that interleave do and do concurrent in test/semantics/doconcurrent07.f90

lib/semantics/check-do-concurrent.cc
kiranchandramohan: I have added the suggested comment.

lib/semantics/check-do-concurrent.cc
kiranchandramohan: I have added the name of the do concurrent construct and also the name of the exit target in the error message. And this simplifies and increases the accuracy of testing. Thanks for the suggestion.

lib/semantics/check-do-concurrent.cc
kiranchandramohan: Made this change.

lib/semantics/check-do-concurrent.cc
kiranchandramohan: OK. done.

lib/semantics/check-do-concurrent.cc
kiranchandramohan: This was a stray variable. I removed it.

lib/semantics/check-do-concurrent.cc
kiranchandramohan: Added braces. Changed addName function to return void and its callers now return true.

kiranchandramohan:
Thanks @sscalpone. I have now pushed some changes addressing the review comments provided by @klausler , @schweitzpgi and @tskeith.

klausler:
Thanks for resolving all of my comments.

PR #346, opened 2019-03-20
Simplify indirection.h, remove needless classes

klausler:
While working on development in another branch, I was able to simplify `lib/common/indirection.h` by removing the template classes `OwningPointer` and `ForwardReference` in favor of using other techniques to allow `std::unique_ptr<>` and `Indirection<>` on forward-referenced undefined types.  It would have been great if I had figured out these techniques earlier and avoided creating the needless template classes in the first place, but at least I'm cleaning them up now.

This PR extracts those simplifications from that development branch as a stand-alone change (pretty much).  Many more changes are coming, and they'll be easier to review without these in the way.

The documentation is also updated.

### DIFF
--- a/lib/parser/unparse.cc
+++ b/lib/parser/unparse.cc
@@ -836,7 +836,7 @@ public:
     Walk(std::get<2>(x.t));  // right
   }
   void Unparse(const DefinedOpName &x) {  // R1003, R1023, R1414, & R1415
-    Put('.'), Walk(x.v), Put('.');
+    Walk(x.v);
   }
   void Unparse(const AssignmentStmt &x) {  // R1032
     Walk(x.t, " = ");
--- a/lib/semantics/assignment.h
+++ b/lib/semantics/assignment.h
@@ -17,6 +17,7 @@
 
 #include "semantics.h"
 #include "../common/indirection.h"
+#include "../evaluate/expression.h"
 
 namespace Fortran::parser {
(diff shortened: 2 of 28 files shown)


lib/semantics/assignment.h
```diff
@@ -31,12 +32,23 @@ struct ForallStmt;
 struct ForallConstruct;
 }
 
+namespace Fortran::evaluate {
+void CheckPointerAssignment(parser::ContextualMessages &, const Symbol &,
+    const evaluate::Expr<evaluate::SomeType> &);
+}
```
tskeith: It looks like this declaration isn't supposed to be part of this change?

klausler: I left it in, it's just a prototype.

klausler:
Bonus extra with this PR: a fix to the unparser that gets f18 running again when emitting code for pgf90 or for parse/unparse/reparse/unparse fixed-point testing.

lib/parser/unparse.cc
```diff
@@ -836,7 +836,7 @@ class UnparseVisitor {
     Walk(std::get<2>(x.t));  // right
   }
   void Unparse(const DefinedOpName &x) {  // R1003, R1023, R1414, & R1415
-    Put('.'), Walk(x.v), Put('.');
+    Walk(x.v);
```
tskeith: Thanks for fixing this.

PR #351, opened 2019-03-22
implement the DO WHILE check

schweitzpgi:


lib/semantics/check-do-concurrent.cc
```diff
@@ -450,8 +450,35 @@ class DoConcurrentContext {
 private:
   bool ExpressionHasTypeCategory(const evaluate::GenericExprWrapper &expr,
       const common::TypeCategory &type) {
-    // TODO - implement
-    return false;
+    return std::visit(
```
klausler:
```
auto dynamicType{expr.v.GetType()};
return dynamicType.has_value() && dynamicType->category == type;
```

schweitzpgi: Thanks. I figured there must be a simpler way, but ...

PR #352, opened 2019-03-22
Fix some parsing issues found by Steve

klausler:
Steve reported six failing tests in the gfortran regression test suite; fix them.

The one nontrivial change is to move decimal->binary conversion of integer literal constants (that are actual data values, not kind codes or other values needed in semantics) from the parser into expression semantics, where overflow can be detected properly using the now-known kind of the integer type.  This means that those integer literals in the parse tree are now CharBlocks rather than 64-bit ints, which weren't working so well for 128-bit literals anyway.


lib/parser/token-parsers.h
```diff
@@ -377,6 +377,40 @@ struct BOZLiteral {
...
+      if (auto digits{digitString.Parse(state)}) {
+        result.ExtendToCover(*digits);
+        return result;
+      }
+    }
+    return std::nullopt;
+  }
+};
+
+constexpr struct DigitString64 {
```
tskeith: Is this still needed?

tskeith: Never mind -- I see it is used to declare digitString64.

klausler:
Thanks for the review.  I'll hold off on merging since I'll be away next week.

PR #354, opened 2019-03-25
FIR: continued development work

schweitzpgi:


lib/FIR/flattened.h
```diff
@@ -0,0 +1,264 @@
...
+constexpr ArtificialJump ARTIFICIAL{};
+
+// a source of an absolute control flow edge
+struct GotoOp
+  : public SumTypeCopyMixin<const parser::CycleStmt *, const parser::ExitStmt *,
+        const parser::GotoStmt *, ArtificialJump> {
+  template<typename A>
+  GotoOp(const A &stmt, LabelRef dest, const Location &source)
+    : SumTypeCopyMixin{&stmt}, target{dest}, source{source} {}
+  explicit GotoOp(LabelRef dest) : SumTypeCopyMixin{ARTIFICIAL}, target{dest} {}
```
tskeith: My assumption has been the once the FIR is built, the parse tree is no longer needed. But this seems to be saving pointers to parse tree nodes. Will these still be around when the FIR is completely built? Either way, it would be good to document what we need in terms of parse tree lifetime, though I don't have any idea where that should be.

lib/FIR/flattened.h
```diff
@@ -0,0 +1,264 @@
...
+// target for a control-flow edge
+struct LabelOp {
+  explicit LabelOp(LabelBuilder &builder);
+  LabelOp(const LabelOp &that);
+  LabelOp &operator=(const LabelOp &that);
+  void setReferenced() const;
+  bool isReferenced() const;
+  LabelRef get() const { return label_; }
+  operator LabelRef() const { return get(); }
+  void dump() const;
```
tskeith: Follow our naming conventions, these functions should be `SetReferenced()`, `IsReferenced()`, `Get()` (or `label()`), and `Dump()`.

lib/FIR/afforestation.cc
```diff
@@ -1057,27 +309,60 @@ class FortranIRLowering {
...
+    return evaluate::AsGenericExpr(evaluate::Relate(
+        context.GetContextualMessages(), op, std::move(e1), std::move(e2))
+                                       .value());
+  }
+  parser::Name MakeTemp(Type tempType) {
+    auto name{NewTemporaryName()};
+    auto details{semantics::ObjectEntityDetails{true}};
+    details.set_type(std::move(*tempType));
+    auto *sym{&semanticsContext_.globalScope().MakeSymbol(
+        name, {}, std::move(details))};
```
tskeith: If we ever try to refer to this symbol in a message, it won't work because the name is not part of the original cooked character stream so there is no provenance information. Is that a problem?

lib/FIR/statements.h
```diff
@@ -599,15 +577,26 @@ inline std::list<BasicBlock *> succ_list(BasicBlock &block) {
...
 
+inline AllocateLocalInsn *GetLocal(Statement *stmt) {
+  return std::get_if<AllocateLocalInsn>(&stmt->u);
+}
+
 Addressable_impl *GetAddressable(Statement *stmt);
+
+template<typename A> std::string ToString(const A *a) {
+  std::stringstream ss;
+  ss << std::hex << reinterpret_cast<std::intptr_t>(a);
```
tskeith: `ss << static_cast<const void *>(a);` would work here too.

lib/FIR/statements.h
```diff
@@ -585,8 +561,10 @@ class Statement : public SumTypeMixin<ReturnStmt,  //
     return reinterpret_cast<std::size_t>(&static_cast<Statement *>(nullptr)->u);
   }
   static Statement *From(Stmt_impl *stmt) {
-    return reinterpret_cast<Statement *>(
-        reinterpret_cast<char *>(stmt) - Statement::offsetof_impl());
+    char *cp{reinterpret_cast<char *>(stmt)};
+    auto adjust{Statement::offsetof_impl()};
+    CHECK(adjust == 0);
+    return reinterpret_cast<Statement *>(cp - adjust);
```
tskeith: I still find the confusing and unportable. If you have to get from a Stmt_impl back to the Statement that contains it, why not store a pointer to the Statement in the Stmt_impl?

lib/FIR/flattened.h
schweitzpgi: The flat form is an intermediate step along the lowering path. (This is just being moved and isn't anything new.) It does keep pointers into the parse tree (since it needs to reference the input/the tree to know how to proceed with further lowering), but the flat form is transitory and goes away.

lib/FIR/afforestation.cc
schweitzpgi: Our working premise is that semantics checking will have been fully completed, so hopefully issuing user input sorts of diagnostic messages is also (mostly?) completed.  Still, who can guarantee there will be no further messages?  As to not having provenance information, that seems like a given here. The "symbol" is made up by the compiler so it doesn't make sense for it to have any sort of context in the input source.

lib/FIR/flattened.h
schweitzpgi: Quoting from the project C++Style.md:

**In Brief**
> - Where **LLVM's** C++ style guide is clear on usage, follow it.

**Naming***
> 1. C++ names that correspond to well-known interfaces from the STL and **LLVM** can and should look like their models when the reader can safely assume that they mean the same thing -- e.g., clear() and size() member functions in a class that implements an STL-ish container.

Please fix the style guide so it isn't ambiguous.

lib/FIR/afforestation.cc
klausler: Semantics will have caught any fatal errors, but optimization can generate informational messages.  Any `CharBlock` in the symbol table should be mappable by the provenance system.   Maybe the symbol table should `CHECK()` that if it doesn't already. There is already a means for storing compiler-generated text there (it comes up when the prescanner or preprocessor has to insert a newline or spaces that weren't originally present).

lib/FIR/flattened.h
klausler: I will make the style guide more clear.  Consistency with established precedents (now that we have them) is more important.

lib/FIR/statements.h
```diff
@@ -581,12 +557,14 @@ class Statement : public SumTypeMixin<ReturnStmt,  //
...
+  // g++/clang++ will optimize this to a simple register copy
   static Statement *From(Stmt_impl *stmt) {
-    return reinterpret_cast<Statement *>(
-        reinterpret_cast<char *>(stmt) - Statement::offsetof_impl());
+    static Statement s{nullptr, UnreachableStmt::Create()};
+    auto *result{reinterpret_cast<Statement *>(reinterpret_cast<char *>(stmt) -
+        (reinterpret_cast<char *>(&s.u) - reinterpret_cast<char *>(&s)))};
+    CHECK(result == reinterpret_cast<Statement *>(stmt) &&
+        "expecting pointers to be equal");
+    return result;
```
tskeith: This still results in accessing a Statement through a pointer that is to a Stmt_impl. I don't think that behavior is defined.

If we need to get from a Stmt_impl to the Statement that contains it, why not use a pointer?

lib/FIR/flattened.h
schweitzpgi: I'm going to decline to change the names of all the member functions on the grounds that they conform with the LLVM coding standard from which they came, do not pose any significant challenge to understanding the code, and latitude is granted to the programmer in the F18 style guide.

lib/FIR/afforestation.cc
schweitzpgi: @klausler I didn't know about that feature in the prescanner. Can you provide a pointer to this? Thanks.

lib/FIR/statements.h
schweitzpgi: The actual point with this is to enforce constraints at the point of FIR construction rather than allow the representation to be constructed incorrectly and then stumble across the problem later.  This is a good thing. Particularly while I am working on the implementation. :)

Specifically for this function, this code is an early detection rather than to just use blind free-casting of the pointer, `reinterpret_cast<Statement*>(p)`. This is also a good thing.

Addressing some possible concerns:

The "guts" of a Statement will never appear on their own without a Statement wrapping them.  I am not worried that they might at some point become separable (as standalone "skin" and "guts"), as that would be a very fundamental and extensive alteration of the structure of the IR.

Concern that the layout of a class's data members in memory will change during execution also seems highly unlikely in practice.  C++ compilers do not (generally, to my knowledge) generate code that will rearrange a class's data members during runtime.

On the standards take: the Statement class is not a POD, so "offsetof" is not well-defined; that is true.  However, Statement, while it is composed of mixins, does not use virtual inheritance, and it is "POD enough" that its data members do not become virtualized indirections, etc.

Conclusion:

I think this can be left for now and handled as a separable TODO issue when FIR becomes more stabilized than it is at present.  Following in the same vein as other parts of this project, I anticipate doing experimentation on how the IR construction constraints can be elevated to the type system at some point.

Hope that helps.

lib/FIR/afforestation.cc
```diff
@@ -988,39 +209,70 @@ const parser::Format *FindReadWriteFormat(
...
+  auto len{strlen(cache)};
+  char *name{new char[len]};  // XXX: add these to a pool?
+  memcpy(name, cache, len);
+  return {name, name + len};
+}
+
+static TypeRep GetDefaultIntegerType() {
+  return {semantics::NumericTypeSpec{evaluate::SubscriptInteger::category,
+      AsExpr(evaluate::Constant<evaluate::SubscriptInteger>{
+          evaluate::SubscriptInteger::kind})}};
```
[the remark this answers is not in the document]
schweitzpgi: In surfing around through other code, I found this `SubscriptInteger` being used.  I don't really know if it is "best" way to do this.  But perhaps the name of the function could be improved as well.  I went surfing for a "default", but in the context of how this is *actually* being used, at least one of the cases really is computing a new expression for accessing into array objects.  So that, I'm guessing, should likely be somehow tied to other subscript type manipulations.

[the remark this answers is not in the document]
jeanPerier: Ok, it makes sens to produce SubscriptInteger if you are using them for subscripts, maybe the function could just be renamed `GetSubscriptIntegerType` to avoid any confusion with the default integer that may or may not be a subscript integer.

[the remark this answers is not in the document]
schweitzpgi: done

lib/FIR/afforestation.cc
sscalpone: Not for this check-in but something to think about:  We're in agreement that any errors have been reported before the CFG pass starts; however, like Peter says, the later analyses might offer information to the user so we have to anticipate that messages will pop out that are related to compiler-generated names.  We should think about a way to make these meaningful to the user (instead of, for example, trying to pattern match on such names & keeping the messages quiet).

lib/FIR/statements.h
sscalpone: Please add a comment to the code with a description of the data structures involved and what logically is happening with From().

lib/FIR/afforestation.cc
schweitzpgi: +1

PR #355, opened 2019-03-25
Update C++style.md

klausler:
Now that we have a bunch of code in the source base, adjust guidelines to recommend compatibility with established precedents.

### DIFF
--- a/documentation/C++style.md
+++ b/documentation/C++style.md
@@ -6,12 +6,13 @@ Copyright (c) 2018-2019, NVIDIA CORPORATION.  All rights reserved.
 * Use *clang-format* on all C++ source and header files before
   every merge to master.  All code layout should be determined
   by means of clang-format.
-* Where [LLVM's C++ style guide](https://llvm.org/docs/CodingStandards.html#style-issues)
-is clear on usage, follow it.
-* Otherwise, where a clear precedent exists in the project, follow it.
-* Otherwise, where a good public C++ style guide is relevant and clear,
-  follow it.  [Google's](https://google.github.io/styleguide/cppguide.html)
-  is pretty good and comes with lots of justifications for its rules.
+* Where a clear precedent for resolving a question of style has already
+  been established in the project, please be consistent with it.
+* Where a clear precedent for resolving a question of style has not
+  already been set, consider other style guides.  Being compatible with
+  [LLVM's C++ style guide](https://llvm.org/docs/CodingStandards.html#style-issues)
+  is good, and [Google's](https://google.github.io/styleguide/cppguide.html)
+  is worth a look and comes with lots of justifications for its rules.
 * Reasonable exceptions to these guidelines can be made.
 ## In particular:
 ### Files


sscalpone:
If the code in question is derived from llvm code, and looks more like llvm code, then it makes sense to follow the llvm naming style more directly in that code. (I don't know how this rule would be applied in the code in question; I'm just proposing it as a general rule.)

tskeith:
> If the code in question is derived from llvm code, and looks more like llvm code, then it makes sense to follow the llvm naming style more directly in that code. (I don't know how this rule would be applied in the code in question; I'm just proposing it as a general rule.)

I would say no. I think consistency within f18 should take precedence over consistency with llvm.

schweitzpgi:
> If the code in question is derived from llvm code, and looks more like llvm code, then it makes sense to follow the llvm naming style more directly in that code. (I don't know how this rule would be applied in the code in question; I'm just proposing it as a general rule.)

The FIR code leverages LLVM code.

klausler:
> The FIR code leverages LLVM code.

This is known.  The questions of precedents to follow are not immediately clarified, however, and may well not be the same in all situations.

For example, f18 code is pretty googley when it comes to accessors and mutators of private data members.  I don't think that new classes with accessors and mutators should be different from existing accessors and mutators elsewhere in f18 unless the new class inherits from an llvm class with a clear means of naming its accessors and mutators.

On the other hand, llvmish names like `dump()` might seem gratuitously weird to LLVM programmers if we changed them.

sscalpone:
I think the restriction of "inherits from llvm class" is too strict.  I think new code should be written in the style of the subsystem of which it is part.  So if the new code mostly interacts with llvm-ish code and is mostly in the style of llvm, then it should follow the naming conventions of llvm more directly.

tskeith:
> I think the restriction of "inherits from llvm class" is too strict. I think new code should be written in the style of the subsystem of which it is part. So if the new code mostly interacts with llvm-ish code and is mostly in the style of llvm, then it should follow the naming conventions of llvm more directly.

I think this is too vague to follow. What does "mostly interacts with" mean? If you have code written in the standard style and then add interactions with llvm, do you change the names to follow the llvm conventions? Over what scope -- function, file, namespace?

If you want to use two sets of naming conventions, there has to be easy-to-follow rules about which apply where.

schweitzpgi:
`namespace` boundaries seem reasonable.  It is highly desirable (even mandatory in some sense) to leverage other well-tested libraries and do so in a way that doesn't require ubiquitous and gratuitous case-and-underscore identifier replacement.  For example, the STL doesn't conform to the F18 style convention, and I haven't seen anyone seriously suggest it must undergo a comprehensive identifier replacement to make it conform.  Other libraries and the code that is building on them should be given the same consideration.

klausler:
> `namespace` boundaries seem reasonable. It is highly desirable (even mandatory in some sense) to leverage other well-tested libraries and do so in a way that doesn't require ubiquitous and gratuitous case-and-underscore identifier replacement. For example, the STL doesn't conform to the F18 style convention, and I haven't seen anyone seriously suggest it must undergo a comprehensive identifier replacement to make it conform. Other libraries and the code that is building on them should be given the same consideration.

What do you mean by "ubiquitous and gratuitous case-and-underscore identifier replacement"?

schweitzpgi:
> What do you mean by "ubiquitous and gratuitous case-and-underscore identifier replacement"?

I was parroting your line about the `dump()` example.  Who wants to remember that one set of classes uses `Dump()` while everywhere else in LLVM/clang/etc.-land, you use `dump()`? For another example, the same could apply to using STL iterators. Who would argue that it was an improvement to build custom container classes that use `Iterator`, `ConstIterator`, `Begin`, and `End` to blindly follow the style guide and thus immediately rule out using all the STL algorithms?

tskeith:
> > What do you mean by "ubiquitous and gratuitous case-and-underscore identifier replacement"?
> 
> I was parroting your line about the `dump()` example. Who wants to remember that one set of classes uses `Dump()` while everywhere else in LLVM/clang/etc.-land, you use `dump()`?

Using `Dump()` in f18 code and `dump()` in LLVM code seems perfectly reasonable to me. Using `Dump()` in parts of f18 and `dump()` in other parts less so.

> For another example, the same could apply to using STL iterators. Who would argue that it was an improvement to build custom container classes that use `Iterator`, `ConstIterator`, `Begin`, and `End` to blindly follow the style guide and thus immediately rule out using all the STL algorithms?

This is nonsense. Calling APIs is completely different from writing our own code.

PR #357, opened 2019-03-26
Track current statement location in SemanticsVisitor

tskeith:
Change `SemanticsVisitor` to track the location of the current
statement, if any, so that it's available through
`SemanticsContext::location()`.

Add overloading of `SemanticsContext::Say()` that reports the
message at the location of the current statement if a `CharBlock`
is not provided.

### DIFF
--- a/lib/semantics/mod-file.cc
+++ b/lib/semantics/mod-file.cc
@@ -680,7 +680,7 @@ std::optional<std::string> ModFileReader::FindModFile(
           : "Cannot find module file for submodule '%s' of module '%s'"_err_en_US,
       name.ToString().data(), ancestor.data()}};
   attachments.AttachTo(error);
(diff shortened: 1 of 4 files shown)


sscalpone:
For IF statement processing, for example, the source position was either the logical expression or the following action stmt.  I can imagine in the future wanting to push/pop current location in the context (instead of writing code to create a new temp context or locally saving/restoring the old value).  But for now, this is great.

klausler:
> For IF statement processing, for example, the source position was either the logical expression or the following action stmt. I can imagine in the future wanting to push/pop current location in the context (instead of writing code to create a new temp context or locally saving/restoring the old value). But for now, this is great.

The messaging system has a means for establishing default locations and adjusting them in a scoped way.

PR #360, opened 2019-03-27
Implement semantics for computed GOTO

sscalpone:
Note that a PGI extension will implicitly convert a float to integer (and issue a warning).  This commit does not implement that extension.

### DIFF
--- a/lib/semantics/check-computed-goto.cc
+++ b/lib/semantics/check-computed-goto.cc
@@ -1,0 +1,44 @@
+// Copyright (c) 2019, NVIDIA CORPORATION.  All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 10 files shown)


lib/semantics/check-computed-goto.cc
```diff
@@ -0,0 +1,44 @@
...
+// limitations under the License.
+
+#include "check-computed-goto.h"
+#include "attr.h"
+#include "scope.h"
+#include "semantics.h"
+#include "symbol.h"
+#include "tools.h"
+#include "type.h"
+#include "../evaluate/traversal.h"
```
tskeith: You probably don't need the includes of attr.h, scope.h, symbol.h, type.h, traversal.h.

sscalpone: Thanks!  I removed 'em.

PR #361, opened 2019-03-28
Fix compilation issues of PR #292

jeanPerier:
Fix the following compilation issues introduced by PR #292 :

- Silence GCC 8.1 spurious warning regarding missing return.
- MacOS lack of bessel function support:
j0, j1, jn, y0, y1 and yn are not standard C function but Unix extension and they seem not available on MacOS. Bessel folding default support has been removed. F18 can be built with Libpgmath in order to fold bessel functions.
- Power compilation failure:
Power long double is by default an IBM 128bits fp that is not IEEE compatible. Host.h was correctly detecting this, but an error was raised due to the instantiation of the unused ill-formed type in the related std::conditional type. This was fix by making the unused type valid. Also add guards around runtime function loading to make it clear that they may not be used if the host type do not match F18 scalar types.

- Improved the tests to also test folding of libpgmath only functions when libpgmath is linked with F18.


sscalpone:
MacOS has bessel function support.  Hope over to mojave1 & type "man j0" for example.

tskeith:
> MacOS has bessel function support. Hop over to mojave1 & type "man j0" for example.

Yes `j0` but not `j0f` or `j0l`.

jeanPerier:
> MacOS has bessel function support. Hope over to mojave1 & type "man j0" for example.

From GNU [j0 libc doc](https://www.gnu.org/software/gnulib/manual/html_node/j0f.html#j0f): 

> This function is missing on some platforms: Mac OS X 10.5, Minix 3.1.8, AIX 5.1, HP-UX 11, IRIX 6.5, mingw, MSVC 14.

Also windows defined all these functions with a leading "`_`" ([see doc](https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/bessel-functions-j0-j1-jn-y0-y1-yn?view=vs-2017)).

So, I have a feeling it might be trouble to use them because they are not C standard and I am not sure it is worth macro-ing the code, but we can.

Do you want to add the bessel double version only under a `#ifdef __GLIBC__` ? Maybe we can even risk the long double and float versions under `#ifdef __linux__ `

sscalpone:
Perhaps we should lobby the pgmath team to support standard interfaces for these functions?  Until we know, just make a note that we have to revisit and commit this change when it is ready otherwise.

lib/evaluate/fold.cc
```diff
@@ -392,10 +392,9 @@ Expr<Type<TypeCategory::Integer, KIND>> FoldOperation(FoldingContext &context,
                 std::is_same_v<From, Expr<SomeInteger>> ||
                 std::is_same_v<From, Expr<SomeComplex>>) {
               return Fold(context, ConvertToType<T>(std::move(x)));
-            } else {
-              common::die("int() argument type not valid");
-              return Expr<T>{std::move(funcRef)};  // unreachable
             }
+            common::die("int() argument type not valid");
+            return Expr<T>{std::move(funcRef)};  // unreachable
```
schweitzpgi: Does `return {};` fail to compile here?  I see the unreachable comment, but there is still a fair number of characters that look like they could do something before that.

jeanPerier: Yeah,` return {};` would require a constructor without arguments and Expr<T> do not have such constructor on purpose (they use EVALUATE_UNION_CLASS_BOILERPLATE from lib/evaluate/common.h that explicitly delete the constructor without arguments).
The code should compile without the return anyway because `common::die` is `[no return]` but we would get warnings from gcc....

schweitzpgi: I thought so.  What about wrapping that dummy `Expr<T>` ctor in a function and call it something like `UnreachableFakeValue()`?  It's fine with me to leave it as well.

jeanPerier: That is not a bad idea, I will not make it in this PR but maybe we could have in one of our common header something like:

```
template<typename T>
T UnreachableFakeValue() {
  common::die("Reached statement that is intended to be unreachable");
  return *reinterpret_cast<T*>(nullptr);
}
```
 We could use it here with:
`return UnreachableFakeValue<Expr<T>>();`

It would make the intention clear and offer dynamic checking.

jeanPerier: Anyway, I may be wrong and the [no return] aspect of common::die is sufficient here; gcc 7.2, gcc 8.1 and clang-7 are not giving warning if I completely remove the return. I think I have added this first for the lambda return type deduction but it is not useful anymore because I am specifying the return type explicitly. I will re-check with more compiler and remove the return later if I do not get warnings.

PR #365, opened 2019-03-29
Fix semantics test for "DO".

psteinfeld:
This test fails when the directory in which the tests are run contains a path that matches the pattern "do *[1-9]".  In my case, I had a directory called "do1".  The "*" is not necessary in the pattern.  Now the test passes unless the directory contains a " ".

### DIFF
--- a/test/semantics/canondo04.f90
+++ b/test/semantics/canondo04.f90
@@ -13,7 +13,7 @@
 ! limitations under the License.
 
 ! RUN: ${F18} -funparse-with-symbols %s 2>&1 | ${FileCheck} %s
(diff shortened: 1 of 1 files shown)


tskeith:
I'm not sure we even need the checking part of this test any more. ResolveNamesVisitor will report an internal error if it sees a `LabelDoStmt` so if this test compiles successfully the DO loops were rewritten.

psteinfeld:
Jean, I like your suggestion, but I'm not sure how to separate path names from the rest of the compiler output.

Note that this problem will not happen if the test does not produce compilation errors or the "CHECK-NOT" pattern contains a space.

PR #368, opened 2019-03-29
Rework .mod file writing for subprogram interfaces

tskeith:
And a couple of small name resolution changes.

### DIFF
--- a/test/semantics/modfile23.f90
+++ b/test/semantics/modfile23.f90
@@ -18,6 +18,7 @@ module m1
   integer(8) :: i
   type t1
     sequence
+    integer :: j
   end type
   type t2
   end type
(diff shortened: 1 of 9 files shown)


test/semantics/modfile23.f90
```diff
@@ -0,0 +1,184 @@
...
+! WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+! See the License for the specific language governing permissions and
+! limitations under the License.
+
+! Test that subprogram interfaces get all of the symbols that they need.
+
+module m1
+  integer(8) :: i
+  type t1
+    sequence
```
jeanPerier: Is t1 a legal type ? From C740 in 7.5.2.3:

> If SEQUENCE appears, the type shall have at least one component [...]

Maybe it does not matter for what you are testing here though.

tskeith: It doesn't matter for this test, but I don't want the test to fail when we start checking that constraint, so I'll fix it. Thanks for catching it.

PR #370, opened 2019-04-01
Fix #363: Adjust member function names in integer.h to avoid confusion.

### LINKED ISSUE #363: F18 DSHIFTR implementation swaps I and J arguments compare to the Standard
DSHIFTR(1, 16, 3) currently returns 0 when it should return 2**29 + 2 according to Fortran 18 Standard section 16.9.66.

It seems the implementation of DSHIFTR on F18 integer scalar type swaps I and J.
Section 16.9.66 defines DSHIFTR as:
`IOR (SHIFTL (I, BIT_SIZE (I)−SHIFT), SHIFTR (J, SHIFT)).`

In `lib/evaluate/integer.h`, the implementation of `constexpr Integer Integer::DSHIFTR(const Integer &fill, int count)`
is `SHIFTR(count).IOR(fill.SHIFTL(bits - count))`.
Considering `*this` maps to `I` and `fill` to `J`, it look like it is reversed.

If modified as suggested, `Real<W, P, IM>::Multiply(const Real &y, Rounding rounding)`, that uses DHSIFTR, should be modified to continue giving a correct result.

klausler:


klausler:
Jean, does this look okay to you?

PR #371, opened 2019-04-01
Fix #350: insert parens into formatted expressions only when required

### LINKED ISSUE #350: Extraneous parentheses when Expr is converted to Fortran
In some cases when an `evaluate::Expr` is converted to Fortran by using
`ExpressionBase::AsFortran()`, the result has extra parentheses around it.

This doesn't hurt anything, except that I am testing `.mod` file generation
by compiling the generated `.mod` files and expecting to get the same file
back out. In these cases the result is not identical because each time an
extra set of parentheses is added.

Example -- note declaration of `a` in `m.mod`:
```
$ cat m.f90
!mod$ v1 sum:f38582d2dfb50d39
module m
contains
subroutine s(n,a)
integer(4)::n
character((n+1_4),1)::a
end
end

$ f18 -fdebug-semantics -fparse-only m.f90

$ cat m.mod
!mod$ v1 sum:4c9142df0dc70a4c
module m
contains
subroutine s(n,a)
integer(4)::n
character(((n+1_4)),1)::a
end
end

```

klausler:
Gather the `AsFortran()` formatting member functions of the representational class templates in `lib/evaluate` together into one new `formatting.cc` source file.  Adjust the formatting of dyadic operations so that parentheses are added only where Fortran syntax would require them in order to parse the output correctly.  Add an overload of `operator<<` so that dumping to an `ostream` is more natural.


lib/evaluate/formatting.h
```diff
@@ -0,0 +1,55 @@
...
+template<typename A, bool COPYABLE>
+auto operator<<(
+    std::ostream &o, const Fortran::common::Indirection<A, COPYABLE> &x)
+    -> decltype(o << x.value()) {
+  return o << x.value();
+}
+
+template<typename A>
+auto operator<<(std::ostream &o, const std::optional<A> &x)
+    -> decltype(o << *x) {
```
tskeith: Why use `decltype` for the return type? Won't it always be `std::ostream&`?

klausler: Only if the contents (`typename A`) of the `optional` are themselves formattable; otherwise, SFINAE applies and this template is not instantiated for `std::optional<A>`.

lib/semantics/mod-file.cc
```diff
@@ -437,13 +437,13 @@ void PutTypeParam(std::ostream &os, const Symbol &symbol) {
 
 void PutInit(std::ostream &os, const MaybeExpr &init) {
   if (init) {
-    init->AsFortran(os << '=');
+    os << '=' << init;
```
tskeith: For writing .mod files I prefer staying with the `AsFortran` call. It provides a more precise guarantee about what it does: write the expression as Fortran. Whereas `operator<<` just means output this expression somehow.

test/semantics/modfile13.f90
```diff
@@ -38,9 +38,9 @@ subroutine s(n, a, b, c, d)
 !contains
 !  subroutine s(n,a,b,c,d)
 !    integer(4)::n
-!    character((n+1_4),1)::a
-!    character((n+2_4),2)::b
-!    character((n+3_4),1)::c
+!    character(n+1_4,1)::a
+!    character(n+2_4,2)::b
+!    character(n+3_4,1)::c
```
tskeith: Excellent, thank you.

lib/semantics/mod-file.cc
klausler: OK.

Issue #376, opened 2019-04-02
Internal Error when using structure constructor


kiranchandramohan:
The following Internal Error is observed when using a structure constructor in an internal subroutine. The error vanishes if the subroutine is inlined.
```
f18 up_test.f90 -fdebug-semantics

fatal internal error: CHECK(type != nullptr) failed at /home/kircha02/f18/lib/semantics/expression.cc(1791)
Aborted
```
Testcase
```
Program mn
  Type counter
    integer :: val = 10
  End Type
  call s()
Contains
  Subroutine s
    Type(counter) :: c
    c = counter()
    print *, c%val
  End Subroutine
End Program
```

klausler:
The semantic analysis of structure constructors is under active development when I am able to give attention to them.  Do not expect structure constructor semantic analysis to work correctly today.  Depending on the rate of new distractions, I hope to get back to them this week.

sscalpone:
Hi @kiranchandramohan ; thanks for the test case.  Right now, buts are best limited to those dealing with parsing and unparsing.  Stay tuned for when other parts of the compiler are ready for bug reports.

klausler:
Actually, I can pretty easily extract a fix from work in progress; will do.

kiranchandramohan:
Thanks @klausler for fixing this quickly.
@sscalpone I will await further info before filing bugs other than parsing and unparsing.

Issue #377, opened 2019-04-02
g++ 7.2 fails to compile


schweitzpgi:
I'm seeing the following error when building with g++ 7.2

f18/lib/evaluate/formatting.cc:220:51: error: no matching function for call to ?GetPrecedence(const Fortran::evaluate::NullPointer&)?
           [](const auto &x) { return GetPrecedence(x); },
                                      ~~~~~~~~~~~~~^~~

(Same code compiles with g++ 8.2, however.)

PR #378, opened 2019-04-02
Fix bugs #376, #377

klausler:
Extracts and improves some development code to avoid an internal error in expression semantics when rewriting a misparsed function call as a structure constructor (now that it's clear that's what it is), and tweaks some new code from yesterday to dodge a bogus compilation error from gcc 7.2.0.

### DIFF
--- a/lib/semantics/scope.cc
+++ b/lib/semantics/scope.cc
@@ -185,12 +185,17 @@ bool Scope::CanImport(const SourceName &name) const {
   }
 }
 
-const Scope *Scope::FindScope(const parser::CharBlock &source) const {
+const Scope *Scope::FindScope(parser::CharBlock source) const {
+  return const_cast<const Scope *>(
+      const_cast<Scope *>(this)->FindScope(source));
+}
+
+Scope *Scope::FindScope(parser::CharBlock source) {
   if (!sourceRange_.Contains(source)) {
     return nullptr;
   }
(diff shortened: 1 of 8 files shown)


lib/semantics/scope.cc
```diff
@@ -185,12 +185,17 @@ bool Scope::CanImport(const SourceName &name) const {
   }
 }
 
-const Scope *Scope::FindScope(const parser::CharBlock &source) const {
+const Scope *Scope::FindScope(parser::CharBlock source) const {
+  return const_cast<const Scope *>(
+      const_cast<Scope *>(this)->FindScope(source));
```
tskeith: I don't think the outer const_cast is needed.

lib/semantics/scope.cc
```diff
@@ -185,12 +185,17 @@ bool Scope::CanImport(const SourceName &name) const {
...
+}
+
+Scope *Scope::FindScope(parser::CharBlock source) {
```
tskeith: Why change to passing source by value?

klausler: CharBlocks are tiny values and the value semantics are more clear than const references, so I often make this change whenever I notice the usage.  Like an int, a CharBlock is just a context-free value.  Using a const reference to a simple value raises questions like "why does this have to be in memory?" and "will its address be taken?  do I have to worry about lifetimes?".

PR #383, opened 2019-04-03
Update OptionComparison.md based on feedback from Kiran.

psteinfeld:
Kiran provided several corrections, mostly related to Flang.

### DIFF
--- a/documentation/OptionComparison.md
+++ b/documentation/OptionComparison.md
@@ -1214,7 +1214,10 @@ IBM Fortran's options allow the source line length to be specified with the opti
 *   **GNU:** For both "ffixed-line-length-_n_" and "ffree-line-length-_n_" options, characters are ignored after the specified length.  The default for fixed is 72.  The default for free is 132.  For free, you can specify 'none' as the length, which means that all characters in the line are meaningful.
 *   **IBM:** For **fixed**, the default is 72.  For **free**, there's no default, but the maximum length for either form is 132.
 *   **Intel:** The default is 72 for **fixed** and 132 for **free**.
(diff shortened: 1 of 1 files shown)


documentation/OptionComparison.md
```diff
@@ -1214,7 +1214,7 @@ IBM Fortran's options allow the source line length to be specified with the opti
 *   **GNU:** For both "ffixed-line-length-_n_" and "ffree-line-length-_n_" options, characters are ignored after the specified length.  The default for fixed is 72.  The default for free is 132.  For free, you can specify 'none' as the length, which means that all characters in the line are meaningful.
 *   **IBM:** For **fixed**, the default is 72.  For **free**, there's no default, but the maximum length for either form is 132.
 *   **Intel:** The default is 72 for **fixed** and 132 for **free**.
-*   **PGI, Flang:** "-Mextend" instructs the compiler to accept 132-column source code; otherwise it accepts 72-column code.  It's unclear from the documentation whether this applies to fixed, free, or both source formats.
+*   **PGI, Flang:** By default, flang accepts up to 1000 characters per line in free form.  "-Mextend" instructs the compiler to accept 132-column source code; otherwise it accepts 72-column code.  It's unclear from the documentation whether this applies to fixed, free, or both source formats.
```
tskeith: This is my understanding of PGI's behavior (and probably flang's too):
- in free form, it is an error if the line is longer than 1000 characters
- in fixed form by default, characters after column 72 are ignored
- in fixed form with -Mextend, characters after column 132 are ignored

kiranchandramohan: Thanks @psteinfeld for making the changes. While you are right about the documentation not specifying Mextend for fixed or free, in practice it only applies to fixed form.

In Flang, an error message is produced in fixed form if there are more than 72 columns or more than 132 with Mextend.
Errors out with "F90-S-0026-Unmatched quote".

psteinfeld: Thanks, @kiranchandramohan and @tskeith.  I'll update the document.

PR #384, opened 2019-04-04
Support coarrays in name resolution

tskeith:
A coarray is represented as a `Symbol` with `ObjectEntityDetails` that
has a non-empty coshape. The coshape is represented using the same type
(`ArrayShape`) as the shape is, so the fact that it is a coshape is
determined from context.

Move code for analyzing shapes to `resolve-names-utils.cc` and
generalize it for coshapes.

In `symbol.cc` add dumping of coshapes. Simplify some of the functions
by adding some `Dump*` functions to handle common cases.

In `mod-file.cc` generalize the code for writing shapes to also write
coshapes. Fix a bug in `PutShapeSpec()`.

### DIFF
--- a/lib/semantics/resolve-names-utils.h
+++ b/lib/semantics/resolve-names-utils.h
@@ -18,10 +18,13 @@
 // Utility functions and class for use in resolve-names.cc.
 
 #include "symbol.h"
+#include "type.h"
 #include "../parser/message.h"
 
 namespace Fortran::parser {
 class CharBlock;
+struct ArraySpec;
+struct CoarraySpec;
 struct DefinedOpName;
 struct GenericSpec;
 struct Name;
@@ -30,6 +33,7 @@ struct Name;
 namespace Fortran::semantics {
 
 using SourceName = parser::CharBlock;
+class SemanticsContext;
 
 // Record that a Name has been resolved to a Symbol
 Symbol &Resolve(const parser::Name &, Symbol &);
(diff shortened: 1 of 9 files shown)


lib/semantics/resolve-names-utils.h
```diff
@@ -64,6 +68,12 @@ class GenericSpecInfo {
   void Analyze(const parser::GenericSpec &);
 };
 
+// Analyze a parser::ArraySpec or parser::CoarraySpec into the provide ArraySpec
```
klausler: "provided"

But why not return an `ArraySpec` as a value?

tskeith: I initially had it that way but changed it because this way was slightly simpler. I'm happy to change it back if you think returning a value is better.

klausler: It would be more clear, IMO, by removing any question about whether the result argument is meant to be updated in place, reassigned wholesale, or whatever.

tskeith:
This change doesn't attempt to do any of the semantic checks that aren't needed to generate symbol information for coarrays.

It also doesn't resolve codimension-decls in CHANGE TEAM statements.

klausler:
> This change doesn't attempt to do any of the semantic checks that aren't needed to generate symbol information for coarrays.
> 
> It also doesn't resolve codimension-decls in CHANGE TEAM statements.

Do you expect that the various restrictions on the use of coarrays in declarations will be checked in the name resolution pass with code to be added later, or should expression semantics do that checking?

tskeith:
> Do you expect that the various restrictions on the use of coarrays in declarations will be checked in the name resolution pass with code to be added later, or should expression semantics do that checking?

It probably should be done in name resolution.

lib/semantics/resolve-names-utils.h
tskeith: Done

PR #385, opened 2019-04-05
fix some bugs with use of std::move

schweitzpgi:
also anchors the virtual dtor in statements.cc

### DIFF
--- a/lib/FIR/mixin.h
+++ b/lib/FIR/mixin.h
@@ -30,7 +30,7 @@ namespace Fortran::FIR {
 
 // implementation of a (moveable) sum type (variant)
 template<typename... Ts> struct SumTypeMixin {
-  template<typename A> SumTypeMixin(A &&x) : u{std::move(x)} {}
+  template<typename A> SumTypeMixin(A &&x) : u{std::forward<A>(x)} {}
   using SumTypeTrait = std::true_type;
   SumTypeMixin(SumTypeMixin &&) = default;
   SumTypeMixin &operator=(SumTypeMixin &&) = default;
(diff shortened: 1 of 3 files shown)


lib/FIR/mixin.h
```diff
@@ -61,7 +60,7 @@ template<typename... Ts> struct SumTypeCopyMixin {
 
 // implementation of a (moveable) product type (tuple)
 template<typename... Ts> struct ProductTypeMixin {
-  template<typename A> ProductTypeMixin(A &&x) : t{std::move(x)} {}
+  ProductTypeMixin(Ts... x) : t{std::forward<Ts>(x)...} {}
```
klausler: You used rvalue reference arguments in the two cases above, but not here.

schweitzpgi: Right. This seemed to be what the reference was suggesting to do. I should be ok that some initializers of the tuple are moved and some copied as context appropriate.

lib/FIR/mixin.h
```diff
@@ -73,7 +72,7 @@ template<typename... Ts> struct ProductTypeMixin {
 
 // implementation of a (moveable) maybe type
 template<typename T> struct MaybeMixin {
-  MaybeMixin(T &&x) : o{std::move(x)} {}
+  MaybeMixin(T &&x) : o{std::forward<T>(x)} {}
```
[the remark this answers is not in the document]
schweitzpgi: OK, got it. Thanks.

lib/FIR/mixin.h
schweitzpgi: By the same argument as the MaybeMixin, it seems `std::forward` is not required here.

Thanks for the details. I'll play around with this some more.

schweitzpgi: The double ctor call was because the class needed a copy ctor (not the default) that didn't make copies of its arguments (i.e., used const T&).  In testing many permutations, `std::forward` is a good choice here as it allows a tuple to hold references and to be constructed with a mix of copies and moves as specified.  For example, `D d{std::move(e), f, &g, h->q()};`

PR #386, opened 2019-04-05
Shape analysis

klausler:
Implement analysis of array shapes (vector-valued integer expressions denoting the extents of the dimensions on an array object or expression).  Use these shapes to properly validate compilation-time constraints on the SHAPE= argument to the RESHAPE intrinsic, resolving a TODO that prevented a compilation of a test case noticed by Steve.  Represent references to fields in descriptors from expressions, which are needed to represent shapes of deferred- and assumed-shape objects.  Fix bugs encountered while testing.

### DIFF
--- a/lib/common/template.h
+++ b/lib/common/template.h
@@ -237,7 +237,7 @@ std::optional<std::tuple<A...>> AllElementsPresent(
 
 // std::vector<std::optional<A>> -> std::optional<std::vector<A>>
 // i.e., inverts a vector of optional values into an optional vector that
-// has a value of if all of the original elements were present.
+// will have a value only when all of the original elements are present.
 template<typename A>
 std::optional<std::vector<A>> AllElementsPresent(
     std::vector<std::optional<A>> &&v) {
--- a/lib/common/template.h
+++ b/lib/common/template.h
@@ -157,6 +157,21 @@ struct AreTypesDistinctHelper {
 template<typename... Ts>
 constexpr bool AreTypesDistinct{AreTypesDistinctHelper<Ts...>::value()};
 
+template<typename A, typename... Ts> struct AreSameTypeHelper {
+  using type = A;
+  static constexpr bool value() {
+    if constexpr (sizeof...(Ts) == 0) {
+      return true;
+    } else {
+      using Rest = AreSameTypeHelper<Ts...>;
+      return std::is_same_v<type, typename Rest::type> && Rest::value();
+    }
+  }
+};
+
+template<typename... Ts>
+constexpr bool AreSameType{AreSameTypeHelper<Ts...>::value()};
+
 template<typename> struct TupleToVariantHelper;
 template<typename... Ts> struct TupleToVariantHelper<std::tuple<Ts...>> {
   static_assert(AreTypesDistinct<Ts...> ||
--- a/lib/common/template.h
+++ b/lib/common/template.h
@@ -240,6 +240,10 @@ std::optional<R> MapOptional(
   }
   return std::nullopt;
 }
+template<typename R, typename... A>
+std::optional<R> MapOptional(R (*f)(A &&...), std::optional<A> &&... x) {
+  return MapOptional(std::function<R(A && ...)>{f}, std::move(x)...);
+}
 
 // Given a VISITOR class of the general form
 //   struct VISITOR {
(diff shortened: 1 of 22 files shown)


lib/common/template.h
```diff
@@ -219,6 +235,24 @@ std::optional<std::tuple<A...>> AllElementsPresent(
       std::move(t), std::index_sequence_for<A...>{});
 }
 
+// std::vector<std::optional<A>> -> std::optional<std::vector<A>>
+// i.e., inverts a vector of optional values into an optional vector that
+// has a value of if all of the original elements were present.
```
klausler: Typo here, will fix.

lib/evaluate/shape.h
```diff
@@ -0,0 +1,180 @@
...
+#include "type.h"
+#include "../common/indirection.h"
+#include <optional>
+#include <variant>
+
+namespace Fortran::evaluate {
+
+using ExtentType = SubscriptInteger;
+using ExtentExpr = Expr<ExtentType>;
+using MaybeExtent = std::optional<ExtentExpr>;
```
tskeith: There are lots of other places we use `Expr<SubscriptInteger>` and `std:optional<Expr<SubscriptInteger>>`. Should we use these same type aliases for all of them?

lib/evaluate/shape.h
```diff
@@ -0,0 +1,180 @@
...
+using Shape = std::vector<MaybeExtent>;
+
+// Convert between various representations of shapes
+Shape AsShape(const Constant<ExtentType> &arrayConstant);
+std::optional<Shape> AsShape(ExtentExpr &&arrayExpr);
```
tskeith: What does it mean when this returns a nullopt as opposed to an empty Shape?

lib/evaluate/shape.h
klausler: Maybe.  I used these names here to try to clarify the declarations that use them.  But maybe they're a distraction.  The whole business with `SubscriptInteger` is a messy work-around to a problem that really should be fixed within the Fortran standard (default REAL is single-precision IEEE-754, default INTEGER has to be the same bit size as default REAL, and yet we want to be able to have arrays and loop bounds bigger than 32 bits).

lib/evaluate/shape.h
klausler: A scalar has a valid shape that happens to be empty.  A failure to discern a shape at compilation time results in an absent value.

lib/evaluate/fold.cc
```diff
@@ -473,14 +474,48 @@ Expr<Type<TypeCategory::Integer, KIND>> FoldOperation(FoldingContext &context,
...
+      if (auto shape{GetShape(args[0].value())}) {
+        if (auto shapeExpr{AsShapeArrayExpr(*shape)}) {
+          return Fold(context, ConvertToType<T>(std::move(*shapeExpr)));
+        }
+      }
+    } else if (name == "size") {
+      if (auto shape{GetShape(args[0].value())}) {
+        if (auto &dimArg{args[1]}) {  // DIM= is present, get one extent
+          if (auto dim{ToInt64(dimArg->value())}) {
+            std::int64_t rank = shape->size();
```
tskeith: Use brace init here?

klausler: The result type of `size()` is unsigned and brace initialization elicits a warning about possible truncation.

lib/evaluate/shape.cc
```diff
@@ -0,0 +1,357 @@
...
+    return GetShape(*symbol);
+  } else if (const auto *intrinsic{
+                 std::get_if<SpecificIntrinsic>(&call.proc().u)}) {
+    if (intrinsic->name == "shape" || intrinsic->name == "lbound" ||
+        intrinsic->name == "ubound") {
+      return Shape{MaybeExtent{
+          ExtentExpr{call.arguments().front().value().value().Rank()}}};
+    } else if (intrinsic->name == "reshape") {
+      if (call.arguments().size() >= 2 && call.arguments().at(1).has_value()) {
+      }
```
tskeith: Is this a TODO?

klausler: No, it's something that I intended to get to before review, sorry; will complete it now.

PR #387, opened 2019-04-06
Update the grammar with team-value

tskeith:
Everywhere we had `team-variable` except for the FORM TEAM statement is
now `team-value`, according to the N2162 draft standard. The grammar
and parse tree are updated to reflect that.

The addition of the production for `team-value` caused productions
R1177-R1181 to have their numbers increased by one.

### DIFF
--- a/documentation/f2018-grammar.txt
+++ b/documentation/f2018-grammar.txt
@@ -342,7 +342,7 @@ R924 image-selector ->
        lbracket cosubscript-list [, image-selector-spec-list] rbracket
 R925 cosubscript -> scalar-int-expr
 R926 image-selector-spec ->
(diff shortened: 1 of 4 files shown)


tskeith:
I compared the productions in the standard against those in `f2018-grammar.txt` and there are a few other minor differences (looking at R number and lhs only):
1. R752: `bind-attr` should be `binding-attr`
2. R866: `implicit-name-spec` should be `implicit-none-spec`
3. R929-R942: `stat-variable` is now defined in R942, not R929, so R930-R942 should all be decremented by 1.

Assuming we want to fix these, the first two affect type names and the third doesn't affect code at all. All three affect comments that have productions in them.

PR #388, opened 2019-04-07
Check semantic constraints for the nullify statement.

sscalpone:
The full legality check is up to the user; however, some checks could
be done at compile time and others at runtime.  We choose to skip any
checking at compile time because it would not be complete.

### DIFF
--- a/lib/semantics/check-nullify.cc
+++ b/lib/semantics/check-nullify.cc
@@ -1,0 +1,67 @@
+// Copyright (c) 2019, NVIDIA CORPORATION.  All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "check-nullify.h"
+#include "expression.h"
+#include "tools.h"
+#include "../evaluate/expression.h"
+#include "../parser/message.h"
+#include "../parser/parse-tree.h"
+
+#include "../parser/dump-parse-tree.h"
+#include <iostream>
+
+namespace Fortran::semantics {
+
+void NullifyChecker::Leave(const parser::NullifyStmt &nullifyStmt) {
+  // R938
+  for (const parser::PointerObject &pointerObject : nullifyStmt.v) {
+    // R939
+    std::visit(
+        common::visitors{
+            [&](const parser::Name &name) {
+              auto const *symbol{name.symbol};
+              if (!IsVariableName(*symbol) && !IsProcName(*symbol)) {
+                context_.messages().Say(name.source,
+                    "name must be a variable or procedure pointer name"_err_en_US);
+              } else if (!IsPointer(*symbol)) {  // C951
+                context_.messages().Say(name.source,
+                    "name must have the POINTER attribute"_err_en_US);
+              }
+            },
+            [&](const parser::StructureComponent &structureComponent) {
+              evaluate::ExpressionAnalyzer analyzer{context_};
+              if (MaybeExpr checked{analyzer.Analyze(structureComponent)}) {
+                if (std::optional<evaluate::DataRef> dataRef{
+                        evaluate::ExtractDataRef(std::move(checked))}) {
+                  const Symbol &symbol{dataRef->GetLastSymbol()};
+                  if (!IsPointer(symbol)) {  // C951
+                    context_.messages().Say(structureComponent.component.source,
+                        "component must have the POINTER attribute"_err_en_US);
+                  }
+                }
+              }
+            },
+        },
+        pointerObject.u);
+  }
+  // From 9.7.3.1(1)
+  //   A pointer-object shall not depend on the value,
+  //   bounds, or association status of another pointer-
+  //   object in the same NULLIFY statement.
+  // This restriction is the programmer's responsibilty.
+  // Some dependencies can be found compile time or at
+  // runtime, but for now we choose to skip such checks.
+}
+}  // namespace Fortran::semantics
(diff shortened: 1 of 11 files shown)


lib/semantics/check-nullify.cc
```diff
@@ -0,0 +1,67 @@
...
+
+#include "check-nullify.h"
+#include "expression.h"
+#include "tools.h"
+#include "../evaluate/expression.h"
+#include "../parser/message.h"
+#include "../parser/parse-tree.h"
+
+#include "../parser/dump-parse-tree.h"
+#include <iostream>
```
tskeith: These two includes are probably left over from debugging.

lib/semantics/check-nullify.cc
```diff
@@ -0,0 +1,67 @@
...
+
+namespace Fortran::semantics {
+
+void NullifyChecker::Leave(const parser::NullifyStmt &nullifyStmt) {
+  // R938
+  for (const parser::PointerObject &pointerObject : nullifyStmt.v) {
+    // R939
```
tskeith: I don't see any benefit to including production numbers. Everything you need to know is in NullifyStmt and PointerObject.

lib/semantics/check-nullify.cc
```diff
@@ -0,0 +1,67 @@
...
+            },
+            [&](const parser::StructureComponent &structureComponent) {
+              evaluate::ExpressionAnalyzer analyzer{context_};
+              if (MaybeExpr checked{analyzer.Analyze(structureComponent)}) {
+                if (std::optional<evaluate::DataRef> dataRef{
+                        evaluate::ExtractDataRef(std::move(checked))}) {
+                  const Symbol &symbol{dataRef->GetLastSymbol()};
+                  if (!IsPointer(symbol)) {  // C951
+                    context_.messages().Say(structureComponent.component.source,
+                        "component must have the POINTER attribute"_err_en_US);
```
tskeith: I think that for all of the messages it would be better to indicate why the pointer attribute is required (i.e. because it is in a nullify statement). Maybe "Component in NULLIFY statement must have the POINTER attribute".

lib/semantics/check-nullify.cc
```diff
@@ -0,0 +1,67 @@
...
+                context_.messages().Say(name.source,
+                    "name must have the POINTER attribute"_err_en_US);
+              }
+            },
+            [&](const parser::StructureComponent &structureComponent) {
+              evaluate::ExpressionAnalyzer analyzer{context_};
+              if (MaybeExpr checked{analyzer.Analyze(structureComponent)}) {
+                if (std::optional<evaluate::DataRef> dataRef{
+                        evaluate::ExtractDataRef(std::move(checked))}) {
+                  const Symbol &symbol{dataRef->GetLastSymbol()};
```
tskeith: Can you just use `structureComponent.component.symbol` to get this symbol?

test/semantics/nullify01.f90
```diff
@@ -0,0 +1,27 @@
+INTEGER, PARAMETER :: maxvalue=1024
```
tskeith: Need a copyright

test/semantics/nullify02.f90
```diff
@@ -0,0 +1,28 @@
...
+Nullify(x(2)%p)
+
+!ERROR: name must have the POINTER attribute
+Nullify(pi)
+
+!ERROR: name must have the POINTER attribute
+Nullify(prp)
+
+!ERROR: name must be a variable or procedure pointer name
+Nullify(maxvalue)
```
tskeith: Maybe have a test with more than one object in the nullify statement.

lib/semantics/check-nullify.cc
sscalpone: What's the equivalent of GetLastSymbol?  (It returns 'p' in a%p.)

tskeith: `structureComponent.component.symbol` should be the symbol for `p`.

lib/semantics/check-nullify.cc
sscalpone: Yup.  Gone.

lib/semantics/check-nullify.cc
sscalpone: woo hoo!

lib/semantics/check-nullify.cc
```diff
@@ -19,37 +19,28 @@
...
+                    "name in NULLIFY statement must be a variable or procedure pointer name"_err_en_US);
               } else if (!IsPointer(*symbol)) {  // C951
                 context_.messages().Say(name.source,
-                    "name must have the POINTER attribute"_err_en_US);
+                    "name in NULLIFY statement must have the POINTER attribute"_err_en_US);
               }
             },
             [&](const parser::StructureComponent &structureComponent) {
               evaluate::ExpressionAnalyzer analyzer{context_};
               if (MaybeExpr checked{analyzer.Analyze(structureComponent)}) {
```
tskeith: I don't think the call to Analyze is needed either. The symbol should already be filled in by name resolution or we won't get this far.

sscalpone: The structure component might be complex, like a(i + 4)%b -- I think semantics needs to be called to check the "i + 4" for example, or that the index expression is an integer expression.  Does that happen in name resolution?

tskeith: Expression analysis (in ExprChecker) should analyze all of the top-level Exprs. That would include `i+4` in `nullify(a(i+4)%b)` because it is not a sub-expression of a bigger expression.

But I'm not sure about when `a(i+4)` gets analyzed because it's a `DataRef` that's not part of an `Expr`.

sscalpone: Right now, the StructureComponents in NULLIFY are not handled during the expression analysis phase.  Strictly speaking, it's not an Expr but we can decide to extend the top-level expression analysis to handle this case (and others, e.g. AllocateObject).  Let's talk to Peter about it.

lib/semantics/check-nullify.cc
klausler: Is that data reference really more readable, though?  `GetLastSymbol()` might be overkill but it's probably what I would have used here for clarity and as protection against any data structure changes.

lib/semantics/check-nullify.cc
klausler: It might be best to extend the grammar to accept an expression here, wrapped in a new constraint check.  Then the expression gets analyzed automatically, its representation gets saved for the CFG, and syntax errors get better recovery.

sscalpone: +1

PR #396, opened 2019-04-10
AArch64: Set flushing mode for subnormals on glibc and bionic based systems

pawosm-arm:


sscalpone:
Please re-do this pull request with a branch from master.   Otherwise, it looks good to me.

pawosm-arm:
opened again for different branch.

PR #397, opened 2019-04-10
AArch64: Set flushing mode for subnormals on glibc and bionic based systems

pawosm-arm:


lib/evaluate/host.cc
```diff
@@ -42,6 +42,24 @@ void HostFloatingPointEnvironment::SetUpHostFloatingPointEnvironment(
...
+    currentFenv_.__fpcr &= ~(1U << 24);  // control register
+  }
+#elif defined(__BIONIC__)
+  if (context.flushSubnormalsToZero()) {
+    currentFenv_.__control |= (1U << 24);  // control register
+  } else {
+    currentFenv_.__control &= ~(1U << 24);  // control register
+  }
+#else
+  // TODO other C libraries on AArch64
```
klausler: Why not have an `#error` or `static_assert` here?  Emitting an error message at the user doesn't seem right.  And they'll get spammed with a message every time this gets called.

pawosm-arm: It's just to keep consistency with the code immediately below.

klausler: Somebody's going to have to fix that code immediately below someday, I think, and now they'll have to fix it twice.

jeanPerier: @tskeith had already suggested `#warning`. I had not applied it. We should also change the other `#else` case then.
However, we should also white-list  host architectures that cannot control the option (e.g power (at least power 8), I am not sure for power9).
Then we should have a user warning if the requested subnormal flushing cannot be obtained on the host.
@Paul can you make both messages `#error` or static_assert ? I will handle the power case when adding the test case.

jeanPerier: @klausler, @pawosm-arm sorry for the confusion, I can fix all warnings in my branch PR.

pawosm-arm: @jeanPerier so do I still need to make changes?

jeanPerier: No, I will take care of changing both warning messages to compile time errors while taking care of Power and adding the test case.

PR #398, opened 2019-04-10
Fix issue 392, improve the error message when a DO loop is

psteinfeld:
I wanted to give more information about why the compiler puts out an error message when a DO loop
is terminated by a labeled statement that's not an END DO or CONTINUE.

I also cleaned up the code a little in resolve-labels.cc and added a test for new message.

### DIFF
--- a/lib/semantics/resolve-labels.cc
+++ b/lib/semantics/resolve-labels.cc
@@ -893,17 +893,19 @@ void CheckLabelDoConstraints(const SourceStmtList &dos,
           parser::MessageFormattedText{
               "label '%u' is not in scope"_en_US, SayLabel(label)});
     } else if (!doTarget.labeledStmtClassificationSet.test(
-                   TargetStatementEnum::Do) &&
-        !doTarget.labeledStmtClassificationSet.test(
-            TargetStatementEnum::CompatibleDo)) {
-      errorHandler.Say(doTarget.parserCharBlock,
-          parser::MessageFormattedText{
-              "'%u' invalid DO terminal statement"_err_en_US, SayLabel(label)});
-    } else if (!doTarget.labeledStmtClassificationSet.test(
                    TargetStatementEnum::Do)) {
-      errorHandler.Say(doTarget.parserCharBlock,
-          parser::MessageFormattedText{
-              "'%u' invalid DO terminal statement"_en_US, SayLabel(label)});
+      if (!doTarget.labeledStmtClassificationSet.test(
+              TargetStatementEnum::CompatibleDo)) {
+        errorHandler.Say(doTarget.parserCharBlock,
+            parser::MessageFormattedText{
+                "'%u' DO statements must terminate with END DO or CONTINUE"_err_en_US,
+                SayLabel(label)});
+      } else {
+        errorHandler.Say(doTarget.parserCharBlock,
+            parser::MessageFormattedText{
+                "'%u' Obsolete construct not allowed.  DO statements must terminate with END DO or CONTINUE"_en_US,
+                SayLabel(label)});
+      }
     } else {
       loopBodies.emplace_back(SkipLabel(position), doTarget.parserCharBlock);
     }
--- a/lib/semantics/resolve-labels.cc
+++ b/lib/semantics/resolve-labels.cc
@@ -898,12 +898,12 @@ void CheckLabelDoConstraints(const SourceStmtList &dos,
               TargetStatementEnum::CompatibleDo)) {
         errorHandler.Say(doTarget.parserCharBlock,
             parser::MessageFormattedText{
-                "'%u' DO statements must terminate with END DO or CONTINUE"_err_en_US,
+                "'%u' Obsolete construct.  A DO loop should terminate with END DO or CONTINUE"_err_en_US,
                 SayLabel(label)});
       } else {
         errorHandler.Say(doTarget.parserCharBlock,
             parser::MessageFormattedText{
-                "'%u' Obsolete construct not allowed.  DO statements must terminate with END DO or CONTINUE"_en_US,
+                "'%u' A DO loop should terminate with END DO or CONTINUE statement"_en_US,
                 SayLabel(label)});
       }
     } else {
--- a/lib/semantics/resolve-labels.cc
+++ b/lib/semantics/resolve-labels.cc
@@ -898,7 +898,7 @@ void CheckLabelDoConstraints(const SourceStmtList &dos,
               TargetStatementEnum::CompatibleDo)) {
         errorHandler.Say(doTarget.parserCharBlock,
             parser::MessageFormattedText{
-                "Only an END DO or CONTINUE should be used to terminate a labeled DO loop"_err_en_US,
+                "Only an END DO or CONTINUE must be used to terminate a labeled DO loop"_err_en_US,
                 SayLabel(label)});
       } else {
         errorHandler.Say(doTarget.parserCharBlock,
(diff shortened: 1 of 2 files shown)


lib/semantics/resolve-labels.cc
```diff
@@ -892,18 +892,20 @@ void CheckLabelDoConstraints(const SourceStmtList &dos,
...
+      if (!doTarget.labeledStmtClassificationSet.test(
+              TargetStatementEnum::CompatibleDo)) {
+        errorHandler.Say(doTarget.parserCharBlock,
+            parser::MessageFormattedText{
+                "'%u' DO statements must terminate with END DO or CONTINUE"_err_en_US,
+                SayLabel(label)});
+      } else {
+        errorHandler.Say(doTarget.parserCharBlock,
+            parser::MessageFormattedText{
+                "'%u' Obsolete construct not allowed.  DO statements must terminate with END DO or CONTINUE"_en_US,
```
klausler: This is a warning, so the obsolete construct actually is being allowed.  I would rephrase that warning to something like "A DO loop should be terminated with an END DO or CONTINUE statement."

lib/semantics/resolve-labels.cc
```diff
@@ -892,18 +892,20 @@ void CheckLabelDoConstraints(const SourceStmtList &dos,
...
     } else if (!doTarget.labeledStmtClassificationSet.test(
                    TargetStatementEnum::Do)) {
-      errorHandler.Say(doTarget.parserCharBlock,
-          parser::MessageFormattedText{
-              "'%u' invalid DO terminal statement"_en_US, SayLabel(label)});
+      if (!doTarget.labeledStmtClassificationSet.test(
+              TargetStatementEnum::CompatibleDo)) {
+        errorHandler.Say(doTarget.parserCharBlock,
+            parser::MessageFormattedText{
+                "'%u' DO statements must terminate with END DO or CONTINUE"_err_en_US,
```
klausler: The thing being terminated is a DO loop, not a DO statement.

klausler:
Old Fortran code decks are going to get this message a lot if it's enabled by default.  DO loops with single-line bodies are pretty idiomatic, e.g.
```
      DO 666 J=1,N
666   A(J)=B(J)*X
```
So we should either leave the warning off by default or make it really easy to disable.  Or make our standard conformance checking specific to particular revisions of the language rather than just to 2018.

lib/semantics/resolve-labels.cc
psteinfeld: Thanks.  I'll fix that.

schweitzpgi: This is confusing.  Specifically, the "must" part of the new messages were never the intention.

The original code is testing two things which are being spliced together here.  In the former case, the test is if the DO loop ends with an invalid statement altogether, and reports an _error_ in that case.  In the latter case, the test is for Fortran 2018 compliance that the DO loop ends with a CONTINUE or END DO.  This is a _warning_ only, because of the project-level goal of allowing F18 to compile legal Fortran programs as defined by previous standards of the language.

lib/semantics/resolve-labels.cc
psteinfeld: Sounds good.  I'll change the message.

schweitzpgi:
> Old Fortran code decks are going to get this message a lot if it's enabled by default. DO loops with single-line bodies are pretty idiomatic, e.g.
> 
> ```
>       DO 666 J=1,N
> 666   A(J)=B(J)*X
> ```
> So we should either leave the warning off by default or make it really easy to disable. Or make our standard conformance checking specific to particular revisions of the language rather than just to 2018.

Agreed.  This was written when there was no real facility other than to emit errors and warnings, as far as I recall.  Having a driver that can be more selective about the messages would be great.

lib/semantics/resolve-labels.cc
psteinfeld: I didn't change the analysis in the compiler that distinguishes between the error and warning cases.  Also, there is the third case of not producing a message at all.  This change is simply to improve the message that comes out in the warning and error cases by adding additional information about how to make the message go away.  In the current state of the compiler, only the warning will come out.  For that warning case, the latest version of the code produces the message "A DO loop should terminate with END DO or CONTINUE".  Is this a good change to make at this time?

lib/semantics/resolve-labels.cc
```diff
@@ -898,12 +898,12 @@ void CheckLabelDoConstraints(const SourceStmtList &dos,
               TargetStatementEnum::CompatibleDo)) {
         errorHandler.Say(doTarget.parserCharBlock,
             parser::MessageFormattedText{
-                "'%u' DO statements must terminate with END DO or CONTINUE"_err_en_US,
+                "'%u' Obsolete construct.  A DO loop should terminate with END DO or CONTINUE"_err_en_US,
```
klausler: The construct is obsolete in either case.  Maybe you could use the same text in both branches.

How about "Only an END DO or CONTINUE statement should be used to terminate a labeled DO loop."

schweitzpgi: That seems a reasonable message.  

There's little utility to putting in the message something to the effect that "an assignment (or ...) statement would also be fine, but only if you're not trying to have strict 2018 standard conformance".

psteinfeld: After talking to @schweitzpgi, I realized that the error message might come happen for real errors rather than just obsolete constructs, so we don't want to include that text.

It seems weird to me to have the same text for both the error and warning cases, but I can't think of a better alternative.

tskeith: You could say `Only an END DO or CONTINUE should be used...` for the warning
and `Only an END DO or CONTINUE may be used...` for the error.
"Should" implies it is a recommendation, not a requirement.

vdonaldson: A possible alternative to
  Only <...> may be used
is
  <...> must be used
(I'm currently using the "must" for in new messages.)

psteinfeld: Thanks, @tskeith and @vdonaldson.  I'll update the messages as per @vdonaldson's recommendation.

lib/semantics/resolve-labels.cc
```diff
@@ -892,18 +892,20 @@ void CheckLabelDoConstraints(const SourceStmtList &dos,
...
     } else if (!doTarget.labeledStmtClassificationSet.test(
                    TargetStatementEnum::Do)) {
-      errorHandler.Say(doTarget.parserCharBlock,
-          parser::MessageFormattedText{
-              "'%u' invalid DO terminal statement"_en_US, SayLabel(label)});
+      if (!doTarget.labeledStmtClassificationSet.test(
+              TargetStatementEnum::CompatibleDo)) {
+        errorHandler.Say(doTarget.parserCharBlock,
+            parser::MessageFormattedText{
+                "Only an END DO or CONTINUE must be used to terminate a labeled DO loop"_err_en_US,
```
tskeith: I believe Val's suggestion was "An END DO or CONTINUE must be used...", i.e. omit the "only". I think that's better.

lib/semantics/resolve-labels.cc
```diff
@@ -892,18 +892,20 @@ void CheckLabelDoConstraints(const SourceStmtList &dos,
...
+                SayLabel(label)});
+      } else {
+        errorHandler.Say(doTarget.parserCharBlock,
+            parser::MessageFormattedText{
+                "Only an END DO or CONTINUE should be used to terminate a labeled DO loop"_en_US,
+                SayLabel(label)});
```
tskeith: `SayLabel` was used in the original messages because the label number was included in the message. Now that it's not, `SayLabel(label)` is an unused argument.

PR #399, opened 2019-04-10
add QualifiedStmt class

schweitzpgi:
This starts to introduce a "type wrapper" that is used to enforce construction constraints on FIR statements at compile-time.  Previously, we used a run-time check at the point of construction.

Specifically, a `Statement` is a sum type of operations.  Certain operations require, for correctness, that an input from another `Statement` must be of a certain set of subclasses of `Statement`.  This can be encoded such that `QualifiedStmt<A>` requires that the argument is a qualified `Statement` of subclass `A`.

### DIFF
--- a/lib/FIR/statements.cc
+++ b/lib/FIR/statements.cc
@@ -54,9 +54,8 @@ static std::list<BasicBlock *> SuccBlocks(
   return result.second;
 }
 
-ReturnStmt::ReturnStmt(Statement *exp) : value_{GetApplyExpr(exp)} {
-  CHECK(value_);
-}
+ReturnStmt::ReturnStmt(QualifiedStmt<ApplyExprStmt> exp) : value_{exp} {}
+ReturnStmt::ReturnStmt() : value_{QualifiedStmt<ApplyExprStmt>{nullptr}} {}
 
 SwitchStmt::SwitchStmt(const Value &cond, const ValueSuccPairListType &args)
   : condition_{cond} {
@@ -126,19 +125,25 @@ LoadInsn::LoadInsn(Statement *addr) : address_{addr} {
   CHECK(GetAddressable(addr));
 }
 
-StoreInsn::StoreInsn(Statement *addr, Statement *val)
-  : address_{GetAddressable(addr)} {
+// Store ctors
+StoreInsn::StoreInsn(QualifiedStmt<Addressable_impl> addr, BasicBlock *val)
+  : address_{addr}, value_{val} {
   CHECK(address_);
-  if (auto *value{GetAddressable(val)}) {
-    value_ = value;
-  } else {
-    auto *expr{GetApplyExpr(val)};
-    CHECK(expr);
-    value_ = expr;
-  }
+  CHECK(val);
 }
-StoreInsn::StoreInsn(Statement *addr, BasicBlock *val)
-  : address_{GetAddressable(addr)}, value_{val} {
+StoreInsn::StoreInsn(QualifiedStmt<Addressable_impl> addr, Value val)
+  : address_{addr}, value_{val} {
+  CHECK(address_);
+}
+StoreInsn::StoreInsn(
+    QualifiedStmt<Addressable_impl> addr, QualifiedStmt<ApplyExprStmt> val)
+  : address_{addr}, value_{val} {
+  CHECK(address_);
+  CHECK(val);
+}
+StoreInsn::StoreInsn(
+    QualifiedStmt<Addressable_impl> addr, QualifiedStmt<Addressable_impl> val)
+  : address_{addr}, value_{val} {
   CHECK(address_);
   CHECK(val);
 }
--- a/lib/FIR/statements.h
+++ b/lib/FIR/statements.h
@@ -54,6 +54,23 @@ public:
   using StatementTrait = std::true_type;
 };
 
+// Some uses of a Statement should be constrained.  These contraints are imposed
+// at compile time.
+template<typename A = Stmt_impl> class QualifiedStmt {
+public:
+  QualifiedStmt() = delete;
+  template<typename B, std::enable_if_t<std::is_base_of_v<A, B>, int> = 0>
+  QualifiedStmt(Statement *stmt, const B &) : stmt{stmt} {}
+
+  // create a stub, where stmt == nullptr
+  QualifiedStmt(std::nullptr_t) : stmt{nullptr} {}
+  operator Statement *() const { return stmt; }
+  operator bool() const { return stmt; }
+  operator A *() const;
+
+  Statement *stmt;
+};
+
 // Every basic block must end in a terminator
 class TerminatorStmt_impl : virtual public Stmt_impl {
 public:
@@ -65,15 +82,18 @@ public:
 // Transfer control out of the current procedure
 class ReturnStmt : public TerminatorStmt_impl {
 public:
-  static ReturnStmt Create(Statement *stmt) { return ReturnStmt{stmt}; }
-  static ReturnStmt Create() { return ReturnStmt{nullptr}; }
+  static ReturnStmt Create(QualifiedStmt<ApplyExprStmt> stmt) {
+    return ReturnStmt{stmt};
+  }
+  static ReturnStmt Create() { return ReturnStmt{}; }
   std::list<BasicBlock *> succ_blocks() const override { return {}; }
   bool has_value() const { return value_; }
-  Statement *value() const;
+  Statement *value() const { return value_; }
 
 private:
-  ApplyExprStmt *value_;
-  explicit ReturnStmt(Statement *exp);
+  QualifiedStmt<ApplyExprStmt> value_;
+  explicit ReturnStmt(QualifiedStmt<ApplyExprStmt> exp);
+  explicit ReturnStmt();
 };
 
 // Encodes two-way conditional branch and one-way absolute branch
@@ -265,12 +285,15 @@ protected:
 class ApplyExprStmt : public ActionStmt_impl {
 public:
   static ApplyExprStmt Create(const Expression *e) { return ApplyExprStmt{*e}; }
-  static ApplyExprStmt Create(Expression &&e) { return ApplyExprStmt{e}; }
+  static ApplyExprStmt Create(Expression &&e) {
+    return ApplyExprStmt{std::move(e)};
+  }
 
   Expression expression() const { return expression_; }
 
 private:
   explicit ApplyExprStmt(const Expression &e) : expression_{e} {}
+  explicit ApplyExprStmt(Expression &&e) : expression_{e} {}
 
   Expression expression_;
 };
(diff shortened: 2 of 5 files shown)


lib/FIR/statements.h
```diff
@@ -378,25 +401,26 @@ class LoadInsn : public MemoryStmt_impl {
...
-      std::variant<Value, ApplyExprStmt *, Addressable_impl *, BasicBlock *>;
-  template<typename T> static StoreInsn Create(T *addr, T *value) {
-    return StoreInsn{addr, value};
-  }
-  template<typename T> static StoreInsn Create(T *addr, BasicBlock *value) {
+  using ValueType = std::variant<Value, QualifiedStmt<ApplyExprStmt>,
+      QualifiedStmt<Addressable_impl>, BasicBlock *>;
+
+  template<typename A>
+  static StoreInsn Create(QualifiedStmt<Addressable_impl> addr, A value) {
```
vjayathirtha-nv: shouldn't `A` also be tested with a QualifiefStmt<> ?

vjayathirtha-nv: If not, then is the QualifiedStmt <> test necessary here for `Addressable_impl` since we have that check in the constructors ?

lib/FIR/statements.cc
```diff
@@ -126,19 +125,25 @@ LoadInsn::LoadInsn(Statement *addr) : address_{addr} {
...
-  } else {
-    auto *expr{GetApplyExpr(val)};
-    CHECK(expr);
-    value_ = expr;
-  }
+  CHECK(val);
+}
+StoreInsn::StoreInsn(QualifiedStmt<Addressable_impl> addr, Value val)
+  : address_{addr}, value_{val} {
+  CHECK(address_);
```
vjayathirtha-nv: is a null `val` legal here?

lib/FIR/statements.h
```diff
@@ -378,25 +401,26 @@ class LoadInsn : public MemoryStmt_impl {
...
 private:
-  explicit StoreInsn(Value addr, Value val);
-  explicit StoreInsn(Value addr, BasicBlock *val);
-  explicit StoreInsn(Statement *addr, Statement *val);
-  explicit StoreInsn(Statement *addr, BasicBlock *val);
-
-  Addressable_impl *address_;
+  explicit StoreInsn(QualifiedStmt<Addressable_impl> addr, Value val);
+  explicit StoreInsn(
+      QualifiedStmt<Addressable_impl> addr, QualifiedStmt<ApplyExprStmt> val);
```
vjayathirtha-nv: Formatting for this line and the next one seems different.

lib/FIR/statements.h
```diff
@@ -54,6 +54,23 @@ class Stmt_impl {
   using StatementTrait = std::true_type;
 };
 
+// Some uses of a Statement should be constrained.  These contraints are imposed
+// at compile time.
+template<typename A = Stmt_impl> class QualifiedStmt {
+public:
+  QualifiedStmt() = delete;
+  template<typename B, std::enable_if_t<std::is_base_of_v<A, B>, int> = 0>
+  QualifiedStmt(Statement *stmt, const B &) : stmt{stmt} {}
```
jeanPerier: Should you have a dynamic check that the `*stmt` indeed points to variant containing B?

Or alternatively, you could create the statement on the go to ensure that stmt **is** a B. e.g:
`QualifiedStmt(BasicBlock*p, B &&b) : stmt{new Statement(p, std::forward<B>(b))} {}`

I see that you are only calling the constructor in `QualifiedInsert` and that this constructor would fit there. I would suggest to only allow such constructor if you can to avoid the risk of having `*stmt` and `B` mismatching.

lib/FIR/statements.cc
schweitzpgi: At this point, we can store a `Nothing`.  That may be useful will this gets built out, but should eventually be phased out.

lib/FIR/statements.h
schweitzpgi: I'm not thrilled with this interface, but it gets past the compiler.  The plan is to just pass in the envelope and add enough template magic to type check.

lib/FIR/statements.h
```diff
@@ -54,6 +54,23 @@ class Stmt_impl {
...
+
+  // create a stub, where stmt == nullptr
+  QualifiedStmt(std::nullptr_t) : stmt{nullptr} {}
+  operator Statement *() const { return stmt; }
+  operator bool() const { return stmt; }
+  operator A *() const;
+
+  Statement *stmt;
```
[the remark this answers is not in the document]
schweitzpgi: No.  One does not rewrite an IR node in place.  (All the ctors are private.)

lib/FIR/statements.h
```diff
@@ -265,12 +285,15 @@ class ActionStmt_impl : virtual public Stmt_impl {
...
-  static ApplyExprStmt Create(Expression &&e) { return ApplyExprStmt{e}; }
+  static ApplyExprStmt Create(Expression &&e) {
+    return ApplyExprStmt{std::move(e)};
+  }
 
   Expression expression() const { return expression_; }
 
 private:
   explicit ApplyExprStmt(const Expression &e) : expression_{e} {}
+  explicit ApplyExprStmt(Expression &&e) : expression_{e} {}
```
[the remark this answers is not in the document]
schweitzpgi: done

lib/FIR/statements.h
schweitzpgi: I'll re-run clang-format on this.

lib/FIR/statements.h
```diff
@@ -556,21 +580,12 @@ class Statement : public SumTypeMixin<ReturnStmt,  //
...
-    auto *result{reinterpret_cast<Statement *>(reinterpret_cast<char *>(stmt) -
-        (reinterpret_cast<char *>(&s.u) - reinterpret_cast<char *>(&s)))};
-    CHECK(result == reinterpret_cast<Statement *>(stmt) &&
-        "expecting pointers to be equal");
-    return result;
-  }
 };
 
+template<typename A> inline QualifiedStmt<A>::operator A *() const {
+  return reinterpret_cast<A *>(&stmt->u);
```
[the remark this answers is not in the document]
schweitzpgi: Because it's actually not what I want.  `get` will work if we want an exact type, but not if we want any of a set of subtypes.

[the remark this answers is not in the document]
jeanPerier: The second option should also work with subtypes (if these are things like `TerminatorStmt_impl`, `ActionStmt_impl`...). But I agree it may be a bit bad looking and does more stuff than a `reinterpret_cast`, though these things are mostly checks.

schweitzpgi:
> I think it may be worth improving the fact that a `QualifiedSmt` is actually containing a pointer to a Statement of the right variant,

I am not sure I'm following.  A `QualifiedStmt` _is_ a pointer.  Or rather it is a pointer with some suitable type wrappings for the C++ type system that designate precisely: this wrapped `Statement` does/must/will always be a member of the given set of subtypes.
 
> and I do not know if you can completely get rid of dynamic checks if Statements are mutable. 

They are not.  They are instantiated once.  The intuition is that for a given operation, there can be any number of other operations that use the result.  Morphing that operation into something else arbitrarily will almost surely violate the overarching semantics of the program being represented, since that would imply changing the operation universally in all contexts.

> But I like the idea of having the static representation of what should be inside a Statement.

It's worth pointing out that we've always had the ability to recover what the exact statement implementation type of any generic `Statement` is; i.e., `Statements` have always used parametric polymorphism.  We've also had sets of subtypes for subtype polymorphism, and these sets are built into the implementation type hierarchy.  This layer gives control over subtype constraints.

By analogy, a basic block is a container ship.  Everything shipped to Hamburg goes in rectangular containers (statements) on the ship.  We've know how to open any container and see exactly what's inside it (the exact statement variant).  This gives us the property, whatever is inside a particular container is known to be blue, whether it is a car, a keyboard, or a taco.

schweitzpgi:
> I think it may be worth improving the fact that a `QualifiedSmt` is actually containing a pointer to a Statement of the right variant,

I am not sure I'm following.  A `QualifiedStmt` _is_ a pointer.  Or rather it is a pointer with some suitable type wrappings for the C++ type system that designate precisely: this wrapped `Statement` does/must/will always be a member of the given set of subtypes.
 
> and I do not know if you can completely get rid of dynamic checks if Statements are mutable. 

They are not.  They are instantiated once.  The intuition is that for a given operation, there can be any number of other operations that use the result.  Morphing that operation into something else arbitrarily will almost surely violate the overarching semantics of the program being represented, since that would imply changing the operation universally in all contexts.

> But I like the idea of having the static representation of what should be inside a Statement.

It's worth pointing out that we've always had the ability to recover what the exact statement implementation type of any generic `Statement` is; i.e., `Statements` have always used parametric polymorphism.  We've also had sets of subtypes for subtype polymorphism, and these sets are built into the implementation type hierarchy.  This layer gives control over subtype constraints.

By analogy, a basic block is a container ship.  Everything shipped to Hamburg goes in rectangular containers (statements) on the ship.  We've know how to open any container and see exactly what's inside it (the exact statement variant).  This gives us the property, whatever is inside a particular container is known to be blue, whether it is a car, a keyboard, or a taco.

lib/FIR/statements.h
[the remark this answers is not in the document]
schweitzpgi: github won't let me reply to the general comment, so I'll put my reply here.

> I think it may be worth improving the fact that a `QualifiedSmt` is actually containing a pointer to a Statement of the right variant,

I am not sure I'm following.  A `QualifiedStmt` is ultimately (after type erasure) nothing but a pointer.  Or if you rather, it is a pointer with some suitable type wrappings for the C++ type system that designate precisely: this wrapped `Statement` does/must/will always be a member of the given set of subtypes.
 
> and I do not know if you can completely get rid of dynamic checks if Statements are mutable. 

They are not.  They are instantiated once.  The intuition is that for a given operation, there can be any number of other operations that use the result.  Morphing that operation into something else arbitrarily will almost surely violate the overarching semantics of the program being represented, since that would imply changing the operation universally in all contexts.

> But I like the idea of having the static representation of what should be inside a Statement.

It's worth pointing out that we've always had the ability to recover what the exact statement implementation type of any generic `Statement` is; i.e., `Statements` have always used parametric polymorphism.  We've also had sets of subtypes for subtype polymorphism, and these sets are built into the implementation type hierarchy.  This layer gives control over subtype constraints.

By analogy, a basic block is a container ship.  Everything shipped to Hamburg goes in rectangular containers (statements) on the ship.  We've know how to open any container and see exactly what's inside it (the exact statement variant).  This gives us the property, whatever is inside a particular container is known to be blue, whether it is a car, a keyboard, or a taco.

schweitzpgi:
I'm having twitchy mouse issues, I guess.

lib/FIR/statements.h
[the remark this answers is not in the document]
schweitzpgi: OK, It still seems like over-engineering to add run-time tests to confirm that the compile-time constraints actually do what they are designed for.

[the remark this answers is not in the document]
klausler: Casting a pointer to a std::variant<> into a pointer to a distinct std::variant<> type assumes that the two variant types use the same tag codes for their shared alternatives.

PR #400, opened 2019-04-10
Removed unused parameter and tweaked the error message one more time.

psteinfeld:
@tskeith pointed out two more changes.  This update does those changes.

### DIFF
--- a/lib/semantics/resolve-labels.cc
+++ b/lib/semantics/resolve-labels.cc
@@ -898,11 +898,11 @@ void CheckLabelDoConstraints(const SourceStmtList &dos,
               TargetStatementEnum::CompatibleDo)) {
         errorHandler.Say(doTarget.parserCharBlock,
             parser::MessageFormattedText{
-                "END DO or CONTINUE must be used to terminate a labeled DO loop"_err_en_US});
+                "A DO loop should terminate with an END DO or CONTINUE"_err_en_US});
       } else {
         errorHandler.Say(doTarget.parserCharBlock,
             parser::MessageFormattedText{
-                "Only an END DO or CONTINUE should be used to terminate a labeled DO loop"_en_US});
+                "A DO loop should terminate with an END DO or CONTINUE"_en_US});
       }
     } else {
       loopBodies.emplace_back(SkipLabel(position), doTarget.parserCharBlock);
--- a/lib/semantics/resolve-labels.cc
+++ b/lib/semantics/resolve-labels.cc
@@ -898,13 +898,11 @@ void CheckLabelDoConstraints(const SourceStmtList &dos,
               TargetStatementEnum::CompatibleDo)) {
         errorHandler.Say(doTarget.parserCharBlock,
             parser::MessageFormattedText{
-                "Only an END DO or CONTINUE must be used to terminate a labeled DO loop"_err_en_US,
-                SayLabel(label)});
+                "Only an END DO or CONTINUE must be used to terminate a labeled DO loop"_err_en_US});
       } else {
         errorHandler.Say(doTarget.parserCharBlock,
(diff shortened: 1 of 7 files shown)


lib/semantics/resolve-labels.cc
```diff
@@ -898,13 +898,11 @@ void CheckLabelDoConstraints(const SourceStmtList &dos,
               TargetStatementEnum::CompatibleDo)) {
         errorHandler.Say(doTarget.parserCharBlock,
             parser::MessageFormattedText{
-                "Only an END DO or CONTINUE must be used to terminate a labeled DO loop"_err_en_US,
-                SayLabel(label)});
+                "Only an END DO or CONTINUE must be used to terminate a labeled DO loop"_err_en_US});
```
klausler: That's awkward English usage.  Try "An END DO or CONTINUE statement must..." or "Only an END DO or CONTINUE statement may...".

psteinfeld: I agree that the English seems awkward.

How about "Either" in both cases: "Either an END DO or CONTINUE must ..." and "Either an END DO or CONTINUE should ..."

tskeith: I think it is good to use similar wording in both cases. Any of these seem fine to me:
"END DO or CONTINUE must/should be used to terminate a labeled DO loop"
"An END DO or CONTINUE must/should be used to terminate a labeled DO loop"
"Either END DO or CONTINUE must/should be used to terminate a labeled DO loop"

klausler: What we really mean is that the terminal statement of a labeled DO loop must be an END DO or CONTINUE statement.  So try wording it like that, with the offending statement of the error message also being the subject of the sentence that's describing what's wrong with it.

lib/semantics/resolve-labels.cc
```diff
@@ -898,13 +898,11 @@ void CheckLabelDoConstraints(const SourceStmtList &dos,
               TargetStatementEnum::CompatibleDo)) {
         errorHandler.Say(doTarget.parserCharBlock,
             parser::MessageFormattedText{
-                "Only an END DO or CONTINUE must be used to terminate a labeled DO loop"_err_en_US,
-                SayLabel(label)});
+                "END DO or CONTINUE must be used to terminate a labeled DO loop"_err_en_US});
```
schweitzpgi: The irony of this message is that code like
```
    DO 10 i = 1,10
10 IF (a.eq.b) THEN
```
can be repaired to compile by disobeying the directive.
```
    DO 10 i = 1,10
10   arr(i) = d
    IF (a.eq.b) THEN
```

psteinfeld: Good point.  There are (at least) three compilation cases depending on which options are in effect.  First, the code (as illustrated by your example) might simply be illegal.  For the other two cases, the code might be something that we want to support but is illegal in the Fortran 2018 standard.  If the -Mstandard option is in effect, we would want to issue a message about END DO and CONTINUE that's a warning.  If the -Werror option is also in effect, that message should be an error.

Doing this differentiation is beyond the scope of what I intended when I created issue 392.  I'll create another issue to do this differentiation.

lib/semantics/resolve-labels.cc
psteinfeld: Actually, after thinking about @schweitzpgi's comment, I don't think that we want to have different messages for warnings and errors.  Rather, I think that, for the case where the code is both valid and does not conform to the standard, and the -Wstandard option is in effect, we should issue a warning message about using END DO or CONTINUE.  If the -Werror option is in effect, that message should be an error.

I'll create a separate issue for differentiating the case where the DO loop termination is invalid versus not conforming to the standard.

PR #403, opened 2019-04-11
Add software subnormal flushing around host library calls for hosts that cannot do it in HW 

jeanPerier:
This PR goes in Parallel with PR #396. Following PR #396 discussion, it adds the possibility to perform the subnormal flushing in software around the call to host library functions so that we can get rid of the clunky user warnings that was here as a TODO for now.

I think keeping the possibility to do hardware flushing on X86_64 and Aarch64 host make sens because it will yield a better fp determinism between host folding and target run-time computations when host and target are the same. As we have seen, some math libraries try to do weird things to circumvent hardware subnormal flushing. e.g some logf returning constant -1.03972076416015625e2_4 for all positive subnormal values instead of -inf.... There is no way to reproduce this kind of behaviors with software flushing around host library call.

Also add test related to subnormal flushing and use of host library functions.

@pawosm-arm, I will rebase this PR and remove your warning when @sscalpone merge #396.

### DIFF
--- a/lib/evaluate/host.cc
+++ b/lib/evaluate/host.cc
@@ -45,24 +45,26 @@ void HostFloatingPointEnvironment::SetUpHostFloatingPointEnvironment(
   }
 #elif defined(__aarch64__)
 #if defined(__GNU_LIBRARY__)
+  hasSubnormalFlushingHardwareControl_ = true;
   if (context.flushSubnormalsToZero()) {
     currentFenv_.__fpcr |= (1U << 24);  // control register
   } else {
     currentFenv_.__fpcr &= ~(1U << 24);  // control register
   }
 #elif defined(__BIONIC__)
+  hasSubnormalFlushingHardwareControl_ = true;
   if (context.flushSubnormalsToZero()) {
     currentFenv_.__control |= (1U << 24);  // control register
   } else {
     currentFenv_.__control &= ~(1U << 24);  // control register
   }
 #else
-  // TODO other C libraries on AArch64
-  context.messages().Say(
-      "TODO: flushing mode for subnormals is not set for this host architecture due to incompatible C library it uses"_en_US);
+  // If F18 is built with other C libraries on AArch64, software flushing will
+  // be performed around host library calls if subnormal flushing is requested
 #endif
 #else
-  // Software flushing will be performed around host library calls if needed.
+  // If F18 is not built on one of the above host architecture, software
+  // flushing will be performed around host library calls if needed.
 #endif
   errno = 0;
   if (fesetenv(&currentFenv_) != 0) {
--- a/lib/evaluate/host.cc
+++ b/lib/evaluate/host.cc
@@ -35,7 +35,7 @@ void HostFloatingPointEnvironment::SetUpHostFloatingPointEnvironment(
     return;
   }
 #if __x86_64__
-  HasSubnormalFlushingHardwareControl_ = true;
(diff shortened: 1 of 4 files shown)


lib/evaluate/host.cc
```diff
@@ -43,9 +44,7 @@ void HostFloatingPointEnvironment::SetUpHostFloatingPointEnvironment(
     currentFenv_.__mxcsr &= ~0x0040;  // operands
   }
 #else
-  // TODO other architectures
-  context.messages().Say(
-      "TODO: flushing mode for subnormals is not set for this host architecture when folding with host runtime functions"_en_US);
+  // Software flushing will be performed around host library calls if needed.
```
pawosm-arm: I guess it's a placeholder for the warnings.

lib/evaluate/host.cc
```diff
@@ -35,6 +35,7 @@ void HostFloatingPointEnvironment::SetUpHostFloatingPointEnvironment(
     return;
   }
 #if __x86_64__
+  HasSubnormalFlushingHardwareControl_ = true;
   if (context.flushSubnormalsToZero()) {
     currentFenv_.__mxcsr |= 0x8000;  // result
```
pawosm-arm: Are you sure it's not GLIBC or Linux specific? fenv_t could be defined differently on various *BSD systems.

lib/evaluate/host.cc
jeanPerier: What do you mean. Would you go for a #error in for unknown architectures/libc? I was planning on letting software flushing be the default then, but that is maybe not conservative enough.

pawosm-arm: ah, ok, I see what you're up to.

lib/evaluate/host.cc
jeanPerier: To be honest, I do not know. I checked Bionic X86_64 after I saw your ifdef for it and it is is also  defining the `__mxcsr` field. I do not want to start reviewing all possible libc. We used this in flang without any libc consideration. 
I guess when someone tries compiling f18 with a X86_64 libc that does not have this, it won't compile here and they can open an issue and we will see what to do then.

lib/evaluate/host.h
```diff
@@ -39,10 +39,14 @@ class HostFloatingPointEnvironment {
...
   void SetUpHostFloatingPointEnvironment(FoldingContext &);
   void CheckAndRestoreFloatingPointEnvironment(FoldingContext &);
+  bool HasSubnormalFlushingHardwareControl() {
+    return HasSubnormalFlushingHardwareControl_;
+  }
 
 private:
   std::fenv_t originalFenv_;
   std::fenv_t currentFenv_;
+  bool HasSubnormalFlushingHardwareControl_{false};
```
klausler: If you want your code to look more like the project style, it would be better to not capitalize the data member or its accessor.

jeanPerier: Thanks, done

jeanPerier:
> Jean -- Will you merge yours and Pawel's changes?

I have rebased with the master containing #397 and removed the user warnings.
@pawosm-arm, are you OK with this?

pawosm-arm:
> > Jean -- Will you merge yours and Pawel's changes?
> 
> I have rebased with the master containing #397 and removed the user warnings.
> @pawosm-arm, are you OK with this?

Yeah, it looks ok, there's also enough comment in the code to find it for those who experience issues with other system platforms.

PR #406, opened 2019-04-12
Reactivate tree-driven expression constraint checking

klausler:
And fix the expression visitor used by semantics so that it doesn't analyze the contents of constrained expressions multiple times.

Delete the now-empty `IF` construct semantic checker.

Be advised: things are are yet finished in the `DO CONCURRENT` checks and tests.  More to come there, but I wanted to get this out for initial review now.


sscalpone:
Nice change!

tskeith:
doconcurrent04 fails for me:
```
Not found: must have INTEGER type, but is REAL(4)
FAIL  /Users/tsk/f18/pmk-fix-expr-constraints/test/semantics/doconcurrent04.f90
```

klausler:
> doconcurrent04 fails for me:
> 
> ```
> Not found: must have INTEGER type, but is REAL(4)
> FAIL  /Users/tsk/f18/pmk-fix-expr-constraints/test/semantics/doconcurrent04.f90
> ```

Yes, that's what I was warning about in the PR message.  Will clean that up today.

klausler:
Added more code to optionally die if there are any failures in expression analysis without concomitant fatal error messages.  Turned it on, found one case in our standard regression tests that fails, fixed it, and then turned it back off.  Please take another look if convenient, this PR has changed quite a bit since it was first approved.

lib/semantics/check-do-concurrent.cc
```diff
@@ -401,7 +401,6 @@ class DoConcurrentContext {
     : messages_{context.messages()} {}
 
   bool operator==(const DoConcurrentContext &x) const { return this == &x; }
-  bool operator!=(const DoConcurrentContext &x) const { return this != &x; }
```
tskeith: Why have `operator==` without `operator!=`?

tskeith:
> Added more code to optionally die if there are any failures in expression analysis without concomitant fatal error messages. Turned it on, found one case in our standard regression tests that fails, fixed it, and then turned it back off. Please take another look if convenient, this PR has changed quite a bit since it was first approved.

Still looks good to me.

lib/semantics/check-do-concurrent.cc
klausler: I realized that classes that were used as part of `Indirection<>` had to define unused `operator!=` because `Indirection<>` itself has an unused `operator!=`.  So I cleaned them up.

PR #408, opened 2019-04-12
various code cleanups from QualifiedSmt reviews

schweitzpgi:
A collection of minor mods inspired by the last code review on QualifiedStmt

### DIFF
--- a/lib/FIR/statements.h
+++ b/lib/FIR/statements.h
@@ -56,11 +56,10 @@ public:
 
 // Some uses of a Statement should be constrained.  These contraints are imposed
 // at compile time.
-template<typename A = Stmt_impl> class QualifiedStmt {
+template<typename A> class QualifiedStmt {
 public:
   QualifiedStmt() = delete;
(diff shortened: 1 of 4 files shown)


lib/FIR/statements.h
```diff
@@ -56,11 +56,10 @@ class Stmt_impl {
 
 // Some uses of a Statement should be constrained.  These contraints are imposed
 // at compile time.
-template<typename A = Stmt_impl> class QualifiedStmt {
+template<typename A> class QualifiedStmt {
 public:
   QualifiedStmt() = delete;
-  template<typename B, std::enable_if_t<std::is_base_of_v<A, B>, int> = 0>
-  QualifiedStmt(Statement *stmt, const B &) : stmt{stmt} {}
+  QualifiedStmt(Statement *stmt) : stmt{stmt} {}
```
jeanPerier: Would it be worth to making the constructor private to enforce MakeQualifiedStmt has to be used ?
Note: I am not completely sure it is possible. But would be nice unless you need it for something else.

PR #410, opened 2019-04-12
Remove ObjectName alternative from Designator

tskeith:
A simple name in a `Designator` is always parsed as a `DataRef`, not
an `ObjectName`. So remove that alternative.

`StmtFunctionStmt::ConvertToAssignment` was creating a `Designator` with
that alternative: change it to do the same thing as the parser.

Add `GetSimpleName` utility functions to check if an `Expr` or `Variable`
represents a simple name. Many of the places that checked for `ObjectName`
in `Designator` are trying to do that.

Clean up includes and forward declarations in `tools.h`.

### DIFF
--- a/lib/parser/parse-tree.cc
+++ b/lib/parser/parse-tree.cc
@@ -45,7 +45,6 @@ CommonStmt::CommonStmt(std::optional<Name> &&name,
 bool Designator::EndsInBareName() const {
   return std::visit(
       common::visitors{
-          [](const ObjectName &) { return true; },
           [](const DataRef &dr) {
             return std::holds_alternative<Name>(dr.u) ||
                 std::holds_alternative<common::Indirection<StructureComponent>>(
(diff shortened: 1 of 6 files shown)


vdonaldson:
I think your GetSimpleName might simplify this function.

- - -

void IoChecker::Enter(const parser::InputItem &spec) {
  hasDataList_ = true;
  if (const parser::Variable * var{std::get_if<parser::Variable>(&spec.u)}) {
    if (const auto *designator{
            std::get_if<common::Indirection<parser::Designator>>(&var->u)}) {
      if (auto *dataref{std::get_if<parser::DataRef>(&designator->value().u)}) {
        if (const parser::Name * name{std::get_if<parser::Name>(&dataref->u)}) {
          if (name->symbol != nullptr) {
            if (auto *details{name->symbol->detailsIf<ObjectEntityDetails>()}) {
              if (details->IsAssumedSize()) {
                // This check may be superseded by C928 or C1002.
                context_.Say(name->source,
                    "variable '%s' is an assumed size array"_err_en_US,
                    name->ToString().data());  // C1231
              }
            }
          }
        }
      }
    }
  }
}

tskeith:
> It would be great if you could add a comment in grammar.h and/or parse-tree.h to the effect that ObjectName does not appear as an alternative despite being in the grammar of the language standard. Thanks for doing this!

grammar.h already has an extensive comment explaining how 6 of the 7 alternatives for designator are parsed as data-ref. I think that's good enough, but if you think we need something added to parse-tree.h, let me know and I will.

PR #412, opened 2019-04-15
Implement semantic checks for return and alt-return.

sscalpone:


lib/semantics/check-return.cc
```diff
@@ -0,0 +1,55 @@
...
+#include "semantics.h"
+#include "tools.h"
+#include "../parser/message.h"
+#include "../parser/parse-tree.h"
+
+namespace Fortran::semantics {
+
+const Scope *FindContainingSubprogram(const Scope &start) {
+  const Scope *scope{&start};
+  while (scope != nullptr) {
```
tskeith: scope is only ever set to the address of a Scope, so it can never be null.

The condition should be `scope->kind() != Scope::Kind::Global`.

lib/semantics/check-return.cc
```diff
@@ -0,0 +1,55 @@
...
+  return nullptr;
+}
+
+void ReturnStmtChecker::Leave(const parser::ReturnStmt &returnStmt) {
+  // R1542 -- expression analysis validates the scalar-int-expr
+  // C1574 The return-stmt shall be in the inclusive scope of a function or subroutine subprogram.
+  // C1575 The scalar-int-expr is allowed only in the inclusive scope of a subroutine subprogram.
+  auto scope{context_.FindScope(*context_.location())};
+  auto subprogramScope{FindContainingSubprogram(scope)};
+  if(subprogramScope == nullptr) {
```
tskeith: This looks like it needs clang-formatting.

lib/semantics/check-return.cc
sscalpone: OK.  I was following FindProgramUnitContaining, which has the same while loop.

lib/semantics/check-return.cc
```diff
@@ -0,0 +1,57 @@
...
+  return nullptr;
+}
+
+void ReturnStmtChecker::Leave(const parser::ReturnStmt &returnStmt) {
+  // R1542 Expression analysis validates the scalar-int-expr
+  // C1574 The return-stmt shall be in the inclusive scope of a function or
+  // subroutine subprogram.
+  // C1575 The scalar-int-expr is allowed only in the inclusive scope of a
+  // subroutine subprogram.
+  auto scope{context_.FindScope(*context_.location())};
```
tskeith: This will copy scope, which isn't necessary. (Scope maybe shouldn't even have a copy constructor.) So I suggest `auto &scope` or `const auto &scope`.

lib/semantics/check-return.cc
```diff
@@ -0,0 +1,57 @@
...
+  auto subprogramScope{FindContainingSubprogram(scope)};
```
tskeith: I think that in auto declarations it's worthwhile making const and pointers and referenences explicit. Unlike the above case the only difference is in readability, but I would use `const auto *subprogramScope`.

lib/semantics/check-return.cc
klausler: I'd be okay with `Scope` having a `Clone()` member function rather than a copy constructor.  Scope copies are pretty rare -- I think I used it in derived type instantiation, maybe, and perhaps we'll need it for inlining.

PR #414, opened 2019-04-15
More coarray name resolution and semantic checks

tskeith:
Add CoarrayChecker to check for valid team-value in CHANGE TEAM,
SYNC TEAM, and image selector. Check that coarray names and selector
names are distinct in CHANGE TEAM.

Resolve the variable in a coarray-association.
Create a scope for the construct entities of a CHANGE TEAM construct.

Add ResolveSelector to resolve a parser::Selector into an Expr and
optional variable name (and a source location for messages). Make use of
ResolveSelector to handle coarray-association, as well as it's previous
use in associate-stmt.

Improve the check for C1157 in select-type-stmt and add a test.

Add a test for "Associate name must have a type".

Move ResolveName, ResolveDataRef, etc. from ResolveNamesVisitor
to DeclarationVisitor so that they are available in ConstructVisitor
as well. Add ResolveVariable and ResolveDesignator.

In the parse tree, change TeamValue from a type alias to a wrapper
class. We already had a wrapper class anyway, ImageSelectorSpec::Team,
so the new TeamValue can be used instead. This allows the member
of ImageSelectorSpec to be treated like other occurrences of TeamValue.


lib/semantics/check-coarray.cc
```diff
@@ -0,0 +1,118 @@
...
+    return symbol.name() == name && symbol.owner().IsModule() &&
+        symbol.owner().name() == module;
+  }
+}
+static bool IsTeamType(const DerivedTypeSpec *derived) {
+  return IsDerivedTypeFromModule(derived, "iso_fortran_env", "team_type");
+}
+
+void CoarrayChecker::Leave(const parser::ChangeTeamStmt &x) {
+  CheckNamesAreUnique(std::get<std::list<parser::CoarrayAssociation>>(x.t));
```
klausler: s/Unique/Distinct/

lib/semantics/check-coarray.h
```diff
@@ -0,0 +1,52 @@
...
+struct ImageSelectorSpec;
+struct SyncTeamStmt;
+struct TeamValue;
+}
+
+namespace Fortran::semantics {
+
+class CoarrayChecker : public virtual BaseChecker {
+public:
+  inline CoarrayChecker(SemanticsContext &context) : context_{context} {}
```
klausler: `inline` is implicit since the function body is in the class definition.

lib/semantics/resolve-names.cc
```diff
@@ -925,12 +936,22 @@ class ConstructVisitor : public DeclarationVisitor {
   void Post(const parser::ExitStmt &x) { CheckRef(x.v); }
 
 private:
-  // This represents: associate-name => expr | variable
-  // expr is set unless there were errors
+  // R1105 selector -> expr | variable
+  // expr is set in either case unless there were errors
+  struct Selector {
+    Selector() : variable{nullptr} {}
```
klausler: It might be more robust to declare `variable` below with `nullptr` as its default initializer.

tskeith: Thanks for the suggestions, I've incorporated all of them.

PR #415, opened 2019-04-15
Capture source locations for Designators and Calls

klausler:
Tim requested source location captures for instances of Variable; Designator and Call are the two parse subtree alternatives for Variable.


klausler:
Found a regression in the NAG tests with unparsing coarray references, will fix that too while I'm here.

PR #416, opened 2019-04-16
Resolve ambiguous I/O unit parse in favor of CharVariable.

klausler:
This tweak should dodge a premature automatic type constraint error message in an ambiguous parsing case that's resolved later once symbol tables are in hand.

### DIFF
--- a/lib/parser/parse-tree.h
+++ b/lib/parser/parse-tree.h
@@ -2497,7 +2497,7 @@ WRAPPER_CLASS(FileUnitNumber, ScalarIntExpr);
 // R1203 internal-file-variable -> char-variable
 struct IoUnit {
   UNION_CLASS_BOILERPLATE(IoUnit);
-  std::variant<FileUnitNumber, Star, CharVariable> u;
+  std::variant<CharVariable, FileUnitNumber, Star> u;
 };
 
 // R1206 file-name-expr -> scalar-default-char-expr
--- a/lib/parser/parse-tree.h
+++ b/lib/parser/parse-tree.h
@@ -55,6 +55,7 @@ CLASS_TRAIT(EmptyTrait)
 CLASS_TRAIT(WrapperTrait)
 CLASS_TRAIT(UnionTrait)
 CLASS_TRAIT(TupleTrait)
+CLASS_TRAIT(ConstraintTrait)
 
 // Some parse tree nodes have fields in them to cache the results of a
 // successful semantic analysis later.  Their types are forward declared
@@ -272,6 +273,7 @@ using Location = const char *;
 // These template class wrappers correspond to the Standard's modifiers
 // scalar-xyz, constant-xzy, int-xzy, default-char-xyz, & logical-xyz.
 template<typename A> struct Scalar {
+  using ConstraintTrait = std::true_type;
   Scalar(Scalar &&that) = default;
   Scalar(A &&that) : thing(std::move(that)) {}
   Scalar &operator=(Scalar &&) = default;
@@ -279,6 +281,7 @@ template<typename A> struct Scalar {
 };
 
 template<typename A> struct Constant {
+  using ConstraintTrait = std::true_type;
   Constant(Constant &&that) = default;
   Constant(A &&that) : thing(std::move(that)) {}
   Constant &operator=(Constant &&) = default;
@@ -286,6 +289,7 @@ template<typename A> struct Constant {
 };
 
 template<typename A> struct Integer {
+  using ConstraintTrait = std::true_type;
   Integer(Integer &&that) = default;
   Integer(A &&that) : thing(std::move(that)) {}
   Integer &operator=(Integer &&) = default;
@@ -293,6 +297,7 @@ template<typename A> struct Integer {
 };
 
 template<typename A> struct Logical {
+  using ConstraintTrait = std::true_type;
   Logical(Logical &&that) = default;
   Logical(A &&that) : thing(std::move(that)) {}
   Logical &operator=(Logical &&) = default;
@@ -300,6 +305,7 @@ template<typename A> struct Logical {
 };
 
 template<typename A> struct DefaultChar {
+  using ConstraintTrait = std::true_type;
   DefaultChar(DefaultChar &&that) = default;
   DefaultChar(A &&that) : thing(std::move(that)) {}
   DefaultChar &operator=(DefaultChar &&) = default;
--- a/lib/parser/parse-tree.h
+++ b/lib/parser/parse-tree.h
@@ -1772,9 +1772,6 @@ struct Variable {
 // Appears only as part of scalar-logical-variable.
 using ScalarLogicalVariable = Scalar<Logical<Variable>>;
 
-// R905 char-variable -> variable
-WRAPPER_CLASS(CharVariable, Variable);
-
 // R906 default-char-variable -> variable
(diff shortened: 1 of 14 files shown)


lib/parser/parse-tree.h
```diff
@@ -2497,7 +2497,7 @@ WRAPPER_CLASS(FileUnitNumber, ScalarIntExpr);
 // R1203 internal-file-variable -> char-variable
 struct IoUnit {
   UNION_CLASS_BOILERPLATE(IoUnit);
-  std::variant<FileUnitNumber, Star, CharVariable> u;
+  std::variant<CharVariable, FileUnitNumber, Star> u;
```
tskeith: It seems misleading to use CharVariable here since any Variable will end up as a CharVariable. There's no enforcement as with DefaultCharExpr, for example.

So I think it would be clearer to eliminate CharVariable and just use Variable here.

klausler: `CharVariable`  is a wrapper around `Variable` that will remain a `CharVariable` only if it is a character variable.

tskeith: I don't understand this. When does it become something other than a CharVariable?

klausler: I'm fixing it in a parse tree rewriter.  More to come.

vdonaldson: As a data point, this code plus my code still has the original problem.

klausler:
The parse tree is now rewritten to resolve the ambiguity once we have types.  An I/O unit will be a `Variable` only if it as a variable with a character type (any kind, they're all legal).  Otherwise it will be a star or a ScalarIntExpr, and in the latter case, the automatic expression constraint checking will apply.

lib/semantics/rewrite-parse-tree.cc
```diff
@@ -105,6 +108,74 @@ bool RewriteMutator::Pre(parser::ExecutionPart &x) {
...
+  return GetType(std::get<parser::ProcedureDesignator>(x.t));
+}
+static DeclTypeSpec *GetType(const parser::FunctionReference &x) {
+  return GetType(x.v);
+}
+static DeclTypeSpec *GetType(const parser::Variable &x) {
+  return std::visit(
+      [](const auto &indirection) { return GetType(indirection.value()); },
+      x.u);
+}
```
tskeith: This set of GetType functions is similar to GetSimpleName that I recently added to tools.h. There the purpose is to start from an Expr or Variable and return a Name. (I don't think all the necessary overloadings are there yet.)

Also, in a change I'm working on I want to do something similar but get the .source data member from the node that has it, going through wrapper classes as well as from Variable into Designator, Call, etc.

I'm wondering if there is a way to have a shared implementation for this sort of thing.

vdonaldson:
Is this intended to be a complete fix?  Assuming I do in fact have this code (and no code of my own):

$ cat k.f90
  character(len=50) abc
  open(unit=abc)
end

$ /home/vdonalds/pk0/Release/tools/f18/f18 -fdebug-semantics -fparse-only k.f90
k.f90:2:13: error: Must have INTEGER type, but is CHARACTER(KIND=1,LEN=,LEN=50_4)
    open(unit=abc)
              ^^^
/home/vdonalds/pk0/Release/tools/f18/f18: semantic errors in k.f90

vdonaldson:
These GetType functions look like they might be useful in other contexts.  If so, is it possible to make them available for inclusion in other files?

tskeith:
> Is this intended to be a complete fix? Assuming I do in fact have this code (and no code of my own):
> 
> $ cat k.f90
> character(len=50) abc
> open(unit=abc)
> end
> 
> $ /home/vdonalds/pk0/Release/tools/f18/f18 -fdebug-semantics -fparse-only k.f90
> k.f90:2:13: error: Must have INTEGER type, but is CHARACTER(KIND=1,LEN=,LEN=50_4)
> open(unit=abc)
> ^^^
> /home/vdonalds/pk0/Release/tools/f18/f18: semantic errors in k.f90

In an OPEN statement the UNIT has to be an integer, right? This fix is for cases like READ where UNIT can be a character variable or an integer expression.

klausler:
> Is this intended to be a complete fix? Assuming I do in fact have this code (and no code of my own):
> 
> $ cat k.f90
> character(len=50) abc
> open(unit=abc)
> end
> 
> $ /home/vdonalds/pk0/Release/tools/f18/f18 -fdebug-semantics -fparse-only k.f90
> k.f90:2:13: error: Must have INTEGER type, but is CHARACTER(KIND=1,LEN=,LEN=50_4)
> open(unit=abc)
> ^^^
> /home/vdonalds/pk0/Release/tools/f18/f18: semantic errors in k.f90

The formatted version of the character type is garbled; I'll fix it.  The error itself is valid.

vdonaldson:
Yes; I should have been more explicit; thanks.

lib/semantics/rewrite-parse-tree.cc
klausler: I'll put the guts of GetType into a new lib/parser/tools.h file.  If it obsoletes GetSimpleName, I'll combine them.

The code in expression analysis that descends into a tree to acquire source locations might be useful to you.

klausler:
Ok, please take another look whenever it's convenient.  Thanks for your reviews.

lib/semantics/tools.h
```diff
@@ -97,11 +97,5 @@ const Symbol *FindExternallyVisibleObject(
 
 bool ExprHasTypeCategory(
     const evaluate::GenericExprWrapper &expr, const common::TypeCategory &type);
-
-// If this Expr or Variable represents a simple Name, return it.
-parser::Name *GetSimpleName(parser::Expr &);
-const parser::Name *GetSimpleName(const parser::Expr &);
-parser::Name *GetSimpleName(parser::Variable &);
-const parser::Name *GetSimpleName(const parser::Variable &);
```
tskeith: Should remove these from tools.cc as well.

lib/semantics/check-coarray.cc
```diff
@@ -56,7 +57,7 @@ void CoarrayChecker::Leave(const parser::FormTeamStmt &x) {
   AnalyzeExpr(context_, std::get<parser::ScalarIntExpr>(x.t));
   const auto &teamVar{std::get<parser::TeamVariable>(x.t)};
   AnalyzeExpr(context_, teamVar);
-  const parser::Name *name{GetSimpleName(teamVar.thing)};
+  const parser::Name *name{parser::Unwrap<parser::Name>(teamVar)};
```
tskeith: It seems like GetLastName is closer to what GetSimpleName did. How should we decide between using GetLastName and Unwrap<Name>?

klausler: `Unwrap` should be doing what `GetSimpleName` was meant to do: if a parse tree contains a node of interest, like a `Name`, possibly wrapped up in constraint checks, `Expr`, &c., `Unwrap` will find it and return it to you, otherwise null.  If you try to unwrap a Name from `X%Y`, it fails.

`GetLastName` finds the name that determines the type of an expression or variable.  It returns a reference since it can't fail.  On `X%Y`, it return `Y`.

klausler:
Tim, are your concerns resolved and suggestions implemented?

tskeith:
> Tim, are your concerns resolved and suggestions implemented?

Yes, it looks good, thanks.

PR #418, opened 2019-04-18
First batch of semantic checks for allocate

jeanPerier:
Implements first batch of semantic checks for allocate statement constraints: C932, C933, C934, C935, C936, C943, C944 and first part of C945.
Checks for remaining constraints left as a TODO will be done once this PR goes through review.

### DIFF
--- a/lib/semantics/check-allocate.cc
+++ b/lib/semantics/check-allocate.cc
@@ -20,6 +20,7 @@
 #include "../evaluate/fold.h"
 #include "../evaluate/type.h"
 #include "../parser/parse-tree.h"
+#include "../parser/tools.h"
 
 namespace Fortran::semantics {
 
@@ -33,33 +34,20 @@ struct AllocateCheckerInfo {
   bool gotMold{false};
 };
 
-static const parser::Name &GetName(
-    const parser::AllocateObject &allocateObject) {
-  return std::visit(
-      common::visitors{
-          [&](const parser::Name &name) -> const parser::Name & {
-            return name;
-          },
-          [&](const parser::StructureComponent &structureComponent)
-              -> const parser::Name & { return structureComponent.component; },
-      },
-      allocateObject.u);
-}
-
-class AllocationChecker {
+class AllocationCheckerHelper {
 public:
-  AllocationChecker(
+  AllocationCheckerHelper(
       const parser::AllocateObject &obj, AllocateCheckerInfo &info)
-    : allocateInfo_{info}, name_{GetName(obj)}, type_{name_.symbol->GetType()},
+    : allocateInfo_{info}, name_{parser::GetLastName(obj)},
+      type_{name_.symbol->GetType()},
       isSubobject_{std::holds_alternative<parser::StructureComponent>(obj.u)} {
     CHECK(type_ != nullptr);
     if (type_->category() == DeclTypeSpec::Category::Character) {
-      hasDefferedTypeParameter_ =
+      hasDeferredTypeParameter_ =
           type_->characterTypeSpec().length().isDeferred();
     } else if (const DerivedTypeSpec * derivedTypeSpec{type_->AsDerived()}) {
-      for (auto it{derivedTypeSpec->parameters().begin()};
-           it != derivedTypeSpec->parameters().end(); ++it) {
-        hasDefferedTypeParameter_ |= it->second.isDeferred();
+      for (const auto &pair : derivedTypeSpec->parameters()) {
+        hasDeferredTypeParameter_ |= pair.second.isDeferred();
       }
       isAbstract_ = derivedTypeSpec->typeSymbol().attrs().test(Attr::ABSTRACT);
     }
@@ -74,36 +62,14 @@ private:
   const parser::Name &name_;
   const DeclTypeSpec *type_;
   bool isSubobject_;
-  bool hasDefferedTypeParameter_{false};
+  bool hasDeferredTypeParameter_{false};
   bool isUnlimitedPolymorphic_{false};
   bool isAbstract_{false};
 };
 
 static std::optional<AllocateCheckerInfo> CheckAllocateOptions(
-    const parser::AllocateStmt &, SemanticsContext &);
-static bool IsTypeCompatible(const DeclTypeSpec &, const DeclTypeSpec &);
-static bool IsTypeCompatible(
-    const DeclTypeSpec &type1, const evaluate::DynamicType &type2);
-static bool HaveSameAssumedTypeParameters(
-    const DeclTypeSpec &, const DeclTypeSpec &);
-static bool HaveCompatibleKindParameters(
-    const DeclTypeSpec &, const DeclTypeSpec &);
-
-void AllocateChecker::Leave(const parser::AllocateStmt &allocateStmt) {
-  if (auto info{CheckAllocateOptions(allocateStmt, context_)}) {
-    for (const parser::Allocation &allocation :
-        std::get<std::list<parser::Allocation>>(allocateStmt.t)) {
-      AllocationChecker allocationChecker{
-          std::get<parser::AllocateObject>(allocation.t), *info};
-      allocationChecker.RunChecks(context_);
-    }
-  }
-}
-
-static std::optional<AllocateCheckerInfo> CheckAllocateOptions(
     const parser::AllocateStmt &allocateStmt, SemanticsContext &context) {
   AllocateCheckerInfo info;
-  evaluate::ExpressionAnalyzer analyzer{context};  // can emit error messages
   bool stopCheckingAllocate{false};  // for errors that would lead to ambiguity
   info.gotTypeSpec =
       std::get<std::optional<parser::TypeSpec>>(allocateStmt.t).has_value();
@@ -122,7 +88,6 @@ static std::optional<AllocateCheckerInfo> CheckAllocateOptions(
               std::visit(
                   common::visitors{
                       [&](const parser::StatVariable &statVariable) {
-                        analyzer.Analyze(statVariable.v);
                         if (info.gotStat) {  // C943
                           context.Say(
                               "STAT may not be duplicated in a ALLOCATE statement"_err_en_US);
@@ -130,7 +95,6 @@ static std::optional<AllocateCheckerInfo> CheckAllocateOptions(
                         info.gotStat = true;
                       },
                       [&](const parser::MsgVariable &msgVariable) {
-                        analyzer.Analyze(msgVariable.v);
                         if (info.gotMsg) {  // C943
                           context.Say(
                               "ERRMSG may not be duplicated in a ALLOCATE statement"_err_en_US);
@@ -141,7 +105,6 @@ static std::optional<AllocateCheckerInfo> CheckAllocateOptions(
                   statOrErr.u);
             },
             [&](const parser::AllocOpt::Source &source) {
-              analyzer.Analyze(source.v);
               if (info.gotSrc) {  // C943
                 context.Say(
                     "SOURCE may not be duplicated in a ALLOCATE statement"_err_en_US);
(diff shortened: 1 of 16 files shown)


lib/semantics/check-allocate.cc
```diff
@@ -0,0 +1,411 @@
...
+    const parser::AllocateObject &allocateObject) {
+  return std::visit(
+      common::visitors{
+          [&](const parser::Name &name) -> const parser::Name & {
+            return name;
+          },
+          [&](const parser::StructureComponent &structureComponent)
+              -> const parser::Name & { return structureComponent.component; },
+      },
+      allocateObject.u);
```
tskeith: `parser::GetLastName` does what you want for Name and StructureComponent. So it might make sense to extend it to handle AllocateObject as well and then you wouldn't need this function.

lib/semantics/check-allocate.cc
```diff
@@ -0,0 +1,411 @@
...
+  }
+
+  bool RunChecks(SemanticsContext &context);
+
+private:
+  AllocateCheckerInfo &allocateInfo_;
+  const parser::Name &name_;
+  const DeclTypeSpec *type_;
+  bool isSubobject_;
+  bool hasDefferedTypeParameter_{false};
```
tskeith: "Deffered" -> "Deferred"

lib/semantics/check-allocate.cc
```diff
@@ -0,0 +1,411 @@
...
+          [&](const parser::Name &name) -> const parser::Name & {
+            return name;
+          },
+          [&](const parser::StructureComponent &structureComponent)
+              -> const parser::Name & { return structureComponent.component; },
+      },
+      allocateObject.u);
+}
+
+class AllocationChecker {
```
tskeith: I think it's confusing having a class named AllocationCheck in a file where we're defining class AllocateChecker. Maybe AllocateCheckerHelper or something like that is better.

lib/semantics/check-allocate.cc
```diff
@@ -0,0 +1,411 @@
...
+          std::get<parser::AllocateObject>(allocation.t), *info};
+      allocationChecker.RunChecks(context_);
+    }
+  }
+}
+
+static std::optional<AllocateCheckerInfo> CheckAllocateOptions(
+    const parser::AllocateStmt &allocateStmt, SemanticsContext &context) {
+  AllocateCheckerInfo info;
+  evaluate::ExpressionAnalyzer analyzer{context};  // can emit error messages
```
tskeith: I don't think it's necessary to analyze the Exprs and Variables -- ExprChecker should have done it already.

lib/semantics/check-allocate.cc
```diff
@@ -0,0 +1,411 @@
...
+  }
+
+  if (stopCheckingAllocate) {
+    return std::nullopt;
+  }
+
+  if (info.gotSrc || info.gotMold) {
+    CHECK(parserSourceExpr);
+    CHECK(parserSourceExpr->typedExpr);  // TODO: Can we reach this spot without
+                                         // a valid expression?
```
tskeith: As things stand you are correct, if the expression is invalid ExprChecker would have found the error and this checker would never be called. I want to change it so that the other checkers are called even if there are expression errors. So it's better to check that typedExpr is present.

lib/semantics/check-allocate.cc
```diff
@@ -0,0 +1,411 @@
...
+        "%s in ALLOCATE statement must have the ALLOCATABLE or POINTER attribute"_err_en_US,
+        (isSubobject_ ? "component" : "name"));
+    return false;
+  }
+  bool gotSourceExprOrTypeSpec{allocateInfo_.gotMold ||
+      allocateInfo_.gotTypeSpec || allocateInfo_.gotSrc};
+  if (hasDefferedTypeParameter_ && !gotSourceExprOrTypeSpec) {
+    // C933
+    context.Say(name_.source,
+        "Either type-spec or source-expr shall appear in ALLOCATE when allocatable object has a deferred type parameters"_err_en_US);
```
tskeith: I think "must" is better than "shall" in all of these error messages. They use "shall" in the Fortran standard, but as an English word it is ambiguous. It can mean "must" or "should" or "will", so we should avoid it.

lib/semantics/check-allocate.cc
```diff
@@ -0,0 +1,411 @@
...
+      }
+    }
+  }
+  return false;
+}
+
+static bool IsTypeCompatible(
+    const DeclTypeSpec &type1, const DeclTypeSpec &type2) {
+  if (type1.category() == DeclTypeSpec::Category::ClassStar) {
+    // TypeStar does not make sens in allocate context because assumed type
```
tskeith: "sens" -> "sense"

lib/semantics/check-allocate.cc
```diff
@@ -0,0 +1,411 @@
...
+  AllocationChecker(
+      const parser::AllocateObject &obj, AllocateCheckerInfo &info)
+    : allocateInfo_{info}, name_{GetName(obj)}, type_{name_.symbol->GetType()},
+      isSubobject_{std::holds_alternative<parser::StructureComponent>(obj.u)} {
+    CHECK(type_ != nullptr);
+    if (type_->category() == DeclTypeSpec::Category::Character) {
+      hasDefferedTypeParameter_ =
+          type_->characterTypeSpec().length().isDeferred();
+    } else if (const DerivedTypeSpec * derivedTypeSpec{type_->AsDerived()}) {
+      for (auto it{derivedTypeSpec->parameters().begin()};
```
klausler: `for (const auto &pair : derivedTypeSpec->parameters())`

lib/semantics/check-allocate.cc
```diff
@@ -0,0 +1,411 @@
...
+  AllocateCheckerInfo &allocateInfo_;
+  const parser::Name &name_;
+  const DeclTypeSpec *type_;
+  bool isSubobject_;
+  bool hasDefferedTypeParameter_{false};
+  bool isUnlimitedPolymorphic_{false};
+  bool isAbstract_{false};
+};
+
+static std::optional<AllocateCheckerInfo> CheckAllocateOptions(
```
klausler: Why prototype static subroutines?  If they are independent, define them here.  If they are mutually recursive, make them member functions (probably `static`) of your helper class.

lib/semantics/check-allocate.h
```diff
@@ -0,0 +1,34 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef FORTRAN_SEMANTICS_CHECK_ALLOCATE_H_
+#define FORTRAN_SEMANTICS_CHECK_ALLOCATE_H_
+
+#include "semantics.h"
```
klausler: You probably don't have to include this file if you forward declare `SemanticsContext` in namespace `semantics` below.

lib/semantics/check-allocate.cc
```diff
@@ -0,0 +1,411 @@
...
+      return type1.characterTypeSpec().length().isAssumed() ==
+          type2LengthIsAssumed;
+    }
+    // It is possible to reach this if type1 is unlimited polymorphic
+    return !type2LengthIsAssumed;
+  } else if (const DerivedTypeSpec * derivedType2{type2.AsDerived()}) {
+    int type2AssumedParametersCount{0};
+    int type1AssumedParametersCount{0};
+    for (auto it{derivedType2->parameters().begin()};
+         it != derivedType2->parameters().end(); ++it) {
```
tskeith: Consider using a range-based for loop here and below:
`for (auto pair : derivedType2->parameters())`

lib/semantics/check-allocate.cc
```diff
@@ -0,0 +1,411 @@
...
+      type2AssumedParametersCount += it->second.isAssumed();
+    }
+    if (const DerivedTypeSpec *
+        derivedType1{
+            type1.AsDerived()}) {  // type1 may be unlimited polymorphic
```
tskeith: If the comment can go on its own line, clang-format will do a better job.

lib/semantics/check-allocate.cc
```diff
@@ -0,0 +1,411 @@
...
+          if (const ParamValue *
+              param{derivedType2->FindParameter(it->first)}) {
+            if (!param->isAssumed()) {
+              return false;  // type1 has an assumed param that is not assumed
+                             // in type2
+            }
+          } else {
+            return false;  // type1 has an assumed param that is not a type
+                           // param of type2.
+          }
```
tskeith: Consider writing this if statement like this:
```
const ParamValue *param{derivedType2->FindParameter(it->first)};
if (!param || !param->isAssumed()) {
  return false;
}
```

lib/semantics/check-allocate.cc
jeanPerier: Done, thanks for pointing to this function.

lib/semantics/check-allocate.cc
jeanPerier: Fixed

lib/semantics/check-allocate.cc
jeanPerier: I went for `Allocation` because it matches the standard name for the construct R931 which relates to the checks. But I agree it is ambiguous like this, so I have changed it to `AllocationCheckerHelper`.

lib/semantics/check-allocate.cc
jeanPerier: OK, I took this from other check-xx.cc. I saw that you removed these since and I could not find tests to "use" this anyway. Removed as suggested.

lib/semantics/check-allocate.cc
jeanPerier: Thanks for the heads up on PR #422. I modified this part to use `GetExpr`. As I understood from PR #424, if `GetExpr` is returning a null pointer, it means that the expression was analyzed but there was an error reported and allocate tests should stop without emitting new errors regarding this. Is this correct?

lib/semantics/check-allocate.cc
jeanPerier: Sounds easier and more consistent with other error messages, thanks. Applied.

lib/semantics/check-allocate.cc
jeanPerier: Fixed

lib/semantics/check-allocate.cc
jeanPerier: Looks way simpler. Applied.

lib/semantics/check-allocate.cc
jeanPerier: That was only a forward declaration to avoid defining them before the core functions that use them so that the reader gets to the core functions first, but it has no functional purpose and if you find this weird, I'd rather remove it. I reorganized the function order and removed these.

lib/semantics/check-allocate.h
jeanPerier: Yes, but I do not think it is possible to forward declare `BaseChecker` and inherit from it. `BaseChecker` is declared in `semantics.h`.

lib/semantics/check-allocate.cc
jeanPerier: Applied

lib/semantics/check-allocate.cc
jeanPerier: Applied

lib/semantics/check-allocate.cc
jeanPerier: That looked weird. Fixed

jeanPerier:
Thanks for the reviews and pieces of advice. I have applied all of them but for the removal of `#inlcude "semantics.h"`

lib/semantics/check-allocate.cc
tskeith: Right

PR #419, opened 2019-04-18
More work on classes to represent characteristics of procedures.

klausler:
Extracted from a larger accumulation of work for easier code review.

The Fortran 2018 standard defines a sheaf of "characteristics" that are used to compare procedures against interfaces to test compatibility of actual procedure arguments, procedure pointer targets, &c.  The data structures in characteristics.{h,cc} are intended to hold the results of this analysis.

### DIFF
--- a/lib/evaluate/characteristics.cc
+++ b/lib/evaluate/characteristics.cc
@@ -13,6 +13,11 @@
 // limitations under the License.
 
 #include "characteristics.h"
+#include "intrinsics.h"
+#include "tools.h"
+#include "type.h"
+#include "../common/indirection.h"
+#include "../semantics/symbol.h"
 #include <ostream>
 #include <sstream>
 #include <string>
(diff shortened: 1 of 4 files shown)


lib/evaluate/characteristics.h
```diff
@@ -26,38 +26,88 @@
...
   DECLARE_CONSTRUCTORS_AND_ASSIGNMENTS(DummyProcedure)
-  common::CopyableIndirection<Procedure> explicitProcedure;
-  common::EnumSet<Attr, 32> attrs;
+  explicit DummyProcedure(Procedure &&);
   bool operator==(const DummyProcedure &) const;
+  static std::optional<DummyProcedure> Characterize(
+      const semantics::Symbol &, const IntrinsicProcTable &);
   std::ostream &Dump(std::ostream &) const;
+  common::CopyableIndirection<Procedure> procedure;
+  common::EnumSet<Attr, 32> attrs;
```
tskeith: Why not symbols::Attr_enumSize instead of 32?

klausler: Sure, will do here and below.

lib/evaluate/characteristics.cc
```diff
@@ -96,6 +226,189 @@ std::ostream &Procedure::Dump(std::ostream &o) const {
...
+      }
+      if (symbol.attrs().test(semantics::Attr::INTENT_OUT)) {
+        CHECK(result.intent == common::Intent::Default);
+        result.intent = common::Intent::Out;
+      }
+      if (symbol.attrs().test(semantics::Attr::INTENT_INOUT)) {
+        CHECK(result.intent == common::Intent::Default);
+        result.intent = common::Intent::InOut;
+      }
+      // TODO: acquire coshape when symbol table represents it
```
tskeith: Return result here?

lib/evaluate/characteristics.h
tskeith: I missed that Attr is a nested enum, so it should just be Attr_enumSize.

lib/evaluate/characteristics.cc
klausler: Oh, sure.  Thanks.

PR #422, opened 2019-04-19
Continue semantics checks after expression errors

tskeith:
Change statement semantics to continue with checks after an error
in expression analysis. This allows the compiler to report more
compilation errors.

Now the best practice in statement checker classes is to use GetExpr to
get an analyzed expression from a parse tree node and FindSourceLocation
to get a source location. The result of GetExpr must be checked for
non-null because expression analysis might have failed on the expression.


klausler:
Does this change distinguish between failures of expression semantics to analyze expressions due to bugs or TODOs in that analysis and failures of expression semantics due to errors detected in the program?  It would be great if a missing Expr would trigger an internal error if no fatal user error has been emitted.

lib/parser/parse-tree.cc
```diff
@@ -86,14 +86,17 @@ bool DoConstruct::IsDoConcurrent() const {
...
-    const Name &name, std::list<Expr> &&subscripts) {
+static Designator MakeArrayElementRef(const Name &name,
+    const parser::CharBlock &source, std::list<Expr> &&subscripts) {
   ArrayElement arrayElement{DataRef{Name{name}}, std::list<SectionSubscript>{}};
   for (Expr &expr : subscripts) {
     arrayElement.subscripts.push_back(SectionSubscript{
         Scalar{Integer{common::Indirection{std::move(expr)}}}});
   }
-  return Designator{DataRef{common::Indirection{std::move(arrayElement)}}};
+  auto result{
```
klausler: This replacement of `source` could be done in the caller or in a wrapper; it seems odd to do it here.

lib/semantics/check-arithmeticif.cc
```diff
@@ -19,8 +19,8 @@
 
 namespace Fortran::semantics {
 
-bool IsNumericExpr(const evaluate::GenericExprWrapper &expr) {
-  auto dynamicType{expr.v.GetType()};
+bool IsNumericExpr(const SomeExpr &expr) {
```
klausler: Thanks.  `GenericExprWrapper` usage should always be minimized, it's an ugly hack.

lib/semantics/expression.cc
```diff
@@ -1888,8 +1888,17 @@ MaybeExpr ExpressionAnalyzer::Analyze(const parser::Expr &expr) {
 }
 
 MaybeExpr ExpressionAnalyzer::Analyze(const parser::Variable &variable) {
-  FixMisparsedFunctionReference(context_, variable.u);
-  return Analyze(variable.u);
+  if (variable.typedExpr) {
```
klausler: Since this is pretty much what's done for expressions, maybe it could become a subroutine.

lib/semantics/tools.h
```diff
@@ -97,9 +96,39 @@ const Symbol *FindExternallyVisibleObject(
...
+      return Get(x.thing);
+    } else if constexpr (WrapperTrait<T>) {
+      return Get(x.v);
+    } else {
+      return nullptr;
+    }
+  }
+};
+
+template<typename T> const SomeExpr *GetExpr(const T &x) {
```
klausler: Idea: this could be a member template function of `SemanticsContext` and it could ensure on failure that the user program is already known to have fatal errors.

lib/semantics/check-do-concurrent.cc
```diff
@@ -420,29 +417,6 @@ class DoConcurrentContext {
         parser::Walk(
             std::get<parser::Block>(doConstruct.t), doConcurrentLabelEnforce);
         EnforceConcurrentLoopControl(*concurrent);
-      } else if (auto *loopBounds{
-                     std::get_if<parser::LoopBounds<parser::ScalarIntExpr>>(
-                         &optionalLoopControl->u)}) {
-        // C1120 - FIXME? may be checked before we get here
```
schweitzpgi: What is the project's plan for auditing the code base for coverage of all Fortran standard constraints?

sscalpone: Re: constraint checking coverage -- probably a conversation best had outside of a code review.

lib/semantics/tools.h
tskeith: What about introducing a special value of evaluate::Expr (or of GenericExprWrapper) to indicate the expression is erroneous? Then GetExpr could assert that every typedExpr has a value.

lib/semantics/check-do-concurrent.cc
tskeith: None that I know of.

lib/semantics/tools.h
klausler: We wouldn't want to put an error recovery token into `Expr<>`, it would add an alternative that would have to be checked for and propagated throughout the expression representation and folding code. Maybe `GenericExprWrapper` or as a flag next to the `typedExpr` member(s) in the parse tree.

sscalpone:
> Now the best practice....
Is this information replicated in the source code or a documentation file?

lib/semantics/expression.cc
tskeith: OK, I've made this change.

lib/semantics/tools.h
tskeith: I'll do this in a separate PR.

lib/semantics/expression.cc
```diff
@@ -1856,27 +1856,29 @@ void FixMisparsedFunctionReference(
...
-MaybeExpr ExpressionAnalyzer::Analyze(const parser::Expr &expr) {
-  if (expr.typedExpr) {
+// Common handling of parser::Expr and Parser::Variable
+template<typename PARSED>
+MaybeExpr ExpressionAnalyzer::ExprOrVariable(const PARSED &x) {
+  if (x.typedExpr) {
     // Expression was already checked by ExprChecker
-    return std::make_optional<Expr<SomeType>>(expr.typedExpr->v);
+    return std::make_optional<Expr<SomeType>>(x.typedExpr->v);
   } else {
```
sscalpone: Do you need the else?

PR #424, opened 2019-04-20
Distinguish between unanalyzed and error expressions

tskeith:
When an Expr or Variable is analyzed, always fill in a GenericExprWrapper
for it. That now holds the result of expression analysis which is
std::nullopt when there is an error.

After the ExprChecker pass has finished, the typedExpr data member of
Variables and top-level Exprs should be filled in. Assert that is the
case when we access them.

### DIFF
--- a/lib/FIR/afforestation.cc
+++ b/lib/FIR/afforestation.cc
@@ -25,9 +25,13 @@
 
 namespace Fortran::FIR {
 namespace {
-Expression *ExprRef(const parser::Expr &a) { return &a.typedExpr.get()->v; }
+Expression *ExprRef(const parser::Expr &a) {
+  CHECK(a.typedExpr);
+  CHECK(a.typedExpr->v);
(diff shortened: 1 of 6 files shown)


lib/FIR/afforestation.cc
```diff
@@ -25,9 +25,13 @@
 
 namespace Fortran::FIR {
 namespace {
-Expression *ExprRef(const parser::Expr &a) { return &a.typedExpr.get()->v; }
+Expression *ExprRef(const parser::Expr &a) {
```
klausler: I know this PR maybe shouldn't mess with APIs, but I think that this function should return a reference rather than a pointer.

tskeith: Good idea. I've made that change and it does simplify things.
I also realized there should be an overloading that takes parser::Variable so I added that too.

PR #429, opened 2019-04-25
Continue semantic checking after name resolution error

tskeith:
When an error occurs in name resolution, continue semantic processing
in order to detect other errors. This means we can no longer assume
that every `parser::Name` has a symbol even after name resolution
completes. In `RewriteMutator`, only report internal error for unresolved
symbol if there have been no fatal errors.

Add `Error` flag to `Symbol` to indicate that an error occurred related
to it. Once we report an error about a symbol we should avoid reporting
any more to prevent cascading errors. Add `HasError()` and `SetError()`
to simplify working with this flag.

Change some places that we assume that a `parser::Name` has a non-null
symbol. There are probably more.

`resolve-names.cc`: Set the `Error` flag when we report a fatal error
related to a symbol. (This requires making some symbols non-const.)
Remove `CheckScalarIntegerType()` as `ExprChecker` will take care of
those constraints if they are expressed in the parse tree. One exception
to that is the name in a `ConcurrentControl`. Explicitly perform that
check using `EvaluateExpr()` and constraint classes so we get consistent
error messages.

In expression analysis, when a constraint is violated (like `Scalar<>`
or `Integer<>`), reset the wrapped expression so that we don't assume it
is valid. A `GenericExprWrapper` holding a std::nullopt indicates error.
Change `EnforceTypeConstraint()` to return false when the constraint
fails to enable this.

check-do-concurrent.cc: Reorganize the Gather*VariableNames functions
into one to simplify the task of filtering out unresolved names. Remove
`CheckNoDuplicates()` and `CheckNoCollisions()` as those checks is
already done in name resolution when the names are added to the scope.

### DIFF
--- a/lib/semantics/assignment.cc
+++ b/lib/semantics/assignment.cc
@@ -270,7 +270,7 @@ void AssignmentContext::Analyze(const parser::ConcurrentHeader &header) {
       std::get<std::list<parser::ConcurrentControl>>(header.t)) {
     const parser::Name &name{std::get<parser::Name>(control.t)};
     bool inserted{forall_->activeNames.insert(name.source).second};
-    CHECK(inserted || HasError(name));
+    CHECK(inserted || context_.HasError(name));
   }
 }
 
--- a/lib/semantics/assignment.cc
+++ b/lib/semantics/assignment.cc
@@ -15,6 +15,7 @@
 #include "assignment.h"
 #include "expression.h"
 #include "symbol.h"
+#include "tools.h"
 #include "../common/idioms.h"
 #include "../evaluate/expression.h"
 #include "../evaluate/fold.h"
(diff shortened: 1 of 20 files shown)


tskeith:
> When we currently have a `CHECK(p != nullptr)`, and now we don't, I worry about missing compiler bugs because the new code assumes that we're in fatal error recovery. I suggest that you think about having some means for ensuring that a fatal error has been detected before allowing the compiler to proceed with missing or invalid data, like `CHECK(p != nullptr || semanticsContext.AnyFatalError())`, possibly packaged as a member function of `SemanticsContext`. Then we maintain the ability to crash as soon as we can detect an internal inconsistency while still being able to report more non-cascaded errors messages.

Thanks for the suggestion. I've moved HasError and SetError into SemanticsContext so we can do that consistency check.

jeanPerier:
Does this change imply there are other places where we may now get nullptr `Symbol` and it is not an internal compiler error ? For instance, can we now get a null `Symbol` pointer in `DerivedTypeDetails::OrderParameterDeclarations` result list if the related component yielded an error during name resolution?

tskeith:
> Does this change imply there are other places where we may now get nullptr `Symbol` and it is not an internal compiler error ? For instance, can we now get a null `Symbol` pointer in `DerivedTypeDetails::OrderParameterDeclarations` result list if the related component yielded an error during name resolution?

The intention is that this applies only to Name::symbol.

PR #430, opened 2019-04-25
Bug fix: resolve ComponentArraySpec

tskeith:
This used to work but broke due to coarray spec changes.
The fix is to extend AnalyzeArraySpec to work on ComponentArraySpec
and to call it when we encounter one.

modfile24.f90 tested this case but had the wrong expected results.

Thanks to Jean for finding this.

### DIFF
--- a/test/semantics/modfile24.f90
+++ b/test/semantics/modfile24.f90
@@ -63,7 +63,7 @@ end
 !Expect: m3.mod
 !module m3
 ! type::t
-!  real(4)::c[1_8:10_8,1_8:*]
+!  real(4)::c(1_8:5_8)[1_8:10_8,1_8:*]
 !  complex(4)::d[1_8:5_8,1_8:*]
 ! end type
 ! real(4),allocatable::e[:,:,:]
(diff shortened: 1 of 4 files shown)


test/semantics/modfile24.f90
```diff
@@ -63,7 +63,7 @@ subroutine s(a, b, n)
 !Expect: m3.mod
 !module m3
 ! type::t
-!  real(4)::c[1_8:10_8,1_8:*]
```
jeanPerier: I think component `c` might be violating C746. As I happen to be working on coarray implications in ALLOCATE, I fall upon it today. It applies only for derived type components:

>  If a coarray-spec appears, it shall be a deferred-coshape-spec-list and the component shall have the ALLOCATABLE attribute.

So you might get an error here when we check more coarray constraints.
As I understand it you should have: `real, allocatable :: c(:)[:,:]`

tskeith: Thanks for catching that -- I'll fix it.

PR #433, opened 2019-04-29
Fix source positions during parse tree rewriting

tskeith:
When a StmtFunctionStmt was rewritten as an array element assignment
the subscripts were not getting correct source locations. They need
to be copied from the function args.

Also, the entire array element expression (e.g. `a(i)`) did not have a
source position. This was tricky because there is no source position
in the original parse that matches what we need. So we take the source
position from the beginning of the function name to the end of the last
arg and extend it one more character to include the closing parenthesis.

Change `ActualArgToExpr()` to return `Expr` rather than
`std::optional<Expr>` because callers always call `.value()` on the
result anyway.

Add `WithSource()` utility to update the `source` data member of a
parse tree node.

This bug shows up as incorrect source positions for error messages. For
example, in this program the error (due to real subscript) did not have
a source position:
```
real :: a(10), x, y
a(x) = y
end
```

### DIFF
--- a/lib/parser/parse-tree.cc
+++ b/lib/parser/parse-tree.cc
@@ -86,36 +86,41 @@ bool DoConstruct::IsDoConcurrent() const {
   return control && std::holds_alternative<LoopControl::Concurrent>(control->u);
 }
 
-static Designator MakeArrayElementRef(const Name &name,
-    const parser::CharBlock &source, std::list<Expr> &&subscripts) {
+static Designator MakeArrayElementRef(
+    const Name &name, std::list<Expr> &&subscripts) {
   ArrayElement arrayElement{DataRef{Name{name}}, std::list<SectionSubscript>{}};
   for (Expr &expr : subscripts) {
     arrayElement.subscripts.push_back(SectionSubscript{
         Scalar{Integer{common::Indirection{std::move(expr)}}}});
   }
-  auto result{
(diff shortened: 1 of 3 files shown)


lib/parser/parse-tree.cc
```diff
@@ -163,13 +168,19 @@ Statement<ActionStmt> StmtFunctionStmt::ConvertToAssignment() {
...
+    subscripts.push_back(WithSource(arg.source,
+        Expr{common::Indirection{
+            WithSource(arg.source, Designator{DataRef{Name{arg}}})}}));
+    source.ExtendToCover(arg.source);
   }
-  auto variable{Variable{common::Indirection{
-      MakeArrayElementRef(funcName, funcName.source, std::move(subscripts))}}};
+  // extend source to include closing paren
+  CHECK(*source.end() == ')');
+  source = CharBlock{source.begin(), source.end() + 1};
```
tskeith: This is a kludge but I can't find any other way to get the source location of the new Designator.

We are converting the LHS of a stmt-function-stmt (e.g. `f(i) = 1`) into a Variable. There is no source location for the LHS. We could add one during parsing, but it would not be the location for the entire node. So we would probably want to split the StmtFunctionNode into LHS and RHS so we have a place to put the LHS source position. This seems like a lot of trouble to go to just to avoid this kludge.

PR #434, opened 2019-04-30
Tests for C1131, DO loop names

psteinfeld:
Making sure that the names match between the DO loop statement and the END DO statement.

The code for the semantic check was already implemented, but there were no tests for it.

### DIFF
--- a/test/semantics/canondo08.f90
+++ b/test/semantics/canondo08.f90
@@ -1,0 +1,46 @@
+! Copyright (c) 2019, NVIDIA CORPORATION.  All rights reserved.
+!
+! Licensed under the Apache License, Version 2.0 (the "License");
+! you may not use this file except in compliance with the License.
+! You may obtain a copy of the License at
+!
+!     http://www.apache.org/licenses/LICENSE-2.0
+!
+! Unless required by applicable law or agreed to in writing, software
+! distributed under the License is distributed on an "AS IS" BASIS,
+! WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+! See the License for the specific language governing permissions and
+! limitations under the License.
+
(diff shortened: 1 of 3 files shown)


test/semantics/canondo08.f90
```diff
@@ -0,0 +1,46 @@
...
+
+! RUN: ${F18} -funparse-with-symbols %s 2>&1 | ${FileCheck} %s
+! CHECK-NOT: 34:.*name required but missing
+! CHECK-NOT: 34:.*DO construct name unexpected
+! CHECK-NOT: 32:.*should be
+! CHECK-NOT: 32:.*unnamed DO statement
+! CHECK: 39:.*name required but missing
+! CHECK: 37:.*should be
+! CHECK: 44:.*DO construct name unexpected
+! CHECK: 42:.*unnamed DO statement
```
tskeith: I guess the other canondo* tests are written this way, but it seems to me that using `test_errors.sh` would be better. It would fail if there unexpected errors other than the ones checked for on line 32 and line 34. It wouldn't fail if a blank line were added.

schweitzpgi:
The tests really should use LLVM's `FileCheck` rather than ad hoc shell scripts.

sscalpone:
clang doesn't appear to use filecheck; looks like they rolled their own.
https://clang.llvm.org/doxygen/classclang_1_1VerifyDiagnosticConsumer.html#details

See for example
llvm/tools/clang/test/SemaCXX/access-base-class.cpp

test/semantics/canondo08.f90
psteinfeld: Thanks, @tskeith.  I should have mentioned this flaky use of line numbers in test in my pull request.

I'll take a look at using test_errors.sh.

PR #439, opened 2019-05-03
Structure constructor semantics, part 1 of 2

klausler:
These are the changes to `lib/evaluate` that have accumulated while I wrapped up work on structure constructor expression semantics in `lib/semantics`.  Those changes will follow this PR.

I tried to split this work up into smaller changes, but they're interdependent.

- `Expr<SomeType>` can now hold a procedure designator or procedure pointer returned from a function reference.  Some conveniences were added to make it easier to distinguish instances of `Expr<SomeType>` that (like these new possibilities, plus BOZ literals and `NULL()`) don't have types in the Fortran sense from those that do.
- The representation of procedure characteristics has been worked over and extended as I've been using it in semantics.
- Intrinsic function table probes now return characteristics of the specific intrinsics they find, and these characteristics are retained in the `SpecificIntrinsic` structure used to represent the call in the expression.
- Some bugs in intrinsics were found and fixed along the way; conversions of BOZ literals to Fortran types should not convert the BOZ as if it were an integer.
- `DynamicType` was rearranged a little, and now has static member functions for creating its instances from symbol table entries and types, rather than associated functions that would have to become friends of the class when I get around to privatizing its members.

Some changes are attached that need to be made to compile `lib/semantics` with these expression changes.  Some of them are final, others are temporary updates that will be replaced with my next PR that finishes up structure constructor semantics once this work is merged.

### DIFF
--- a/lib/evaluate/call.cc
+++ b/lib/evaluate/call.cc
@@ -13,6 +13,7 @@
 // limitations under the License.
 
 #include "call.h"
+#include "characteristics.h"
 #include "expression.h"
 #include "tools.h"
 #include "../common/idioms.h"
@@ -61,9 +62,17 @@ bool ActualArgument::operator==(const ActualArgument &that) const {
       isAlternateReturn == that.isAlternateReturn && u_ == that.u_;
 }
 
+SpecificIntrinsic::SpecificIntrinsic(
+    IntrinsicProcedure n, characteristics::Procedure &&chars)
+  : name{n}, characteristics{new characteristics::Procedure{std::move(chars)}} {
+}
+
+DEFINE_DEFAULT_CONSTRUCTORS_AND_ASSIGNMENTS(SpecificIntrinsic)
+
+SpecificIntrinsic::~SpecificIntrinsic() {}
+
 bool SpecificIntrinsic::operator==(const SpecificIntrinsic &that) const {
-  return name == that.name && type == that.type && rank == that.rank &&
-      attrs == that.attrs;
+  return name == that.name && characteristics == that.characteristics;
 }
 
 ProcedureDesignator::ProcedureDesignator(Component &&c)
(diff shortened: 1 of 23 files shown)


klausler:
I didn't include some new tests with this pull request; they'll follow with the next one, since they depend on the work in `lib/semantics` too.

lib/evaluate/shape.h
```diff
@@ -22,6 +22,7 @@
 #include "tools.h"
 #include "type.h"
 #include "../common/indirection.h"
+#include "../parser/message.h"
```
tskeith: I don't know if it matters, but you could use a forward declaration of ContextualMessages and move the include to shape.cc.

lib/evaluate/call.cc
```diff
@@ -111,6 +143,19 @@ const Symbol *ProcedureDesignator::GetSymbol() const {
...
 
+parser::CharBlock ProcedureDesignator::GetName() const {
+  return std::visit(
+      common::visitors{
+          [](const SpecificIntrinsic &i) -> parser::CharBlock {
+            return i.name;
+          },
+          [](const Symbol *sym) -> parser::CharBlock { return sym->name(); },
+          [](const common::CopyableIndirection<Component> &c)
+              -> parser::CharBlock { return c.value().GetLastSymbol().name(); },
```
tskeith: I don't think the return type is necessary in the last two cases.

And in the first case, I think explicit construction (`return parser::CharBlock{i.name};`) is clearer, since the type has to be named somewhere in the lambda.

lib/evaluate/shape.h
klausler: Sure.

lib/evaluate/call.cc
klausler: Ok.

PR #441, opened 2019-05-03
Support std::string and CharBlock in formatted messages

klausler:
Proposed messaging API extension: support `%B` and `%S` formatting codes in `MessageFormattedText` messages that implement `parser::CharBlock` and `std::string` arguments.  This should allow the replacement of most usage of `parser::CharBlock::ToString()` and `std::string::data()` or `std::string::c_str()` in messaging.  This PR only uses the new formatting codes in `lib/parser`; a later change, if this turns to to be accepted as a good idea, will sweep through the code and update the messages.

### DIFF
--- a/lib/parser/message.cc
+++ b/lib/parser/message.cc
@@ -34,17 +34,20 @@ std::ostream &operator<<(std::ostream &o, const MessageFixedText &t) {
 }
 
 static bool NeedsSpecialFormatting(const char *p) {
+  bool result{false};
+  bool tooLate{false};
   while (*p != '\0') {
     if (*p++ == '%') {
       if (*p == 'S' || *p == 'B') {
-        return true;
+        CHECK(!tooLate);
+        result = true;
       } else if (*p != '%' && *p != 's' && *p != 'd' &&
           !((*p == 'z' || *p == 'j') && (p[1] == 'd' || p[1] == 'u'))) {
-        return false;
+        tooLate = true;
       }
     }
   }
-  return false;
+  return result;
 }
 
 // Some standard formatting codes (e.g., %d) are handled here
--- a/lib/parser/message.cc
+++ b/lib/parser/message.cc
@@ -33,6 +33,56 @@ std::ostream &operator<<(std::ostream &o, const MessageFixedText &t) {
   return o;
 }
 
+static bool NeedsSpecialFormatting(const char *p) {
+  while (*p != '\0') {
+    if (*p++ == '%') {
+      if (*p == 'S' || *p == 'B') {
+        return true;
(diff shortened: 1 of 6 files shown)


lib/parser/message.cc
```diff
@@ -33,29 +33,82 @@ std::ostream &operator<<(std::ostream &o, const MessageFixedText &t) {
...
+    } else {
+      ++p;
+      switch (*p++) {
+      case '%': result += '%'; break;
+      case 's': result += va_arg(ap, const char *); break;
+      case 'd': result += std::to_string(va_arg(ap, int)); break;
+      case 'S': result += va_arg(ap, std::string); break;
+      case 'B': result += va_arg(ap, CharBlock).ToString(); break;
+      default:
+        if (*p == 'd') {
```
vdonaldson: Could other cases such as 'u' occur here?

klausler: Sure.  Any formatting accepted by the underlying `vsnprintf()` can appear, and if it's not handled by the cases in this code, it'll end up being handled in `vsnprintf()` below.

vdonaldson: I'm puzzled as to why %jd is different than $ju

klausler: The difference is that once we hit a format that this code doesn't special-case, it goes off to `vsnprintf()` to deal with the rest of the formatting, so any instances of the special `%B` and `%S` formats after that point won't be expanded.  So there's some special casing here of those format codes that might appear before a `%B` or `%S` so that we don't have to call `vsnprintf()` for them.  Is there an instance of `%ju` formatting anywhere?

vdonaldson: I have a %ju in new code.   There might also be future instances.

lib/parser/message.cc
```diff
@@ -33,29 +33,92 @@ std::ostream &operator<<(std::ostream &o, const MessageFixedText &t) {
...
 }
 
+static bool NeedsSpecialFormatting(const char *p) {
+  while (*p != '\0') {
+    if (*p++ == '%') {
+      if (*p == 'S' || *p == 'B') {
+        return true;
+      } else if (*p != '%' && *p != 's' && *p != 'd' &&
+          !((*p == 'z' || *p == 'j') && (p[1] == 'd' || p[1] == 'u'))) {
+        return false;
```
tskeith: If we hit this case, shouldn't we keep going to see if there is %S or %B after this code we don't handle? Then we could assert rather than outputting a bad message.

klausler: Done.

jeanPerier:
4mins too late :). Never mind then, I am fine with the merged solution also.

PR #443, opened 2019-05-06
Process specification parts before execution parts

tskeith:
Change the order in which names are resolved. Before resolving names
in the execution part of a subprogram we need to know the interface
of contained subprograms. This is because the type of some construct
entities can depend on the return type of contained functions, e.g.
```
  associate(x => f())
  end associate
contains
  function f()...
```

To do this, we now build a tree rooted at each program unit with
child nodes corresponding to subprograms contained in the parent.
This provides flexibility in choosing an order to resolve names.
The current implementation processes all specification parts before
any execution parts. This ensures contained subprogram interfaces
are know before analyzing constructs like ASSOCIATE.

Resolving a specification part involves first adding
`SubprogramNameDetails` symbols for each contained subprogram, then
processing the statement that introduces the program unit (`ModuleStmt`,
`SubroutineStmt`, etc.), then visiting all of the statements in the
specification part.

If it proves necessary, we can add a phase to do implicit declarations
in the execution part before processing the specification part of
contained subprograms.


lib/semantics/resolve-names-utils.cc
```diff
@@ -286,4 +286,158 @@ Bound ArraySpecAnalyzer::GetBound(const parser::SpecificationExpr &x) {
...
+    for (const auto &prefix : *prefixes) {
+      if (std::holds_alternative<parser::PrefixSpec::Module>(prefix.u)) {
+        return true;
+      }
+    }
+  }
+  return false;
+}
+
+ProgramTree::Kind ProgramTree::GetKind() const {
```
klausler: This could be implemented as a `constexpr` templatized variable, too.

```
template<typename A> constexpr ProgramTree::Kind /*or whatever the type is */ KindOfStmt;
template<> constexpr ProgramTree::Kind<parser::Statement<parser::ProgramStmt>> KindOfStmt{ProgramTree::Kind::Program};
// &c.
return std::visit([](const auto *x){ return KindOfStmt<std::decay_t<decltype(*x)>>; }, stmt_);
```
which may or may not be as clear but would be fewer lines of code.

lib/semantics/resolve-names.cc
```diff
@@ -332,13 +329,14 @@ class ImplicitRulesVisitor : public DeclTypeSpecVisitor {
...
 
 protected:
-  void PushScope();
-  void PopScope();
-  void ClearScopes() { implicitRules_.reset(); }
+  void BeginScope(const Scope &);
+  void SetScope(const Scope &);
 
 private:
+  // scope -> implicit rules for that scope
```
klausler: Why not use a data member in `Scope`?  Is it because this mapping can change, or is emphemeral?

lib/semantics/resolve-names-utils.h
```diff
@@ -76,6 +67,69 @@ ArraySpec AnalyzeArraySpec(
 ArraySpec AnalyzeCoarraySpec(
     SemanticsContext &context, const parser::CoarraySpec &);
 
-}
+// A tree of program units and their contained subprograms.
```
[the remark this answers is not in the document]
tskeith: OK, I'll do that.

lib/semantics/resolve-names.cc
tskeith: The latter. I want to avoid putting things in Scope that aren't needed after semantic analysis.

lib/semantics/resolve-names-utils.cc
tskeith: My attempt at using this approach is below. I don't think it's clearer or shorter.
```
template<typename A> constexpr ProgramTree::Kind KindOfStmt;
template<>
constexpr ProgramTree::Kind KindOfStmt<parser::Statement<parser::ProgramStmt>>{
    ProgramTree::Kind::Program};
template<>
constexpr ProgramTree::Kind KindOfStmt<parser::Statement<parser::FunctionStmt>>{
    ProgramTree::Kind::Function};
template<>
constexpr ProgramTree::Kind
    KindOfStmt<parser::Statement<parser::SubroutineStmt>>{
        ProgramTree::Kind::Subroutine};
template<>
constexpr ProgramTree::Kind
    KindOfStmt<parser::Statement<parser::MpSubprogramStmt>>{
        ProgramTree::Kind::MpSubprogram};
template<>
constexpr ProgramTree::Kind KindOfStmt<parser::Statement<parser::ModuleStmt>>{
    ProgramTree::Kind::Module};
template<>
constexpr ProgramTree::Kind
    KindOfStmt<parser::Statement<parser::SubmoduleStmt>>{
        ProgramTree::Kind::Submodule};

ProgramTree::Kind ProgramTree::GetKind() const {
  return std::visit(
      [](const auto *x) { return KindOfStmt<std::decay_t<decltype(*x)>>; },
      stmt_);
```

lib/semantics/resolve-names-utils.h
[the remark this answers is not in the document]
tskeith: I've moved ProgramTree to `program-tree.{h,cc}`

tskeith:
> Is there still a place-holding subprogram name "details" variant for Symbol that is no longer needed?

We still need `SubprogramNameDetails`. For example, in `procedure(p) :: x`, `p` may be the name of an internal subprogram. So we need still need a symbol for `p` before we have processed its specification part.

PR #444, opened 2019-05-06
Replace formatting of CharBlock & string

klausler:
Based on a suggestion from Jean.  Allows for formatting of CharBlock and std::string with regular '%s' formats.

### DIFF
--- a/lib/parser/message.cc
+++ b/lib/parser/message.cc
@@ -64,12 +64,21 @@ const char *MessageFormattedText::Convert(const std::string &s) {
   return conversions_.front().c_str();
 }
 
+const char *MessageFormattedText::Convert(std::string &s) {
+  conversions_.emplace_front(s);
+  return conversions_.front().c_str();
+}
+
 const char *MessageFormattedText::Convert(std::string &&s) {
   conversions_.emplace_front(std::move(s));
   return conversions_.front().c_str();
 }
 
 const char *MessageFormattedText::Convert(const CharBlock &x) {
+  return Convert(x.ToString());
+}
+
+const char *MessageFormattedText::Convert(CharBlock &x) {
   return Convert(x.ToString());
 }
 
@@ -80,6 +89,8 @@ const char *MessageFormattedText::Convert(CharBlock &&x) {
 const char *MessageFormattedText::Convert(const Name &n) {
   return Convert(n.source);
 }
+
+const char *MessageFormattedText::Convert(Name &n) { return Convert(n.source); }
 
 const char *MessageFormattedText::Convert(Name &&n) {
   return Convert(n.source);
--- a/lib/parser/message.cc
+++ b/lib/parser/message.cc
@@ -14,7 +14,6 @@
 
 #include "message.h"
 #include "char-set.h"
-#include "parse-tree.h"
 #include "../common/idioms.h"
 #include <algorithm>
 #include <cstdarg>
@@ -84,16 +83,6 @@ const char *MessageFormattedText::Convert(CharBlock &x) {
 
 const char *MessageFormattedText::Convert(CharBlock &&x) {
   return Convert(x.ToString());
-}
-
-const char *MessageFormattedText::Convert(const Name &n) {
-  return Convert(n.source);
-}
-
-const char *MessageFormattedText::Convert(Name &n) { return Convert(n.source); }
-
-const char *MessageFormattedText::Convert(Name &&n) {
-  return Convert(n.source);
 }
 
 std::string MessageExpectedText::ToString() const {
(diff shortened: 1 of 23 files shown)


lib/parser/message.cc
```diff
@@ -107,21 +43,23 @@ MessageFormattedText::MessageFormattedText(MessageFixedText text, ...)
...
   va_start(ap, text);
-  if (NeedsSpecialFormatting(p)) {
-    string_ = SpecialFormatting(p, ap);
-  } else {
-    char buffer[256];
-    vsnprintf(buffer, sizeof buffer, p, ap);
-    string_ = buffer;
-  }
+  char buffer[256];
+  vsnprintf(buffer, sizeof buffer, p, ap);
```
sscalpone: Why 256?  A call to vnsprintf will return the required size for the result if the first arg is null.

klausler: Stack allocation is cheap.  I'll extend the code to handle overflow.

lib/parser/message.h
```diff
@@ -67,14 +68,17 @@ constexpr MessageFixedText operator""_err_en_US(
...
+// following arguments.  CharBlock and std::string values are also
+// accepted; they are automatically converted to char pointers
+// that are suitable for '%s' formatting.
 class MessageFormattedText {
 public:
-  MessageFormattedText(MessageFixedText, ...);
+  template<typename... A>
+  MessageFormattedText(const MessageFixedText &text, A &&... x)
+    : isFatal_{text.isFatal()} {
+    Format(text, Convert(std::move(x))...);
```
jeanPerier: Maybe the `std::move(x)` should be a  `std::forward<A>(x)` due to the perfect forwarding context. That said, `Convert()` are not "stealing" any storage from there arguments from what I see, so there should not be any issue. But it also means that maybe taking rvalue args does not bring much here in the end compare to the simpler `const A&`.

lib/parser/message.cc
```diff
@@ -107,21 +43,23 @@ MessageFormattedText::MessageFormattedText(MessageFixedText text, ...)
   }
   va_list ap;
   va_start(ap, text);
```
tskeith: I get a warning here from clang:
```
../../lib/parser/message.cc:45:16: warning: passing an object of reference type to 'va_start' has undefined behavior [-Wvarargs]
  va_start(ap, text);
               ^
```

klausler:
Ok, I'll try to add `Name` support as well.

lib/parser/message.h
```diff
@@ -84,8 +88,14 @@ class MessageFormattedText {
...
 private:
-  std::string string_;
+  void Format(MessageFixedText text, ...);
+  template<typename A> A Convert(A &&x) { return std::forward<A>(x); }
+  const char *Convert(const std::string &x) { return x.c_str(); }
+  const char *Convert(CharBlock);
+
   bool isFatal_{false};
+  std::string string_;
+  std::vector<std::string> conversions_;  // preserves created strings
```
jeanPerier: References/pointers to `std::vector` elements are not guaranteed to be valid after new elements are inserted/emplaced. In fact, I caught this because libc++ is resizing the container before inserting the second name in message generated at `lib/semantics/expression.cc:1330`. This resulted in rubbish in the first name such as  `Externally visible object 'ø<80>^C' must ...`
Anyway, using a `std::list` fixes the issue and is guaranteed to preserve the references through its life-time: [cpp ref source](https://en.cppreference.com/w/cpp/container#Iterator_invalidation).

lib/parser/message.h
```diff
@@ -84,8 +88,14 @@ class MessageFormattedText {
   std::string MoveString() { return std::move(string_); }
 
 private:
-  std::string string_;
+  void Format(MessageFixedText text, ...);
+  template<typename A> A Convert(A &&x) { return std::forward<A>(x); }
+  const char *Convert(const std::string &x) { return x.c_str(); }
```
jeanPerier: I tried to see if this PR is fixing #446. Unfortunately, it is not. The issue is that when getting `MessageFormattedText{text, s}` where s is a `std::string`  that is not const, the compiler instantiates the template version with `A=std::string&` rather than selecting the `const std::string&` version as it would have done if there was no template version.... Sight, mixing templates and overloads is amazingly complex.
Three solutions I see:
- add other overloads (empirical and endless fight IMHO)
- make it only one function with `auto` return type and `if constexpr` on types
- add an SFINAE barrier on the template version

I favor the last one (tested to fix #446 with clang-7.1 and its libc++):
```
using DisableForTypes = std::enable_if_t<(... && !std::is_same_v<A, B>)>;
  template<typename A,
      typename = DisableForTypes<std::decay_t<A>, std::string, CharBlock>>
  A Convert(A &&x) {
    return std::forward<A>(x);
  }
```

klausler:
`Name` support doesn't seem like the best idea until I feel more confident about using forwarding in templates, so I'm backing it out.

PR #449, opened 2019-05-07
Handle remaining cases in expression tree traversal

klausler:
Pete noticed some odd behavior; this PR fills in the gaps, and removes the dangerous "no-op by default" base case so that we know all cases are handled.

### DIFF
--- a/lib/evaluate/variable.h
+++ b/lib/evaluate/variable.h
@@ -273,6 +273,8 @@ struct DataRef {
 // In the F2018 standard, substrings of array sections are parsed as
 // variants of sections instead.
 class Substring {
+  using Parent = std::variant<DataRef, StaticDataObject::Pointer>;
+
 public:
   CLASS_BOILERPLATE(Substring)
   Substring(DataRef &&parent, std::optional<Expr<SubscriptInteger>> &&lower,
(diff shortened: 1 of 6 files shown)


lib/semantics/check-do-concurrent.cc
```diff
@@ -339,7 +339,8 @@ static CS GatherReferencesFromExpression(const parser::Expr &expression) {
       explicit CollectSymbols(int) {}
       void Handle(const Symbol *symbol) { result().push_back(symbol); }
     };
-    return evaluate::Visitor<CollectSymbols>{0}.Traverse(*expression.typedExpr);
+    return evaluate::Visitor<CollectSymbols>{0}.Traverse(
+        expression.typedExpr->v);
```
tskeith: I think it is clearer if we avoid `typedExpr` and use `GetExpr()` instead:
```
  if (const auto *expr{GetExpr(expression)}) {
    ...
    return evaluate::Visitor<CollectSymbols>{0}.Traverse(*expr);
  } else ...
```

Also, we should just get rid of the "else" part. If there isn't an analyzed expression we should just skip it. We don't need an alternate implementation.

lib/evaluate/variable.h
```diff
@@ -288,7 +290,12 @@ class Substring {
   }
 
   Expr<SubscriptInteger> lower() const;
+  Expr<SubscriptInteger> *lower();
   Expr<SubscriptInteger> upper() const;
+  Expr<SubscriptInteger> *upper();
```
jeanPerier: In `variable.h` classes, the following accessors of optional members are returning `std::optional` instead of a pointer:

- `Triplet::upper()`
- `Triplet::lower()`
- `CoarrayRef::stat()`
- `CoarrayRef::team()`

Should theses accessors be changed to return pointer instead so that they follow the same pattern as `Substring::lower()` and `Substring::upper()`?

I did not get far enough when experimenting with folding using the mutable visitor framework, but it seems to me that returning optional is an issue for the mutable `Descend` because it ends up descending in copies of the member rather than the member themselves.

klausler: Yes, that's a bad interface design.  I'll clean it up before merging.

PR #452, opened 2019-05-08
Added a test for C1121, DO scalar-mask-expr cannot reference IMPURE procedure

psteinfeld:
scalar-mask-expr cannot reference an IMPURE procedure.

### DIFF
--- a/test/semantics/dosemantics02.f90
+++ b/test/semantics/dosemantics02.f90
@@ -1,0 +1,33 @@
+! Copyright (c) 2019, NVIDIA CORPORATION.  All rights reserved.
+!
+! Licensed under the Apache License, Version 2.0 (the "License");
+! you may not use this file except in compliance with the License.
+! You may obtain a copy of the License at
+!
+!     http://www.apache.org/licenses/LICENSE-2.0
+!
+! Unless required by applicable law or agreed to in writing, software
+! distributed under the License is distributed on an "AS IS" BASIS,
+! WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+! See the License for the specific language governing permissions and
+! limitations under the License.
+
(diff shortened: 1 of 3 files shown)


test/semantics/dosemantics02.f90
```diff
@@ -0,0 +1,33 @@
...
+! Unless required by applicable law or agreed to in writing, software
+! distributed under the License is distributed on an "AS IS" BASIS,
+! WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+! See the License for the specific language governing permissions and
+! limitations under the License.
+
+! C1121 -- any procedure referenced in a concurrent header must be pure
+
+! RUN: ${F18} -funparse-with-symbols %s 2>&1 | ${FileCheck} %s
+! CHECK: image control statement not allowed in DO CONCURRENT
```
sscalpone: What's this?

psteinfeld: Oops.  This is a copy/paste leftover.

Thanks for catching it.  I'll remove it.

test/semantics/dosemantics02.f90
```diff
@@ -0,0 +1,33 @@
...
+
+SUBROUTINE do_concurrent_c1121(i,n)
+  IMPLICIT NONE
+  INTEGER :: i, n, flag
+!ERROR: concurrent-header mask expression cannot reference an impure procedure
+  DO CONCURRENT (i = 1:n, random() < 3)
```
tskeith: As a future enhancement, it would be friendly to mention the name of the impure procedure and associate the error with it rather than the whole statement.

psteinfeld: Good point.  I'll add a TODO in the code and add this later.

PR #454, opened 2019-05-08
Information on additional compiler debug options

psteinfeld:
When looking at the f18 driver, I noticed that some options were implemented, but not described in the overview document, so I added them.  Let me know if you think that they should not be publicized.


psteinfeld:
> The f18 driver program is temporary scaffolding code and some of its flags are just for our use now and may not survive into whatever the real driver ends up being. Please don't document any flag that we don't want to support forever.

Exactly.  I'm not sure which of these options were intentionally left undocumented.  Perhaps all of them were.  I'm asking for feedback on this.

PR #455, opened 2019-05-09
Change parse tree to allow DO loop over REAL

tskeith:
It is a common extension to allow DO loops with REAL variable and
bounds. The parse tree currently parses those with the variable
as `scalar-int-variable-name` and the bounds as `scalar-int-expr`.
That causes the INTEGER constraint to be enforced automatically.

Change the grammar and parse tree to treat them as `scalar-variable-name`
and `scalar-expr`. This allows the name and expression to be REAL,
but we will have to verify that they aren't any other type (in a future
change).

To accomplish this, add a template parameter to `LoopBounds` for the type
of the variable name (always `Scalar<Name>` or `Scalar<Integer<Name>>`).
We sometimes need names for the instantiations of `LoopBounds` so add
type aliases like `LoopControl::Bounds` for each one.


lib/parser/grammar.h
```diff
@@ -959,9 +960,16 @@ TYPE_PARSER(construct<BOZLiteralConstant>(BOZLiteral{}))
 // R1124 do-variable -> scalar-int-variable-name
 constexpr auto doVariable{scalar(integer(name))};
 
+// NOTE: In loop-control we allow REAL name and bounds too.
```
klausler: Rather than use `is_same_v` tests in the template function body, it would be perhaps more clear to just define two overloads (if that's all that are used), or have one general template and one specialization.

lib/parser/parse-tree.h
```diff
@@ -1208,24 +1208,28 @@ WRAPPER_CLASS(ArrayConstructor, AcSpec);
 // R1124 do-variable -> scalar-int-variable-name
 using DoVariable = Scalar<Integer<Name>>;
 
-template<typename A> struct LoopBounds {
+template<typename A, typename B> struct LoopBounds {
```
klausler: Maybe use `DOVAR` and `EXPR` or something like that.

tskeith:
> Can/should REAL-valued implied DO loops be used in DATA statements?

None of the other compilers I tried support this. (PGI, ifort, gfortran, NAG, xlf).

I found one place they disagree: xlf allows a real `ac-do-variable` in the execution part but not in the specification part. The others don't allow it at all.

So it allows this:
```
  real :: a, x(8)
  x = [(a, a=1.0, 8.0)]
```
but not this:
```
  real :: a, x(8) = [(a, a=1.0, 8.0)]
```

lib/parser/grammar.h
tskeith: Good idea, that is cleaner.

lib/parser/parse-tree.h
tskeith: OK. I changed them to `VAR` and `BOUND`.

sscalpone:
Please update the extensions document.

tskeith:
> Please update the extensions document.

Done

vdonaldson:
Could the grammar changes under a **deprecated** combinator?

klausler:
I'm surprised that XLF would support this relic in a new context like an array constructor.  I don't think that that's a precedent that we need to follow; the work-around is trivial.

tskeith:
> Could the grammar changes be placed under a **deprecated** combinator?

That would be nice but I'm not sure it can be done. If the parser sees `do a = b, c` it can't tell if it is the deprecated case (one or more of the variables is REAL, the others INTEGER), the illegal case (one or more is neither REAL nor INTEGER), or the legal case (all are INTEGER).

We could make a constraint wrapper (like `Integer<>`) named something like `IntegerOrDeprecatedReal<>`. Then `EnforceTypeConstraint()` would be extended to do the enforcement (OK if INTEGER, warning if REAL, error otherwise). But I don't think that is worthwhile unless we have other places we want to use it as well.

sscalpone:
It's likely that we'll want to print a message that says the feature is deprecated or an extension.  Adding the info to the grammar is a notationally convenient place to put these hooks.  But I agree with Tim that it's not worthwhile for a single case.  In this case, we can add the check in semantics.  (We need a way to remember that we want to do this check in semantics.)

tskeith:
> (We need a way to remember that we want to do this check in semantics.)

I created issue #458 for this.

PR #457, opened 2019-05-09
Keep track of test that produce confusing parse messages

psteinfeld:
In the process of creating tests for DO semantics, I created a test that compiles and executes without
error on GNU Fortran.  f18 fails to parse it, which is correct in that the program is not standard Fortran.
But the error messages seemed confusing to me and in need of improvement.

I want to keep track of this test so that, if and when we decide to improve the parser's messages, it can
serve as a test case.

### DIFF
--- a/test/semantics/CMakeLists.txt
+++ b/test/semantics/CMakeLists.txt
@@ -197,6 +197,13 @@ set(IF_TESTS
   if_stmt01.f90
 )
 
+# These tests show situations where we'd like the parser to produce
+# better error messages.  We don't run these tests normally
+# See issue 456
+#set(PARSE_TESTS
+#  parserquirk01.f90
+#)
+
 foreach(test ${ERROR_TESTS})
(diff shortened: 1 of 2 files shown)


tskeith:
I don't think parsing tests should be in `test/semantics`.

vdonaldson:
The gfortran behavior is a bug for not diagnosing C1122.

psteinfeld:
> The gfortran behavior is a bug for not diagnosing C1122.

Yes, this is really a bug in gfortran.  gfortran compiles this test even using the "-std=f2018" option., which
is supposed to force standards conformance.

psteinfeld:
> 
> 
> I don't think parsing tests should be in `test/semantics`.

This seemed flaky to me, too.  It's unclear to me that putting these files into source control is the best
way to handle this.  Alternatively, I could just attach files to the issue.

sscalpone:
Putting the src file in the issue means you have to d/l the file before you try it.

Do we need a "parser" test directory?

psteinfeld:
Since there were no approvals for this pull request, I'll just add my test case directly to the issue.

PR #461, opened 2019-05-13
correct text: pointer can point to allocatable

kweide:
Minor text change to eliminate wrong statement.

### DIFF
--- a/documentation/FortranForCProgrammers.md
+++ b/documentation/FortranForCProgrammers.md
@@ -260,8 +260,7 @@ little like C++17's `std::visit()` on a discriminated union.
 Pointers
 --------
 Pointers are objects in Fortran, not data types.
-Pointers can point to data, arrays, and subprograms, but not to other pointers
-or to an allocatable.
+Pointers can point to data, arrays, and subprograms, but not to other pointers.
 A pointer can only point to data that has the `TARGET` attribute.
 Outside of the pointer assignment statement (`P=>X`) and some intrinsic
 functions and cases with pointer dummy arguments, pointers are implicitly
--- a/documentation/FortranForCProgrammers.md
+++ b/documentation/FortranForCProgrammers.md
@@ -261,6 +261,7 @@ Pointers
 --------
 Pointers are objects in Fortran, not data types.
 Pointers can point to data, arrays, and subprograms, but not to other pointers.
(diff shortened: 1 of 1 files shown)


documentation/FortranForCProgrammers.md
```diff
@@ -260,8 +260,7 @@ little like C++17's `std::visit()` on a discriminated union.
 Pointers
 --------
 Pointers are objects in Fortran, not data types.
-Pointers can point to data, arrays, and subprograms, but not to other pointers
-or to an allocatable.
+Pointers can point to data, arrays, and subprograms, but not to other pointers.
```
klausler: What was intended by the original text is that one cannot construct a pointer in Fortran that points to the *implementation* of an allocatable object as such, as opposed to the data held within the allocatable.  One can certainly have a pointer that points to those contents, so long as the allocatable has the TARGET attribute, just as with any other data object.

kweide:
How about:

  Pointers can point to data, arrays, and subprograms, but not to other pointers. A pointer to an `ALLOCATABLE` points to to the `ALLOCATABLE's` data.

klausler:
That's great (except that `points to to` should be `points to`).

kweide:
indeed,

klausler:
Please retain the sense of the original text; pointers can point to the data managed by an allocatable, but cannot point to the allocatable itself (i.e., one cannot deallocate or reallocate an allocatable by means of a pointer to it, nor do the automatic reallocation semantics apply when an assignment is made to an allocatable through an intervening pointer).

kweide:
Please reformulated as you see fit. Maybe you should jut copy the text in the parentheses of your last comment.

It seems that you want to address some false expectations that C or C++ programmers may have. Your are probably in a better position to address those than I, since I am more familiar with Fortran than C or C++ and thus do not share those expectation, and even find them somewhat hard to understand.

I just want to point out that the sentence "Pointers cannot point to an allocatable" is a false statement, if the terms in that sentence are taken to have their Fortran meanings.

klausler:
A PR is out for review that extends and corrects the text.

sscalpone:
Hi @kweide, does #464 address your concerns?

kweide:
Yes, looks fine to me.

PR #462, opened 2019-05-15
Update description of semantic analysis

tskeith:
Semantics.md described the plans for semantic analysis as of some
time ago. Update it to document what has been implemented.

### DIFF
--- a/documentation/Semantics.md
+++ b/documentation/Semantics.md
@@ -4,8 +4,15 @@ Copyright (c) 2018-2019, NVIDIA CORPORATION.  All rights reserved.
 
 # Semantic Analysis
 
-The semantic analysis pass will take a parse tree for a syntactically
-correct Fortran program and determine whether it is legal by enforcing
+The semantic analysis pass determines if a syntactically correct Fortran
+program is is legal by enforcing the constraints of the language.
+
+The input is a parse tree with a `Program` node at the root;
+and a "cooked" character stream, a contiguous stream of characters
+containing a normalized form of the Fortran source.
+
+The semantic analysis pass takes a parse tree for a syntactically
+correct Fortran program and determines whether it is legal by enforcing
 the constraints of the language.
 
 If the program is not legal, the results of the semantic pass will be a list of
@@ -15,7 +22,7 @@ If the program is legal, the semantic pass will produce a (possibly modified)
 parse tree for the semantically correct program with each name mapped to a symbol
 and each expression fully analyzed.
 
-All user errors are detected during semantic analysis.
+All user errors are detected either prior to or during semantic analysis.
 After it completes successfully the program should compile with no error messages.
 There may still be warnings or informational messages.
 
@@ -32,11 +39,17 @@ There may still be warnings or informational messages.
    Fix incorrect parses based on symbol information
 5. [Expression analysis](#expression-analysis) -
    Analyze all expressions in the parse tree and fill in `Expr::typedExpr` and
-   `Variable::typedExpr` with analyzed expressions
+   `Variable::typedExpr` with analyzed expressions; fix incorrect parses
+   based on the result of this analysis
 6. [Statement semantics](#statement-semantics) -
    Perform remaining semantic checks on the execution parts of subprograms
 7. [Write module files](#write-module-files) -
    If no errors have occurred, write out `.mod` files for modules and submodules
+
+If phase 1 or phase 2 encounter an error on any of the program units,
+compilation terminates. Otherwise, phases 3-6 are all performed even if
+errors occur.
+Module files are written (phase 7) only if there are no errors.
 
 ### Validate labels
 
@@ -108,7 +121,7 @@ Any remaining expressions are analyzed in this phase.
 For each `Variable` and top-level `Expr` (i.e. one that is not nested below
 another `Expr` in the parse tree) the analyzed form of the expression is saved
 in the `typedExpr` data member. After this phase has completed, the analyzed
-expression can be accessed using `GetExpr()` in `lib/semantics/tools.cc`.
+expression can be accessed using `semantics::GetExpr()`.
 
 This phase also corrects mis-parses based on the result of expression analysis:
 - An expression like `a(b)` is parsed as a function reference but may need
--- a/documentation/Semantics.md
+++ b/documentation/Semantics.md
@@ -48,7 +48,7 @@ Perform semantic checks related to labels and branches:
 
 ### Rewrite DO loops
 
-This phases normalizes the parse tree by removing all unstructures `DO` loops
+This phase normalizes the parse tree by removing all unstructured `DO` loops
 and replacing them with `DO` constructs.
 
 ### Name resolution
@@ -75,10 +75,10 @@ Types are also owned by scopes.
 
 Name resolution happens on the parse tree in this order:
 1. Process the specification of a program unit:
-  1. Create a new scope for the unit
-  2. Create a symbol for each contained subprogram containing just the name
-  3. Process the opening statement of the unit (`ModuleStmt`, `FunctionStmt`, etc.)
-  4. Process the specification part of the unit
+   1. Create a new scope for the unit
+   2. Create a symbol for each contained subprogram containing just the name
+   3. Process the opening statement of the unit (`ModuleStmt`, `FunctionStmt`, etc.)
+   4. Process the specification part of the unit
 2. Apply the same process recursively to nested subprograms
 3. Process the execution part of the program unit
 4. Process the execution parts of nested subprograms recursively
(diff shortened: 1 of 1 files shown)


documentation/Semantics.md
```diff
@@ -1,206 +1,139 @@
...
+
+Perform semantic checks related to labels and branches:
+- check that any labels that are referenced are defined and in scope
+- check branches into loop bodies
+- check that labeled `DO` loops are properly nested
+- check labels in data transfer statements
+
+### Rewrite DO loops
+
+This phases normalizes the parse tree by removing all unstructures `DO` loops
```
ichoyjx: unstructured ?

documentation/Semantics.md
```diff
@@ -1,206 +1,139 @@
...
+3. Process the execution part of the program unit
+4. Process the execution parts of nested subprograms recursively
+
+After the completion of this phase, every `Name` corresponds to a `Symbol`
+unless an error occurred.
+
+### Rewrite parse tree
+
+The parser cannot build a completely correct parse tree without symbol information.
+This phases correct mis-parses based on symbols:
```
ichoyjx: phase

documentation/Semantics.md
```diff
@@ -1,206 +1,139 @@
...
+  to be rewritten to an array element reference (if `a` is an object entity)
+  or to a structure constructor (if `a` is a derive type)
+- An expression like `a(b:c)` is parsed as an array section but may need to be
+  rewritten as a substring if `a` is an object with type CHARACTER
+
+### Statement semantics
+
+Multiple independent checkers driven by the `SemanticsVisitor` framework
+perform the remaining semantic checks.
+By this phase, and names and expressions that can be successfully resolved
```
ichoyjx: ~~and~~

documentation/Semantics.md
```diff
@@ -1,206 +1,139 @@
...
+have been. But there may be names without symbols or expressions without
```
ichoyjx: "have been" ?

tskeith: I meant "By this phase, **all** names and expressions that can be successfully resolved have been."

documentation/Semantics.md
```diff
@@ -1,206 +1,139 @@
...
+All `Symbol` objects are owned by the symbol table data structures.
+They should be accessed as `Symbol *` or `Symbol &` outside of the symbol
+table classes as they can't be created, copied, or moved.
+The `Symbol` class has functions and data common across all symbols, and a
+`details` field that contains more information specific to that type of symbol.
+Many symbols also have types, represented by `DeclTypeSpec`.
+Types are also owned by scopes.
+
+Name resolution happens on the parse tree in this order:
+1. Process the specification of a program unit:
```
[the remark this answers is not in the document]
tskeith: Thanks. I used a markdown previewer in Chrome and it looked okay, but I forgot to check on github.

tskeith:
Thanks for the comments, @ichoyjx and @jeanPerier. I've made the changes suggested.

documentation/Semantics.md
```diff
@@ -1,206 +1,139 @@
 <!--
-Copyright (c) 2018, NVIDIA CORPORATION.  All rights reserved.
+Copyright (c) 2018-2019, NVIDIA CORPORATION.  All rights reserved.
 -->
 
-The semantic pass will determine whether the input program is a legal Fortran
-program.
+# Semantic Analysis
+
+The semantic analysis pass will take a parse tree for a syntactically
```
psteinfeld: This sentence uses future tense, and the rest of the document uses present tense.  How about "The semantic pass determintes"?

documentation/Semantics.md
```diff
@@ -1,206 +1,139 @@
...
-- Array intrinsics
+If the program is legal, the semantic pass will produce a (possibly modified)
+parse tree for the semantically correct program with each name mapped to a symbol
+and each expression fully analyzed.
+
+All user errors are detected during semantic analysis.
+After it completes successfully the program should compile with no error messages.
+There may still be warnings or informational messages.
+
+## Phases of Semantic Analysis
```
psteinfeld: It would be good to know what happens in the compiler before semantic analysis -- options are processed (some of which affect semantic analysis), a "cooked" character stream is constructed, the program is parsed, and source provenance information is generated.

documentation/Semantics.md
```diff
@@ -1,206 +1,139 @@
...
-TBD
-
-### Constant Expression Evaluation
-- Scalars
-- Array intrinsics
+If the program is legal, the semantic pass will produce a (possibly modified)
+parse tree for the semantically correct program with each name mapped to a symbol
+and each expression fully analyzed.
+
+All user errors are detected during semantic analysis.
```
psteinfeld: I would say "All user errors are detected either prior to or during semantic analysis".

documentation/Semantics.md
```diff
@@ -1,206 +1,139 @@
...
+After it completes successfully the program should compile with no error messages.
+There may still be warnings or informational messages.
+
+## Phases of Semantic Analysis
+
+1. [Validate labels](#validate-labels) -
```
psteinfeld: Should we refer to the files in which these actions take place?  I see that you referred so some source files.

psteinfeld: This overview describes the sequence in the case where there are no errors.  It would be good to know what causes semantic analysis to stop.  For example, is the parse tree rewritten if there are errors in name resolution?

psteinfeld:
Thanks for doing this!

documentation/Semantics.md
tskeith: This is in Overview.md. Do you think we should mention that here?

documentation/Semantics.md
```diff
@@ -1,206 +1,139 @@
...
+   Check all constraints on labels and branches
+2. [Rewrite DO loops](#rewrite-do-loops) -
+   Convert all occurrences of `LabelDoStmt` to `DoConstruct`.
+3. [Name resolution](#name-resolution) -
+   Analyze names and declarations, build a tree of Scopes containing Symbols,
+   and fill in the `Name::symbol` data member in the parse tree
+4. [Rewrite parse tree](#rewrite-parse-tree) -
+   Fix incorrect parses based on symbol information
+5. [Expression analysis](#expression-analysis) -
```
vjayathirtha-nv: Point 5 should also mention that it fixes certain incorrect parses right?

documentation/Semantics.md
psteinfeld: I just looked at Overview.md.  There's a lot of information that's duplicated in this document.

I do think that it's good for the reader of this document to have an idea of what the inputs are to semantic analysis even if that repeats some information.  Also, it looks to me like Overview.md should be rewritten once this document is updated.  For example, both documents describe label validation and name resolution.

vjayathirtha-nv:
The earlier version had some TBDs and/or placeholders like `Utility Routines` , `Diagnostic Output`, etc
Should we keep those and in general placeholders for things that need to be thought about?

documentation/Semantics.md
tskeith: I agree that Overview.md will need some work.

I'll add a little more to this document about the inputs.

documentation/Semantics.md
tskeith: I think I'll remove the reference to `tools.cc` and stick to just C++ names.

tskeith: I'll describe how it works now, though I think it is likely to change.

documentation/Semantics.md
tskeith: Right

tskeith:
> The earlier version had some TBDs and/or placeholders like `Utility Routines` , `Diagnostic Output`, etc
> Should we keep those and in general placeholders for things that need to be thought about?

I don't think this document is a good place to record TODOs.

documentation/Semantics.md
```diff
@@ -1,206 +1,152 @@
...
+   `Variable::typedExpr` with analyzed expressions; fix incorrect parses
+   based on the result of this analysis
+6. [Statement semantics](#statement-semantics) -
+   Perform remaining semantic checks on the execution parts of subprograms
+7. [Write module files](#write-module-files) -
+   If no errors have occurred, write out `.mod` files for modules and submodules
+
+If phase 1 or phase 2 encounter an error on any of the program units,
+compilation terminates. Otherwise, phases 3-6 are all performed even if
+errors occur.
```
tskeith: We need to think about whether this is the right thing to do. Maybe we can keep going if some of the branch and label constraints aren't met. But if we are unable to convert a label DO to a DO construct, things may be too messed up to proceed.

Also, what if there are multiple program units and one has an error? Should we continue with the others? If the one with errors is a module, other program units that USE it probably shouldn't be analyzed further.

PR #468, opened 2019-05-22
Add location to folding warnings and related tests and prevent spurious "re-folding"

jeanPerier:
So far, non-fatal messages created while folding were printed without location information. This looked weird.

This PR:

- Add location information (simply by setting the context location before calling fold)

- Update parser node typedExpr expressions after folding to avoid doing several times the same work (and generating several times the same warnings...). So far, folding was done at least twice from scratch for all parser::Constant<A> expression.

- Add tests for operation folding "edge cases" (the one that generate warnings). Both checks expected values and warning messages

- Fix parentheses handling in intrinsic function folding (`acos((0.5_4))`) was not folding properly because a bare `UnwrapExpr<Constant<T>>` was used instead of new `GetConstantValue<T>`

### DIFF
--- a/test/evaluate/folding03.f90
+++ b/test/evaluate/folding03.f90
@@ -15,261 +15,289 @@
 ! Test operation folding edge case (both expected value and messages)
 ! These tests make assumptions regarding real(4) and integer(4) extrema.
 
-program test
-  integer(4), parameter :: i4_neg_max = int(Z'FFFFFFFF')
-  integer(4), parameter :: i4_nmax = -2147483647_4 - 1_4 !TODO bug that -2147483648_4 cannot be written ?
-  integer(4), parameter :: i4_nmax2 = -2147483648_4
-  integer(4), parameter :: i4_pmax1 = 2147483647_4
-  integer(4), parameter :: i4_pmax2 = 2147483648_4
+#define TEST_ISNAN(v) logical, parameter :: test_##v =.NOT.(v.EQ.v)
 
-print*, "i4_neg_max ", i4_neg_max
-print*, "i4_nmax ", i4_nmax
-print*, "i4_nmax2 ", i4_nmax2
-print*, "i4_pmax1 ", i4_pmax1
-print*, "i4_pmax2 ", i4_pmax2
+! TODO: 2 behaviors to clarify:
 
-end program
+!  Biggest negative literal integer(4): -(2**31) or -(2**31-1) ?
+!   -> Fortran standard says representable value set must be in [-HUGE, HUGE].
+!      For f18 integer(4) that means [-(2**31-1), 2**31-1].
+!   -> Current f18 behavior is to emit an error for -2147483648_4 literal (-2**31), but not for
+!      -2147483647_4 -1_4 which is folded to -2147483648_4 without errors/warnings.
+!  It seems to me that biggest negative literal and representable value should
+!  be the same.
+!   -> pgfortran and ifort are OK with -2147483648_4 literal.
+!   -> gfortran gives an error for both -2147483648_4 unless -fno-range-check is used.
+!      It never raises errors for -2147483647_4 -1_4 and fold it to -2147483648_4.
 
-!module integer_tests
-  ! These test assume a 
-!  integer(4), parameter :: i4_neg_max = int(Z'FFFFFFFF')
-!  integer(4), parameter :: i4_pmax = 2147483647_4
-!  integer(4), parameter :: i4_nmax = -2147483647_4 - 1_4 !TODO bug that -2147483648_4 cannot be written ?
-! ! Or maybe it is a bug that it -2147483648_4 can be reached because it is not
-! ! representable in the 16.4 bit model... ??
-!
-!  ! TODO: Should we test something for 0/0 hw seems to handle this differently (0 or pinf)?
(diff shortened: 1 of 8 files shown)


test/evaluate/folding03.f90
```diff
@@ -0,0 +1,305 @@
...
+! WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+! See the License for the specific language governing permissions and
+! limitations under the License.
+
+! Test operation folding edge case (both expected value and messages)
+! These tests make assumptions regarding real(4) and integer(4) extrema.
+
+#define TEST_ISNAN(v) logical, parameter :: test_##v =.NOT.(v.EQ.v)
+
+! TODO: 2 behaviors to clarify:
```
jeanPerier: Testing of edge-cases raised two questions detailed as a TODO in folding03.f90 test cases: I think integer divide by zero should be made a fatal error (I have not made the change so far). Also, is -2147483648_4 literal not supported on purpose ?

klausler: Unfortunately, Fortran doesn't have signed integer literals in expressions, so `-2147483648_4` is actually a negation operation `-(2147483648_4)`.

Division by zero during folding should (I think) be a warning, not a fatal error, since it can be possible that the value of the expression is not needed at runtime.

test/evaluate/folding03.f90
```diff
@@ -0,0 +1,305 @@
...
+!  Biggest negative literal integer(4): -(2**31) or -(2**31-1) ?
+!   -> Fortran standard says representable value set must be in [-HUGE, HUGE].
+!      For f18 integer(4) that means [-(2**31-1), 2**31-1].
+!   -> Current f18 behavior is to emit an error for -2147483648_4 literal (-2**31), but not for
+!      -2147483647_4 -1_4 which is folded to -2147483648_4 without errors/warnings.
+!  It seems to me that biggest negative literal and representable value should
+!  be the same.
+!   -> pgfortran and ifort are OK with -2147483648_4 literal.
+!   -> gfortran gives an error for both -2147483648_4 unless -fno-range-check is used.
+!      It never raises errors for -2147483647_4 -1_4 and fold it to -2147483648_4.
```
klausler: As I noted in a comment above, `-2147483647_4` parses to a unary negation `-(2147483647_4)`.  If you want to allow it, you'll need to change the grammar.

test/evaluate/folding03.f90
```diff
@@ -0,0 +1,305 @@
...
+
+! Proposal: add option to control literal range. Make -2147483648_4 accepted by
+! default?
+
+! How to handle 0_4/0_4:
+!  -> raise an error (pgfortran, gfortran, ifort)
+!  -> warning and return 2147483647_4 (currently f18)
+!  -> warning and return 0_4 (maybe power HW (not tested))
+! Actually, gfortran, pgfortran and ifort forbid integer divide by zero in all
+! case.
```
klausler: Division by zero in the computation of the value of an integer `PARAMETER` or other constant expression should be fatal, I think.  So should overflow (including `MAX_NEGATIVE_INTEGER/(-1)`) in division and other operations.  But these exceptional cases shouldn't necessarily be fatal in folded expressions outside constant expressions, just warnings.

test/evaluate/folding03.f90
jeanPerier: Thanks for explaining, I had not noticed that `literal-constant` was the union of the literal constants without signs. The funny thing is that `data i /-2147483648_4/ ` is then ok because from the Fortran standard says `signed-int-literal-constant` in R845.
Anyway, I do not think this edge case is worth modifying the grammar without knowing if it is actually useful.

test/evaluate/folding03.f90
jeanPerier: Right, then the folding errors have to be handled a bit differently than they currently are. The folding library does not know if the expression being folded is for a constant or not. So, it should report division by zero and overflows without judging of the severity and let semantics make it fatal or not, alternatively, the folding context could be configured before folding to indicate if these cases should be fatal or not. I would go for the second option to avoid having to add code handling errors every time Fold is called.
Should I do this now or should we do this when improving the error/warning system to classify warnings and give the option to make them fatal ?

klausler: Folding should emit a warning and then return the original expression (possibly reduced to something canonical like 0/0).  In contexts where a constant expression is required, the fact that the result is not a constant will elicit the fatal error.

jeanPerier: > Folding should emit a warning and then return the original expression (possibly reduced to something canonical like 0/0)

OK, that sounds even easier. Here is the list of operations that I will then modify in fold.cc to return the original expression:
- Integer division
   - division by zero
   - overflow
-  Integer addition, subtraction, multiplication, power, negation
    - overflows

Real/Complex operations stay unchanged (NaN and +/-inf are here to propagate folding issues).

Do you agree with the list?

klausler: I don't think that we should return the original expression in case of integer overflow -- just return the truncated value.  That matches what would have taken place at runtime.

jeanPerier: OK, new commits make integer division by zero in constants a fatal error.
I first went the wrong path of expecting `Constant<T` in such expression.... I should have read `IsConstantExpr` comment more carefully... I took the opportunity to remove what looked like to me an useless refold in  `EvaluateConvertedExpr` (resolve-names.cc).

PR #469, opened 2019-05-22
Prevent possible infinite loop in parameter folding

jeanPerier:
The following program creates an infinite loop in folding that segfaults f18:

```
subroutine()
  real, parameter :: x = x
end subroutine 
```

The infinite loop starts in `GetParameterValue` that calls `Fold` on the init expression that ends-up recalling `GetParameterValue` with the same arguments and ...

To avoid this, set the `typeExpr` of the symbol in `GetParameterValue` to `std::nullopt` before calling `Fold` on the init expression.


jeanPerier:
> Do you have a test that causes the infinite loop?

Sure, not sure where is the best place to put it, so I've added it in `resolve37.f90`.

PR #470, opened 2019-05-24
Re-write OpenMP 4.5 grammar doc

ichoyjx:


documentation/OpenMP-4.5-grammar.txt
```diff
@@ -1,114 +1,396 @@
...
-2.16 declare-reduction -> DECLARE REDUCTION
+
+2.13.6 atomic -> ATOMIC [seq_cst[,]] READ | WRITE | UPDATE | CAPTURE [[,]seq_cst]
+2.13.6 atomic -> ATOMIC [seq_cst]
+
+2.13.7 flush -> FLUSH [(variable-name-list)]
+
+2.13.8 ordered -> ORDERED ordered-construct-clause [[[,] ordered-construct-clause]...]
+       ordered-construct-clause -> depend-clause
+
```
psteinfeld: I would expect to find the production for section 2.13.9 here.

documentation/OpenMP-4.5-grammar.txt
```diff
@@ -1,114 +1,396 @@
...
+2.14.1 cancel -> CANCEL construct-type-clause [ [,] if-clause]
+       construct-type-clause -> PARALLEL |
+                                SECTIONS |
+                                DO |
+                                TASKGROUP
+
+2.14.2 cancellation-point -> CANCELLATION POINT construct-type-clause
+
+2.15.2 threadprivate -> THREADPRIVATE (variable-name-list)
+
```
psteinfeld: I would expect to see the productions for section 2.15.3 here.

documentation/OpenMP-4.5-grammar.txt
```diff
@@ -1,114 +1,396 @@
...
+2.16 declare-reduction -> DECLARE REDUCTION (reduction-identifier : type-list : combiner)
```
psteinfeld: Is this production missing the "[initializer-clause]"?

ichoyjx:
> Be advised: A parser based on the older grammar would have to check validity of clauses for each directive during semantic analysis; a parser based on this new grammar will emit syntax errors for unexpected clauses instead. To get the best error reporting, you'll want to add error recovery to the grammar as well so that an error message like "you can't have this clause on this directive" will be generated.

That's very clear, thanks. From the current status, I think most of the validity check will happen in Semantics.

klausler:
> > Be advised: A parser based on the older grammar would have to check validity of clauses for each directive during semantic analysis; a parser based on this new grammar will emit syntax errors for unexpected clauses instead. To get the best error reporting, you'll want to add error recovery to the grammar as well so that an error message like "you can't have this clause on this directive" will be generated.
> 
> That's very clear, thanks. From the current status, I think most of the validity check will happen in Semantics.

Just so you understand, a parser based on this new grammar will not be able to construct a directive that has an invalid clause on it, so semantics won't be able to see it and complain about it.  Instead, the parser will be generating a syntax error on the directive.  If you think that most of the validity checks for invalid clauses should happen in semantics, you'll want to use the original grammar that allows any clause on any directive.  If you are willing to implement custom error recovery in the parser, use the new grammar.

documentation/OpenMP-4.5-grammar.txt
```diff
@@ -1,114 +1,396 @@
...
-2.15.4.2 copyprivate -> COPYPRIVATE (variable-name-list)
-2.15.5.1 map -> MAP ([[ALWAYS[,]] (TO | FROM | TOFROM | ALLOC | RELEASE |
+
+2.7.1 collapse-clause -> COLLAPSE (scalar-constant)
+
+2.7.1 ordered-clause -> ORDERED [(scalar-constant)]
+
+2.7.1 nowait-clause -> NOWAIT
+
+2.8.1 aligned-clause -> ALIGNED (variable-name-list[ : scalar-constant])
```
psteinfeld: The specification states that the "alignment" should be a constant positive integer expression.  Is "scalar-constant" the correct non-terminal to specify this?

sscalpone:
Does the parser allow clauses to appear in directives that are not legal?

klausler:
> Does the parser allow clauses to appear in directives that are not legal?

Prior to this change, yes.  After this change, no. 

Prior to this change, semantics had the responsibility of catching inappropriate clauses.  After this change, semantics will not see them, the parse tree would not be able to represent them, and the parser would fail to parse them.

documentation/OpenMP-4.5-grammar.txt
```diff
@@ -1,114 +1,396 @@
...
+2.15.3.4 firstprivate-clause -> FIRSTPRIVATE (variable-name-list)
+
+2.15.3.5 lastprivate-clause -> LASTPRIVATE (variable-name-list)
+
+2.15.3.6 reduction-clause -> REDUCTION (reduction-identifier: variable-name-list)
+         reduction-identifier -> + | - | * | .AND. | .OR. | .EQV. | .NEQV. |
+                                 MAX | MIN | IAND | IOR | IEOR
+
+2.15.3.7 linear-clause -> LINEAR (linear-list[ : linear-step])
+         linear-list -> list | modifier(list)
```
psteinfeld: This production includes the non-terminal "list".  But I don't see "list" defined anywhere.

documentation/OpenMP-4.5-grammar.txt
```diff
@@ -1,114 +1,396 @@
...
+         modifier -> REF | VAL | UVAL
+
+2.15.4.1 copyin-clause -> COPYIN (variable-name-list)
+
+2.15.4.2 copyprivate-clause -> COPYPRIVATE (variable-name-list)
+
+2.15.5.1 map -> MAP ([ [ALWAYS[,]] (TO | FROM | TOFROM | ALLOC | RELEASE |
```
psteinfeld: Should that "(" be in front of "TO" and the ")" follow the "DELETE"?

documentation/OpenMP-4.5-grammar.txt
ichoyjx: This `list` is like `variable-name-list`.

documentation/OpenMP-4.5-grammar.txt
ichoyjx: Yes, we don't have `ScalarPositiveIntConstantExpr` defined in the parser. The check can be deferred to Semantics.

klausler: Fortran doesn't have a constraint for positive values in its grammar.  It does have constraints for scalar, constant, and integer, so you can use those at least, and then check for a positive value in semantics.

documentation/OpenMP-4.5-grammar.txt
ichoyjx: Yes you're right, I will add that.

documentation/OpenMP-4.5-grammar.txt
ichoyjx: There is no production in Section 2.15.3. Clauses in 2.15.3.x were listed in the `Clauses` section at the end.

ichoyjx:
> > Does the parser allow clauses to appear in directives that are not legal?
> 
> Prior to this change, yes. After this change, no.
> 
> Prior to this change, semantics had the responsibility of catching inappropriate clauses. After this change, semantics will not see them, the parse tree would not be able to represent them, and the parser would fail to parse them.

@klausler This txt file change is mainly for user's reference. Are you saying that with my change, we have to change the parser to check the validities of all the productions listed on this txt file?

klausler:
> > > Does the parser allow clauses to appear in directives that are not legal?
> > 
> > 
> > Prior to this change, yes. After this change, no.
> > Prior to this change, semantics had the responsibility of catching inappropriate clauses. After this change, semantics will not see them, the parse tree would not be able to represent them, and the parser would fail to parse them.
> 
> @klausler This txt file change is mainly for user's reference. Are you saying that with my change, we have to change the parser to check the validities of all the productions listed on this txt file?

These documents specify the grammar of the language recognized by the parser and represented by the parse tree.  Are you modifying the grammar but not the parser or the parse tree?  If so, then I completely misunderstood what you are trying to accomplish.

documentation/OpenMP-4.5-grammar.txt
```diff
@@ -1,114 +1,396 @@
...
+2.5 parallel -> PARALLEL [parallel-clause[ [,] parallel-clause]...]
+    parallel-clause -> if-clause |
+                       num-threads-clause |
+                       default-clause |
+                       private-clause |
+                       firstprivate-clause |
+                       shared-clause |
+                       copyin-clause |
+                       reduction-clause |
+                       proc-bind-clause
```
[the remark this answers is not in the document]
ichoyjx: I didn't add all the `end directive` except they may accept clauses. On the second thought, it's better to add them for completion as you suggested. Thanks.

documentation/OpenMP-4.5-grammar.txt
```diff
@@ -1,114 +1,396 @@
-# Copyright (c) 2018, NVIDIA CORPORATION.  All rights reserved.
+# Copyright (c) 2018-2019, NVIDIA CORPORATION.  All rights reserved.
 # OpenMP 4.5 Specifications
 
-2 omp-directive -> sentinel directive-name [clause [,] clause...]
+2 omp-directive -> sentinel directive-name [clause[ [,] clause]...]
```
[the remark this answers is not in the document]
ichoyjx: `directive-name` is all the directives under `# directive-name` section.

documentation/OpenMP-4.5-grammar.txt
```diff
@@ -1,114 +1,396 @@
...
+       target-update-clause -> motion-clause |
+                               if-clause |
+                               device-clause |
+                               nowait-clause |
+                               depend-clause
+       motion-clause -> to-clause |
+                        from-clause
+
+2.10.6 declare-target -> DECLARE TARGET (extended-list)
+2.10.6 declare-target -> DECLARE TARGET [declare-target-clause[ [,] declare-target-clause]...]
```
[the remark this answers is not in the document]
ichoyjx: Okay.

documentation/OpenMP-4.5-grammar.txt
ichoyjx: you are right, i will delete them

documentation/OpenMP-4.5-grammar.txt
```diff
@@ -1,114 +1,396 @@
...
 2.13.5 taskgroup -> TASKGROUP
-2.13.6 atomic -> ATOMIC
-2.13.7 flush -> FLUSH
-2.13.8 ordered -> ORDERED
-2.14.1 cancel -> CANCEL
-2.14.2 cancellation-point -> CANCELLATION POINT
-2.15.2 threadprivate -> THREADPRIVATE
-2.16 declare-reduction -> DECLARE REDUCTION
+
+2.13.6 atomic -> ATOMIC [seq_cst[,]] READ | WRITE | UPDATE | CAPTURE [[,]seq_cst]
```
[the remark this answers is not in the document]
ichoyjx: Yes you're right, I will change that.

klausler:
Please resolve my last comment.  Is this new grammar supposed to specify what the parser parses and the parse tree represents?  If not, will there still be a grammar file that does specify the language accepted by the parser?

PR #471, opened 2019-05-28
Fix UTF-8 issues and add implement some character intrinsic functions folding

jeanPerier:
- Fix UTF-8 support in prescanner, semantic analysis and unparsing. Add small UTF-8 literal tests.

- Add support for character related intrinsic functions IACHAR, ICHAR, CHAR, ACHAR, ADJUSTL, ADJUSTR. Add related tests, included with multi-byte character literals.


lib/parser/prescan.cc
```diff
@@ -545,7 +545,7 @@ void Prescanner::QuotedCharacterLiteral(
   bool escape{false};
   bool escapesEnabled{features_.IsEnabled(LanguageFeature::BackslashEscapes)};
   while (true) {
-    char ch{*at_};
+    unsigned char ch{static_cast<unsigned char>(*at_)};
```
klausler: `char32_t` might be better since that's what `EmitQuotedChar`'s argument is declared as.  But did this change fix anything?

lib/evaluate/character.h
```diff
@@ -0,0 +1,102 @@
...
+
+namespace Fortran::evaluate {
+
+template<int KIND> class CharacterUtils {
+  using Character = Scalar<Type<TypeCategory::Character, KIND>>;
+  using CharT = typename Character::value_type;
+
+public:
+  static constexpr bool IsValidCharacterCode(std::uint64_t code) {
+    if constexpr (KIND == 1) {
```
sscalpone: Do we want to hard-code these kind values or move them to a config file?  Is the const-ness of these functions important?

jeanPerier: Regarding the constexpr-ness, it is not required by call-site, I did put it because I could and this follows our style guideline to do so when possible.
So, yes, I think it is possible here to move the mapping between the kind and the encoding into some kind of config file and have the semantic or folding context initialized with the mapping.

However, I think the semantic analysis phase is already making some assumptions regarding the encoding behind the character kinds. Also, making the size of the encoding configurable would be troublesome (1 is one byte, 2 is two byte and 4 is four). So people could only configure 2 to be UCS-2 or the EUC-JP variant or another 2-byte encoding.
 
By config file, do you mean a config file to be read by f18 before every compilation of a Fortran program or a config file that is used once during the compilation of f18 ?

lib/evaluate/character.h
```diff
@@ -0,0 +1,102 @@
...
+      return IsValidASCII(code);
+    } else if constexpr (KIND == 2) {
+      return IsValidInternalEUC_JP(code);
+    } else if constexpr (KIND == 4) {
+      return IsValidUCS4(code);
+    } else {
+      static_assert(true, "bad character kind");
```
tskeith: Should this be static_assert(false...)?

lib/evaluate/character.h
```diff
@@ -0,0 +1,102 @@
...
+    if (pos != Character::npos && pos != str.length() - 1) {
+      auto delta{str.length() - 1 - pos};
+      return Character{Character(delta, Space()) + str.substr(0, pos + 1)};
+    }
+    // else empty or only spaces, or no trailing spaces
+    return str;
+  }
+
+private:
+  static constexpr bool IsValidASCII(std::uint64_t code) { return code < 128; }
```
tskeith: This results in a warning if you use `ACHAR(129)`, for example. Why not accept any code < 256?

lib/evaluate/character.h
sscalpone: By config file, I just mean that all implementation-dependent constants ought to be collected in one place.  In a C++ source file is fine with me.

jeanPerier: Then `lib/evaluate/type.h` might be the good place to map the character KIND to encodings, since it already gathers this kind of choice for numerical types. Does that sounds right to you ?

sscalpone: OK w/ me.

lib/evaluate/character.h
jeanPerier: I went for simple ASCII because the standard was not asking for more.
Also, currently the parser/semantic analysis of character literal with ASCII >= 128 creates a `character(kind=1, len=2)`. So `len("ÿ")` (ascii 255) returns 2 and so does `len("<80>") ` (ascii 128).
Which means that `achar(255).EQ.("ÿ")` would be `false` today if it was allowed by achar.

I think it is probably safe to support extended ASCII as 7-bit systems are probably not the target today, but then it requires modification in the parser or character literal analysis.
@klausler and @tskeith, should I modify this ?

tskeith: Other compilers that I tried all get `len("ÿ")` to be 2 and `len(achar(255))` to be 1. That's true for PGI, gfortran, ifort, nag, and xlf. None of them warn about `achar(255)`.

I find it surprising that `"ÿ"` is two characters but I guess that the few current Fortran programmers that do things like might not.

schweitzpgi: 0xFF is not a valid codepoints in UTF-8.

Do the 3 and 4 byte encodings of UTF-8 also report as LEN=2?

Interesting however that LEN reports the storage in bytes rather than the symbolic length of the string for all these implementation. My cursory look at the standard just led to "character length", which seems ambiguous in a superficial glance.

klausler: `LEN` is length in characters.

`LEN(ACHAR())` is defined to be 1.

Quoted character literals of kind 1 can hold UTF-8 sequences or other non-ASCII source characters, but each byte is treated as a distinct 'character' in the literal.  This allows UTF-8 file names on OPEN statements to appear naturally as literals.  Character literals of kinds 2 and 4 undergo conversion from Kanji/UTF-8 to their 16/32-bit encodings.  The scanner actually cares about source encoding only when scanning Hollerith literals (the character count must be correctly interpreted).  We might have to support other source encodings in the future (e.g., Latin-1) and that will be easy so long as they're one-byte extensions of seven-bit ASCII.

Character variables (not quoted literals) of kind 1 can be abused to hold arbitrary byte sequences.  `ACHAR(I,kind=1)` should just take the least-order 8 bits of I; kinds 2 and 4 should take 16 and 32.  Or, in other words, `IACHAR(ACHAR())` and `ACHAR(IACHAR())` should be no-ops when the integer and character kind values are the same.  No UTF-8 semantics should apply to these.

lib/evaluate/character.h
```diff
@@ -0,0 +1,102 @@
...
+  static constexpr bool IsValidInternalEUC_JP(std::uint64_t code) {
+    std::uint16_t hi{static_cast<std::uint16_t>(code >> 8)};
+    std::uint16_t lo{static_cast<std::uint16_t>(code & 0xff)};
+    return IsValidASCII(code) ||
+        (code < 0xffff &&
+            ((0xa1 <= hi && hi <= 0Xfe && 0xa1 <= lo && lo <= 0Xfe) ||
```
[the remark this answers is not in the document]
jeanPerier: It was. Check for code validity removed following other comment.

lib/parser/prescan.cc
jeanPerier: Yes `char32_t` is more appropriate for `ch`, I have changed that.
The cast to `unsigned char` changed something (at least on the machine I was testing were `char` is signed). Take "你" for instance (U+20320), its UTF-8 encoding is` 0xE4 0xBD 0xA0`. Without the cast, the first byte `0XE4` ended up interpreted as `-28` as a `char`, that was casted to `4294967267` as a `char32_t` in `EmitQuotedChar` instead of `228`. As a result, `EmitQuotedChar` was emitting `"\367\277\277\244"` instead of `"\344"` and "你" was pre-scanned as `"\367\277\277\244\367\277\276\275\367\277\276\240"` instead of `"\344\275\240"`

lib/evaluate/character.h
jeanPerier: I tracked back from moving the KIND mapping in `lib/evaluate/type.h` following comments from Tim and Peter below. I removed checks for encoding validity so there is no need for any kind of mapping here anymore.

lib/evaluate/character.h
jeanPerier: Yes, it should have been thanks. Removed this part as part of other comment below.

lib/evaluate/character.h
jeanPerier: Thanks for your inputs. Checking more with other compilers behavior, and based on your comments it will bring more harm than good to do any checks related to the encoding here.

I have removed all validity checks in `char` and `achar` which now behave like pgfortran and xlf by simply keeping the lower bits fitting into the character underlying type.

I also changed a bit the signed/unsigned cast to ensure `ACHAR(IACHAR())` is a no-op when possible.

schweitzpgi: You might consider an optional warning for values that are not in the range specified in 16.9.3 para 5.

jeanPerier: I agree it would be nice to do so but I am not comfortable regarding how to emit optional warnings in f18 so I would rather not emit a warning at all using the fact that 16.9.3 also says "otherwise, the result is processor dependent" when the value is outside of the specified range. f18 processor dependent behavior in this case is similar to pgfortran and xlf. I could also check a little bit that pgfortran kanji `ncharacter` `nchar(x)` works as f18 `char(x, 2)`, though pgfortran cannot fold `nchar` and I still do not know make `ncharacter` literals .

All that to say, if nobody has any other objections, I will merge this PR as it is now.

PR #472, opened 2019-05-29
Roll up some fixes to problems found running more correctness tests

klausler:
- Implement commonly used intrinsics in constant expressions: SELECTED_INT_KIND, SELECTED_REAL_KIND, PRECISION, RANGE, RADIX
- Implement folding of ANY and ALL when applied to whole arrays
- Implement folding of structure components and subscripted structure components as applied to derived type parameters
- Make naming more consistent for utility functions that return pointers into their arguments (use "Unwrap" rather than "Get")
- Allow for use of alternative module file name suffixes
- Adjust the sequence of analysis in semantics for when we default undeclared entities to be objects (not procedures)


klausler:
These changes get the "not building" count in the particular test suite I'm running down from 682 to 597 (with semantics checking enabled).

klausler:
ping

lib/evaluate/fold.cc
```diff
@@ -215,13 +239,13 @@ static inline Expr<TR> FoldElementalIntrinsicHelper(FoldingContext &context,
             // same. Shouldn't this be checked elsewhere so that this is also
             // checked for non constexpr call to elemental intrinsics function?
             context.messages().Say(
-                "arguments in elemental intrinsic function are not conformable"_err_en_US);
+                "Arguments in elemental intrinsic function are not conformable"_err_en_US);
```
vdonaldson: I see that most compilers tend to start messages with caps, but our messages sometimes do and sometimes don't.  Do we need a suggestion somewhere to favor caps?  The new message at line 579 below may be an exception.

klausler: Tim's messages tend to start with caps and I've been converting others as I run across them.

lib/evaluate/call.cc
```diff
@@ -42,15 +42,15 @@ ActualArgument &ActualArgument::operator=(Expr<SomeType> &&expr) {
 }
 
 std::optional<DynamicType> ActualArgument::GetType() const {
-  if (const auto *expr{GetExpr()}) {
+  if (const auto *expr{UnwrapExpr()}) {
```
psteinfeld: I'd find this easier to understand if "auto" were replaced by the actual type.

klausler: You bet.

lib/evaluate/constant.cc
```diff
@@ -66,27 +74,37 @@ static ConstantSubscript SubscriptsToOffset(
...
+  return elements;
 }
 
 template<typename T>
 auto Constant<T>::At(const ConstantSubscripts &index) const -> Element {
   return Base::values_.at(SubscriptsToOffset(index, Base::shape_));
 }
 
+template<typename T>
+auto Constant<T>::Reshape(ConstantSubscripts &&dims) const -> Constant {
```
psteinfeld: Here's another case where I don't understand what type "auto" represents by looking at this file.

klausler: It's the function result type, which is a `Constant` of the same type as the class.  I could have written this using `Constant<T> Constant<T>::Reshape(...)` but I prefer the use of `auto` whenever an outlined member function's type is best defined in terms of type definitions in the class.

lib/evaluate/fold.cc
psteinfeld: I've also seen other kinds of inconsistencies.  It would be good to agree on message guidance.

lib/evaluate/constant.cc
psteinfeld: Thanks for the explanation.  Sounds good.

lib/evaluate/fold.cc
vdonaldson: I've changed some messages to start with caps.  Watch out for tests ..

lib/evaluate/type.cc
```diff
@@ -219,4 +220,76 @@ bool SomeKind<TypeCategory::Derived>::operator==(
...
+    if (rOK) {
+      return -4;
+    } else {
+      return -2;
+    }
+  } else {
+    if (rOK) {
+      return -1;
+    } else {
+      return -3;
```
schweitzpgi: What do the magic negative values in this function mean?

klausler: They are specified in the definition of the intrinsic.

PR #477, opened 2019-06-04
Roll up some minor fixes + tweak name resolution pass order

klausler:
Another batch of fixes to problems found in testing f18.

The nontrivial change is in name resolution; executable parts are now skimmed for calls to dummy procedures so that they can be known to be procedures, so that they don't get converted to objects by default.

Some standard modules have been partially populated so that tests depending on them can be run.  More work is needed to complete their contents, and to write additional modules.

The bit encodings of `.TRUE.` and `.FALSE.` in `LOGICAL` are now parameterized and their defaults have been set so that `LOGICAL(1)` uses `1` for `.TRUE.` so that kind 1 can serve as `C_BOOL` in standard module `ISO_C_BINDING`, where it is required by the standard to be the same as `_Bool`(`bool`) in C.

Now that object lifetimes are better understood, the various classes that refer to the sole instance of `parser::AllSources` now do so with a simple reference instead of dynamic allocation and reference counting.


lib/semantics/resolve-names.cc
```diff
@@ -4669,6 +4671,51 @@ bool ResolveNamesVisitor::Pre(const parser::ProgramUnit &x) {
...
+
+  SemanticsContext &context_;
+  Scope &scope_;
+};
+
+void ExecutionPartSkimmer::NoteCall(
+    Symbol::Flag flag, const parser::Call &call) {
+  auto &designator{std::get<parser::ProcedureDesignator>(call.t)};
+  if (const auto *name{std::get_if<parser::Name>(&designator.u)}) {
+    if (Symbol * symbol{scope_.FindSymbol(name->source)}) {
```
tskeith: `FindSymbol` could find the symbol in an outer subprogram if it's not in `scope_`. It could be that doesn't matter because it will never have `EntityDetails`. But it might be safer to use `scope_.find(name->source)` instead so that it looks only in `scope_`.

klausler: A host's dummy procedure can be called from a contained subprogram, and that might be the only call, so the skimming of the contained subprogram's executable part needs to be able to access its host's scope.

PR #478, opened 2019-06-04
Semantic checks for DO controls of type REAL

### LINKED ISSUE #458: Check type of do variable and bounds in DO loop
The grammar no longer constrains the do variable and bounds of a DO loop to be INTEGER so `EnforceTypeConstraint()` does not check the types of these.

We need to add a semantic check that looks at the `LoopBounds` of a `LoopControl` and checks the type of `name`, `lower`, `upper`, and `step`. For each, issue a warning if it has a REAL type and an error if it is neither INTEGER nor REAL.

Add an enumerator to `LanguageFeature` to control whether this extension is allowed or not.

psteinfeld:
These changes fix #458, to perform semantic checks on DO variables
and initial, final, and step expressions.  As a new extension, we want to allow
REAL and DOUBLE PRECISION values for them by default.  For expressions that
are not INTEGER, REAL, or DOUBLE PRECISION, we want to issue error messages.

 - There already existed infrastructure for semantic checking of DO loops that
   was partially specific to DO CONCURRENT loops.  Because I re-used some of
   this infrastructure, I renamed some files and classes from "concurrent" to
   "stmt".
 - I added some functions to distinguish among the different kinds of DO
   statements.
 - I added the functions to check-do-stmt.cc to produce the necessary warnings
   and errors.  Note that there are no tests for the warnings since the
   necessary testing infrastructure does not yet exist.
 - I changed test-errors.sh so that additional compilation options can be
   specified in the test source.
 - I added two new tests to test for the various kinds of values that can be
   used for the DO variables and control expressions.  The two tests are
   identical except for the use of different compilation options.
   dosemantics03.f90 specifies the options "-Mstandard -Werror" to produce
   error messages for the use of REAL and DOUBLE PRECISION DO variables and
   controls.  dosemantics04.f90 uses the default options and only produces
   error messages for constructs that are erroneous by default.


test/semantics/test_errors.sh
```diff
@@ -36,8 +36,12 @@ log=$temp/log
 actual=$temp/actual
 expect=$temp/expect
 diffs=$temp/diffs
+options=$temp/options
 
-cmd="$CMD $src"
+# See if there are additional options
+awk '/^ *!OPTIONS: / {gsub (/^! OPTIONS: /, " " ); print}' $src > $options
```
tskeith: We could be more flexible about allowing spaces and using sed makes this simpler. I suggest:
`sed -n 's/^ *! *OPTIONS: *//p' $src > $options`

test/semantics/dosemantics03.f90
```diff
@@ -0,0 +1,250 @@
...
+! are both in effect.  This test turns on the options for standards
+! conformance and turning warnings into errors.  This produces error
+! messages for the cases where REAL and DOUBLE PRECISION variables
+! and expressions are used in the DO controls.
+!
+! This test is just like dosemantics04.f90 but with the options
+! to produce error messages when using REAL and DOUBLE PRECISION DO
+! loop controls.
+
+! RUN: ${F18} -funparse-with-symbols %s 2>&1 | ${FileCheck} %s
```
tskeith: This doesn't do anything when you're using `test_errors.sh`.

lib/semantics/check-do-stmt.h
```diff
@@ -12,30 +12,21 @@
...
+#ifndef FORTRAN_SEMANTICS_CHECK_DO_STMT_H_
+#define FORTRAN_SEMANTICS_CHECK_DO_STMT_H_
 
 #include "semantics.h"
-#include "../common/indirection.h"
+#include "../parser/parse-tree.h"
 
-namespace Fortran::parser {
-struct DoConstruct;
-}
```
tskeith: Why replace the forward declaration with an include? In general it's good for `.h` files to include as little as possible (while still being self-contained). In this case `parse-tree.h` may be included wherever `check-do-stmt.h` is so it doesn't matter, but that's not obvious and it could change.

lib/parser/parse-tree.cc
```diff
@@ -80,9 +80,26 @@ Expr::Expr(Designator &&x)
 Expr::Expr(FunctionReference &&x)
   : u{common::Indirection<FunctionReference>::Make(std::move(x))} {}
 
+const std::optional<LoopControl> &DoConstruct::GetLoopControl() const {
+  NonLabelDoStmt const &doStmt{
```
tskeith: Everywhere else we put `const` before the type so we should stick with that style (or change it everywhere).

lib/semantics/check-do-stmt.cc
```diff
@@ -344,41 +372,108 @@ static CS GatherReferencesFromExpression(const parser::Expr &expression) {
...
+  }
+
+private:
+  using Bounds = parser::LoopControl::Bounds;
+
+  const Bounds &GetBounds(const parser::DoConstruct &doConstruct) {
     auto &doStmt{
         std::get<parser::Statement<parser::NonLabelDoStmt>>(doConstruct.t)};
     auto &optionalLoopControl{
         std::get<std::optional<parser::LoopControl>>(doStmt.statement.t)};
```
tskeith: Use `doConstruct.GetLoopControl()` here now that it is available?

test/semantics/test_errors.sh
psteinfeld: Thanks!  I'll do it.

test/semantics/dosemantics03.f90
psteinfeld: Oops, this is left over from a previous mistake.  I'll remove it, along with a couple of other cases of the same mistake.

lib/semantics/check-do-stmt.cc
```diff
@@ -344,41 +372,108 @@ static CS GatherReferencesFromExpression(const parser::Expr &expression) {
...
+  }
+
+  void CheckDoControl(parser::CharBlock sourceLocation, bool isReal) {
+    if ((isReal) && (!context_.warnOnNonstandardUsage())) {
+      // No messages for the default case
+    } else if (isReal && context_.warnOnNonstandardUsage() &&
+        (!context_.warningsAreErrors())) {
+      messages_.Say(sourceLocation,
+          "DO controls not of type INTEGER are a non-standard "
+          "extension"_en_US);
```
tskeith: The only other place we use `warnOnNonstandardUsage()` (in `expression.cc`) the message just says what is wrong, it doesn't say it is non-standard. We should consistently do it one way or the other.

klausler: They're not standard *any more*.  They're an obsolete feature, not an extension.

lib/semantics/check-do-stmt.cc
```diff
@@ -498,21 +595,15 @@ class DoConcurrentContext {
     CheckDefaultNoneImpliesExplicitLocality(localitySpecs);
   }
 
+  SemanticsContext &context_;
   parser::Messages &messages_;
```
tskeith: Now that this class has `context_` it probably doesn't need `messages_`. Replace `messages_.Say()` with `context_.Say()` and other occurrences of `messages_` with `context_.messages()`.

lib/semantics/check-do-stmt.cc
```diff
@@ -344,41 +372,108 @@ static CS GatherReferencesFromExpression(const parser::Expr &expression) {
...
+// Find a DO statement and enforce semantics checks on its body
+class DoStmtContext {
 public:
-  DoConcurrentContext(SemanticsContext &context)
-    : messages_{context.messages()} {}
+  DoStmtContext(SemanticsContext &context)
+    : context_{context}, messages_{context.messages()} {}
 
-  bool operator==(const DoConcurrentContext &x) const { return this == &x; }
+  bool operator==(const DoStmtContext &x) const { return this == &x; }
```
tskeith: I think the `operator==` was only needed because of the use of this class in `Indirection<>` so this is probably not necessary anymore.

lib/semantics/check-do-stmt.h
psteinfeld: I don't understand the underlying principle here.  The advantage of the include is that it tells me where DoConstruct is defined.

What's the advantage of making .h files include as little as possible?

tskeith: The advantage is potentially faster compilation. If a `.cc` file includes `check-do-concurrent.h` and not `parse-tree.h`, it will compile faster with the forward declaration because the compiler won't need to analyze `parse-tree.h`.

sscalpone:
If you use # with an issue number, you will get a hyperlink to the issue, like #458.

lib/parser/parse-tree.cc
psteinfeld: I'm surprised that this doesn't cause a syntax error.  Thanks for catching it.

lib/semantics/check-do-stmt.cc
psteinfeld: Yep.  Thanks for spotting this.

lib/semantics/check-do-stmt.cc
psteinfeld: Sounds good.  Thanks for pointing this out.

lib/semantics/check-do-stmt.cc
psteinfeld: Actually, there were a few classes that were using parser::Messages.  I've purged all of them from the file in favor of SemanticsContext.  Thanks for pointing out this simplification opportunity.

lib/semantics/check-do-stmt.cc
psteinfeld: Apparently so.  I was able to delete it with no ill effects.

lib/semantics/check-do-stmt.h
psteinfeld: Not a choice I would have made.  But I will now.

klausler:
If possible, please give this obsolete feature a name in the enumeration of language extensions, and enable/disable its support based on the set of active features.

psteinfeld:
> 
> 
> If possible, please give this obsolete feature a name in the enumeration of language extensions, and enable/disable its support based on the set of active features.

I added a name to the enumeration and I changed the code to make sure that the feature is enabled
and to check to see if we should issue a warning.  I did **not**, however, implement a new driver option for it, since we already have the -Mstandard option.  Also, none of the other compilers I looked at has a
separate option for REAL DO controls.

klausler:
You shouldn't have to check whether warnings are errors while creating the warnings.  There's a check elsewhere (don't remember where) that fails compilation if there's any fatal errors or if there's any messages at all and `-Werror` is in effect.

psteinfeld:
> 
> 
> You shouldn't have to check whether warnings are errors while creating the warnings. There's a check elsewhere (don't remember where) that fails compilation if there's any fatal errors or if there's any messages at all and `-Werror` is in effect.

In this instance, I wanted to decide whether to use the "en_US" or "err_en_US" suffix on the end of the message.  I don't know of any other way to handle this currently.

klausler:
> > You shouldn't have to check whether warnings are errors while creating the warnings. There's a check elsewhere (don't remember where) that fails compilation if there's any fatal errors or if there's any messages at all and `-Werror` is in effect.
> 
> In this instance, I wanted to decide whether to use the "en_US" or "err_en_US" suffix on the end of the message. I don't know of any other way to handle this currently.

If it's a warning unless `-Werror` is in effect, just use `_en_US`.

psteinfeld:
> 
> 
> > > You shouldn't have to check whether warnings are errors while creating the warnings. There's a check elsewhere (don't remember where) that fails compilation if there's any fatal errors or if there's any messages at all and `-Werror` is in effect.
> > 
> > 
> > In this instance, I wanted to decide whether to use the "en_US" or "err_en_US" suffix on the end of the message. I don't know of any other way to handle this currently.
> 
> If it's a warning unless `-Werror` is in effect, just use `_en_US`.

Yep.  I believe that's what the code currently does.  When implementing this, I assumed that, by default, we want to allow REAL DO controls.  Thus, in the current implementation, with no options specified, no warnings will be produced by REAL DO controls with no compiler options specified.  The perhaps surprising behavior happens when you specify only the -Werror option.  In that case, REAL DO controls are still allowed, and the compilation succeeds with no messages.

psteinfeld:
> 
> 
> From R1119, a _do-construct_ consists of a _do-stmt_, a _block_, and an _end-do_. Current check-do-concurrent.{h,cc} code checks the full _do-construct_ part of that variant, not just the _do-stmt_ part, and there are at least a couple of normal DO constraints like that as well. So 'check-do-stmt.{h,cc}' may be a little inaccurate. As long as file names are being updated, would something like 'check-do.[h,cc}' be more appropriate? If that argument is worth consideration, there may be other instances of this elsewhere, such as _DoStmtChecker_.

Thanks, Val.  This seems like a reasonable thing to do, especially since I'm just getting started on DO loop semantics.  There may turn out to be some situations where this distinction would make the code clearer as I add more semantic checking code.

lib/semantics/check-do-stmt.cc
```diff
@@ -344,41 +372,101 @@ static CS GatherReferencesFromExpression(const parser::Expr &expression) {
...
+private:
+  using Bounds = parser::LoopControl::Bounds;
+
+  const Bounds &GetBounds(const parser::DoConstruct &doConstruct) {
+    auto &loopControl{doConstruct.GetLoopControl().value()};
+    return std::get<Bounds>(loopControl.u);
+  }
+
+  void CheckDoControl(parser::CharBlock sourceLocation, bool isReal) {
+    bool warn = context_.warnOnNonstandardUsage() ||
```
klausler:
```
  if (isReal && context_.ShouldWarn(parser::LanguageFeature::RealDoControls)) {
    context_.say(sourceLocation, "DO controls should be INTEGER"_en_US);
  }
```

That's it.  You don't have to check for a disabled feature (it's not one of those few) and you don't have to check whether warnings should be errors -- they will be converted to errors automatically if that's the case.

psteinfeld: Got it.  I'll make the changes.

lib/semantics/check-do-stmt.cc
```diff
@@ -344,41 +372,101 @@ static CS GatherReferencesFromExpression(const parser::Expr &expression) {
...
+      // No messages for the default case
+    } else if (isReal && enabled && warn && !context_.warningsAreErrors()) {
+      context_.Say(sourceLocation, "DO controls should be INTEGER"_en_US);
+    } else {
+      context_.Say(sourceLocation, "DO controls should be INTEGER"_err_en_US);
+    }
+  }
+
+  void CheckDoVariable(const parser::ScalarName &scalarName) {
+    const DeclTypeSpec *symType{scalarName.thing.symbol->GetType()};
```
jeanPerier: You should check that you actually got a type (`symType !=nullptr`). I can segfault this PR with the following program:


```
subroutine foo()
end subroutine

program test
  ! segfaults semantic checks
  do foo = 1, 10, 3
    print *, "segfault semantic checks"
  end do
end program
```
This is because some symbols, such as subroutine, have no type.

Also, you should probably check that `scalarName.thing.symbol` is indeed a varaiable (and also not a procedure pointer). The following program goes through semantic when it probably should not:
```
function ibar()
end function

program test
  procedure(ibar), pointer :: ibar2 => NULL()

  ! Is accepted when it should not
  do ibar = 1, 10, 3
    print *, "something"
  end do
  ! Same
  Do ibar2 = 1, 10, 3
    print *, "something"
  end do
end program
```

lib/semantics/check-do-stmt.cc
```diff
@@ -344,41 +372,101 @@ static CS GatherReferencesFromExpression(const parser::Expr &expression) {
...
+    if (symType->IsNumeric(TypeCategory::Integer)) {
+      return;  // No warnings or errors for INTEGER
+    }
+    const parser::CharBlock &sourceLocation{scalarName.thing.source};
+    CheckDoControl(sourceLocation, symType->IsNumeric(TypeCategory::Real));
+  }
+
+  void CheckDoExpression(const parser::ScalarExpr &scalarExpression) {
+    const evaluate::Expr<evaluate::SomeType> *expr{GetExpr(scalarExpression)};
```
jeanPerier: You should also check that `expr != nullptr`, it may be null if the expression analysis failed. For instance, the following program segfault this PR:

```
program test
  integer i
  ! Expr may be invalid, like here (literal cannot be represented)
  do i = -2147483648_4, 10_4, 3_4
    print *, "segfault"
  end do
end program
```

lib/semantics/check-do-stmt.cc
jeanPerier: Also, regarding `scalarName.thing.symbol`, maybe you should check that it is not null, though I am not completely sure it can.
@klausler and @tskeith , should `GetUltimate()` be used here also in case the do-variable is coming from use association (I am not quite sure it is legal, but cannot find why not either)?

tskeith: Those are both good points. It may be that`symbol` is null if the do variable is not declared and IMPLICIT NONE is in effect. And I also cannot find any indication that the do variable may not be use-associated.

psteinfeld: Wow!  Thanks for catching this.  I'll add a fix and a test.

lib/semantics/check-do-stmt.cc
psteinfeld: Again, thanks for catching this.  You're a genius!  I'll add a fix and a test.

lib/semantics/check-do-stmt.cc
psteinfeld: Thanks to both of you for pointing this out.  @tskeith, I wrote a test, and you're correct that a
variable that's not declared causes a seg fault in my current implementation.

I'm not sure if I understand the references to use association and their implications for using
`GetUltimate()`.  I don't see anything in the standard that prohibits using a variable from a
`MODULE` as a DO variable.  I wrote the following test program, and everything seems to work
as expected:
`MODULE share
  INTEGER :: intvar
  REAL :: realvar
END MODULE share

PROGRAM use_assoc
  USE share
  DO intvar = 1, 10
    PRINT *, "hello"
  END DO
  DO realvar = 1, 10
    PRINT *, "hello"
  END DO
END PROGRAM use_assoc
`

lib/semantics/check-do-stmt.h
```diff
@@ -12,30 +12,24 @@
...
-extern template class Fortran::common::Indirection<
-    Fortran::semantics::DoConcurrentContext>;
 
 namespace Fortran::semantics {
-class DoConcurrentChecker : public virtual BaseChecker {
+
+class DoStmtChecker : public virtual BaseChecker {
 public:
-  explicit DoConcurrentChecker(SemanticsContext &);
```
vjayathirtha-nv: The definition for this and other DoConcurrentChecker class members also need to be removed or renamed right?

They exist in chec-do-concurrent.cc file.

vjayathirtha-nv: You can disregard this comment. The diff of check-do-concurrent.cc file was not showing before, I had to manually load the diff.

lib/semantics/check-do-stmt.cc
tskeith: As long as it works correctly it should be fine. Some tests would be a good idea.

It works in this case because `Symbol::GetType()` follows use associations to get the type from the ultimate symbol. To query some properties of use-associated symbols you have to call `GetUltimate()` on them and look at that symbol.

psteinfeld: Thanks, @tskeith .  I'll add a test.

psteinfeld:
I think I've addressed all previous comments.  Please let me know if you recommend more changes.

lib/semantics/check-do-stmt.cc
```diff
@@ -406,31 +410,50 @@ class DoStmtContext {
...
   void CheckDoVariable(const parser::ScalarName &scalarName) {
-    const DeclTypeSpec *symType{scalarName.thing.symbol->GetType()};
-    if (symType->IsNumeric(TypeCategory::Integer)) {
-      return;  // No warnings or errors for INTEGER
-    }
     const parser::CharBlock &sourceLocation{scalarName.thing.source};
-    CheckDoControl(sourceLocation, symType->IsNumeric(TypeCategory::Real));
+    const Symbol *symbol{scalarName.thing.symbol};
+    if (!symbol) {
+      SayBadDoControl(sourceLocation);
```
tskeith: Do you have a test that covers this case? I would expect that name resolution would have already produced an error about this name so you can just ignore it here.

lib/semantics/check-do-stmt.cc
```diff
@@ -406,31 +410,50 @@ class DoStmtContext {
...
     const evaluate::Expr<evaluate::SomeType> *expr{GetExpr(scalarExpression)};
-    if (ExprHasTypeCategory(*expr, TypeCategory::Integer)) {
-      return;  // No warnings or errors for INTEGER
-    }
     const parser::CharBlock &sourceLocation{
         scalarExpression.thing.value().source};
-    CheckDoControl(
-        sourceLocation, ExprHasTypeCategory(*expr, TypeCategory::Real));
+    if (!expr) {
+      SayBadDoControl(sourceLocation);
```
tskeith: Same thing here: expression analysis should have reported an error.

test/semantics/dosemantics03.f90
```diff
@@ -106,6 +123,52 @@ PROGRAM do_issue_458
...
+
+! Array DO variable
+!ERROR: Must be a scalar value, but is a rank-1 array
+  DO avar = 1, 10, 3
+    PRINT *, "plvar is: ", plvar
+  END DO
+
+! Undeclared DO variable
+!ERROR: No explicit type declared for 'undeclared'
+!ERROR: DO controls should be INTEGER
```
tskeith: See previous comments. The second error is superfluous.

lib/semantics/check-do-stmt.cc
psteinfeld: dosemantics03.f90 has a test for this.  The problem happens when "IMPLICIT NONE" is in effect, and
you use an undeclared variable as the DO variable.  The failing test starts at line 150.

lib/semantics/check-do-stmt.cc
psteinfeld: There are tests for this in dosemantics03.f90.  This situation happens when you have an invalid value for an expression.  In this case, other error messages also appear.  So I really didn't need to emit an additional message, but I **did** need to check for a null value for expr.  Otherwise, the
compiler seg faults.

I added tests for the initial, final, and step expressions at lines 222, 261, and 300.

Both of these cases were pointed out by Jean.

test/semantics/dosemantics02.f90
```diff
@@ -14,8 +14,6 @@
 
 ! C1121 -- any procedure referenced in a concurrent header must be pure
 
-! RUN: ${F18} -funparse-with-symbols %s 2>&1 | ${FileCheck} %s
```
vjayathirtha-nv: any reason this was removed?

test/semantics/dosemantics03.f90
psteinfeld: Thanks, Tim.  I just pushed a fix for this.

PR #480, opened 2019-06-05
Fix two bugs

klausler:
Don't require explicit shapes on objects in COMMON blocks when the objects are pointers or allocatables.

Perform semantics checks in ALLOCATABLE statements on the ultimate symbols found by USE association, if any.

These two bugs came up in the same regression test, and each affects several tests.

### DIFF
--- a/lib/semantics/resolve-names.cc
+++ b/lib/semantics/resolve-names.cc
@@ -3384,8 +3384,7 @@ void DeclarationVisitor::Post(const parser::CommonBlockObject &x) {
     return;  // error was reported
   }
   commonBlockInfo_.curr->get<CommonBlockDetails>().add_object(symbol);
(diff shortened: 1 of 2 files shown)


lib/semantics/resolve-names.cc
```diff
@@ -3384,7 +3384,8 @@ void DeclarationVisitor::Post(const parser::CommonBlockObject &x) {
     return;  // error was reported
   }
   commonBlockInfo_.curr->get<CommonBlockDetails>().add_object(symbol);
-  if (!IsExplicit(details->shape())) {
+  if (!symbol.attrs().HasAny({Attr::POINTER, Attr::ALLOCATABLE}) &&
```
tskeith: Could use `IsAllocatableOrPointer(symbol)` from `tools.h`.

PR #481, opened 2019-06-05
Add OpenBSD to get_ldd function.

ibara:
All the open source BSDs have ldd. But I've only ever tried with OpenBSD.


ibara:
Closing as this was fixed in #483.

PR #483, opened 2019-06-06
Fix issue #482 related to folding warnings

### LINKED ISSUE #482: folding02 test fails on OpenBSD
Hi --

Not sure if this is known or expected, but test 12, folding02, fails on OpenBSD. I ran test_folding.sh manually (with #481 applied) and got the following output:
```
 $ ./test_folding.sh folding02.f90                                                                                                              
Assuming no libpgmath support
folding test failed:

actual at 129: in a macro defined here
actual at 129: in a macro defined here
actual at 162: invalid argument on folding function with host runtime
actual at 164: invalid argument on folding function with host runtime
FAIL
```
f18 is indeed built without libpgmath (but I could build with if that matters).

Thanks!

jeanPerier:
**Fixes:**
- This PR fixes #482 by removing floating point status flag check in constant folding when compiling f18 with clang (justification below). Instead, the floating point results are checked for issues.

- It also removes the OS dependent check in `test_folding.sh` and uses cmake instead to convey if folding test should assumes `libpgmath` is used for constant folding or not. This is to avoid having to modify this script every-time f18 is build with a new OS.

**Detail regarding #482 (actually not linked to OpenBSD, but related to clang++ -02 and libc++):**

The issue was that intrinsic function folding with host run-time function was always checking `<cfenv>` floating point environment flags to emit warnings regarding invalid arguments, overflow, div by zero.... 
These flags are meaningless when `lib/evaluate` is compiled with clang++ because clang++ optimizations are not guaranteed to be consistent regarding the floating point environment.
For instance, when compiling f18 with clang++, libc++ (and no libpgmath support) in release mode on X86, a division by zero/invalid arg flag was signalling after valid calls to `std::comple<float>` `tan`and `tanh`. This is because clang vectorizes a division without masking the insignificant `xmm` slots that were set to zero (hence the div by zero hardware status flag).

### DIFF
--- a/test/evaluate/CMakeLists.txt
+++ b/test/evaluate/CMakeLists.txt
@@ -139,7 +139,7 @@ set(TEST_LIBPGMATH "-pgmath=false")
 if (LIBPGMATH_DIR)
   find_library(LIBPGMATH pgmath PATHS ${LIBPGMATH_DIR})
   if(LIBPGMATH)
-  set(TEST_LIBPGMATH "-pgmath=true")
+    set(TEST_LIBPGMATH "-pgmath=true")
   endif()
 endif()
 
--- a/test/evaluate/CMakeLists.txt
+++ b/test/evaluate/CMakeLists.txt
@@ -135,6 +135,15 @@ add_test(RESHAPE reshape-test)
 add_test(ISO-binding ISO-Fortran-binding-test)
 add_test(folding folding-test)
 
+set(TEST_LIBPGMATH "-pgmath=false")
+if (LIBPGMATH_DIR)
+  find_library(LIBPGMATH pgmath PATHS ${LIBPGMATH_DIR})
+  if(LIBPGMATH)
+  set(TEST_LIBPGMATH "-pgmath=true")
+  endif()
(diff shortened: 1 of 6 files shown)


test/evaluate/CMakeLists.txt
```diff
@@ -135,6 +135,15 @@ add_test(RESHAPE reshape-test)
 add_test(ISO-binding ISO-Fortran-binding-test)
 add_test(folding folding-test)
 
+set(TEST_LIBPGMATH "-pgmath=false")
+if (LIBPGMATH_DIR)
+  find_library(LIBPGMATH pgmath PATHS ${LIBPGMATH_DIR})
+  if(LIBPGMATH)
+  set(TEST_LIBPGMATH "-pgmath=true")
```
tskeith: This seems like a much better way to manage the tests.

This line should be indented two more spaces.

jeanPerier: Fixed, thanks. I wish we had a clang-format equivalent for `CMakeLists.txt` and  `.sh`

PR #484, opened 2019-06-06
Clean up subscript-sized int definitions

klausler:
After some discussion with Tim, define a couple of typedef aliases for std::int64_t, use them, and remove the means by which large array address calculations could be disabled.

### DIFF
--- a/lib/evaluate/fold.cc
+++ b/lib/evaluate/fold.cc
@@ -267,7 +267,7 @@ static inline Expr<TR> FoldElementalIntrinsicHelper(FoldingContext &context,
     }
     // Build and return constant result
     if constexpr (TR::category == TypeCategory::Character) {
-      auto len{static_cast<LengthCIntType>(
+      auto len{static_cast<ConstantSubscript>(
           results.size() ? results[0].length() : 0)};
       return Expr<TR>{Constant<TR>{len, std::move(results), std::move(shape)}};
     } else {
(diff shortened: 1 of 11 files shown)


lib/evaluate/fold.cc
```diff
@@ -966,8 +966,8 @@ static std::optional<Constant<SubscriptInteger>> GetConstantSubscript(
             }
           },
           [](Triplet &triplet) -> std::optional<Constant<SubscriptInteger>> {
-            std::optional<std::int64_t> lbi{1}, ubi;
-            std::optional<std::int64_t> stride{ToInt64(triplet.stride())};
+            std::optional<ConstantSubscript> lbi{1}, ubi;
+            std::optional<ConstantSubscript> stride{ToInt64(triplet.stride())};
```
tskeith: Using `ToInt64` here seems to break the abstraction. If `ConstantSubscript` were a wider type than `std::int64_t` we wouldn't get all of the bits.

Ideally we would want `std::optional<ConstantSubscript> ToConstantSubscript(...)`, as well as `ToLengthCIntType()` and `ToSubscriptCIntType()`, but that seems excessive.

klausler: Tell me what you want.

tskeith: What do you think of changing `ToInt64` to `std::optional<std::intmax_t> ToIntMax(...)`? Then we know whatever `SubscriptCIntType` is, we won't lose precision in this kind of initialization.

If you don't think that is worth the trouble, this is fine as it is.

klausler: Braced initialization would become errors as those 128-bit integer values were used to initialize `ConstantSubscript` values that were 64 bits.  So I'd have to put casts in, or an intervening conversion function call, or change `ConstantSubscript` to be 128 bits.

klausler:
Changing names to the preferred `ConstantSubscript` so that the same name can be used everywhere.

klausler:
I updated the change; please take another look whenever convenient.

tskeith:
> I updated the change; please take another look whenever convenient.

Looks good to me.

PR #488, opened 2019-06-07
More fixes from test suite shakedown

klausler:
Calls to intrinsic functions that resolved to what Fortran calls "specific" intrinsics were being overly strict in their argument checks.

Expressions & variables appearing as selectors in `associate` and `select type/rank` statements are now processed using tools from the evaluate library.  This dodges a segfault that affected all the tests for these constructs.  (They still don't build due to a still-missing intrinsic.)

The order of processing for name resolution has been tweaked again, along with the conversion of unknown entities to objects, so that `associate(x => internalFunction())` now works (this was the original impetus for the program tree based processing order).

### DIFF
--- a/documentation/Extensions.md
+++ b/documentation/Extensions.md
@@ -98,3 +98,10 @@ Extensions and legacy features deliberately not supported
 * Use of host FORMAT labels in internal subprograms (PGI-only feature)
 * ALLOCATE(TYPE(derived)::...) as variant of correct ALLOCATE(derived::...) (PGI only)
 * USE association of a procedure interface within that same procedure's definition
+* After "TYPE,EXTENDS(T1)::T2;...", the nonstandard structure constructor
+  T2(T1(x)) is accepted by PGI/GNU/Intel.  Use T2(T1=T1(x)) or T2(x) instead.
+* NULL() as a structure constructor expression for an ALLOCATABLE component (PGI).
+* Conversion of LOGICAL to INTEGER.
+* IF (integer expression) THEN ... END IF  (PGI/Intel)
+* Comparsion of LOGICAL with ==/.EQ. rather than .EQV. (also .NEQV.) (PGI/Intel)
+* Procedure pointers in COMMON blocks (PGI/Intel)
--- a/lib/evaluate/intrinsics.cc
+++ b/lib/evaluate/intrinsics.cc
@@ -1422,7 +1422,7 @@ std::optional<SpecificCall> IntrinsicProcTable::Implementation::Probe(
         bool ok{false};
         if (const auto &arg{specificCall->arguments[0]}) {
           if (const auto *expr{arg->UnwrapExpr()}) {
-            if (const Symbol * symbol{IsWholeSymbolDataRef(*expr)}) {
+            if (const Symbol * symbol{UnwrapWholeSymbolDataRef(*expr)}) {
               ok = symbol->attrs().test(semantics::Attr::OPTIONAL);
             }
           }
--- a/lib/evaluate/intrinsics.cc
+++ b/lib/evaluate/intrinsics.cc
@@ -641,6 +641,7 @@ static const IntrinsicInterface genericIntrinsicFunction[]{
 struct SpecificIntrinsicInterface : public IntrinsicInterface {
   const char *generic{nullptr};
   bool isRestrictedSpecific{false};
+  bool forceResultType{false};
 };
 
 static const SpecificIntrinsicInterface specificIntrinsicFunction[]{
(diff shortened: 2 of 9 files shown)


documentation/Extensions.md
```diff
@@ -98,3 +98,10 @@ Extensions and legacy features deliberately not supported
 * Use of host FORMAT labels in internal subprograms (PGI-only feature)
 * ALLOCATE(TYPE(derived)::...) as variant of correct ALLOCATE(derived::...) (PGI only)
 * USE association of a procedure interface within that same procedure's definition
+* After "TYPE,EXTENDS(T1)::T2;...", the nonstandard structure constructor
+  T2(T1(x)) is accepted by PGI/GNU/Intel.  Use T2(T1=T1(x)) or T2(x) instead.
```
schweitzpgi: Should this line also begin with `*`?

lib/evaluate/intrinsics.cc
```diff
@@ -1432,7 +1422,7 @@ std::optional<SpecificCall> IntrinsicProcTable::Implementation::Probe(
         bool ok{false};
```
schweitzpgi: Would it make sense to make the semantics checking of "PRESENT" a function? It seems like it may need to be checked in more than one place.

documentation/Extensions.md
klausler: The leading asterisk in markdown indicates a new bullet point, which would not look good in the middle of a sentence.

lib/evaluate/intrinsics.cc
klausler: How would that happen?

sscalpone:
Does the doc Semantics.md need updating to capture the change to order of processing?

klausler:
> Does the doc Semantics.md need updating to capture the change to order of processing?

No, it remains accurate as written at its level of detail.

lib/evaluate/tools.h
```diff
@@ -72,9 +72,16 @@ template<typename A> bool IsVariable(const FunctionRef<A> &funcRef) {
     return false;
   }
 }
-template<typename A> bool IsVariable(const Expr<A> &expr) {
+template<typename T> bool IsVariable(const Expr<T> &expr) {
   return std::visit([](const auto &x) { return IsVariable(x); }, expr.u);
 }
+template<typename A> bool IsVariable(const std::optional<A> &x) {
```
tskeith: Is there a convention you are following as to when to use `T` and when to use `A`?

lib/evaluate/tools.h
```diff
@@ -88,9 +95,16 @@ template<typename A> bool IsAssumedRank(const Designator<A> &designator) {
...
-template<typename A> bool IsAssumedRank(const Expr<A> &expr) {
+template<typename T> bool IsAssumedRank(const Expr<T> &expr) {
   return std::visit([](const auto &x) { return IsAssumedRank(x); }, expr.u);
 }
+template<typename A> bool IsAssumedRank(const std::optional<A> &x) {
+  if (x.has_value()) {
+    return IsAssumedRank(*x);
+  } else {
+    return false;
+  }
```
tskeith: This is just personal preference, but I would write this as `return x.has_value() && IsAssumedRank(*x)`.

lib/evaluate/tools.h
klausler: I use `A` when it can be any C++ type in the compiler, and `T` when it can be any Fortran type in the program.  So `optional<A>` but `Expr<T>`.  I'm probably not entirely consistent with these.

klausler: I think that I use `A` because `a` is idiomatic for an arbitrary type variable in Haskell.

lib/evaluate/tools.h
klausler: I would prefer that myself, but I've been trying to be more verbose for long-term readability; I don't really like using `if (x != nullptr)` over `if (x)` either, but maybe it helps somebody in the future understand the code better.

Why wouldn't you use `return x && IsAssumedRank(*x);`?

Or maybe `IsAssumedRank()` should handle `std::optional<>` arguments.  I'll try that.

tskeith: I guess in this case I don't find the more verbose form to be more readable.

I'm fine with `return x && IsAssumedRank(*x);` too.

tskeith:
Do you have a test case that triggered the segfault?

lib/evaluate/tools.h
klausler: Whatever you like.  I'll change it here and elsewhere.

klausler:
> Do you have a test case that triggered the segfault?

Yes, see slack.

tskeith:
Here's an example that passes with this fix and crashed before. Can we add it as a test somewhere?
```
subroutine s(a)
  implicit none
  integer :: a(2), v(2), j
  associate(i => a(v))
    j = i
  end associate
end subroutine
```

klausler:
> Here's an example that passes with this fix and crashed before. Can we add it as a test somewhere?
> 
> ```
> subroutine s(a)
>   implicit none
>   integer :: a(2), v(2), j
>   associate(i => a(v))
>     j = i
>   end associate
> end subroutine
> ```

As a general rule, I have not been duplicating test cases from existing regression test suites as I fix the bugs that they expose, since we'll be running those regression tests suites themselves soon.  Should this bug be an exception?

PR #489, opened 2019-06-07
Remove the FIR directory

schweitzpgi:
These changes remove the FIR directory and source files and clean up the f18 project so it no longer refers to the FIR library.  Removing this will improve build times.


klausler:
Is this code being moved, or about to be replaced, or just being thrown away, or what?

tools/f18/CMakeLists.txt
```diff
@@ -12,11 +12,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-if(LINK_WITH_FIR)
```
tskeith: `LINK_WITH_FIR` is in the top-level cmake file too, so that can be eliminated.

tools/f18/CMakeLists.txt
```diff
@@ -26,8 +21,6 @@ target_link_libraries(f18
   FortranParser
   FortranEvaluate
   FortranSemantics
-  ${FORTRAN_FIR_LIB}
-  ${LLVM_COMMON_LIBS}
```
tskeith: If `LLVM_COMMON_LIBS` is going to be needed for the FIR replacement I think it would be fine to leave it in, even though it's not currently necessary to be able to link.

tools/f18/CMakeLists.txt
schweitzpgi: I left it as past experience leads me to believe no one will want to build/use the new middle end for a while.

tools/f18/CMakeLists.txt
schweitzpgi: I don't think it serves any purpose to leave it in. LLVM is unused by the f18 program and not needed for linking it. 

The new middle end will explicitly require sets of MLIR and LLVM libraries in its cmake specs.

schweitzpgi:
> Is this code being moved, or about to be replaced, or just being thrown away, or what?

Wheeling the trash bins to the curb.

PR #490, opened 2019-06-07
Dodge a GCC-specific valgrind warning

klausler:
When f18 is built with gcc 8.3.0 or 9.1.0 (& possibly others), valgrind detects and warns that a conditional jump or move has been determined from uninitialized data in an instance of `RecoveryParser`, viz. the one built to emit an error and recover if a REDIMENSION statement appears.  I had thought that the problem was peculiar to gcc 9.1.0, but it's not -- I just usually do valgrind testing and performance measurement on f18 as built by clang, and had missed it with earlier versions of gcc.

I tracked it down to the use of a superfluous `>> ok` in the `statement()` parser; when the superfluous term is removed, the valgrind warning goes away.

So here's that change, plus some clean-up that accumulated along the way.

### DIFF
--- a/lib/parser/basic-parsers.h
+++ b/lib/parser/basic-parsers.h
@@ -49,7 +49,7 @@ namespace Fortran::parser {
 
 // fail<A>("..."_err_en_US) returns a parser that never succeeds.  It reports an
 // error message at the current position.  The result type is unused,
-// but might have to be specified at the point of call for satisfy
+// but might have to be specified at the point of call to satisfy
 // the type checker.  The state remains valid.
 template<typename A> class FailParser {
 public:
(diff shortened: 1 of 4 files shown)


lib/parser/basic-parsers.h
```diff
@@ -295,26 +295,24 @@ template<typename PA, typename... Ps> class AlternativesParser {
...
-    if constexpr (J <= sizeof...(Ps)) {
-      ParseState prevState{std::move(state)};
-      state = backtrack;
-      const auto &parser{std::get<J>(ps_)};
-      static_assert(std::is_same_v<resultType,
-          typename std::decay_t<decltype(parser)>::resultType>);
-      result = parser.Parse(state);
-      if (!result.has_value()) {
-        state.CombineFailedParses(std::move(prevState));
+  void ParseRest(std::optional<resultType> &result, ParseState &state, ParseState &backtrack) const {
```
tskeith: This probably needs formatting, judging from the length of this line.

Issue #491, opened 2019-06-11
Incorrect parse error in fixed form test


kiranchandramohan:
Section 6.3.3.2 of the Fortran standard says that the character '!' initiates a comment except in character context or in position 6.  Also if the first nonblank character on a line except in position 6 is '!' then the line is a comment line.
In the next Section, it is clarified that the character '!' in position 6 indicates a continuation.

Consider the program "strange" given below. The third line has '!' in position 6 indicating a continuation but since it also has another '!' in position 7 the rest of the line is a comment. The fourth line also has '!' in position 6 indicating continuation and hence the string there forms part of the print statement in line 2.

Flang and Gfortran accepts this program but F18 rejects this program with a parse error. Sample compilations are below.
```
$ cat ffce.f
      program strange
      print *,
     !!disappear
     !'appear'
      end
$ flang ffce.f
$ ./a.out
 appear
$ gfortran ffce.f
$ ./a.out
 appear
$ f18 ffce.f
ffce.f:2:8: error: expected execution part construct
        print *,
         ^
in the context: execution part
in the context: main program
f18: could not parse ffce.f
```

klausler:
It seems that f18 properly recognizes the line as a continuation, but then incorrectly treats the comment as extending to the end of the next continuation line.

klausler:
Should now be fixed; please confirm and close when convenient for you.  Thanks so much for the bug report!

kiranchandramohan:
Thanks @klausler for the quick fix. I can confirm that the issue is fixed.

PR #492, opened 2019-06-11
Fix bug #491 (and replace a misleading name)

klausler:
Deal with inline `!` comments in fixed-form source that appear in continuation lines.

Also rename `lineStart_` to `nextLine_` because it (now) always points to the beginning of the *next* line.  This is a clean-up I keep forgetting to do.

### DIFF
--- a/lib/parser/prescan.cc
+++ b/lib/parser/prescan.cc
@@ -304,16 +304,9 @@ void Prescanner::NextChar() {
     } else {
       mightNeedSpace = *at_ == '\n';
     }
-    while (*at_ == '\n' || *at_ == '&') {
-      bool continued{inFixedForm_ ? FixedFormContinuation(mightNeedSpace)
-                                  : FreeFormContinuation()};
-      if (continued) {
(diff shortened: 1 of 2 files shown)


lib/parser/prescan.cc
```diff
@@ -282,26 +282,38 @@ void Prescanner::SkipToEndOfLine() {
...
+      if (continued) {
+        if (MustSkipToEndOfLine()) {
+          SkipToEndOfLine();
         }
-      } else if (!FreeFormContinuation()) {
-        return;
+        mightNeedSpace = false;
+      } else {
+        break;
       }
```
tskeith: The logic *might* be clearer like this:
```
if (!continued) {
   break;
}
if (MustSkipToEndOfLine()) { ...
```

klausler: Please take another look and see whether clarity has been improved for you.

tskeith: > Please take another look and see whether clarity has been improved for you.

Yes, thank you.

PR #494, opened 2019-06-12
Resolve and check names in equivalence sets

tskeith:
Collect sets of `parser::EquivalenceObject` to process at the end of
the specification part. This is so that names mentioned in the
EQUIVALENCE statement don't trigger implicit declarations.

The `EquivalenceSets` class performs most of the numerous checks
on objects that can be in equivalence sets at all and objects that
can be in them together. It also merges sets when the same object
appears in more than one.

Once equivalence sets are checked they are added to the `Scope`.
Further checks will be necessary after the size and alignment of
variables are computed.

Add `FindUltimateComponent` to simplify checks on ultimate components
of derived types. Use it to implement `HasCoarrayUltimateComponent`
and checks on equivalence objects.

Make `ExpressionAnalyzer::Analyze(Designator)` public so that
`parser::EquivalenceObject` can be analyzed.

Add `GetDefaultKind`, `doublePrecisionKind`, and `quadPrecisionKind`
to `SemanticsContext` so that `defaultKinds_` does not need to be
accessed directly.

### DIFF
--- a/lib/semantics/expression.cc
+++ b/lib/semantics/expression.cc
@@ -459,16 +459,15 @@ MaybeExpr ExpressionAnalyzer::Analyze(const parser::RealLiteralConstant &x) {
   // letter used in an exponent part (e.g., the 'E' in "6.02214E+23")
   // should agree.  In the absence of an explicit kind parameter, any exponent
   // letter determines the kind.  Otherwise, defaults apply.
-  auto &defaults{context_.defaultKinds()};
-  int defaultKind{defaults.GetDefaultKind(TypeCategory::Real)};
+  int defaultKind{context_.GetDefaultKind(TypeCategory::Real)};
   const char *end{x.real.source.end()};
   std::optional<int> letterKind;
   for (const char *p{x.real.source.begin()}; p < end; ++p) {
     if (parser::IsLetter(*p)) {
       switch (*p) {
-      case 'e': letterKind = defaults.GetDefaultKind(TypeCategory::Real); break;
-      case 'd': letterKind = defaults.doublePrecisionKind(); break;
-      case 'q': letterKind = defaults.quadPrecisionKind(); break;
+      case 'e': letterKind = context_.GetDefaultKind(TypeCategory::Real); break;
+      case 'd': letterKind = context_.doublePrecisionKind(); break;
+      case 'q': letterKind = context_.quadPrecisionKind(); break;
       default: Say("Unknown exponent letter '%c'"_err_en_US, *p);
       }
       break;
@@ -1947,7 +1946,7 @@ Expr<SubscriptInteger> ExpressionAnalyzer::AnalyzeKindSelector(
 }
 
 int ExpressionAnalyzer::GetDefaultKind(common::TypeCategory category) {
-  return context_.defaultKinds().GetDefaultKind(category);
+  return context_.GetDefaultKind(category);
 }
 
 DynamicType ExpressionAnalyzer::GetDefaultKindOfType(
@@ -2012,7 +2011,7 @@ bool ExpressionAnalyzer::EnforceTypeConstraint(parser::CharBlock at,
             parser::ToUpperCaseLetters(type->AsFortran()));
         return false;
       } else if (defaultKind) {
-        int kind{context_.defaultKinds().GetDefaultKind(category)};
+        int kind{context_.GetDefaultKind(category)};
         if (type->kind() != kind) {
           Say(at, "Must have default kind(%d) of %s type, but is %s"_err_en_US,
               kind, parser::ToUpperCaseLetters(EnumToString(category)),
--- a/lib/semantics/expression.cc
+++ b/lib/semantics/expression.cc
@@ -459,15 +459,16 @@ MaybeExpr ExpressionAnalyzer::Analyze(const parser::RealLiteralConstant &x) {
   // letter used in an exponent part (e.g., the 'E' in "6.02214E+23")
   // should agree.  In the absence of an explicit kind parameter, any exponent
   // letter determines the kind.  Otherwise, defaults apply.
-  int defaultKind{context_.GetDefaultKind(TypeCategory::Real)};
+  auto &defaults{context_.defaultKinds()};
+  int defaultKind{defaults.GetDefaultKind(TypeCategory::Real)};
   const char *end{x.real.source.end()};
   std::optional<int> letterKind;
   for (const char *p{x.real.source.begin()}; p < end; ++p) {
     if (parser::IsLetter(*p)) {
       switch (*p) {
-      case 'e': letterKind = context_.GetDefaultKind(TypeCategory::Real); break;
-      case 'd': letterKind = context_.doublePrecisionKind(); break;
-      case 'q': letterKind = context_.quadPrecisionKind(); break;
+      case 'e': letterKind = defaults.GetDefaultKind(TypeCategory::Real); break;
+      case 'd': letterKind = defaults.doublePrecisionKind(); break;
+      case 'q': letterKind = defaults.quadPrecisionKind(); break;
       default: Say("Unknown exponent letter '%c'"_err_en_US, *p);
       }
       break;
--- a/lib/semantics/resolve-names-utils.cc
+++ b/lib/semantics/resolve-names-utils.cc
@@ -15,7 +15,9 @@
 #include "resolve-names-utils.h"
 #include "expression.h"
 #include "semantics.h"
+#include "tools.h"
 #include "../common/idioms.h"
+#include "../common/indirection.h"
 #include "../evaluate/fold.h"
 #include "../evaluate/tools.h"
 #include "../evaluate/type.h"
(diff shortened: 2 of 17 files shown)


lib/semantics/scope.cc
```diff
@@ -25,6 +25,29 @@ namespace Fortran::semantics {
...
+}
+
+std::string EquivalenceObject::AsFortran() const {
+  std::stringstream ss;
+  ss << symbol.name().ToString();
+  if (!subscripts.empty()) {
+    char sep{'('};
+    for (auto subscript : subscripts) {
+      ss << sep << subscript;
+      sep = ',';
```
jeanPerier: Should the separator be `:` for characters ?
Currently, `-fdebug-dump-symbols` dump of  is `equivalence (c(5:10), d)` where `c` and `d` are scalar characters is `Equivalence Sets: (c(5,10),d)` .

lib/semantics/scope.h
```diff
@@ -31,6 +31,22 @@ namespace Fortran::semantics {
...
+// and the constant lower bound of a substring or indices of an array element.
+struct EquivalenceObject {
+  EquivalenceObject(Symbol &symbol, std::vector<ConstantSubscript> subscripts)
+    : symbol{symbol}, subscripts{subscripts} {}
+  bool operator==(const EquivalenceObject &) const;
+  bool operator<(const EquivalenceObject &) const;
+  std::string AsFortran() const;
+
+  Symbol &symbol;
+  std::vector<ConstantSubscript> subscripts;  // for array elem or substring
```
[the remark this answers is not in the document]
tskeith: I had been assuming that was not valid, but I think you're right. I can't find anything that prohibits it. I will need to record the substring lower bound separately from the array subscripts.

lib/semantics/resolve-names-utils.cc
```diff
@@ -286,4 +288,356 @@ Bound ArraySpecAnalyzer::GetBound(const parser::SpecificationExpr &x) {
...
+      if (obj == *obj2) {
+        continue;  // already there
+      }
+      context_.Say(source,
+          "'%s' and '%s' cannot have the same first storage unit"_err_en_US,
+          obj2->AsFortran(), obj.AsFortran());
+    } else {
+      dst.push_back(obj);
+    }
+    objectToSet_[obj] = dstIndex;
```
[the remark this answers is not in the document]
tskeith: Yes

lib/semantics/scope.h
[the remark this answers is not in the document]
klausler: C8109 specifically mentions substrings; they're valid.

lib/semantics/expression.cc
```diff
@@ -459,16 +459,15 @@ MaybeExpr ExpressionAnalyzer::Analyze(const parser::RealLiteralConstant &x) {
   // letter used in an exponent part (e.g., the 'E' in "6.02214E+23")
   // should agree.  In the absence of an explicit kind parameter, any exponent
   // letter determines the kind.  Otherwise, defaults apply.
-  auto &defaults{context_.defaultKinds()};
-  int defaultKind{defaults.GetDefaultKind(TypeCategory::Real)};
+  int defaultKind{context_.GetDefaultKind(TypeCategory::Real)};
```
klausler: I don't think that hiding `context_.defaultKinds()` improves readability apart from saving a line of code, and it might actually make it harder to find the API for its class.

lib/semantics/resolve-names-utils.cc
```diff
@@ -286,4 +288,356 @@ Bound ArraySpecAnalyzer::GetBound(const parser::SpecificationExpr &x) {
...
+    }
+    auto subscripts{currObject_.subscripts};
+    if (subscripts.empty() && symbol.IsObjectArray()) {
+      // record a whole array as its first element
+      for (const ShapeSpec &spec : symbol.get<ObjectEntityDetails>().shape()) {
+        if (auto &lbound{spec.lbound().GetExplicit()}) {
+          if (auto subscript{evaluate::ToInt64(*lbound)}) {
+            subscripts.push_back(*subscript);
+          }
+        }
```
klausler: else error, unless you've checked elsewhere.

Alternatively, if you run that designator through expression semantics, you could build up a call to the `LBOUND` intrinsic around that expression, fold it, and use the elements of its constant array result.

lib/semantics/resolve-names-utils.cc
```diff
@@ -286,4 +288,356 @@ Bound ArraySpecAnalyzer::GetBound(const parser::SpecificationExpr &x) {
...
+      MergeInto(source, sets_[*it], dstIndex);
+    }
+  }
+  currSet_.clear();
+}
+
+// Report an error if sym1 and sym2 cannot be in the same equivalence set.
+bool EquivalenceSets::CheckCanEquivalence(
+    const parser::CharBlock &source, const Symbol &sym1, const Symbol &sym2) {
+  parser::MessageFixedText msg{"", 0};
```
klausler: Does default construction not work for `MessageFixedText`?  If not, it might be better to add a default constructor to that class and use it.

lib/semantics/resolve-names.cc
```diff
@@ -3398,6 +3400,40 @@ void DeclarationVisitor::Post(const parser::CommonBlockObject &x) {
...
+  }
+  return false;  // don't implicitly declare names yet
+}
+
+void DeclarationVisitor::CheckEquivalenceSets() {
+  EquivalenceSets equivSets{context()};
+  for (const auto *set : equivalenceSets_) {
+    const auto &source{set->front().v.value().source};
+    if (set->size() <= 1) {  // R871
+      Say(source, "Equivalence set must have more than one object"_err_en_US);
```
klausler: Could be a warning, it's benign to have a singleton.

lib/semantics/scope.h
```diff
@@ -122,6 +138,8 @@ class Scope {
     return symbols_.emplace(name, &symbol);
   }
 
+  const std::list<EquivalenceSet> &equivalenceSets() const;
```
klausler: Or `map<const Symbol *, EquivalenceSet *>`.

lib/semantics/scope.h
```diff
@@ -31,6 +31,22 @@ namespace Fortran::semantics {
...
+  EquivalenceObject(Symbol &symbol, std::vector<ConstantSubscript> subscripts)
+    : symbol{symbol}, subscripts{subscripts} {}
+  bool operator==(const EquivalenceObject &) const;
+  bool operator<(const EquivalenceObject &) const;
+  std::string AsFortran() const;
+
+  Symbol &symbol;
+  std::vector<ConstantSubscript> subscripts;  // for array elem or substring
+};
+using EquivalenceSet = std::vector<EquivalenceObject>;
```
klausler: Or `map<const Symbol *, EquivalenceObject*>`.

lib/semantics/tools.cc
```diff
@@ -382,4 +354,26 @@ const Symbol *HasEventOrLockPotentialComponent(
...
+  for (const auto &pair : *scope) {
+    const Symbol &component{*pair.second};
+    const DeclTypeSpec *type{component.GetType()};
+    if (!type) {
+      continue;
+    }
+    const DerivedTypeSpec *derived{type->AsDerived()};
+    bool isUltimate{IsAllocatableOrPointer(component) || !derived};
+    if (const Symbol *
+        result{!isUltimate ? FindUltimateComponent(*derived, predicate)
```
klausler: Consider reversing the predicate to avoid the negation, for readability.

lib/semantics/tools.cc
```diff
@@ -382,4 +354,26 @@ const Symbol *HasEventOrLockPotentialComponent(
   }
   return nullptr;
 }
+
+const Symbol *FindUltimateComponent(const DerivedTypeSpec &derivedTypeSpec,
```
klausler: Alternatively, each `DerivedTypeSpec` could have a list of ultimate components, and you could use `find_if()` on it.  (But I think that this traversal with a predicate is better, just wanted to raise the possibility in case you like it.)

lib/semantics/resolve-names-utils.cc
tskeith: It should have been checked already so I will change it to an assert.

lib/semantics/expression.cc
tskeith: OK

lib/semantics/resolve-names-utils.cc
tskeith: No, it doesn't have a default constructor. I think that using optional is a better way to do it anyway, so I'll change it to that.

lib/semantics/resolve-names.cc

lib/semantics/scope.h
tskeith: A symbol may be in more than one equivalence set.

klausler: Oh, I get it, you're not merging them as you go along.  Ok.

lib/semantics/tools.cc
tskeith: It is currently in the form `!a ? x : b ? y : z`.
Inverting the first condition makes it `a ? b ? y : z : x`.
I think the latter is less readable.

tskeith:
> Looks good. Will storage offsets and total group sizes be calculated later? How about interaction with COMMON (such as checking for the rule that an EQUIVALENCEd object in COMMON has to have a zero offset in its EQUIVALENCE group)?

Yes, this doesn't attempt to do anything with sizes and offsets.

PR #497, opened 2019-06-13
Use -fno-rtti when compiling f18.

sscalpone:
Our coding conventions prohibits rtti; enforce with the -fno-rtti compilation flag.


sscalpone:
I tested this on linux 86 with gcc 8.3 static and shared libraries and clang 7+ gcc8.3

CMakeLists.txt
```diff
@@ -91,6 +91,7 @@ if(CMAKE_COMPILER_IS_GNUCXX OR (CMAKE_CXX_COMPILER_ID MATCHES "Clang"))
    endif()
    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++17")
    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall -pedantic")
+   set (CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fno-rtti" )
```
tskeith: There shouldn't be a space after `set`.

tskeith:
It works fine on MacOS with gcc and clang.

PR #498, opened 2019-06-13
Clause 13 (Input/output editing) syntax and semantic checking

vdonaldson:
Diagnose:
 - all clause syntax errors (independent of normal parsing)
 - all clause constraints
 - use of non-standard extensions under -Mstandard
 - obvious program requirement violations

The code:
 - is invoked for both format statement and I/O statement constant formats
 - should be useable by the runtime I/O library for runtime formats
 - is able to recover from errors
 - is able to diagnose multiple errors in a single format
 - has accurate markers for errors
 - can process (fixed character size) Hollerith strings
 - generates reasonable error messages for known error scenarios
 - should not require C++ run-time library support

The code is templatized for use with fixed size character kind={1,2,4} input,
but only the kind=1 variant is actually exercised.


lib/semantics/check-io.cc
```diff
@@ -15,12 +15,62 @@
...
 namespace Fortran::semantics {
 
 // TODO: C1234, C1235 -- defined I/O constraints
 
+void IoChecker::Enter(
+    const parser::Statement<common::Indirection<parser::FormatStmt>> &stmt) {
+  if (!stmt.label.has_value()) {
+    context_.Say("Format statement must be labeled"_err_en_US);  // C1301
+  }
+  parser::CharBlock stmtCharBlock{stmt.source};
```
tskeith: Is there any need to make a copy of the CharBlock? Why not just use `stmt.source` below?

lib/semantics/check-io.cc
```diff
@@ -15,12 +15,62 @@
...
+  auto parenIndex{stmtCharBlock.ToString().find('(')};
```
tskeith: You can probably use `std::memchr` directly on the start and size of the CharBlock and avoid creating the `std::string` temporary. That would also avoid depending on the C++ runtime, if that is a goal.

lib/semantics/check-io.cc
```diff
@@ -15,12 +15,62 @@
...
+  parser::CharBlock formatCharBlock{parser::CharBlock(
+      stmtCharBlock.begin() + parenIndex, stmtCharBlock.size() - parenIndex)};
```
tskeith: No need for the second `parser::CharBlock` in this statement.

lib/semantics/check-io.cc
```diff
@@ -15,12 +15,62 @@
...
+  default: CRASH_NO_CASE;
+  }
+  for (int i{0}; i < errorCount; ++i) {
+    common::FormatError *error{formatErrors + i};
+    parser::MessageFormattedText msg{
+        parser::MessageFixedText(error->text_, strlen(error->text_)),
+        error->arg_};
+    parser::CharBlock errorCharBlock{parser::CharBlock(
+        formatCharBlock.begin() + error->offset_, error->length_)};
+    context_.Say(errorCharBlock, "%s"_err_en_US, msg.string());
```
tskeith: `Say(errorCharBlock, msg)` should be good enough if `isFatal` is set correctly in msg.

lib/semantics/check-io.cc
```diff
@@ -96,15 +146,73 @@ void IoChecker::Enter(const parser::FileUnitNumber &spec) {
...
+            const std::optional<std::string> constantFormat{
+                GetConstExpr<std::string>(format)};
+            if (!constantFormat) {
+              return;
+            }
+            // validate constant format -- 12.6.2.2
+            static constexpr int maxFormatErrors{3};
+            common::FormatError formatErrors[maxFormatErrors];
+            int errorCount{0};
+            switch (context_.GetDefaultKind(TypeCategory::Character)) {
```
tskeith: This looks a lot like the other place FormatValidator is called. Can the code be shared?

test/semantics/io07.f90
```diff
@@ -0,0 +1,96 @@
...
+
+4001 format(2(X))
+
+     !ERROR: List repeat specifier must be positive
+     !ERROR: DT descriptor repeat specifier must be positive
+4101 format(0(X), 0dt)
+
+6001 format(((I0, B0)))
+
+     !ERROR: A descriptor 'w' value must be positive
```
tskeith: It would be clearer if the edit descriptors were in quotes too. Also, some messages say "descriptor" and some say "edit descriptor". I think it would be better to use "edit descriptor" everywhere.

tskeith:
I don't think that `common` is the right place for code shared between the compiler and runtime. My understanding is that it is for utilities that are generally useful anywhere.

We should have a new directory for this shared code. If there are restrictions on it (e.g. about using the C++ runtime library) they should be documented there. It could depend on `common` if that is helpful and `common` meets the requirements of the runtime.

lib/common/format.h
```diff
@@ -0,0 +1,164 @@
...
+struct FormatError {
+  static constexpr int maxArgLength_{25};
+
+  const char *text_;  // error text; may have one %s argument
+  char arg_[maxArgLength_ + 1];  // optional %s argument value
+  int offset_;  // offset to error marker
+  int length_;  // length of error marker
+};
+
+template<typename CHAR = char> class FormatValidator {
```
tskeith: It would be good to document what this class does.

lib/common/format.h
```diff
@@ -0,0 +1,164 @@
...
+
+namespace Fortran::common {
+
+struct FormatError {
+  static constexpr int maxArgLength_{25};
+
+  const char *text_;  // error text; may have one %s argument
+  char arg_[maxArgLength_ + 1];  // optional %s argument value
+  int offset_;  // offset to error marker
+  int length_;  // length of error marker
```
tskeith: Public data members shouldn't have trailing underscores.

lib/common/format.cc
```diff
@@ -0,0 +1,616 @@
...
+
+template<typename CHAR> CHAR FormatValidator<CHAR>::NextChar() {
+  for (++cursor_; cursor_ < end_; ++cursor_) {
+    if (*cursor_ != ' ') {
+      return toupper(*cursor_);
+    }
+  }
+  cursor_ = end_;
+  return ' ';
+};
```
tskeith: This semicolon is not necessary.

lib/common/format.cc
```diff
@@ -0,0 +1,616 @@
...
+      if (!allowed && knrValue >= 0) {
+        AppendError("Repeat specifier before %s descriptor", knrToken, name);
+      } else if (knrValue == 0) {
+        AppendError(  // C1304
+            "%s descriptor repeat specifier must be positive", knrToken, name);
+      }
+    };
+
+    // Return the predicate "w value is present" to control further processing.
+    [[maybe_unused]] auto check_w = [&](bool required = true) -> bool {
```
tskeith: Why "maybe_unused"?

lib/common/format.h
```diff
@@ -0,0 +1,164 @@
...
+};
+
+template<typename CHAR = char> class FormatValidator {
+public:
+  explicit FormatValidator(const CHAR *format, size_t length, IoStmtKind stmt,
+      bool warnOnNonstandardUsage, FormatError *errors, int maxErrors)
+    : format_{format}, end_{format ? format + length : nullptr}, stmt_{stmt},
+      warnOnNonstandardUsage_{warnOnNonstandardUsage}, errors_{errors},
+      maxErrors_{maxErrors}, cursor_{format - 1} {}
```
tskeith: It seems like error reporting could be simplified by passing in a function to call to report errors.

lib/common/format.cc
```diff
@@ -0,0 +1,616 @@
...
+        NextToken();
+        return true;
+      }
+      if (required) {
+        AppendError("Expected %s descriptor 'w' value");  // C1306
+      }
+      return false;
+    };
+
+    auto check_m = [&]() -> void {
```
tskeith: Why are these lambdas rather than member functions? It obscures the fact that local variables are being accessed and even modified.

lib/common/format.h
```diff
@@ -0,0 +1,164 @@
...
+  char arg_[maxArgLength_ + 1];  // optional %s argument value
+  int offset_;  // offset to error marker
+  int length_;  // length of error marker
+};
+
+template<typename CHAR = char> class FormatValidator {
+public:
+  explicit FormatValidator(const CHAR *format, size_t length, IoStmtKind stmt,
+      bool warnOnNonstandardUsage, FormatError *errors, int maxErrors)
+    : format_{format}, end_{format ? format + length : nullptr}, stmt_{stmt},
```
klausler: The initialization of `end_` makes me think that you expect `format` to sometimes be a null pointer.  Can it be null, though?

lib/common/format.h
```diff
@@ -0,0 +1,164 @@
...
+    RParen,
+    Comma,
+    Point,
+    Sign,
+    UnsignedInteger,  // value in integerValue_
+    String,  // char-literal-constant or hollerith constant
+  };
+
+  struct Token {
+    explicit Token() : kind_{TokenKind::None}, offset_{0}, length_{1} {}
```
klausler: I'd put those initial values on the declarations of the data members themselves, and then remove the explicit constructor.

lib/common/format.cc
```diff
@@ -0,0 +1,616 @@
...
+    if (*laCursor_ != ' ') {
+      return toupper(*laCursor_);
+    }
+  }
+  laCursor_ = end_;
+  return ' ';
+};
+
+template<typename CHAR> void FormatValidator<CHAR>::Advance(TokenKind tk) {
+  cursor_ = laCursor_;
```
vjayathirtha-nv: From my understanding of this logic, for every `NextChar()` call there must be a call to `LookAheadChar()` before calling `Advanced()` in order for it to work correctly. It would be better to add this information next to the definition and/or declaration of Advance(), or use an alternate logic which doesn't require this.

lib/common/format.cc
```diff
@@ -0,0 +1,616 @@
...
+
+namespace Fortran::common {
+
+template<typename CHAR> CHAR FormatValidator<CHAR>::NextChar() {
+  for (++cursor_; cursor_ < end_; ++cursor_) {
+    if (*cursor_ != ' ') {
+      return toupper(*cursor_);
+    }
+  }
+  cursor_ = end_;
```
vjayathirtha-nv: The loop ends when `cursor_` is equal to `end_`, this is not necessary.

lib/common/format.cc
```diff
@@ -0,0 +1,616 @@
...
+  return ' ';
+};
+
+template<typename CHAR> CHAR FormatValidator<CHAR>::LookAheadChar() {
+  for (laCursor_ = cursor_ + 1; laCursor_ < end_; ++laCursor_) {
+    if (*laCursor_ != ' ') {
+      return toupper(*laCursor_);
+    }
+  }
+  laCursor_ = end_;
```
vjayathirtha-nv: same as `cursor_`

lib/common/format.cc
```diff
@@ -0,0 +1,616 @@
...
+  case '8':
+  case '9': {
+    int64_t lastValue;
+    const CHAR *lastCursor;
+    integerValue_ = 0;
+    bool overflow{false};
+    do {
+      lastValue = integerValue_;
+      lastCursor = cursor_;
+      integerValue_ = 10 * integerValue_ + c - '0';
```
vjayathirtha-nv: does `c - '0'` work the same for all `CHAR` types?

lib/common/format.h
vdonaldson: The possibility of a null format depends on caller code.  For
  write(*,fmt='')
format is currently a non-null pointer and *format is 0.  For
 100 format ()
format is a non-null pointer and *format is '('.  Allowing for format==nullptr is just extra caution for now, pending future runtime I/O library calls to this code.  If format is a nullptr, that would generate an 'Empty format expression' error, which is what happens with the write statement above.

lib/common/format.cc
vdonaldson: I've added a comment before the Advance definition.

lib/common/format.cc
vdonaldson: Done; thanks.

lib/common/format.cc
vdonaldson: Done; thanks.

lib/common/format.cc
vdonaldson: Yes, I  believe so.  I don't know how to exercise this code for kind=2/4 characters with an external program, but I have invoked it a few times with forced strings.  Those tests look good, including tests with integer values.

lib/common/format.cc
```diff
@@ -0,0 +1,616 @@
...
+  default:
+    if (cursor_ >= end_) {
+      suppressErrorCascade_ = false;
+      AppendError("Unterminated format expression");
+    }
+    token_.set_kind(TokenKind::None);
+    break;
+  }
+
+  token_.set_length(cursor_ - format_ - token_.offset() + 1);
```
vjayathirtha-nv: This overrides for instance line 206 

`token_.set_length(cursor_` - format_ - token_.offset());`

Is that expected behavior?

lib/common/format.cc
```diff
@@ -0,0 +1,616 @@
...
+    } else if (laChar == ')') {
+      AppendError("Unexpected ',' before ')' in format expression");
+    }
+    break;
+  }
+  case '\'':
+  case '"':
+    for (++cursor_; cursor_ < end_; ++cursor_) {
+      if (*cursor_ == c) {
+        if (cursor_ >= end_ || *(cursor_ + 1) != c) {
```
vjayathirtha-nv: `cursor_` can never be >= `end_`  at this point in the body of the loop.

vdonaldson: Yes.  I caught that one as well.

lib/common/format.cc
vdonaldson: Yes.  I rewrote this yesterday to provide better error markers, which is what this does.  The code is now a little different, but this pattern is the same.

lib/common/format.cc
tskeith: > `cursor_` can never be >= `end_` at this point in the body of the loop.

Maybe that should be `cursor_ == end_ -  1` because otherwise `*(cursor_ + 1)` accesses one past the end of the format in that case.

Also, `*(cursor_ + 1)` is usually written `cursor_[1]`.

lib/common/format.cc
vdonaldson: I responded too soon without actually testing this change.   For an unterminated format, cursor_ may be end_+1 after the loop, rather than end_.  This assignment avoids a bogus duplicate error in one of my new tests.  There are other ways to deal with this issue, but this looks like a good way to do that.  I have reinstated the assignments in these two functions.

vjayathirtha-nv: Ah, I see. So `NextChar` could be called with `cursor_` == `end_` and in that case this assignment is required. Could you leave a comment about this?

lib/common/format.cc
vdonaldson: I will change to
        if (cursor_ + 1 < end_ && *(cursor_ + 1) != c) {
I prefer *(cursor_ + 1) as a companion to *cursor_ rather than cursor_[1], all the more so with the addition of the left side of the conjunction.

klausler: `if (auto *nc{cursor_+1}; nc < end_ && *nc != c) { ... `

lib/common/format.cc
vdonaldson: Ok.  I put in a comment, then took it out, since the test catches this case.  I'll put it back.

lib/common/format.cc
vdonaldson: That's a mix of elegance and .. je ne sais pas.  I'll go with it.

lib/common/format.cc
vdonaldson: The intent was to do what a (void) cast on a call does when running lint - indicate that the result is not always used.  But it looks like it is instead designed for functions that are not called at all.  I'll remove it.

lib/semantics/check-io.cc
vdonaldson: It may be premature to consider doing this until more is known about handling multiple character kinds.  I've added a note about that in the TODO comments.

lib/common/format.h
vdonaldson: The administrative code for caller processing of errors is minimal.  As the code is now, I don't see much benefit to extracting single use functionality into a separate function.  If that changes as the code evolves, this could be reconsidered.

lib/semantics/check-io.cc
vdonaldson: Done; thanks.

lib/semantics/check-io.cc
vdonaldson: I've updated this code.

lib/semantics/check-io.cc
vdonaldson: Done; thanks.

lib/semantics/check-io.cc
vdonaldson: Done; thanks.

test/semantics/io07.f90
vdonaldson: Done; thanks.  Doing this suggested other changes as well.

lib/common/format.h
vdonaldson: Done.

lib/common/format.h
vdonaldson: Done.

lib/common/format.cc
vdonaldson: Done.

lib/common/format.cc
vdonaldson: There are trade-offs in doing this, but fewer drawbacks than I expected.  Done.

lib/common/format.h
vdonaldson: Ok; done.

vdonaldson:
I don't see an associated place to respond to Tim's comment that some new directory might be more appropriate for format checking code than lib/common, so I'll do that here.  The issue is that the code is intended for use both in the compiler and at runtime.  My recollection is that Peter originally suggested considering either lib/common or runtime, and of those two I prefer lib/common.  Given that the compiler calls come from lib/semantics code, and constraints are being checked, I think that lib/semantics would be a better place than either of those.  I'm insufficiently familiar with the global organization of the project to understand why a runtime library make file couldn't reach over and borrow source code from a lib subdirectory.  Since I don't have this background, I prefer deferring this choice, perhaps until either the runtime library code needs access, and/or some other code is identified that needs to be shared.

lib/common/format.h
tskeith: > The administrative code for caller processing of errors is minimal.

The caller has to create a fixed-size array and create a validator where 3 of the 6 arguments are related to error processing. Then it has to loop through the errors and do something with them. That doesn't seem minimal to me.

> As the code is now, I don't see much benefit to extracting single use functionality into a separate function. If that changes as the code evolves, this could be reconsidered.

I don't understand what single use functionality you are referring to.

vdonaldson: The code has three parts - memory allocation, the Check call, and error reporting.  I believe memory allocation needs to be in the caller.  These two instances of that code in each of the Enter functions can probably be the same, but that is 3 (could be less) lines of code.   The Check calls and reporting code are different in the two Enter instances, and the kind=2/4 variants are just placeholders for currently unknown code.  When the eventual code is better understood, it may or may not make sense to put some of this code into a common function.

lib/common/format.cc
```diff
@@ -0,0 +1,620 @@
...
+    bool overflow{false};
+    do {
+      lastValue = integerValue_;
+      lastCursor = cursor_;
+      integerValue_ = 10 * integerValue_ + c - '0';
+      if (lastValue > integerValue_) {
+        overflow = true;
+      }
+      c = NextChar();
+    } while (isdigit(c));
```
schweitzpgi: Is this `std::isdigit`?

vdonaldson: Are there multiple instances of this?  If yes, how do I know which one this is?  Do I need an std:: prefix?

lib/common/format.cc
```diff
@@ -0,0 +1,620 @@
...
+    for (++cursor_; cursor_ < end_; ++cursor_) {
+      if (*cursor_ == c) {
+        if (auto nc{cursor_ + 1}; nc < end_ && *nc != c) {
+          token_.set_kind(TokenKind::String);
+          break;
+        }
+        ++cursor_;
+      }
+    }
+    token_.set_length(cursor_ - format_ - token_.offset() + (cursor_ < end_));
```
schweitzpgi: Adding a bool to an integer. Might want to consider a ternary (?1:0) to make the intent explicit.

vdonaldson: Ok:

 token_.set_length(
        cursor_ - format_ - token_.offset() + (cursor_ < end_ ? 1 : 0));

klausler: The language explicitly defines `true` as being 1 when cast to an integral type.  It's probably less ugly without the redundant reimplementation of the conversion, IMO.

sscalpone: I'd add a new statement with an 'if' that adds one.  It's clear and not ugly.  The boolean-as-integer expr inline is hard to see.

vdonaldson: Ok.  My mild preference is for the original.  I don't mind the ternary op too much, and I understand the point of the suggestion.  I'll switch back.

vdonaldson: Of these suggestions, my preference is for the original.  If I get two votes for something else, or another suggestion that looks better, I'll go with it.

lib/common/format.h
tskeith: I'm suggesting an interface to `FormatValidator` like this:
```
    common::FormatValidator<char> validator{formatCharBlock.begin(),
        formatCharBlock.size(), IoStmtKind::None,
        [&](int offset, int length, const char *text) {
          // report error
        }};
    validator.Check();
```

lib/common/format.h
```diff
@@ -0,0 +1,181 @@
...
+  void check_r(bool allowed = true);
+  bool check_w(bool required = true);
+  void check_m();
+  bool check_d(bool required = true);
+  void check_e();
+
+  const CHAR *const format_;  // format text
+  const CHAR *const end_;  // one-past-last of format_ text
+  IoStmtKind stmt_;
+  bool warnOnNonstandardUsage_;
```
tskeith: How can it warn on non-standard usage when there is no way to report warnings?

vdonaldson: Like this (from test io10.f90):

$ cat a.f90
  write(*, '(EN,$)')
end

$ f18 -fparse-only -fdebug-semantics a.f90
$ f18 -fparse-only -fdebug-semantics -Mstandard a.f9
a.f90:1:16: error: Expected 'EN' edit descriptor 'w' value
    write(*, '(EN,$)')
                 ^
a.f90:1:17: error: Non-standard '$' edit descriptor
    write(*, '(EN,$)')
                  ^

tskeith: That's an error, not a warning.

vdonaldson: It's now a warning; thanks.

lib/common/format.cc
schweitzpgi: Discussed offline.

vdonaldson: I'll replace isdigit(c) with c >= '0' && c <= '9'

lib/semantics/check-io.cc
```diff
@@ -15,12 +15,62 @@
...
+#include "../common/format.h"
 #include "../parser/tools.h"
 
 namespace Fortran::semantics {
 
 // TODO: C1234, C1235 -- defined I/O constraints
 
+void IoChecker::Enter(
+    const parser::Statement<common::Indirection<parser::FormatStmt>> &stmt) {
+  if (!stmt.label.has_value()) {
```
sscalpone: Does resolve-labels check that format statements are labeled?  If not, should this check be moved there?

vdonaldson: No, resolve-labels does not currently check that a format statement has a label.

I can move the subset of this function that does this constraint check to a Pre function in resolve-labels.cc, but doing that seems to silently suppress the call to this new function, suppressing the rest of the checks in this new function.  This behavior may be related to the comment near the beginning of semantics.cc -- _No two checkers may have the same Enter or Leave function._

klausler: An unlabeled `FORMAT` statement is irrelevant to the program and perhaps shouldn't even be validated.

Ditto for unreferenced labeled `FORMAT`, although we can't figure that out completely yet until we can handle `ASSIGN`.

lib/common/format.h
vdonaldson: There is now a FormatErrorReporter class derived from your suggestions (thanks!) that does at least some of this.

lib/semantics/check-io.cc
tskeith: > I can move the subset of this function that does this constraint check to a Pre function in resolve-labels.cc, but doing that seems to silently suppress the call to this new function, suppressing the rest of the checks in this new function.

If `ValidateLabels()` detects an error, semantic checking stops. This is because phases after that depend on label DO statements being rewritten to DO constructs. For that reason it would be better to leave this check outside of `resolve-labels.cc`.

lib/common/format.h
vdonaldson: It looks like it is ok to do address arithmetic on a nullptr, as long as the result is not dereferenced.  I have therefore changed the constructor to check that format is not null.

PR #504, opened 2019-06-19
Expressions in a DO CONCURRENT header should not reference names declared in the header

psteinfeld:
C1123 requires that expressions in a DO  CONCURRENT header not reference names declared in the
 same header.  To implement this, I changed the "Pre" function for ConcurrentHeader to walk the
subtree in an order that caused the variables associated with the declared names to be created before
any of the expressions were walked.

I added the test dosemantics04.f90 to test this.

### DIFF
--- a/lib/semantics/resolve-names.cc
+++ b/lib/semantics/resolve-names.cc
@@ -875,7 +875,6 @@ private:
 class ConstructVisitor : public DeclarationVisitor {
 public:
   bool Pre(const parser::ConcurrentHeader &);
-  void Post(const parser::ConcurrentHeader &);
   bool Pre(const parser::LocalitySpec::Local &);
   bool Pre(const parser::LocalitySpec::LocalInit &);
   bool Pre(const parser::LocalitySpec::Shared &);
@@ -885,7 +884,6 @@ public:
   bool Pre(const parser::DataStmtObject &);
   bool Pre(const parser::DoConstruct &);
   void Post(const parser::DoConstruct &);
-  void Post(const parser::ConcurrentControl &);
   bool Pre(const parser::ForallConstruct &);
   void Post(const parser::ForallConstruct &);
   bool Pre(const parser::ForallStmt &);
@@ -970,6 +968,8 @@ private:
   void SetTypeFromAssociation(Symbol &);
   void SetAttrsFromAssociation(Symbol &);
   Selector ResolveSelector(const parser::Selector &);
+  void ResolveControlExpression(const parser::ConcurrentControl &control);
+  void ResolveIndexName(const parser::Name &name);
 };
 
 // Walk the parse tree and resolve names to symbols.
@@ -3886,12 +3886,78 @@ ParamValue DeclarationVisitor::GetParamValue(const parser::TypeParamValue &x) {
 
 // ConstructVisitor implementation
 
-bool ConstructVisitor::Pre(const parser::ConcurrentHeader &) {
-  BeginDeclTypeSpec();
-  return true;
+void ConstructVisitor::ResolveIndexName(const parser::Name &name) {
+  auto *prev{FindSymbol(name)};
+  if (prev) {
+    if (prev->owner().kind() == Scope::Kind::Forall ||
+        prev->owner() == currScope()) {
+      SayAlreadyDeclared(name, *prev);
+      return;
+    }
+    name.symbol = nullptr;
+  }
+  auto &symbol{DeclareObjectEntity(name, {})};
+
+  if (symbol.GetType()) {
+    // type came from explicit type-spec
+  } else if (!prev) {
+    ApplyImplicitRules(symbol);
+  } else if (!prev->has<ObjectEntityDetails>() && !prev->has<EntityDetails>()) {
+    Say2(name, "Index name '%s' conflicts with existing identifier"_err_en_US,
+        *prev, "Previous declaration of '%s'"_en_US);
+    return;
+  } else {
+    if (auto *type{prev->GetType()}) {
+      symbol.SetType(*type);
+    }
+    if (prev->IsObjectArray()) {
+      SayWithDecl(name, *prev, "Index variable '%s' is not scalar"_err_en_US);
+      return;
+    }
+  }
+  EvaluateExpr(parser::Scalar{parser::Integer{common::Clone(name)}});
 }
-void ConstructVisitor::Post(const parser::ConcurrentHeader &) {
+
+void ConstructVisitor::ResolveControlExpression(
+    const parser::ConcurrentControl &control) {
+  Walk(std::get<1>(control.t));
+  Walk(std::get<2>(control.t));
+  const auto &optionalStep{std::get<3>(control.t)};
+  if (optionalStep.has_value()) {
+    Walk(optionalStep.value());
+  }
+}
+
+// We need to make sure that all of the index-names get declared before the
+// expressions in the loop control are evaluated so that references to the
+// index-names in the expressions are correctly detected.
+bool ConstructVisitor::Pre(const parser::ConcurrentHeader &header) {
+  BeginDeclTypeSpec();
+
+  // Process the type spec, if present
+  auto &typeSpec{std::get<std::optional<parser::IntegerTypeSpec>>(header.t)};
+  if (typeSpec.has_value()) {
+    SetDeclTypeSpec(MakeNumericType(TypeCategory::Integer, typeSpec->v));
+  }
+
+  // Process the index-name nodes in the ConcurrentControl nodes
+  auto &controls{std::get<std::list<parser::ConcurrentControl>>(header.t)};
+  for (auto &control : controls) {
+    ResolveIndexName(std::get<parser::Name>(control.t));
+  }
+
+  // Process the expressions in ConcurrentControls
+  for (auto &control : controls) {
+    ResolveControlExpression(control);
+  }
+
+  // Resolve the names in the scalar-mask-expr, if present
+  auto &maskExpr{std::get<std::optional<parser::ScalarLogicalExpr>>(header.t)};
+  if (maskExpr.has_value()) {
+    Walk(maskExpr.value());
+  }
   EndDeclTypeSpec();
+  return false;
 }
 
 bool ConstructVisitor::Pre(const parser::LocalitySpec::Local &x) {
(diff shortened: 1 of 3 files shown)


lib/semantics/resolve-names.cc
```diff
@@ -3886,12 +3886,78 @@ ParamValue DeclarationVisitor::GetParamValue(const parser::TypeParamValue &x) {
...
-void ConstructVisitor::Post(const parser::ConcurrentHeader &) {
+
+void ConstructVisitor::ResolveControlExpression(
+    const parser::ConcurrentControl &control) {
+  Walk(std::get<1>(control.t));
+  Walk(std::get<2>(control.t));
+  const auto &optionalStep{std::get<3>(control.t)};
+  if (optionalStep.has_value()) {
+    Walk(optionalStep.value());
+  }
```
tskeith: `Walk(std::get<3>(control.t));` works. There is a `Walk` overload that handles optional.

lib/semantics/resolve-names.cc
```diff
@@ -3886,12 +3886,78 @@ ParamValue DeclarationVisitor::GetParamValue(const parser::TypeParamValue &x) {
...
+}
+
+// We need to make sure that all of the index-names get declared before the
+// expressions in the loop control are evaluated so that references to the
+// index-names in the expressions are correctly detected.
+bool ConstructVisitor::Pre(const parser::ConcurrentHeader &header) {
+  BeginDeclTypeSpec();
+
+  // Process the type spec, if present
+  auto &typeSpec{std::get<std::optional<parser::IntegerTypeSpec>>(header.t)};
```
tskeith: When the deduced type is const, it is a good practice to write these as `const auto &`...

test/semantics/dosemantics04.f90
```diff
@@ -0,0 +1,46 @@
...
+! See the License for the specific language governing permissions and
+! limitations under the License.
+
+! Issue 458 -- semantic checks for a normal DO loop.  The DO variable
+! and the initial, final, and step expressions must be INTEGER if the
+! options for standard conformance and turning warnings into errors
+! are both in effect.  This test turns on the options for standards
+! conformance and turning warnings into errors.  This produces error
+! messages for the cases where REAL and DOUBLE PRECISION variables
+! and expressions are used in the DO controls.
```
tskeith: I don't think this comment belongs here.

test/semantics/dosemantics04.f90
```diff
@@ -0,0 +1,46 @@
...
+  DO CONCURRENT (INTEGER *2 :: i = 1:10, i < j + n) LOCAL(n)
+    PRINT *, "hello"
+  END DO
+
+  DO 30 CONCURRENT (i = 1:n:1, j=1:n:2, k=1:n:3, a<3) LOCAL (a)
+    PRINT *, "hello"
+30 END DO
+
+!ERROR: concurrent-control expression references index-name
+  DO CONCURRENT (i = j:3, j=1:3)
```
tskeith: This isn't a comment for this pull request, but it would be good if the error message could be improved to indicate what the expression is and what the name is.

lib/semantics/resolve-names.cc
```diff
@@ -3886,12 +3886,78 @@ ParamValue DeclarationVisitor::GetParamValue(const parser::TypeParamValue &x) {
...
+      SayWithDecl(name, *prev, "Index variable '%s' is not scalar"_err_en_US);
+      return;
+    }
+  }
+  EvaluateExpr(parser::Scalar{parser::Integer{common::Clone(name)}});
 }
-void ConstructVisitor::Post(const parser::ConcurrentHeader &) {
+
+void ConstructVisitor::ResolveControlExpression(
+    const parser::ConcurrentControl &control) {
```
tskeith: `ResolveControlExpressions` (plural) might be a better name since it is doing all of them.

lib/semantics/resolve-names.cc
psteinfeld: Thanks.  I'll do that.

lib/semantics/resolve-names.cc
psteinfeld: Cool.  I'll make that change.

You'd think that there would be a way to iterate from 1 to three calling ```Walk(std::get<i>(control.t))```.  But I can't figure out how to do it.

test/semantics/dosemantics04.f90
psteinfeld: Ay yi yi.  God save me from copy/paste errors.  Or, more accurately, thanks for catching this.

Tim saved me again from a copy/paste error.

lib/semantics/resolve-names.cc
psteinfeld: Sounds good.  I'll make that change.  I'll also comment the calls to ```Walk(...)``` to clarify what's being walked.

lib/semantics/resolve-names.cc
```diff
@@ -3886,12 +3886,78 @@ ParamValue DeclarationVisitor::GetParamValue(const parser::TypeParamValue &x) {
...
+  }
+
+  // Process the expressions in ConcurrentControls
+  for (auto &control : controls) {
+    ResolveControlExpression(control);
+  }
+
+  // Resolve the names in the scalar-mask-expr, if present
+  auto &maskExpr{std::get<std::optional<parser::ScalarLogicalExpr>>(header.t)};
+  if (maskExpr.has_value()) {
```
vjayathirtha-nv: I suppose this check is also not needed

psteinfeld: Wow!  Great catch, @vjayathirtha-nv .  You're a genius!

I'll fix that.

lib/semantics/resolve-names.cc
```diff
@@ -3886,12 +3886,78 @@ ParamValue DeclarationVisitor::GetParamValue(const parser::TypeParamValue &x) {
...
+  // Process the type spec, if present
+  auto &typeSpec{std::get<std::optional<parser::IntegerTypeSpec>>(header.t)};
+  if (typeSpec.has_value()) {
+    SetDeclTypeSpec(MakeNumericType(TypeCategory::Integer, typeSpec->v));
+  }
+
+  // Process the index-name nodes in the ConcurrentControl nodes
+  auto &controls{std::get<std::list<parser::ConcurrentControl>>(header.t)};
+  for (auto &control : controls) {
+    ResolveIndexName(std::get<parser::Name>(control.t));
```
vjayathirtha-nv: For `ResolveIndexName`  calls, the std::get<> is done while passing the parameter, whereas for `ResolveControlExpression` the std::get<> is within the definition of the function, this seems inconsistent to me. I don't have a preference among these two approaches, but we should have the same for both.

psteinfeld: Actually, your suggestion was the way I originally coded it.  I changed it to eliminate the declaration and use of a variable.  But on reflection, you're correct.  The code reads better with similar signatures for both functions.

Thanks for pointing this out!  I'll change it.

lib/semantics/resolve-names.cc
```diff
@@ -3886,12 +3886,78 @@ ParamValue DeclarationVisitor::GetParamValue(const parser::TypeParamValue &x) {
...
+    ResolveControlExpression(control);
+  }
+
+  // Resolve the names in the scalar-mask-expr, if present
+  auto &maskExpr{std::get<std::optional<parser::ScalarLogicalExpr>>(header.t)};
+  if (maskExpr.has_value()) {
+    Walk(maskExpr.value());
+  }
   EndDeclTypeSpec();
+  return false;
```
schweitzpgi: Since you're stopping at `ConcurrentHeader`, it must be the case there is no other constraints that need to be checked concerning the syntactic substructure, correct?

lib/semantics/resolve-names.cc
```diff
@@ -875,7 +875,6 @@ class DeclarationVisitor : public ArraySpecVisitor,
 class ConstructVisitor : public DeclarationVisitor {
 public:
   bool Pre(const parser::ConcurrentHeader &);
```
vdonaldson: Are there any plans to replace these checker-specific Pre/Post calls with semantics-general Enter/Leave functions?  I believe that would be more uniform, and avoid this extra set of calls specific only to do checking.  If so, that might best be done sooner rather than later, so that any new additions use the new framework.

tskeith:
You changed the title of the PR -- it would be good to make the same change to the commit message.

lib/semantics/resolve-names.cc
psteinfeld: @schweitzpgi, I'm not sure whether or not that's the case.  I'm just plugging through all of the constraints one at a time fixing problems and making sure that they're checked and tested.  I might be adding more checks when I deal with additional constraints.  The reason that I chose ```ConcurrentHeader``` is that ```ConcurrentHeader``` defines the scope for the ```index-name```s of the DO CONCURRENT.

lib/semantics/resolve-names.cc
psteinfeld: @vdonaldson , you might be right.  In particular, if you look at the various semantic checking files in .../lib/semantics, you'll see a fair amount of variation.  For me personally, I plan to keep working on individual DO statement constraints.  As I'm working on the code, I'm also learning a lot more about it.  So I feel like I'll be in a much better position to decide whether and how to refactor the code after finishing my current goal of completing semantic checking for DO loops.

lib/semantics/resolve-names.cc
tskeith: This applies to `controls` and `control` too.

psteinfeld:
> 
> 
> You changed the title of the PR -- it would be good to make the same change to the commit message.

Thanks, @tskeith .  Are you saying that it's best practice to make the first line of the commit message match the title of the pull request?  Also, are you asking that I go back and change the message of my previous commit?

tskeith:
> Thanks, @tskeith . Are you saying that it's best practice to make the first line of the commit message match the title of the pull request? Also, are you asking that I go back and change the message of my previous commit?

It doesn't matter if they match. I'm just saying that you improved the title of the pull request and it would be nice to have the same improvement in the commit message, since that is the permanent record of the change. So, yes, change the commit message if it's easy (it may not be now that there is another commit).

lib/semantics/resolve-names.cc
psteinfeld: Thanks.  I'll change that.  I found another case in the code where ```const``` could be added, too.

lib/semantics/resolve-names.cc
schweitzpgi: It appeared that this recursed into instances of `ConcurrentHeader` previously, but `return false;` would eliminate that recursion. It's possible that is fine. I don't know.

psteinfeld: I believe that returning ```false``` is OK.  My new code explicitly walks the grammatical non-terminals of a ```ConcurrentHeader```.  Let me know if there's a test case that is not covered by this code.

PR #508, opened 2019-06-19
Fix a couple bugs with SELECT TYPE(p) from regression tests

klausler:
When a variable name appears in `SELECT TYPE` with no `associate-name =>`, the simple whole variable needs to be checked out as a variable (not parameter), converted if necessary to an object from an unknown entity, and then introduced into the per-guard scopes later.

### DIFF
--- a/test/semantics/test_errors.sh
+++ b/test/semantics/test_errors.sh
@@ -18,7 +18,7 @@
 
 PATH=/usr/bin:/bin
 srcdir=$(dirname $0)
-CMD="${F18:-../../../tools/f18/f18} -fdebug-resolve-names -fparse-only"
+CMD="${F18:-../../../tools/f18/f18} -fdebug-resolve-names -fparse-only -Mstandard"
 
 if [[ $# != 1 ]]; then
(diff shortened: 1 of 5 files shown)


klausler:
Another fix was added (allowing EQUIVALENCE of numeric with character as an extension with a warning, which every compiler but NAG supports).

test/semantics/test_errors.sh
```diff
@@ -18,7 +18,7 @@
 
 PATH=/usr/bin:/bin
 srcdir=$(dirname $0)
-CMD="${F18:-../../../tools/f18/f18} -fdebug-resolve-names -fparse-only"
+CMD="${F18:-../../../tools/f18/f18} -fdebug-resolve-names -fparse-only -Mstandard"
```
tskeith: Does this do anything useful? It turns on some warnings, but we don't check warnings.

Anyway, if it's going to be the default it can be removed from dosemantics03.f90.

klausler: Removed.

PR #512, opened 2019-06-21
Semantic Checks for Block Construct Implementation Proposal

CarolineConcatto:
This file contains implementation proposal for semantic checks for Block construct.
The file also has some questions related to the standard and if they should be implemented as semantic checks.

### DIFF
--- a/documentation/BlockConstruct.md
+++ b/documentation/BlockConstruct.md
@@ -1,0 +1,53 @@
+<!--
+Copyright (c) 2019, Arm Ltd.  All rights reserved.
+-->
+# Design proposal for Block Construct statement semantic analysis
+# The implementation of Block construct is based on the Section 11.1.4 of
+# Fortran 2018 standards Document
+
+The semantic checks of Block Construct in Section 11.1.4 are going to be
+implemented in these new files:
+ check-block.cc and check-block.h
+
+The new files check-block.•   implements/have BlockChecker class.
+The new class should contain (as others) one private field, a reference to the
+current semantic context
+
+The following sections contain:
+A. Implementation proposal
+and
+B. Questions.
+
+A. Implementation:
+1. Implement C1107 with a similar approach to Do by using the method:
+    void Post(const parser::ZZZStmt &).
+Where ZZZ in ZZZStmt are the invalid statements on block specification part.
+This method should be inside the BlockChecker class.
+
+2. Implement C1108  as a method in BlockChecker class.
+For this method, when in Block and if a SaveStmt is found the SavedEntity list
+will have its type checked for Common Statement.
+In case a  Common Statement is found  a message error will be printed.
+
+3. Check C1109 will not be implemented inside BlockChecker.
+C1109 is implemented in resolve-labels.cc as:
+    void CheckName(const parser::BlockConstruct &blockConstruct)
+
+B. Questions:
+1. Is Point number 2 related to Semantic Checks?
+Because I do not intend to implement Point 2 of section 11.1.4.
+This point is related to Asynchrounous and Volatile statements inside Block.
+I do not know if this should be implemented inside the BlockChecker class
+as a checker.
+
+2. Is Point number 3 related to  Semantic Checks?
+I do no intend  to implement Point 3 of section 11.1.4.
+This point is related to the evaluations of the specification expression inside
+block construct.
+I do not know if this should be implemented inside the BlockChecker class
+
+3. Point number 4 in section 11.1.4 checks if a GOTO statement inside a BLOCK
+construnct statment points to and end-block-statement within the block.
+So should this check be handled inside the Block Construct statement or inside
+GOTO statement.
+Or should the Go To statements handle the check?


documentation/BlockConstruct.md
```diff
@@ -0,0 +1,53 @@
...
+3. Check C1109 will not be implemented inside BlockChecker.
+C1109 is implemented in resolve-labels.cc as:
+    void CheckName(const parser::BlockConstruct &blockConstruct)
+
+B. Questions:
+1. Is Point number 2 related to Semantic Checks?
+Because I do not intend to implement Point 2 of section 11.1.4.
+This point is related to Asynchrounous and Volatile statements inside Block.
+I do not know if this should be implemented inside the BlockChecker class
+as a checker.
```
tskeith: This should probably be done as part of name resolution. We need to add a way to record the fact that a data object has the ASYNCHRONOUS or VOLATILE attribute for only part of its lifetime.

documentation/BlockConstruct.md
```diff
@@ -0,0 +1,53 @@
...
+
+2. Is Point number 3 related to  Semantic Checks?
+I do no intend  to implement Point 3 of section 11.1.4.
+This point is related to the evaluations of the specification expression inside
+block construct.
+I do not know if this should be implemented inside the BlockChecker class
```
tskeith: We need to check that expressions used as specification expressions follow the rules for them. That check will probably be done for all specification expressions together, so I agree it shouldn't be done in the BlockChecker.
Other than that, this point just describes how blocks are executed, not any semantic constraints.

documentation/BlockConstruct.md
```diff
@@ -0,0 +1,53 @@
...
+
+3. Point number 4 in section 11.1.4 checks if a GOTO statement inside a BLOCK
+construnct statment points to and end-block-statement within the block.
+So should this check be handled inside the Block Construct statement or inside
+GOTO statement.
+Or should the Go To statements handle the check?
```
tskeith: I believe this is already checked in `resolve-labels.cc`.

documentation/BlockConstruct.md
```diff
@@ -0,0 +1,53 @@
...
+-->
+# Design proposal for Block Construct statement semantic analysis
+# The implementation of Block construct is based on the Section 11.1.4 of
+# Fortran 2018 standards Document
+
+The semantic checks of Block Construct in Section 11.1.4 are going to be
+implemented in these new files:
+ check-block.cc and check-block.h
+
+The new files check-block.•   implements/have BlockChecker class.
```
tskeith: Something is messed up on this line after `check-block.`

documentation/BlockConstruct.md
```diff
@@ -0,0 +1,53 @@
+<!--
+Copyright (c) 2019, Arm Ltd.  All rights reserved.
+-->
+# Design proposal for Block Construct statement semantic analysis
+# The implementation of Block construct is based on the Section 11.1.4 of
+# Fortran 2018 standards Document
```
tskeith: The above two lines probably shouldn't be part of the header, only the line starting "Design proposal...".

documentation/BlockConstruct.md
CarolineConcatto: It should be check-block.*

documentation/BlockConstruct.md
psteinfeld: One great way to figure out if something's already been implemented is to write a test.  In my work on semantic checking for DO loops, I've discovered that some constraints that the standard groups with DO loops are handled by name resolution in resolve-names.[cc,h].

documentation/BlockConstruct.md
sscalpone: Tim, do you have a recommendation for how we should record the fact that a data objects has an attribute for only part of its lifetime?

tskeith: I think when we want to set one of these attributes in a block we would need to create a new symbol with HostAssocDetails. The attribute would be set on that symbol without affecting the symbol outside the block. Both symbols would represent the same Fortran entity.

CarolineConcatto:
@psteinfeld maybe  I am wrong, but it looks like that Block Construct(11.1.4) lists 3 constraints, from C1107 through  C1109.
And Change Team (11.1.5) lists 7 constraints, from C1110 through C116.
I will do your suggestion and make tests for each constraint.

psteinfeld:
> 
> 
> @psteinfeld maybe I am wrong, but it looks like that Block Construct(11.1.4) lists 3 constraints, from C1107 through C1109.
> And Change Team (11.1.5) lists 7 constraints, from C1110 through C116.
> I will do your suggestion and make tests for each constraint.

Oops!  You're correct.  Sorry for the mistake.

CarolineConcatto:
So it looks like that C1107 is already implemented on lib/semantics/resolve-names.cc. It is in this method:
DeclarationVisitor::CheckNotInBlock(const char *stmt) 
and we have some tests for this too (resolve40.f90 and resolve42.f90).
But these tests do not cover all the prohibitive statement. So I am thinking to create more test for all of them and add to my patch as well.

psteinfeld:
> 
> 
> So it looks like that C1107 is already implemented on lib/semantics/resolve-names.cc. It is in this method:
> DeclarationVisitor::CheckNotInBlock(const char *stmt)
> and we have some tests for this too (resolve40.f90 and resolve42.f90).
> But these tests do not cover all the prohibitive statement. So I am thinking to create more test for all of them and add to my patch as well.

This sounds correct and good.  Please add comments where appropriate saying that the code and tests apply to constraint C1107.

sscalpone:
The code pr has been merged!

PR #513, opened 2019-06-21
Test for constraint C1127

psteinfeld:
The DEFAULT ( NONE ) locality-spec shall not appear more than once in a given
concurrent-locality.

Note that, based on feedback, the nature of this pull request has changed from adding a test to downgrading the message produced by the compiler from an error to a non-error message when encountering multiple DEFAULT ( NONE ) locality-spec's.


test/semantics/dosemantics05.f90
```diff
@@ -0,0 +1,31 @@
...
+  IMPLICIT NONE
+  INTEGER :: ivar
+
+! This one works
+  DO CONCURRENT (ivar = 1:10) DEFAULT (NONE)
+    PRINT *, "ivar is: ", ivar
+  END DO
+
+!ERROR: only one DEFAULT(NONE) may appear
+  DO CONCURRENT (ivar = 1:10) DEFAULT (NONE) DEFAULT (NONE)
```
klausler: This can be a warning.  It's benign.

sscalpone: And please document as an extension.

test/semantics/dosemantics05.f90
```diff
@@ -0,0 +1,31 @@
...
+! limitations under the License.
+
+! C1127 -- The DEFAULT (NONE) locality-spec shall not appear more than once in a
+! given concurrent-locality.
+PROGRAM dosemantics05
+  IMPLICIT NONE
+  INTEGER :: ivar
+
+! This one works
+  DO CONCURRENT (ivar = 1:10) DEFAULT (NONE)
```
vjayathirtha-nv: Whilst PGI handles it as expected. Both GNU and Intel fortran compiler threw errors. 

```
test.f90(6): error #5082: Syntax error, found IDENTIFIER 'DEFAULT' when expecting one of: <END-OF-STATEMENT> ;
  DO CONCURRENT (ivar = 1:10) DEFAULT (NONE)
------------------------------^
```

vjayathirtha-nv: This error is separate from the expected one for the next DO CONCURRENT loop

test/semantics/dosemantics05.f90
vdonaldson: I vote that it is better to follow the standard in this case.  Why add an unnecessary laxness for a new feature without a track record of needing it, with the associated hassles (like documentation)?

test/semantics/dosemantics05.f90
tskeith: It looks like gfortran, ifort, nag, and xlf all don't support this feature yet. Only PGI does.

test/semantics/dosemantics05.f90
tskeith: I agree with Val. There are no implementations that I could find that allow this.

test/semantics/dosemantics05.f90
vjayathirtha-nv: I see

test/semantics/dosemantics05.f90
psteinfeld: I also agree with Val.  If I wrote this code, I'd want to be notified that there was something wrong.  On the other hand, if
we're warning for non-standard usage, I would be turning on the option to do that and to treat warnings as errors.

klausler: I'm not saying that we shouldn't emit a message.  It just doesn't need to be fatal by default.  Why should it be?  Nothing bad is going to happen if compilation proceeds.

psteinfeld: Actually, this sounds right to me.  If there's no objection, I'll change the code to check to make this an error only in the case where the -Mstandard and -Werror options are in effect.

klausler: You shouldn't have to change this code to accomplish that result.

psteinfeld: @klausler , I'm not sure what you mean by not having to change the code.  This pull request only adds a test.  If we don't change the current behavior of the compiler, it will issue an error message when two ```DEFAULT (NONE)``` clauses appear in a DO CONCURRENT header.  If I understand your previous comments correctly, that's what you want to change, and that's what I'm proposing to do.

klausler: I was responding to your "I'll change the code to check to make this an error only in the case where the -Mstandard and -Werror options are in effect.", which I understood to mean that you're changing the code that emits the message.

PR #521, opened 2019-06-23
Check constraints on passed-object dummy argument

tskeith:
The passed-object dummy argument cannot be checked until the
interfaces of contained subprograms are known. To accomplish this,
add `FinishSpecificationPart` pass to run after all specification
parts have been analyzed but before any of the execution parts.
This visits all derived types defined in each scope and performs
the checks on each procedure component and procedure binding.

Add a flag to `ParamValue` to distinguish kind from len parameters.

Fix some tests that had errors we now detect.

### DIFF
--- a/lib/semantics/resolve-names.cc
+++ b/lib/semantics/resolve-names.cc
@@ -270,7 +270,7 @@ protected:
 
 private:
   MaybeExpr bindName_;  // from BIND(C, NAME="...")
-  std::optional<SourceName> passName_;  // from PASS(...)
+  const SourceName *passName_{nullptr};  // from PASS(...)
 };
 
 // Find and create types from declaration-type-spec nodes.
@@ -1159,10 +1159,9 @@ Attrs AttrsVisitor::GetAttrs() {
   return *attrs_;
 }
 Attrs AttrsVisitor::EndAttrs() {
-  CHECK(attrs_);
-  Attrs result{*attrs_};
+  Attrs result{GetAttrs()};
   attrs_.reset();
-  passName_.reset();
+  passName_ = nullptr;
   bindName_.reset();
   return result;
 }
@@ -1216,7 +1215,7 @@ bool AttrsVisitor::Pre(const parser::IntentSpec &x) {
 }
 bool AttrsVisitor::Pre(const parser::Pass &x) {
   if (x.v) {
-    passName_ = x.v->source;
+    passName_ = &x.v->source;
     MakePlaceholder(*x.v, MiscDetails::Kind::PassName);
   } else {
     attrs_->set(Attr::PASS);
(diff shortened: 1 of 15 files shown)


lib/semantics/resolve-names.cc
```diff
@@ -1047,6 +1048,10 @@ class ResolveNamesVisitor : public virtual ScopeHandler,
   void AddSubpNames(const ProgramTree &);
   bool BeginScope(const ProgramTree &);
   void ResolveSpecificationParts(ProgramTree &);
+  void FinishSpecificationParts(const ProgramTree &);
+  void FinishDerivedType(Scope &);
```
psteinfeld: Can this be ```void FinishDerivedType(const Scope &);```

lib/semantics/resolve-names.cc
```diff
@@ -4850,6 +4847,146 @@ bool ResolveNamesVisitor::BeginScope(const ProgramTree &node) {
   }
 }
 
+// Perform checks that need to happen after all of the specification parts
+// but before any of the execution parts.
+void ResolveNamesVisitor::FinishSpecificationParts(const ProgramTree &node) {
+  if (!node.scope()) {
+    return;  // error occurred creating scope
```
schweitzpgi: Should this be an assert or are we expecting scope creation errors?

lib/semantics/resolve-names.cc
```diff
@@ -4850,6 +4847,146 @@ bool ResolveNamesVisitor::BeginScope(const ProgramTree &node) {
...
+  }
+  SetScope(*node.scope());
+  for (Scope &childScope : currScope().children()) {
```
psteinfeld: Can ```childScope``` be declared ```const Scope &childScope```?

lib/semantics/resolve-names.cc
```diff
@@ -4850,6 +4847,146 @@ bool ResolveNamesVisitor::BeginScope(const ProgramTree &node) {
...
+  for (std::size_t i{0}; i < symbols.size(); ++i) {
+    if (symbols[i] && symbols[i]->name() == name) {
+      return i;
+    }
+  }
+  return -1;
+}
+
+// Perform checks on procedure bindings of this type
+void ResolveNamesVisitor::FinishDerivedType(Scope &scope) {
```
psteinfeld: Similarly, I think the declaration here can be ```FinishDerivedType(const Scope &scope)```.

lib/semantics/resolve-names.cc
tskeith: We can fail to create a scope, for example, if a submodule names a parent module that doesn't exist. There is no good way to create a meaningful scope in that case.

lib/semantics/resolve-names.cc
tskeith: It would still compile if scope were declared const, but it doesn't seem right to me. I think you should only be able to get to const symbols through a const scope, and this function needs mutable symbols.

This may be a short-coming in the interface of Scope, but I don't know how to fix it.

psteinfeld: Sounds good.  Thanks for the explanation.

PR #522, opened 2019-06-23
Add a set of tests which compile the predefined module files.

sscalpone:
Each test simply tries to compile the module source file. Modify the script test_errors.sh to work with a source file that is specified as a full path. Also, change how the test temp directory is created to use a random name in case any of the source file base names are duplicated.

### DIFF
--- a/test/semantics/test_errors.sh
+++ b/test/semantics/test_errors.sh
@@ -24,12 +24,13 @@ if [[ $# != 1 ]]; then
   echo "Usage: $0 <fortran-source>"
   exit 1
 fi
-src=$srcdir/$1
+case $1 in
+(/*) src=$1 ;;
+(*) src=$srcdir/$1 ;;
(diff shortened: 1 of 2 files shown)


test/semantics/test_errors.sh
```diff
@@ -24,12 +24,13 @@ if [[ $# != 1 ]]; then
...
+case $1 in
+(/*) src=$1 ;;
+(*) src=$srcdir/$1 ;;
+esac
 [[ ! -f $src ]] && echo "File not found: $src" && exit 1
 
-temp=temp-$1
-rm -rf $temp
-mkdir $temp
+temp=`mktemp -d -p .`
```
tskeith: mktemp on MacOS doesn't have `-p`. It looks like this works there and on Linux: `mktemp -d ./tmp.XXXXXX`.

PR #523, opened 2019-06-24
Fix bug in writing PRIVATE subprograms to .mod file

### LINKED ISSUE #519: PRIVATE and PROTECTED subprograms misformatted in module files
Module subprograms with PRIVATE or PROTECTED attributes are formatted with PRIVATE/PROTECTED prefixes on the subprograms; these are not valid prefixes, and the module file fails to parse later when used.

Maybe these attributes should be accepted as extension prefixes.

tskeith:
Subprograms can't have the PRIVATE prefix on them when they are defined.
So if they are private, add a private-stmt for them.

Fixes #519

Note that issue #519 also mentions the PROTECTED attribute on subprograms.
This is not allowed but we don't detect it. This should be done as part of an
overall plan for enforcing constraints on attributes.

### DIFF
--- a/lib/semantics/mod-file.cc
+++ b/lib/semantics/mod-file.cc
@@ -52,6 +52,7 @@ static void PutBound(std::ostream &, const Bound &);
 static std::ostream &PutAttrs(std::ostream &, Attrs,
     const MaybeExpr & = std::nullopt, std::string before = ","s,
     std::string after = ""s);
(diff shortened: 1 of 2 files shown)


tskeith:
> Does an explicit `PUBLIC` attribute also need any treatment?

PUBLIC is okay because that is the default accessibility in the .mod file (we don't change the default) so the PUBLIC attribute is omitted from everything. Module m3 in modfile01.f90 has an example of this.

PR #526, opened 2019-06-25
Compile and install the predefined module files.

sscalpone:
Change the CMakeFile.txt in tools/f18 to build the predefined modules with the just-built compiler. The mod files are created in the new "include" subdirectory of the binary target directory.  The mod files are installed to ${CMAKE_INSTALL_PREFIX}/include.

The f18 driver is already installed in ${CMAKE_INSTALL_PREFIX}/bin.  This change change the location of the f18 binary build to a sibling directory of "include" called "bin" instead of at the top level of the binary destination directory. This change is in anticipation of changing the driver to find the include directory using a path relative to the location of f18.

Update the test scripts to find f18 in the bin subdirectory.

Remove the simple predefined module tests from test/semantics because they are compiled as part of the build and don't need to be recompiled as part of a test.

### DIFF
--- a/tools/f18/CMakeLists.txt
+++ b/tools/f18/CMakeLists.txt
@@ -12,6 +12,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+file(MAKE_DIRECTORY "${CMAKE_CURRENT_BINARY_DIR}/bin")
+file(MAKE_DIRECTORY "${CMAKE_CURRENT_BINARY_DIR}/include")
+
 add_executable(f18
   f18.cc
   dump.cc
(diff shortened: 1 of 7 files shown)


sscalpone:
You can specify the install directory at cmake time with -DCMAKE_INSTALL_PREFIX.  Use "make install" to create and populate the install directory.

sscalpone:
This commit changes the location of the f18 binary after a build.

tools/f18/CMakeLists.txt
```diff
@@ -32,4 +35,30 @@ target_link_libraries(f18-parse-demo
...
+
+set(MODULES
+  "ieee_arithmetic"
+  "ieee_exceptions"
+  "ieee_features"
+  "iso_c_binding"
+  "iso_fortran_env"
+)
+
+# Create module files directly from the top-level module source directory
```
sscalpone: Assumes that the module source file name matches the module file name.  Is this a safe assumption for now?

tskeith: It's a good assumption as long as we make the source file names match the module names. I don't think there is any reason not to do that.

sscalpone:
@klausler I considered adding a CMakeLists.txt to the modules source directory to build the mod file.  I chose not to do that because conceptually the compiled module files are part of the user interface, and in this case the user interface is tools/f18. Also, I did not want the module directory to depend on a particular tool (f18), or a tool in a remote directory, or a tool that will probably change over time.  I did consider building a special tool in the module source directory that simply creates the mod files but that seemed like overkill.

klausler:
> @klausler I considered adding a CMakeLists.txt to the modules source directory to build the mod file. I chose not to do that because conceptually the compiled module files are part of the user interface, and in this case the user interface is tools/f18. Also, I did not want the module directory to depend on a particular tool (f18), or a tool in a remote directory, or a tool that will probably change over time. I did consider building a special tool in the module source directory that simply creates the mod files but that seemed like overkill.

The driver in `tools/f18/f18.cc` is a temporary placeholder that will not persist as long as the predefined modules will.  With your scheme, those predefined module files are going to get rebuilt independently for every build compiler / version / host architecture / build configuration combination.  But none of those parameters should affect the contents of the module files -- when they vary, it's because they will depend on the *target* architecture in a cross-compilation environment.  For example, the array of supported REAL kind parameters will include kind=10 on x86-64 but not on any other target.

sscalpone:
> The driver in tools/f18/f18.cc is a temporary placeholder that will not persist as long as the predefined modules will. With your scheme, those predefined module files are going to get rebuilt independently for every build compiler / version / host architecture / build configuration combination. But none of those parameters should affect the contents of the module files -- when they vary, it's because they will depend on the target architecture in a cross-compilation environment. For example, the array of supported REAL kind parameters will include kind=10 on x86-64 but not on any other target.

Agree with all that.  I see building the modules with each configuration to be an advantage.  I'm not seeing an objection other than the transient nature of this commit.  I don't have a better idea for a long-term implementation.  I'm not arguing; I just don't have the answer.

PR #530, opened 2019-06-26
Common descriptor interface in front-end and runtime: RESHAPE example

jeanPerier:
Define an interface for descriptors and modify RESHAPE implementation to be templatized with a descriptor type compliant with this interface.
The interface comprises both types, functions and constants to be defined by descriptor implementations.
The purpose is to share the implementation of these intrinsic function between the front-end and the runtime by instantiating them with the relevant descriptor. It is also a goal to have an interface that is high level enough to make future changes in the runtime descriptor format easier. 

In the evaluate library, a `ConstantDescriptor` class is added to wrap `Constant<T>` with the descriptor interface.
The runtime `Descriptor` class is extended to provide the new descriptor interface.

The interface is not enforced through inheritance to avoid virtual functions, however, a validation class `DescriptorInterface` is added to both describe and validate the compliance of a descriptor implementation to this interface.

### DIFF
--- a/lib/common/abstract-descriptor.h
+++ b/lib/common/abstract-descriptor.h
@@ -1,0 +1,269 @@
+// Copyright (c) 2019, NVIDIA CORPORATION.  All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef FORTRAN_COMMON_ABSTRACT_DESCRIPTOR_H_
+#define FORTRAN_COMMON_ABSTRACT_DESCRIPTOR_H_
+
+#include "constexpr-bitset.h"
+
+namespace Fortran::common {
+
+// DescriptorInterface both serves to describe and validate the interface
+// that descriptor implementations must comply with in order to be used to
+// instantiate Fortran transformational intrinsic abstract implementation.
+// To validate the compliance of a descriptor implementation DESC with the
+// interface, one must compile in a test file:
+//
+//  template class DescriptorInterface<DESC>;
+//  template class Transformational<DescriptorInterface<DESC>>;
+//
+// The first line ensures DESC has all properties required by the descriptor
+// interface. The second line ensures the Abstract Fortran Runtime
+// implementation is only using descriptor properties that are part of the
+// descriptor interface. These are static tests only and it does not validate
+// the descriptor implementation logic.
+// Note that inheritance and abstract functions were not selected to make the
+// interface to avoid runtime overhead with virtual functions.
+
+template<typename DESC> class DescriptorInterface : private DESC {
+public:
+  // A descriptor implementation must define the following type
+
+  // basic integer types
+  using SubscriptValue = typename DESC::SubscriptValue;
+  using SizeValue = typename DESC::SizeValue;  // may or may not be unsigned
+  using Rank = typename DESC::Rank;
+  using Attribute = typename DESC::Attribute;
+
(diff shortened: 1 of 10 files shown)


lib/common/abstract-descriptor.h
```diff
@@ -0,0 +1,269 @@
...
+  // information may not be useful and this may be a no-op)
+  void SetCompilerCreatedFlag() { DESC::SetCompilerCreatedFlag(); }
+
+  // TODO: improve interface for attributes
+  static Attribute AllocatableAttribute() {
+    return DESC::AllocatableAttribute();
+  }
+};
+
+namespace safestd {
```
klausler: This seems likely to be useful in other situations and should be in its own header.

lib/common/abstract-descriptor.h
```diff
@@ -0,0 +1,269 @@
...
+#define FORTRAN_COMMON_ABSTRACT_DESCRIPTOR_H_
+
+#include "constexpr-bitset.h"
+
+namespace Fortran::common {
+
+// DescriptorInterface both serves to describe and validate the interface
+// that descriptor implementations must comply with in order to be used to
+// instantiate Fortran transformational intrinsic abstract implementation.
+// To validate the compliance of a descriptor implementation DESC with the
```
klausler: If you used the CRTP paradigm instead, this validation might not be necessary.

lib/common/abstract-descriptor.h
```diff
@@ -0,0 +1,269 @@
...
+// interface to avoid runtime overhead with virtual functions.
+
+template<typename DESC> class DescriptorInterface : private DESC {
+public:
+  // A descriptor implementation must define the following type
+
+  // basic integer types
+  using SubscriptValue = typename DESC::SubscriptValue;
+  using SizeValue = typename DESC::SizeValue;  // may or may not be unsigned
+  using Rank = typename DESC::Rank;
```
klausler: Rank is a value in [0..15] and `int` would work just fine.  No need to abstract it.

lib/common/abstract-descriptor.h
```diff
@@ -0,0 +1,269 @@
...
+        static_cast<DescriptorInterface<DESC> *>(&(*DESC::CreateWithSameTypeAs(
+            static_cast<const DESC &>(source), rank, attribute)))};
+  }
+
+  // Allocate a descriptor for which the type is fully defined:
+  //  - set bounds information (the lower bounds are set to to 1)
+  //  - get storage for the entity described by the descriptor and
+  //    make it available to the descriptor.
+  //  The data is not initialized (even for derived type).
+  void Allocate(const SubscriptArray &extents) { DESC::Allocate(extents); }
```
klausler: Will the results of transformational intrinsics require dynamic allocation for their results?

jeanPerier: I did not consider this here, but I think they should not so that we can avoid temporaries as much as possible.
I will try to split the logical and memory part of reshape to not have to allocate every results.
Then it would make sens to implement the transformational as member function of descriptor (that operates directly on their results). Do you agree ?

lib/common/abstract-descriptor.h
jeanPerier: Right, I considered it and I did not like using CRTP only for interface purposes, but maybe it could also be a better place for transformational if I change them to be descriptor member functions in order to split the memory and logic aspects (and not expect to dynamic memory allocation for every transformational results as per your other comments).

lib/common/abstract-descriptor.h
jeanPerier: I abstracted it because of `CFI_rank_t` that aliases `unsigned char` in include/flang/ISO_Fortran_binding.h. However, I agree it is a bit overkill and that the interface  should not have to care about the types used by the implementation to store the information.

I have updated the PR to remove `Rank` and use `int` directly as suggested.

lib/common/abstract-descriptor.h
klausler: We can avoid temporaries in the runtime for many intrinsics using a more complicated API.  Most of the intrinsic functions and intrinsic operations can be viewed as functions of indices into their results.

jeanPerier:
More needs to be done about intrinsic lowering before it is clear what runtime functions will be needed with descriptors (see PR #534). Until then, this PR does not really make sens and might go in the wrong direction.
Folding of transformational intrinsic functions will be performed directly on `Constant<T>` (see PR #544).

PR #531, opened 2019-06-26
Roll up more minor features & fixes to semantics for bugs found by regression testing

klausler:
Support some non-standard specific names for intrinsic functions (DCABS).

Support some more standard intrinsic functions with special semantics (ASSOCIATED).

Support common misinterpretation of the standard for parent components in structure constructors (T(P(1)) instead of T(P=P(1)) or just T(1) when type T extends type P).

Extend checking of pointer targets in pointer initializations.  Fix a test to use `=> NULL()`.

Support lower bounds other than 1 on `Constant<>` arrays so that subscripted `PARAMETER`s with other lower bounds are folded correctly when they appear in expressions.

Refactoring: a typedef for a variant (`SymbolOrComponent`) is now a proper class.

Fix up consequences of recent changes to name resolution so that function calls aren't converted to objects, and are implicitly typed.


module/ieee_arithmetic.f18.mod
```diff
@@ -0,0 +1,115 @@
+﻿!mod$ v1 sum:377fafa870889c4f
```
sscalpone: Do you mean to check-in the mod file?  I ask because it is a generated file.

klausler:
No, I do not intend to check any generated files into the tree.  I don’t think that I did, but I’ll go look and make sure.

sscalpone:
Sorry you saw that comment about the generated files. I deleted it almost immediately after adding it.

lib/evaluate/constant.cc
```diff
@@ -29,17 +29,18 @@ std::size_t TotalElementCount(const ConstantSubscripts &shape) {
   return size;
 }
 
-bool IncrementSubscripts(
-    ConstantSubscripts &indices, const ConstantSubscripts &shape) {
+bool IncrementSubscripts(ConstantSubscripts &indices,
```
vdonaldson: Jean has similar new functions in PR 530, although apparently with different argument types.  Is there any connection or overlap?

lib/evaluate/formatting.h
```diff
@@ -32,6 +32,8 @@
 
 namespace Fortran::evaluate {
 
+extern bool formatForPGF90;
```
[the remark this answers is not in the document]
klausler: It is a hack that will persist as long as we're using pgf90 in the f18 driver for code generation.

lib/evaluate/intrinsics.cc
```diff
@@ -198,6 +200,8 @@ static constexpr IntrinsicDummyArgument MatchingDefaultKIND{"kind",
 static constexpr IntrinsicDummyArgument SubscriptDefaultKIND{"kind",
     {IntType, KindCode::kindArg}, Rank::scalar,
     Optionality::defaultsToSubscriptKind};
+static constexpr IntrinsicDummyArgument RequiredDIM{
+    "dim", {IntType, KindCode::dimArg}, Rank::scalar};
```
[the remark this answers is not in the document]
klausler: Thanks.  My "build everywhere with everything" script must have a bug with exit codes, since I see the same failure in the logs but the overall script passed.  I'll fix it up.

lib/evaluate/fold.cc
```diff
@@ -2282,4 +2281,73 @@ std::optional<std::int64_t> ToInt64(const Expr<SomeType> &expr) {
     return std::nullopt;
   }
 }
+
+// Object pointer initialization checking predicate IsInitialDataTarget().
+// This code determines whether an expression is allowable as the static
+// data address used to initialize a pointer with "=> x".  See C765.
+// The caller is responsible for checking the base object symbol's
+// characteristics (TARGET, SAVE, &c.) since this code can't use GetUltimate().
+template<typename A> bool IsInitialDataTarget(const A &) { return false; }
```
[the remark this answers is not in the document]
klausler: That was my first approach, and it turned out to be not amenable as a solution.  I don't remember why.

lib/semantics/resolve-names.cc
```diff
@@ -4476,6 +4473,94 @@ const parser::Name *DeclarationVisitor::FindComponent(
...
+        if (pointer.Rank() != expr.Rank()) {
+          Say(pointer.name(),
+              "Pointer '%s' initialized with target of different rank"_err_en_US);
+        }
+        return;
+      }
+    }
+  }
+  Say(pointer.name(),
+      "Pointer '%s' initialized with invalid data target designator"_err_en_US);
```
tskeith: I think we should say why the initial data target is invalid, like with Rank above. This doesn't give the user much to go on.

Also, the message should be associated with the source location of the designator. In the example below the error is associated with the first line (because that's where the symbol for `p` is first created); it should be the third.
```
pointer :: p
real :: x
real :: p => x
end
```

lib/evaluate/variable.cc
```diff
@@ -362,6 +348,12 @@ int Component::Rank() const {
   return base().Rank();
 }
 
+int NamedEntity::Rank() const {
+  return std::visit(common::visitors{[](const Symbol *s) { return s->Rank(); },
+                        [](const Component &c) { return c.Rank(); }},
```
tskeith: If you add a comma after the second lambda, clang-format won't make such a mess of this.

lib/evaluate/intrinsics.cc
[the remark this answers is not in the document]
klausler: Build problems have been resolved.

lib/semantics/resolve-names.cc
klausler: Done, with a test.

tskeith: Looks good, thanks.

Issue #527, opened 2019-06-26
Build failure on Ubuntu 18.04 gcc 7.4


peterwaller-arm:
Similar to https://github.com/flang-compiler/f18/issues/312, f18 build fails on the latest Ubuntu LTS with the default compiler (gcc 7.4.0).

Tested on master as of 620441c03e21ae28aed3864b9581cd6cc1705c65 (25 Jun 2019).

There are a lot of errors. One of them is the same proximate error as #312:

```
/usr/include/c++/7/ext/new_allocator.h:136:4: error: use of deleted function ‘Fortran::parser::DataStmtValue::DataStmtValue(const Fortran::parser::DataStmtValue&)’
  { ::new((void *)__p) _Up(std::forward<_Args>(__args)...); }
    ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
In file included from /src/llvm-project/f18/lib/parser/user-state.h:25:0,
                 from /src/llvm-project/f18/lib/parser/parse-state.h:28,
                 from /src/llvm-project/f18/lib/parser/instrumented-parser.h:19,
                 from /src/llvm-project/f18/lib/parser/parsing.h:20,
                 from /src/llvm-project/f18/lib/parser/parsing.cc:15:
/src/llvm-project/f18/lib/parser/parse-tree.h:1420:252: note: declared here
   TUPLE_CLASS_BOILERPLATE(DataStmtValue);
```

# [Click for full build log (make -j1)](https://github.com/flang-compiler/f18/files/3329264/error.log)

# Solution

Through bisection I found my way to edd18355be574122aaa9abf58c15d8c50fb085a1 ("Remove needless braces"). If I revert this commit, master builds. I intend to submit this as a PR shortly.

peterwaller-arm:
After reverting edd1835 I see a lot of test failures with ctest that I don't yet understand, so I've not yet sent the PR.

```
76% tests passed, 44 tests failed out of 186

Total Test time (real) =  43.25 sec

The following tests FAILED:
        136 - symbol01.f90 (Failed)
        137 - symbol02.f90 (Failed)
        138 - symbol03.f90 (Failed)
        139 - symbol04.f90 (Failed)
        140 - symbol05.f90 (Failed)
        141 - symbol06.f90 (Failed)
        142 - symbol07.f90 (Failed)
        143 - symbol08.f90 (Failed)
        144 - symbol09.f90 (Failed)
[...]

PR #532, opened 2019-06-27
Revert "Remove needless braces"

peterwaller-arm:
This reverts commit edd18355be574122aaa9abf58c15d8c50fb085a1.

This fixes the build for me on Ubuntu 18.04 under gcc 7.4.0.

Please see #527 for detailed discussion.

### DIFF
--- a/lib/parser/basic-parsers.h
+++ b/lib/parser/basic-parsers.h
@@ -76,7 +76,7 @@ public:
   using resultType = A;
   constexpr PureParser(const PureParser &) = default;
   constexpr explicit PureParser(A &&x) : value_(std::move(x)) {}
-  std::optional<A> Parse(ParseState &) const { return value_; }
+  std::optional<A> Parse(ParseState &) const { return {value_}; }
 
 private:
   const A value_;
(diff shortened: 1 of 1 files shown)


klausler:
Can you clarify whether the braces are required by C++, or are they required to work around a bug in a particular version of GCC?

peterwaller-arm:
@klausler, apologies, at this stage I did not dig deeper to understand the failure, just applied a revert since it was a rapid and reasonably uncontroversial thing to do.

As you asked, I have further localized the issue to the `return {std::move(...)}` cases. The other cases do compile at least.

klausler:
> @klausler, apologies, at this stage I did not dig deeper to understand the failure, just applied a revert since it was a rapid and reasonably uncontroversial thing to do.
> 
> As you asked, I have further localized the issue to the `return {std::move(...)}` cases. The other cases do compile at least.

Excellent.  That narrows down the area of potential problems to avoid in those older GCC versions.  Thanks for digging further.

peterwaller-arm:
You're welcome :)

PR #533, opened 2019-06-27
Fixes for DO constraint C1124

psteinfeld:
The constraint says that a variable-name in a locality-spec shall be the name
of a variable in the innermost executable construct or scoping unit that
includes the DO CONCURRENT statement.  This check was already being made in
resolve-names.cc for LOCAL and LOCAL_INIT locality specs but not for the
SHARED locality spec.  Also, there was some code in check-do.cc that was
intended to be used to enforce this constraint at some time in the future.

I added code to resolve-names.cc to extend the checking to the SHARED locality
spec and removed the unused code from check-do.cc.  I also extended the
existing tests in resolve35.f90 to exercise the new code.

### DIFF
--- a/lib/semantics/resolve-names.cc
+++ b/lib/semantics/resolve-names.cc
@@ -432,6 +432,7 @@ public:
   void SayAlreadyDeclared(const SourceName &, Symbol &);
   void SayAlreadyDeclared(const parser::Name &, Symbol &);
   void SayWithDecl(const parser::Name &, Symbol &, MessageFixedText &&);
+  void SayBadLocality(const parser::Name &, Symbol &);
   void SayDerivedType(const SourceName &, MessageFixedText &&, const Scope &);
   void Say2(const SourceName &, MessageFixedText &&, const SourceName &,
       MessageFixedText &&);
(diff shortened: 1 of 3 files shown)


lib/semantics/resolve-names.cc
```diff
@@ -3969,18 +3974,25 @@ bool ConstructVisitor::Pre(const parser::LocalitySpec::LocalInit &x) {
...
+      if (prev->owner() == currScope()) {
+        SayAlreadyDeclared(name, *prev);  // C1125 and C1126
+      } else {
+        if (!IsVariableName(*prev)) {
+          SayBadLocality(name, *prev);  // C1124
+        } else {
+          auto &symbol{MakeSymbol(name, HostAssocDetails{*prev})};
+          symbol.set(Symbol::Flag::LocalityShared);
+        }
+      }
```
tskeith: I think that this kind of cascading `if` is better written like this:
```
    if ...
    } else if ...
    } else if ...
    } else ...
    }
```

psteinfeld: Thanks, Tim.  I'll fix this.

PR #534, opened 2019-06-27
Jot down thoughts on array expr and intrinsic evaluation for Jean

klausler:
Jean, this is stil quite rough, but I'm hoping that it will convey the basic idea to you of how array expressions and transformational intrinsic functions can be implemented in a way that separates storage allocation for the results (if any) and the computation of the elements of those results.  I've used this approach successfully in the (long ago) past.  I'll edit and improve this note more as time allows.

### DIFF
--- a/documentation/ArrayComposition.md
+++ b/documentation/ArrayComposition.md
@@ -93,6 +93,33 @@ Partial (axial) reductions can be similarly composed.
 The application of `SUM(A,DIM=2)` to the index `J` is the
 complete reduction `SUM(A(J,:))`.
 
+More completely:
+* Reductions to scalars (`SUM(X)` without `DIM=`) become
+  runtime calls; the result needs no dynamic allocation,
+  being a scalar.
+* Axial reductions (`SUM(X,DIM=d)`) applied to indices `(J,K)`
+  become scalar values like `SUM(X(J,K,:))` if `d=3`.
+* Location reductions to indices (`MAXLOC(X)` without `DIM=`)
+  do not require dynamic allocation, since their results are
+  either scalar or small vectors of length `RANK(X)`.
+* Axial location reductions (`MAXLOC(X,DIM=)`, &c.)
+  are handled like other axial reductions like `SUM(DIM=)`.
+* `TRANSPOSE(M)` exchanges the two components of the index tuple.
+* `RESHAPE(A,SHAPE=s)` without `ORDER=` must precompute the shape
+  vector `S`, and then use it to linearize indices into offsets
+  in the storage order of `A` (whose shape must also be captured).
+  These conversions can involve division and/or modulus, which
+  can be optimized into a fixed-point multiplication using the
+  usual technique.
+* `RESHAPE` with `ORDER=` is similar, but must permute the
+  components of the index tuple; it generalizes `TRANSPOSE`.
+* `CSHIFT` applies addition and modulus.
+* `EOSHIFT` applies addition and a conditional move (`SELECT`).
+* `PACK` and `UNPACK` are likely to require a runtime call.
+* `MATMUL(A,B)` can become `DOT_PRODUCT(A(J,:),B(:,K))`, but
+  might benefit from calling a highly optimized runtime
+  routine.
+
 Determination of rank and shape
 ===============================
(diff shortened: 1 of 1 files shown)


documentation/ArrayComposition.md
```diff
@@ -0,0 +1,130 @@
...
+The code that evaluates the expression, or that implements a transformational
+intrinsic in the runtime library, can be designed with an API that includes
+a pointer to the destination array as an argument.
+
+Statements like `ALLOCATE(A,SOURCE=expression)` should thus be capable
+of evaluating their array expressions directly into the newly-allocated
+storage for the allocatable array.
+The implementation would generate code to calculate the shape, use it
+to allocate the memory and populate the descriptor, and then drive a
+loop nest around the expression to populate the array.
```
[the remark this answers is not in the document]
klausler: Yes, I do think that the compiler should create the loop nest; it's basically a `FORALL`.  This would enable program restructuring transformations, loop optimizations, multi-version code generation, automatic GPU acceleration, &c.

documentation/ArrayComposition.md
```diff
@@ -0,0 +1,130 @@
...
+
+Transformational intrinsic functions as function composition
+============================================================
+Many of the transformational intrinsic functions listed above
+can, when their array arguments are viewed as functions over their
+index tuples, be seen as compositions of those functions with
+functions of the "incoming" indices.
+
+For example, the application of `TRANSPOSE(A + 1.0)` to the index
+tuple `(J,K)` becomes `A(K,J) + 1.0`.
```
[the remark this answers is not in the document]
klausler: Reductions (total or axial) have to be done before the loop nest (or level of the nest) that needs them.

```
A(:) = A(:) + SUM(B*C,DIM=2)
```

becomes essentially

```
FORALL(J=1:N) A(J)=A(J)+SUM(B(J,:)*C(J,:))
```

where `SUM` expands to a nested reduction loop within the outer parallel loop.

The `MapOperation` code in `fold.cc` is just a means of implementing folding of elemental intrinsic array operations.

klausler:
> Thank you so much for taking the time to write this! This clarifies quite a few things for me.
> Reading at this, I am under the impression that the Fortran runtime library should not care
> directly about these intrinsics, for which it could only provide memory when asked to and maybe set up descriptors,
> and that the nested loop part can/will be driven through code generation in the front-end.
> What is your take on this? Should I first come with a runtime interface for all of the intrinsics in this document (without implementing it) and then we could see what we do in code-gen/runtime?

Thanks for taking the time to read it; I'm sorry it was in such a rough state, and I'll still try to clean it up more and expand it before merging.  We'll need to have a runtime library API for these transformational intrinsic functions if we can't achieve an implementation of them via rewriting the calls in the front-end.

documentation/ArrayComposition.md
```diff
@@ -0,0 +1,130 @@
...
+
+This note attempts to describe the motivation for and design of an
+implementation of Fortran 90 (and later) array expression evaluation that
+minimizes the use of dynamically allocated temporary storage for
+the results of calls to transformational intrinsic functions.
+
+The transformational intrinsic functions of Fortran of interest to
+us here include:
+
+* Reductions to scalars (`SUM(X)`, also `ALL`, `ANY`, `COUNT`, `IALL`,
```
vjayathirtha-nv: If we want to include F2008 features then there's `NORM2`

klausler: Do you think that a `NORM2` can/should be computed inline in a compiler-generated loop instead of in a runtime library function?

vjayathirtha-nv: Performance wise it could better inline. However, actually implementing it inline might be convoluted and may not necessarily be the best solution. 
Is this document only talking about inline-able intrinsics?

klausler: Yes, the topic is how to implement some transformational intrinsic functions in generated code rather than in a runtime library.  I can see how a naive `NORM2` as `SQRT(SUM(X**2))` would work inline, but not an overflow-avoiding algorithm as recommended by the standard.

vjayathirtha-nv: Agreed.

PR #535, opened 2019-06-28
Extirpate NCHARACTER type, NC"" literals, and EUC-JP Hollerith

klausler:
These legacy PGI extensions are reported to be unused today in Japan.


tskeith:
The TODO can be removed from this line in `module/iso_c_binding.f90`:
`integer, parameter :: c_char = 1 ! TODO: Kanji mode`

PR #536, opened 2019-07-01
Structural checks for DO construct

ichoyjx:
Most restrictions in 2.7.1 Loop construct, nesting checks are still on TODO list.

`seenClauses` (`EnumSet`) is dropped. A `multimap clauseInfo` is added to save the encountering clause and its parser tree node pointer (`parser::OmpClause`) and to do the checks after walking through all the clauses.

Error msg examples:
```
omp-clause-validity01.f90:93:12: error: When SCHEDULE(AUTO) is specified, CHUNK_SIZE must not be specified
    !$omp do schedule(auto, 2) schedule(runtime, 2)
             ^^^^^^^^^^^^^^^^^
omp-clause-validity01.f90:93:30: error: At most one SCHEDULE clause can appear on the DO directive
    !$omp do schedule(auto, 2) schedule(runtime, 2)
                               ^^^^^^^^^^^^^^^^^^^^
omp-clause-validity01.f90:93:30: error: When SCHEDULE(RUNTIME) is specified, CHUNK_SIZE must not be specified
    !$omp do schedule(auto, 2) schedule(runtime, 2)
                               ^^^^^^^^^^^^^^^^^^^^
omp-clause-validity01.f90:100:12: error: A modifier may not be specified in a LINEAR clause on the DO or SIMD directive
    !$omp do linear(ref(b))
             ^^^^^^^^^^^^^^
omp-clause-validity01.f90:107:12: error: The NONMONOTONIC modifier can only be specified with SCHEDULE(DYNAMIC) or SCHEDULE(GUIDED)
    !$omp do schedule(NONMONOTONIC:static) ordered
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
omp-clause-validity01.f90:107:12: error: The NONMONOTONIC modifier cannot be specified if an ORDERED clause is specified
    !$omp do schedule(NONMONOTONIC:static) ordered
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
omp-clause-validity01.f90:120:12: error: A LINEAR clause or an ORDERED clause with a parameter can be specified on a loop directive but not both
    !$omp do ordered(1) private(b) linear(b) linear(a)
             ^^^^^^^^^^
```

### DIFF
--- a/lib/semantics/check-omp-structure.cc
+++ b/lib/semantics/check-omp-structure.cc
@@ -40,15 +40,14 @@ void OmpStructureChecker::CheckAllowed(const OmpClause &type) {
         parser::ToUpperCaseLetters(GetContext().directiveSource.ToString()));
     return;
   }
-  if (GetContext().allowedOnceClauses.test(type) &&
-      GetContext().seenClauses.test(type)) {
+  if (GetContext().allowedOnceClauses.test(type) && SeenClause(type)) {
     context_.Say(GetContext().clauseSource,
         "At most one %s clause can appear on the %s directive"_err_en_US,
         EnumToString(type),
         parser::ToUpperCaseLetters(GetContext().directiveSource.ToString()));
     return;
   }
-  SetContextSeen(type);
+  SetContextClauseInfo(type);
 }
 
 void OmpStructureChecker::Enter(const parser::OpenMPLoopConstruct &x) {
@@ -114,9 +113,8 @@ void OmpStructureChecker::Enter(const parser::OmpLoopDirective::Do &) {
   // reserve for nesting check
   SetContextDirectiveEnum(OmpDirective::DO);
 
-  OmpClauseSet allowed{OmpClause::PRIVATE, OmpClause::LASTPRIVATE,
-      OmpClause::LINEAR, OmpClause::REDUCTION, OmpClause::SCHEDULE,
-      OmpClause::COLLAPSE, OmpClause::ORDERED};
+  OmpClauseSet allowed{OmpClause::PRIVATE, OmpClause::FIRSTPRIVATE,
+      OmpClause::LASTPRIVATE, OmpClause::LINEAR, OmpClause::REDUCTION};
   SetContextAllowed(allowed);
 
   OmpClauseSet allowedOnce{
@@ -124,8 +122,48 @@ void OmpStructureChecker::Enter(const parser::OmpLoopDirective::Do &) {
   SetContextAllowedOnce(allowedOnce);
 }
 
+void OmpStructureChecker::Leave(const parser::OmpClauseList &) {
+  // 2.7 Loop Construct Restriction
+  if (GetContext().directive == OmpDirective::DO) {
+    if (SeenClause(OmpClause::SCHEDULE)) {
+      // only one schedule clause is allowed
+      auto it{GetContext().clauseInfo.find(OmpClause::SCHEDULE)};
+      auto *clause{it->second};
+      const auto &schedClause{std::get<parser::OmpScheduleClause>(clause->u)};
+      if (ScheduleModifierHasType(schedClause,
+              parser::OmpScheduleModifierType::ModType::Nonmonotonic)) {
+        if (SeenClause(OmpClause::ORDERED)) {
+          context_.Say(clause->source,
+              "The NONMONOTONIC modifier cannot be specified "
+              "if an ORDERED clause is specified"_err_en_US);
+        }
+        if (ScheduleModifierHasType(schedClause,
+                parser::OmpScheduleModifierType::ModType::Monotonic)) {
+          context_.Say(clause->source,
+              "The MONOTONIC and NONMONOTONIC modifiers "
+              "cannot be both specified"_err_en_US);
+        }
+      }
+    }
+
+    if (SeenClause(OmpClause::ORDERED) && SeenClause(OmpClause::LINEAR)) {
+      // only one ordered clause is allowed
+      auto it{GetContext().clauseInfo.find(OmpClause::ORDERED)};
+      auto *clause{it->second};
+      const auto &orderedClause{
+          std::get<parser::OmpClause::Ordered>(clause->u)};
+      if (orderedClause.v.has_value()) {
+        context_.Say(clause->source,
+            "A LINEAR clause or an ORDERED clause with a parameter "
+            "can be specified on a loop directive but not both"_err_en_US);
+      }
+    }
+  }
+}
+
 void OmpStructureChecker::Enter(const parser::OmpClause &x) {
   SetContextClauseSource(x.source);
+  SetContextClause(&x);
 }
 
 void OmpStructureChecker::Enter(const parser::OmpClause::Defaultmap &) {
@@ -234,8 +272,18 @@ void OmpStructureChecker::Enter(const parser::OmpDependClause &) {
 void OmpStructureChecker::Enter(const parser::OmpIfClause &) {
   CheckAllowed(OmpClause::IF);
 }
-void OmpStructureChecker::Enter(const parser::OmpLinearClause &) {
+void OmpStructureChecker::Enter(const parser::OmpLinearClause &x) {
   CheckAllowed(OmpClause::LINEAR);
+
+  // 2.7 Loop Construct Restriction
+  if (GetContext().directive == OmpDirective::DO ||
+      GetContext().directive == OmpDirective::SIMD) {
+    if (std::holds_alternative<parser::OmpLinearClause::WithModifier>(x.u)) {
+      context_.Say(GetContext().clauseSource,
+          "A modifier may not be specified in a LINEAR clause "
+          "on the DO or SIMD directive"_err_en_US);
+    }
+  }
 }
 void OmpStructureChecker::Enter(const parser::OmpMapClause &) {
   CheckAllowed(OmpClause::MAP);
@@ -246,7 +294,47 @@ void OmpStructureChecker::Enter(const parser::OmpProcBindClause &) {
 void OmpStructureChecker::Enter(const parser::OmpReductionClause &) {
   CheckAllowed(OmpClause::REDUCTION);
 }
-void OmpStructureChecker::Enter(const parser::OmpScheduleClause &) {
+
+bool OmpStructureChecker::ScheduleModifierHasType(
+    const parser::OmpScheduleClause &x,
+    const parser::OmpScheduleModifierType::ModType &type) {
+  const auto &modifier{std::get<0>(x.t)};
+  if (modifier.has_value()) {
+    const auto &modType1{std::get<0>(modifier->t)};
+    const auto &modType2{std::get<1>(modifier->t)};
+    if ((modType1.v).v == type ||
+        (modType2.has_value() && (modType2->v).v == type)) {
+      return true;
+    }
+  }
+  return false;
+}
+void OmpStructureChecker::Enter(const parser::OmpScheduleClause &x) {
   CheckAllowed(OmpClause::SCHEDULE);
+
+  // 2.7 Loop Construct Restriction
+  if (GetContext().directive == OmpDirective::DO) {
+    const auto &kind{std::get<1>(x.t)};
+    const auto &chunk{std::get<2>(x.t)};  // chunk_size
+    if ((kind == parser::OmpScheduleClause::ScheduleType::Runtime ||
+            kind == parser::OmpScheduleClause::ScheduleType::Auto) &&
+        chunk.has_value()) {
+      context_.Say(GetContext().clauseSource,
+          "When SCHEDULE(%s) is specified, "
+          "CHUNK_SIZE must not be specified"_err_en_US,
+          parser::ToUpperCaseLetters(
+              parser::OmpScheduleClause::EnumToString(kind)));
+    }
+
+    if (ScheduleModifierHasType(
+            x, parser::OmpScheduleModifierType::ModType::Nonmonotonic)) {
+      if ((kind != parser::OmpScheduleClause::ScheduleType::Dynamic &&
+              kind != parser::OmpScheduleClause::ScheduleType::Guided)) {
+        context_.Say(GetContext().clauseSource,
+            "The NONMONOTONIC modifier can only be specified with "
+            "SCHEDULE(DYNAMIC) or SCHEDULE(GUIDED)"_err_en_US);
+      }
+    }
+  }
 }
 }
--- a/lib/semantics/check-omp-structure.cc
+++ b/lib/semantics/check-omp-structure.cc
@@ -40,7 +40,7 @@ void OmpStructureChecker::CheckAllowed(const OmpClause &type) {
         parser::ToUpperCaseLetters(GetContext().directiveSource.ToString()));
     return;
   }
-  if (GetContext().allowedOnceClauses.test(type) && SeenClause(type)) {
+  if (GetContext().allowedOnceClauses.test(type) && FindClause(type)) {
     context_.Say(GetContext().clauseSource,
         "At most one %s clause can appear on the %s directive"_err_en_US,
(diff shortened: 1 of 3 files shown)


lib/semantics/check-omp-structure.cc
```diff
@@ -246,7 +294,47 @@ void OmpStructureChecker::Enter(const parser::OmpProcBindClause &) {
...
+            kind == parser::OmpScheduleClause::ScheduleType::Auto) &&
+        chunk.has_value()) {
+      context_.Say(GetContext().clauseSource,
+          "When SCHEDULE(%s) is specified, "
+          "CHUNK_SIZE must not be specified"_err_en_US,
+          parser::ToUpperCaseLetters(
+              parser::OmpScheduleClause::EnumToString(kind)));
+    }
+
+    if (ScheduleModifierHasType(
```
sscalpone: How about calling the function HasScheduleModifierType?  Is that an accurate name?

sscalpone:
Did you format this code with clang-format?

ichoyjx:
> Did you format this code with clang-format?

yes

lib/semantics/check-omp-structure.cc
ichoyjx: Looks like all the "Has" functions are prefixing "Has", I will change that, thanks.

lib/semantics/check-omp-structure.h
```diff
@@ -132,15 +135,25 @@ class OmpStructureChecker : public virtual BaseChecker {
   void SetContextAllowedOnce(const OmpClauseSet &allowedOnce) {
     ompContext_.back().allowedOnceClauses = allowedOnce;
   }
-  void SetContextSeen(const OmpClause &seenType) {
-    ompContext_.back().seenClauses.set(seenType);
+  void SetContextClause(const parser::OmpClause *clause) {
```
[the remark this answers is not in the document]
ichoyjx: Nice catch!

ichoyjx:
> What about the following 2.7.1 rule:
> 
> > If both the **collapse** and **ordered** clause with a parameter are specified, the parameter of the **ordered** clause must be greater than or equal to the parameter of the **collapse** clause.
> 
> It could probably be checked in `void OmpStructureChecker::Leave(const parser::OmpClauseList &)`. Is it part of the rules you will validate later?
> Otherwise, look good to me.

It was planned to do the check in resolving names, on the second thought it might be doable here, let me try, thanks!

lib/semantics/check-omp-structure.cc
```diff
@@ -246,7 +294,47 @@ void OmpStructureChecker::Enter(const parser::OmpProcBindClause &) {
...
-void OmpStructureChecker::Enter(const parser::OmpScheduleClause &) {
+
+bool OmpStructureChecker::ScheduleModifierHasType(
+    const parser::OmpScheduleClause &x,
+    const parser::OmpScheduleModifierType::ModType &type) {
+  const auto &modifier{std::get<0>(x.t)};
+  if (modifier.has_value()) {
+    const auto &modType1{std::get<0>(modifier->t)};
+    const auto &modType2{std::get<1>(modifier->t)};
+    if ((modType1.v).v == type ||
```
klausler: Those extra parentheses look strange.

lib/semantics/check-omp-structure.cc
```diff
@@ -246,7 +294,47 @@ void OmpStructureChecker::Enter(const parser::OmpProcBindClause &) {
...
+  }
+  return false;
+}
+void OmpStructureChecker::Enter(const parser::OmpScheduleClause &x) {
   CheckAllowed(OmpClause::SCHEDULE);
+
+  // 2.7 Loop Construct Restriction
+  if (GetContext().directive == OmpDirective::DO) {
+    const auto &kind{std::get<1>(x.t)};
+    const auto &chunk{std::get<2>(x.t)};  // chunk_size
```
klausler: This can be made neater.

'''
if (const auto &chunk{std::get<2>(x.t)}) {
  if (kind == .... || kind == ...) {
    ...
'''

lib/semantics/check-omp-structure.cc
```diff
@@ -114,18 +113,57 @@ void OmpStructureChecker::Enter(const parser::OmpLoopDirective::Do &) {
...
   SetContextAllowedOnce(allowedOnce);
 }
 
+void OmpStructureChecker::Leave(const parser::OmpClauseList &) {
+  // 2.7 Loop Construct Restriction
+  if (GetContext().directive == OmpDirective::DO) {
+    if (SeenClause(OmpClause::SCHEDULE)) {
+      // only one schedule clause is allowed
+      auto it{GetContext().clauseInfo.find(OmpClause::SCHEDULE)};
+      auto *clause{it->second};
```
tskeith: You could replace `SeenClause()` with `FindClause()` that returns `const parser::OmpClause *`.
Then here you could check if the SCHEDULE clause is present and get it all in one step.

Reading this code, it is surprising that you don't check that `it` is valid (i.e. not `end()`) before dereferencing it.

lib/semantics/check-omp-structure.cc
```diff
@@ -246,7 +294,47 @@ void OmpStructureChecker::Enter(const parser::OmpProcBindClause &) {
...
 }
-void OmpStructureChecker::Enter(const parser::OmpScheduleClause &) {
+
+bool OmpStructureChecker::ScheduleModifierHasType(
+    const parser::OmpScheduleClause &x,
+    const parser::OmpScheduleModifierType::ModType &type) {
+  const auto &modifier{std::get<0>(x.t)};
+  if (modifier.has_value()) {
+    const auto &modType1{std::get<0>(modifier->t)};
+    const auto &modType2{std::get<1>(modifier->t)};
```
tskeith: I think you should have explicit types here to improve readability. Either replace the number or auto with the real type. Otherwise you have to look at `OmpScheduleClause` to know what any of these types are.

lib/semantics/check-omp-structure.cc
```diff
@@ -246,7 +294,47 @@ void OmpStructureChecker::Enter(const parser::OmpProcBindClause &) {
...
+    if ((modType1.v).v == type ||
+        (modType2.has_value() && (modType2->v).v == type)) {
```
tskeith: You don't need parentheses around `modType1.v` or `modType2->v`.

lib/semantics/check-omp-structure.cc
klausler: I like having names for predicates that makes them stand out as such, but `HasScheduleModifierType` seems to mean something distinct from `ScheduleModifierHasType`.  The original `ScheduleModifierHasType` reads to me as "Does the schedule modifier have a type?" and the suggested "HasScheduleModifierType" reads as "Does `*this` have a schedule modifier type?"  I don't know which is more accurate.

lib/semantics/check-omp-structure.cc
ichoyjx: Oops... I will fix that, thanks!

lib/semantics/check-omp-structure.cc
```diff
@@ -246,7 +294,47 @@ void OmpStructureChecker::Enter(const parser::OmpProcBindClause &) {
...
+  // 2.7 Loop Construct Restriction
+  if (GetContext().directive == OmpDirective::DO) {
+    const auto &kind{std::get<1>(x.t)};
+    const auto &chunk{std::get<2>(x.t)};  // chunk_size
+    if ((kind == parser::OmpScheduleClause::ScheduleType::Runtime ||
+            kind == parser::OmpScheduleClause::ScheduleType::Auto) &&
+        chunk.has_value()) {
+      context_.Say(GetContext().clauseSource,
+          "When SCHEDULE(%s) is specified, "
+          "CHUNK_SIZE must not be specified"_err_en_US,
```
tskeith: I suggest "When SCHEDULE has %s specified it must not have chunk size specified".
CHUNK_SIZE is not something that appears in the source like SCHEDULE and AUTO.

lib/semantics/check-omp-structure.cc
```diff
@@ -246,7 +294,47 @@ void OmpStructureChecker::Enter(const parser::OmpProcBindClause &) {
...
+          parser::ToUpperCaseLetters(
+              parser::OmpScheduleClause::EnumToString(kind)));
+    }
+
+    if (ScheduleModifierHasType(
+            x, parser::OmpScheduleModifierType::ModType::Nonmonotonic)) {
+      if ((kind != parser::OmpScheduleClause::ScheduleType::Dynamic &&
+              kind != parser::OmpScheduleClause::ScheduleType::Guided)) {
```
tskeith: There are extra parentheses here and up above.

lib/semantics/check-omp-structure.cc
ichoyjx: The original `ScheduleModifierHasType` is more accurate in the context. Thanks Peter.

lib/semantics/check-omp-structure.cc
```diff
@@ -114,18 +113,57 @@ void OmpStructureChecker::Enter(const parser::OmpLoopDirective::Do &) {
   // reserve for nesting check
   SetContextDirectiveEnum(OmpDirective::DO);
 
-  OmpClauseSet allowed{OmpClause::PRIVATE, OmpClause::LASTPRIVATE,
-      OmpClause::LINEAR, OmpClause::REDUCTION, OmpClause::SCHEDULE,
-      OmpClause::COLLAPSE, OmpClause::ORDERED};
+  OmpClauseSet allowed{OmpClause::PRIVATE, OmpClause::FIRSTPRIVATE,
```
vjayathirtha-nv: I think you forgot to update the comment before this Enter() to match whatever is listed here.

lib/semantics/check-omp-structure.cc
```diff
@@ -114,18 +113,57 @@ void OmpStructureChecker::Enter(const parser::OmpLoopDirective::Do &) {
...
+    if (SeenClause(OmpClause::ORDERED) && SeenClause(OmpClause::LINEAR)) {
+      // only one ordered clause is allowed
+      auto it{GetContext().clauseInfo.find(OmpClause::ORDERED)};
+      auto *clause{it->second};
+      const auto &orderedClause{
+          std::get<parser::OmpClause::Ordered>(clause->u)};
+      if (orderedClause.v.has_value()) {
+        context_.Say(clause->source,
+            "A LINEAR clause or an ORDERED clause with a parameter "
+            "can be specified on a loop directive but not both"_err_en_US);
```
tskeith: How about:
"A loop directive may not have both a LINEAR clause and an ORDERED clause with a parameter"

lib/semantics/check-omp-structure.cc
```diff
@@ -114,18 +113,57 @@ void OmpStructureChecker::Enter(const parser::OmpLoopDirective::Do &) {
...
+        if (ScheduleModifierHasType(schedClause,
+                parser::OmpScheduleModifierType::ModType::Monotonic)) {
+          context_.Say(clause->source,
+              "The MONOTONIC and NONMONOTONIC modifiers "
+              "cannot be both specified"_err_en_US);
+        }
+      }
+    }
+
+    if (SeenClause(OmpClause::ORDERED) && SeenClause(OmpClause::LINEAR)) {
```
vjayathirtha-nv: You could use `FindClause()` here as well for `OmpClause::ORDERED`

ichoyjx: yes, `SeenClause` will all be replaced.

lib/semantics/check-omp-structure.cc
ichoyjx: nice catch

lib/semantics/check-omp-structure.cc
vjayathirtha-nv: well, `SeenClause(OmpClause::LINEAR)` here would be cleaner than using `FindClause()` since you don't need access to the return value of find.  According to me it makes more sense to replace only those `SeenClause()`s which later need a `FindClause()`

vjayathirtha-nv: What do you suggest @tskeith ?

tskeith: I think I would replace `SeenClause(OmpClause::LINEAR)` with `FindClause(Omp::LINEAR)`, either with or without an explicit `!= nullptr`. Just so there can be one less function.

If you want to keep `SeenClause()` I would use `FindClause()` to implement it.

lib/semantics/check-omp-structure.h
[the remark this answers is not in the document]
ichoyjx: Addressed in commit [1b57193](https://github.com/flang-compiler/f18/pull/536/commits/1b57193d2377c5121f16075950b64ea77fa72e2e).

lib/semantics/check-omp-structure.cc
ichoyjx: Addressed in commit [1b57193](https://github.com/flang-compiler/f18/pull/536/commits/1b57193d2377c5121f16075950b64ea77fa72e2e).

lib/semantics/check-omp-structure.cc
ichoyjx: Addressed in commit [1b57193](https://github.com/flang-compiler/f18/pull/536/commits/1b57193d2377c5121f16075950b64ea77fa72e2e).

lib/semantics/check-omp-structure.cc
ichoyjx: Addressed in commit [1b57193](https://github.com/flang-compiler/f18/pull/536/commits/1b57193d2377c5121f16075950b64ea77fa72e2e).

lib/semantics/check-omp-structure.cc
ichoyjx: Addressed in commit [1b57193](https://github.com/flang-compiler/f18/pull/536/commits/1b57193d2377c5121f16075950b64ea77fa72e2e).

lib/semantics/check-omp-structure.cc
ichoyjx: Addressed in commit [1b57193](https://github.com/flang-compiler/f18/pull/536/commits/1b57193d2377c5121f16075950b64ea77fa72e2e).

lib/semantics/check-omp-structure.cc
ichoyjx: Addressed in commit [1b57193](https://github.com/flang-compiler/f18/pull/536/commits/1b57193d2377c5121f16075950b64ea77fa72e2e).

lib/semantics/check-omp-structure.cc
ichoyjx: Addressed in commit [1b57193](https://github.com/flang-compiler/f18/pull/536/commits/1b57193d2377c5121f16075950b64ea77fa72e2e).

lib/semantics/check-omp-structure.cc
ichoyjx: Addressed in commit [1b57193](https://github.com/flang-compiler/f18/pull/536/commits/1b57193d2377c5121f16075950b64ea77fa72e2e).

lib/semantics/check-omp-structure.cc
ichoyjx: Addressed in commit [1b57193](https://github.com/flang-compiler/f18/pull/536/commits/1b57193d2377c5121f16075950b64ea77fa72e2e).

lib/semantics/check-omp-structure.cc
ichoyjx: Addressed in commit [1b57193](https://github.com/flang-compiler/f18/pull/536/commits/1b57193d2377c5121f16075950b64ea77fa72e2e). `SeenClause` has been replaced with `FindClause`.

klausler: I would retain `SeenClause()`.  It leaves no doubt about what it means.

PR #538, opened 2019-07-01
Semantics for LOC() & %LOC()

klausler:
Some regression tests that were mostly fixed last week need the non-standard `LOC` intrinsic function to compile completely.  While here, also implement the legacy variant spelling `%LOC` that is already supported in the parser and parse tree.

### DIFF
--- a/lib/evaluate/intrinsics.cc
+++ b/lib/evaluate/intrinsics.cc
@@ -82,6 +82,7 @@ ENUM_CLASS(KindCode, none, defaultIntegerKind,
     effectiveKind,  // for function results: same "kindArg", possibly defaulted
     dimArg,  // this argument is DIM=
     likeMultiply,  // for DOT_PRODUCT and MATMUL
+    subscript,  // address-sized integer
 )
 
 struct TypePattern {
@@ -106,6 +107,7 @@ static constexpr TypePattern DoublePrecision{
     RealType, KindCode::doublePrecision};
 static constexpr TypePattern DoublePrecisionComplex{
     ComplexType, KindCode::doublePrecision};
+static constexpr TypePattern SubscriptInt{IntType, KindCode::subscript};
 
 // Match any kind of some intrinsic or derived types
 static constexpr TypePattern AnyInt{IntType, KindCode::any};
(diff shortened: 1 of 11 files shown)


lib/evaluate/intrinsics.cc
```diff
@@ -1451,6 +1463,14 @@ static bool ApplySpecificChecks(
       messages.Say(
           "Arguments of ASSOCIATED() must be a POINTER and an optional valid target"_err_en_US);
     }
+  } else if (name == "loc") {
+    if (const auto &arg{call.arguments[0]}) {
+      ok = GetLastSymbol(arg->UnwrapExpr()) != nullptr;
+    }
+    if (!ok) {
+      messages.Say(
+          "Argument of LOC() must be an object or procedure"_err_en_US);
```
tskeith: What if the argument is something that has a symbol but is not an object or procedure, like a derived type or module?

klausler: `GetLastSymbol()` should return a null pointer in those cases.

klausler: More: those objects can't be represented in the expression data structure, so `GetLastSymbol()` would never be able to see them.

tskeith: OK, thanks

lib/evaluate/type.h
```diff
@@ -123,15 +129,22 @@ class DynamicType {
...
   DynamicType ResultTypeForMultiply(const DynamicType &) const;
 
   bool IsAssumedLengthCharacter() const;
+  constexpr bool IsPolymorphic() const {
+    return category_ == TypeCategory::Derived && kind_ > 0;
+  }
   constexpr bool IsUnlimitedPolymorphic() const {
-    return isPolymorphic_ && derived_ == nullptr;
+    return category_ == TypeCategory::Derived && derived_ == nullptr &&
+        kind_ == 1;
```
jeanPerier: After reviewing Tim's PR #539, I am not sure the name of this function is accurate. It does not accept `Type(*)` as unlimited polymorphic, but the standard says in 7.3.2:

> An entity that is declared using the TYPE(*) type specifier is assumed-type and is an unlimited polymorphic entity.

Would it make sens to rename this function to `IsClassStar` and/or to change it to accept `Type(*)` ?

klausler: No, it's better to follow the standard and return true for `TYPE(*)`.

PR #540, opened 2019-07-02
2.7.1 DO: Checks about clause arguments that allow Int Expr

ichoyjx:
For the clauses allow constant integer expression as an argument, we can check the validity of the evaluated value in the structural checker. If the argument allows integer variables, the analysis will be deferred.

Example:
```
omp-clause-validity01.f90:81:18: error: The parameter of the NUM_THREADS clause must be a positive integer expression
    !$omp parallel num_threads(1-4)
                   ^^^^^^^^^^^^^^^^
omp-clause-validity01.f90:146:26: error: The parameter of the ORDERED clause must be greater than or equal to the parameter of the COLLAPSE clause
    !$omp do collapse(num) ordered(1+2+3+4)
                           ^^^^^^^^^^^^^^^^
```

### DIFF
--- a/lib/semantics/check-omp-structure.cc
+++ b/lib/semantics/check-omp-structure.cc
@@ -162,14 +162,15 @@ void OmpStructureChecker::Leave(const parser::OmpClauseList &) {
           const auto &collapseClause{
               std::get<parser::OmpClause::Collapse>(clause2->u)};
           // ordered and collapse both have parameters
-          const auto orderedValue{GetIntValue(orderedClause.v)};
-          const auto collapseValue{GetIntValue(collapseClause.v)};
-          if (orderedValue && collapseValue && orderedValue.value() > 0 &&
-              collapseValue.value() > 0 &&
-              orderedValue.value() < collapseValue.value()) {
-            context_.Say(clause->source,
-                "The parameter of the ORDERED clause must be greater than or "
-                "equal to the parameter of the COLLAPSE clause"_err_en_US);
+          if (const auto orderedValue{GetIntValue(orderedClause.v)}) {
+            if (const auto collapseValue{GetIntValue(collapseClause.v)}) {
+              if (*orderedValue > 0 && *orderedValue < *collapseValue) {
+                context_.Say(clause->source,
+                    "The parameter of the ORDERED clause must be greater than "
+                    "or "
+                    "equal to the parameter of the COLLAPSE clause"_err_en_US);
+              }
+            }
           }
         }
       }
@@ -205,11 +206,12 @@ void OmpStructureChecker::Enter(const parser::OmpClause::Untied &) {
 void OmpStructureChecker::Enter(const parser::OmpClause::Collapse &x) {
   CheckAllowed(OmpClause::COLLAPSE);
   // collapse clause must have a parameter
-  const auto v{GetIntValue(x.v)};
-  if (v && v.value() <= 0) {
-    context_.Say(GetContext().clauseSource,
-        "The parameter of the COLLAPSE clause must be "
-        "a constant positive integer expression"_err_en_US);
+  if (const auto v{GetIntValue(x.v)}) {
+    if (*v <= 0) {
+      context_.Say(GetContext().clauseSource,
+          "The parameter of the COLLAPSE clause must be "
+          "a constant positive integer expression"_err_en_US);
+    }
   }
 }
 
@@ -248,11 +250,12 @@ void OmpStructureChecker::Enter(const parser::OmpClause::NumTeams &) {
 }
 void OmpStructureChecker::Enter(const parser::OmpClause::NumThreads &x) {
   CheckAllowed(OmpClause::NUM_THREADS);
-  const auto v{GetIntValue(x.v)};
-  if (v && v.value() <= 0) {
-    context_.Say(GetContext().clauseSource,
-        "The parameter of the NUM_THREADS clause must be "
-        "a positive integer expression"_err_en_US);
+  if (const auto v{GetIntValue(x.v)}) {
+    if (*v <= 0) {
+      context_.Say(GetContext().clauseSource,
+          "The parameter of the NUM_THREADS clause must be "
+          "a positive integer expression"_err_en_US);
+    }
   }
   // if parameter is variable, defer to Expression Analysis
 }
@@ -261,11 +264,12 @@ void OmpStructureChecker::Enter(const parser::OmpClause::Ordered &x) {
   CheckAllowed(OmpClause::ORDERED);
   // the parameter of ordered clause is optional
   if (const auto &expr{x.v}) {
-    const auto v{GetIntValue(expr)};
-    if (v && v.value() <= 0) {
-      context_.Say(GetContext().clauseSource,
-          "The parameter of the ORDERED clause must be "
-          "a constant positive integer expression"_err_en_US);
+    if (const auto v{GetIntValue(expr)}) {
+      if (*v <= 0) {
+        context_.Say(GetContext().clauseSource,
+            "The parameter of the ORDERED clause must be "
+            "a constant positive integer expression"_err_en_US);
+      }
     }
   }
 }
--- a/lib/semantics/check-omp-structure.cc
+++ b/lib/semantics/check-omp-structure.cc
@@ -166,9 +166,9 @@ void OmpStructureChecker::Leave(const parser::OmpClauseList &) {
             if (const auto collapseValue{GetIntValue(collapseClause.v)}) {
               if (*orderedValue > 0 && *orderedValue < *collapseValue) {
                 context_.Say(clause->source,
-                    "The parameter of the ORDERED clause must be greater than "
-                    "or "
-                    "equal to the parameter of the COLLAPSE clause"_err_en_US);
+                    "The parameter of the ORDERED clause must be "
+                    "greater than or equal to "
+                    "the parameter of the COLLAPSE clause"_err_en_US);
               }
             }
           }
--- a/lib/semantics/check-omp-structure.cc
+++ b/lib/semantics/check-omp-structure.cc
@@ -13,6 +13,7 @@
 // limitations under the License.
 
 #include "check-omp-structure.h"
+#include "tools.h"
 #include "../parser/parse-tree.h"
 
 namespace Fortran::semantics {
@@ -161,9 +162,10 @@ void OmpStructureChecker::Leave(const parser::OmpClauseList &) {
           const auto &collapseClause{
               std::get<parser::OmpClause::Collapse>(clause2->u)};
           // ordered and collapse both have parameters
-          const auto orderedValue{GetPosIntValue(orderedClause.v)};
-          const auto collapseValue{GetPosIntValue(collapseClause.v)};
-          if (orderedValue && collapseValue &&
+          const auto orderedValue{GetIntValue(orderedClause.v)};
+          const auto collapseValue{GetIntValue(collapseClause.v)};
+          if (orderedValue && collapseValue && orderedValue.value() > 0 &&
+              collapseValue.value() > 0 &&
               orderedValue.value() < collapseValue.value()) {
             context_.Say(clause->source,
                 "The parameter of the ORDERED clause must be greater than or "
@@ -203,7 +205,8 @@ void OmpStructureChecker::Enter(const parser::OmpClause::Untied &) {
 void OmpStructureChecker::Enter(const parser::OmpClause::Collapse &x) {
   CheckAllowed(OmpClause::COLLAPSE);
   // collapse clause must have a parameter
-  if (!GetPosIntValue(x.v)) {
+  const auto v{GetIntValue(x.v)};
+  if (v && v.value() <= 0) {
     context_.Say(GetContext().clauseSource,
         "The parameter of the COLLAPSE clause must be "
         "a constant positive integer expression"_err_en_US);
@@ -258,8 +261,8 @@ void OmpStructureChecker::Enter(const parser::OmpClause::Ordered &x) {
   CheckAllowed(OmpClause::ORDERED);
   // the parameter of ordered clause is optional
   if (const auto &expr{x.v}) {
-    const auto v{GetPosIntValue(expr)};
-    if (!v) {
+    const auto v{GetIntValue(expr)};
+    if (v && v.value() <= 0) {
       context_.Say(GetContext().clauseSource,
           "The parameter of the ORDERED clause must be "
           "a constant positive integer expression"_err_en_US);
--- a/lib/semantics/check-omp-structure.cc
+++ b/lib/semantics/check-omp-structure.cc
@@ -146,14 +146,28 @@ void OmpStructureChecker::Leave(const parser::OmpClauseList &) {
     }
 
     if (auto *clause{FindClause(OmpClause::ORDERED)}) {
-      if (FindClause(OmpClause::LINEAR)) {
-        // only one ordered clause is allowed
-        const auto &orderedClause{
-            std::get<parser::OmpClause::Ordered>(clause->u)};
-        if (orderedClause.v.has_value()) {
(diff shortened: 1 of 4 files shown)


lib/semantics/check-omp-structure.cc
```diff
@@ -146,15 +146,29 @@ void OmpStructureChecker::Leave(const parser::OmpClauseList &) {
     }
 
     if (auto *clause{FindClause(OmpClause::ORDERED)}) {
```
vjayathirtha-nv: why not name what clause this var is used for?
`orderedClause` here and ...

lib/semantics/check-omp-structure.cc
```diff
@@ -146,15 +146,29 @@ void OmpStructureChecker::Leave(const parser::OmpClauseList &) {
...
+          std::get<parser::OmpClause::Ordered>(clause->u)};
+
+      if (orderedClause.v.has_value()) {
+        if (FindClause(OmpClause::LINEAR)) {
           context_.Say(clause->source,
               "A loop directive may not have both a LINEAR clause and "
               "an ORDERED clause with a parameter"_err_en_US);
         }
+
+        if (auto *clause2{FindClause(OmpClause::COLLAPSE)}) {
```
vjayathirtha-nv: .. `collapseClause` here

lib/semantics/check-omp-structure.cc
vjayathirtha-nv: never mind there's already a var of that name

lib/semantics/check-omp-structure.cc
vjayathirtha-nv: even this var name

lib/semantics/check-omp-structure.h
```diff
@@ -147,6 +149,25 @@ class OmpStructureChecker : public virtual BaseChecker {
...
 
+  template<typename T> std::optional<std::int64_t> GetIntValue(const T &x) {
+    const auto *expr{GetExpr(x)};
+    if (expr && evaluate::ToInt64(*expr).has_value()) {
+      return evaluate::ToInt64(*expr).value();
+    }
+    return std::nullopt;
+  }
+  template<typename T> std::int64_t GetPosIntValue(const T &x) {
+    const auto &v{GetIntValue(x)};
```
klausler: I'm suprised this compiles, if it does.  Your `GetIntValue` routine doesn't return a reference, and I don't think that you can compare an `optional`.

The name of the function should make it very clear that it forces negative values to zero.

lib/semantics/check-omp-structure.h
```diff
@@ -147,6 +149,25 @@ class OmpStructureChecker : public virtual BaseChecker {
...
+    return v && v > 0 ? v.value() : 0;
+  }
+  template<typename T> std::int64_t GetConstPosIntValue(const T &x) {
+    const auto *expr{GetExpr(x)};
+    if (expr && IsConstantExpr(*expr)) {
```
klausler: Make sure that the definition of "constant expression" used by `IsConstantExpr()` is indeed what you want.

```
if (const auto *expr{GetExpr(x)}) {
  if (IsConstantExpr(*expr)) {
    return ...
  }
}
return 0;
```

Also, you had an explicit `evaluate::` namespace on the call to `ToInt64` earlier, but you don't have one here on the call to `IsConstantExpr()`.  Please be consistent if possible.

lib/semantics/check-omp-structure.h
ichoyjx: Thanks Peter, I will fix the reference part. In terms of the function, do you have a suggestion? @klausler

klausler: I would consider returning `std::optional<std::int64_t>` instead and leaving the name alone.

ichoyjx: Okay!

lib/semantics/check-omp-structure.h
```diff
@@ -147,6 +149,25 @@ class OmpStructureChecker : public virtual BaseChecker {
     return nullptr;
   }
 
+  template<typename T> std::optional<std::int64_t> GetIntValue(const T &x) {
```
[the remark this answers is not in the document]
ichoyjx: Addressed in commit [8aa3737](https://github.com/flang-compiler/f18/pull/540/commits/8aa373705b1f740ddf49345affc5a1fa0bbd84d6).

lib/semantics/check-omp-structure.h
```diff
@@ -147,6 +149,25 @@ class OmpStructureChecker : public virtual BaseChecker {
...
+    const auto *expr{GetExpr(x)};
```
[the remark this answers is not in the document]
ichoyjx: Addressed in commit [8aa3737](https://github.com/flang-compiler/f18/pull/540/commits/8aa373705b1f740ddf49345affc5a1fa0bbd84d6). All the Get* functions were changed to return type `std::optional`.

lib/semantics/check-omp-structure.h
ichoyjx: Addressed in commit [8aa3737](https://github.com/flang-compiler/f18/pull/540/commits/8aa373705b1f740ddf49345affc5a1fa0bbd84d6). Thanks.

lib/semantics/check-omp-structure.h
ichoyjx: Addressed in commit [8aa3737](https://github.com/flang-compiler/f18/pull/540/commits/8aa373705b1f740ddf49345affc5a1fa0bbd84d6). After talking with Tim, this `GetConstPosIntValue` function is removed.

lib/semantics/check-omp-structure.h
```diff
@@ -21,7 +21,9 @@
 #define FORTRAN_SEMANTICS_CHECK_OMP_STRUCTURE_H_
 
 #include "semantics.h"
+#include "tools.h"
```
tskeith: This include isn't needed here, only in the .cc file.

lib/semantics/check-omp-structure.h
```diff
@@ -21,7 +21,9 @@
...
 #include "../common/enum-set.h"
+#include "../evaluate/fold.h"
```
tskeith: I don't think this is needed at all.

lib/semantics/check-omp-structure.h
ichoyjx: Yes, I forgot to change that after moving around those functions. Thanks.

lib/semantics/check-omp-structure.cc
```diff
@@ -184,9 +199,17 @@ void OmpStructureChecker::Enter(const parser::OmpClause::Notinbranch &) {
...
 }
-void OmpStructureChecker::Enter(const parser::OmpClause::Collapse &) {
+
+void OmpStructureChecker::Enter(const parser::OmpClause::Collapse &x) {
   CheckAllowed(OmpClause::COLLAPSE);
+  // collapse clause must have a parameter
+  if (!GetPosIntValue(x.v)) {
+    context_.Say(GetContext().clauseSource,
+        "The parameter of the COLLAPSE clause must be "
+        "a constant positive integer expression"_err_en_US);
```
tskeith: I think you'll get two errors here if the expression is not constant or not integer. It would be better just to have one.

ichoyjx: You're right, I don't have a good solution here.
```
do.f90:3:12: error: The parameter of the COLLAPSE clause must be a constant positive integer expression
    !$omp do collapse(1.0) ordered(1+2)
             ^^^^^^^^^^^^^
do.f90:3:21: error: Must have INTEGER type, but is REAL(4)
    !$omp do collapse(1.0) ordered(1+2)
                      ^^^
f18: semantic errors in do.f90
```

lib/semantics/tools.h
```diff
@@ -160,5 +160,18 @@ template<typename T> const SomeExpr *GetExpr(const T &x) {
...
+template<typename T> std::optional<std::int64_t> GetIntValue(const T &x) {
+  if (const auto *expr{GetExpr(x)}) {
+    return evaluate::ToInt64(*expr);
+  } else {
+    return std::nullopt;
+  }
+}
+
+template<typename T> std::optional<std::int64_t> GetPosIntValue(const T &x) {
+  const auto v{GetIntValue(x)};
```
klausler: This would be more readable as

```
if (const auto v{GetIntValue(x)}) {
  if (*v > 0) {
    return v;
  }
}
return std::nullopt;
```

(or at least I think it would be)

I just noticed that you don't consider 0 to be positive.  That might confuse some expectations of readers who think that "positive" means "not negative".  So `GetPosIntValue` is maybe not a good name.  Try something like `GetIntValueIfOverZero`.

lib/semantics/check-omp-structure.cc
tskeith: You should be able to do it like for NUM_THREADS:
```
  const auto v{GetIntValue(x.v)};
  if (v && v.value() <= 0) {
```
`GetIntValue` calls `GetExpr` which should enforce the constraints that are expressed in the parse tree, `ScalarIntConstantExpr` in this case. All you have to check is that it is positive.

lib/semantics/tools.h
tskeith: I'm not sure this function is useful anyway. It might be easier just to use `GetIntValue()` and compare the result to `0` when you need to.

lib/semantics/check-omp-structure.h
ichoyjx: Addressed.

lib/semantics/check-omp-structure.h
ichoyjx: Addressed.

lib/semantics/check-omp-structure.cc
ichoyjx: Addressed in commit [6ad5f9f](https://github.com/flang-compiler/f18/pull/540/commits/6ad5f9f4427af9618137fa8036519c256d2303dc). Non-constant integer expr will be deferred to Expr Analysis. Thanks for the advice.

lib/semantics/tools.h
ichoyjx: Thank you both for the suggestions. I have got rid of this `GetPosIntValue` function in the latest commit. All the checks will analyze the result from `GetIntValue` now.

lib/semantics/check-omp-structure.cc
```diff
@@ -161,9 +162,10 @@ void OmpStructureChecker::Leave(const parser::OmpClauseList &) {
...
               std::get<parser::OmpClause::Collapse>(clause2->u)};
           // ordered and collapse both have parameters
-          const auto orderedValue{GetPosIntValue(orderedClause.v)};
-          const auto collapseValue{GetPosIntValue(collapseClause.v)};
-          if (orderedValue && collapseValue &&
+          const auto orderedValue{GetIntValue(orderedClause.v)};
+          const auto collapseValue{GetIntValue(collapseClause.v)};
+          if (orderedValue && collapseValue && orderedValue.value() > 0 &&
+              collapseValue.value() > 0 &&
               orderedValue.value() < collapseValue.value()) {
```
tskeith: I don't think you need the two `> 0` checks.

ichoyjx: If one or both of them are `<= 0`, it might not be necessary to throw the third error msg:
```
do.f90:4:12: error: The parameter of the COLLAPSE clause must be a constant positive integer expression
    !$omp do collapse(-1) ordered(-2)
             ^^^^^^^^^^^^
do.f90:4:25: error: The parameter of the ORDERED clause must be a constant positive integer expression
    !$omp do collapse(-1) ordered(-2)
                          ^^^^^^^^^^^
do.f90:4:25: error: The parameter of the ORDERED clause must be greater than or equal to the parameter of the COLLAPSE clause
    !$omp do collapse(-1) ordered(-2)
                          ^^^^^^^^^^^
f18: semantic errors in do.f90
```

lib/semantics/check-omp-structure.cc
```diff
@@ -146,15 +147,31 @@ void OmpStructureChecker::Leave(const parser::OmpClauseList &) {
...
           context_.Say(clause->source,
               "A loop directive may not have both a LINEAR clause and "
               "an ORDERED clause with a parameter"_err_en_US);
         }
+
+        if (auto *clause2{FindClause(OmpClause::COLLAPSE)}) {
+          const auto &collapseClause{
+              std::get<parser::OmpClause::Collapse>(clause2->u)};
+          // ordered and collapse both have parameters
+          const auto orderedValue{GetIntValue(orderedClause.v)};
```
klausler: Whenever you have

```
const auto foo{...};
if (foo) {
  ...
}
```

it can be made more readable by placing the declaration into the predicate expression.

```
if (const auto orderedValue{GetIntValue(orderedClause.v)}) {
  if (const auto collapseValue{GetIntValue(collapseClause.v)}) {
    if (*orderedValue > 0 && *orderedValue < *collapseValue) {
      Say("you lose");
    }
  }
}
```

lib/semantics/check-omp-structure.cc
```diff
@@ -220,11 +246,28 @@ void OmpStructureChecker::Enter(const parser::OmpClause::NumTasks &) {
 void OmpStructureChecker::Enter(const parser::OmpClause::NumTeams &) {
   CheckAllowed(OmpClause::NUM_TEAMS);
 }
-void OmpStructureChecker::Enter(const parser::OmpClause::NumThreads &) {
+void OmpStructureChecker::Enter(const parser::OmpClause::NumThreads &x) {
   CheckAllowed(OmpClause::NUM_THREADS);
+  const auto v{GetIntValue(x.v)};
```
klausler:
```
if (const auto v{GetIntValue(x.v)}) {
  if (*v <= 0) {
    Say("you lose");
  }
}
```

lib/semantics/check-omp-structure.cc
```diff
@@ -220,11 +246,28 @@ void OmpStructureChecker::Enter(const parser::OmpClause::NumTasks &) {
...
+  }
+  // if parameter is variable, defer to Expression Analysis
 }
-void OmpStructureChecker::Enter(const parser::OmpClause::Ordered &) {
+
+void OmpStructureChecker::Enter(const parser::OmpClause::Ordered &x) {
   CheckAllowed(OmpClause::ORDERED);
+  // the parameter of ordered clause is optional
+  if (const auto &expr{x.v}) {
+    const auto v{GetIntValue(expr)};
```
klausler: same here

lib/semantics/check-omp-structure.cc
klausler: I think that Tim meant that you don't need to check for `collapseValue.value() > 0` since the next relation would fail anyway if it were not true.

lib/semantics/check-omp-structure.cc
ichoyjx: Thank you!

lib/semantics/check-omp-structure.cc
ichoyjx: I made all the changes you suggested, thanks!

lib/semantics/check-omp-structure.cc
```diff
@@ -146,15 +147,32 @@ void OmpStructureChecker::Leave(const parser::OmpClauseList &) {
...
+
+        if (auto *clause2{FindClause(OmpClause::COLLAPSE)}) {
+          const auto &collapseClause{
+              std::get<parser::OmpClause::Collapse>(clause2->u)};
+          // ordered and collapse both have parameters
+          if (const auto orderedValue{GetIntValue(orderedClause.v)}) {
+            if (const auto collapseValue{GetIntValue(collapseClause.v)}) {
+              if (*orderedValue > 0 && *orderedValue < *collapseValue) {
+                context_.Say(clause->source,
+                    "The parameter of the ORDERED clause must be greater than "
```
klausler: This can be tightened up.

PR #542, opened 2019-07-02
Enforce constraint C1128 for DO CONCURRENT locality-spec's

psteinfeld:
These changes implement most of the requirements for C1128, which says: "A
variable-name that appears in a LOCAL or LOCAL_INIT locality-spec shall not
have the ALLOCATABLE; INTENT (IN); or OPTIONAL attribute; shall not be of
finalizable type; shall not be a nonpointer polymorphic dummy argument; and
shall not be a coarray or an assumed-size array.  A variable-name that is not
permitted to appear in a variable definition context shall not appear in a
LOCAL or LOCAL_INIT locality-spec."

The changes do not implement the checking required to determine whether a
variable can appear in a "variable definition context".

 - I created the function 'PassesLocalityChecks()' to enforce C1128 along with
   C1124, C1125, and C1126.
 - I cleaned up the code to check if a type or symbol is a coarray.
 - I added functions to tools.[h,cc] to test if a symbol is OPTIONAL, INTENT
   IN, finalizable, a coarray, or an assumed size array.  Should these be
   member functions of the type "Symbol"?
 - Since I changed one of the locality related error messages, I needed to
   change the test resolve35.f90.
 - I added the test resolve55.f90 to test all of the checks implemented in this
   update.

### DIFF
--- a/lib/semantics/resolve-names.cc
+++ b/lib/semantics/resolve-names.cc
@@ -3753,10 +3753,9 @@ bool DeclarationVisitor::PassesLocalityChecks(
         name, symbol, "Coarray '%s' not allowed in a locality-spec"_err_en_US);
     return false;
   }
-  const DeclTypeSpec *type{symbol.GetType()};
-  if (type) {
+  if (const DeclTypeSpec * type{symbol.GetType()}) {
     if (type->IsPolymorphic() && symbol.IsDummy() &&
-        (!IsPointer(symbol))) {  // C1128
+        !IsPointer(symbol)) {  // C1128
       SayWithDecl(name, symbol,
           "Nonpointer polymorphic argument '%s' not allowed in a "
           "locality-spec"_err_en_US);
--- a/lib/semantics/resolve-names.cc
+++ b/lib/semantics/resolve-names.cc
@@ -434,7 +434,7 @@ public:
   void SayAlreadyDeclared(const SourceName &, Symbol &);
   void SayAlreadyDeclared(const parser::Name &, Symbol &);
   void SayWithDecl(const parser::Name &, Symbol &, MessageFixedText &&);
-  void SayBadLocality(const parser::Name &, Symbol &);
+  void SayLocalMustBeVariable(const parser::Name &, Symbol &);
   void SayDerivedType(const SourceName &, MessageFixedText &&, const Scope &);
   void Say2(const SourceName &, MessageFixedText &&, const SourceName &,
       MessageFixedText &&);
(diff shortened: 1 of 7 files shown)


lib/semantics/resolve-names.cc
```diff
@@ -3672,8 +3676,62 @@ bool DeclarationVisitor::HandleUnrestrictedSpecificIntrinsicFunction(
...
+        "Finalizable variable '%s' not allowed in a locality-spec"_err_en_US);
+    return false;
+  }
+  if (IsCoarray(symbol)) {  // C1128
+    SayWithDecl(
+        name, symbol, "Coarray '%s' not allowed in a locality-spec"_err_en_US);
+    return false;
+  }
+  const DeclTypeSpec *type{symbol.GetType()};
+  if (type) {
```
ichoyjx: Maybe simply `if (const DeclTypeSpec *type{symbol.GetType()}) {` ?

lib/semantics/tools.cc
```diff
@@ -387,4 +385,33 @@ const Symbol *FindUltimateComponent(const DerivedTypeSpec &derivedTypeSpec,
   return nullptr;
 }
 
+bool IsFinalizable(const Symbol &symbol) {
+  const DeclTypeSpec *type{symbol.GetType()};
+  if (type) {
+    const DerivedTypeSpec *derived{type->AsDerived()};
+    if (derived) {
+      const Scope *scope{derived->scope()};
+      if (scope) {
```
ichoyjx: `type`, `derived`, and `scope`, same as above.

lib/semantics/resolve-names.cc
```diff
@@ -3672,8 +3676,62 @@ bool DeclarationVisitor::HandleUnrestrictedSpecificIntrinsicFunction(
...
+    if (type->IsPolymorphic() && symbol.IsDummy() &&
+        (!IsPointer(symbol))) {  // C1128
```
klausler: Surplus (parentheses)

lib/semantics/tools.cc
psteinfeld: Thanks.  I'll change that.

lib/semantics/resolve-names.cc
psteinfeld: Thanks.  I'll change that.

lib/semantics/resolve-names.cc
psteinfeld: Thanks.  I'll make that change.

lib/semantics/tools.cc
```diff
@@ -319,8 +319,6 @@ bool IsTeamType(const DerivedTypeSpec *derived) {
   return IsDerivedTypeFromModule(derived, "iso_fortran_env", "team_type");
 }
 
-bool IsCoarray(const Symbol &symbol) { return symbol.Corank() > 0; }
```
tskeith: This is a better implementation of `IsCoarray()` than the one that replaces it below. For example, this one returns true for a use-associated coarray and the other one returns false.

psteinfeld: Thanks.  I wondered about that.  I'll make that change.

PR #544, opened 2019-07-03
RESHAPE without shared runtime/front-end descriptor API

jeanPerier:
**EDIT**: 
PR #530 was closed and folding of transformational intrinsic functions will be implemented as done here.
This PR:
- Adds a few helper member functions on `Constant<T>`
- refactors folding to have a common path to all types in `FunctionRef` `FoldOperation` to avoid duplication (its not a big functional change, but this make a big git diff).
- Implements RESHAPE folding in `fold.cc` (because it may emits different user warning messages, it does not really belong as a `Constant<T>` member function).
- Returns a dummy "invalid" intrinsic function reference if user errors prevented folding so that the expression will not get re-folded (which would re-emit error message )
- Adds a note regarding issue #527 workaround (that does not apply in this PR , but was discussed in previous review)

### DIFF
--- a/lib/common/Fortran.h
+++ b/lib/common/Fortran.h
@@ -53,5 +53,7 @@ ENUM_CLASS(IoSpecKind, Access, Action, Advance, Asynchronous, Blank, Decimal,
     Convert,  // nonstandard
     Dispose,  // nonstandard
 )
+
+static constexpr int maxRank{15};
 }
 #endif  // FORTRAN_COMMON_FORTRAN_H_
--- a/lib/common/Fortran.h
+++ b/lib/common/Fortran.h
@@ -54,6 +54,7 @@ ENUM_CLASS(IoSpecKind, Access, Action, Advance, Asynchronous, Blank, Decimal,
     Dispose,  // nonstandard
 )
 
+// Fortran arrays may have up to 15 dimensions (See Fortran 2018 section 5.4.6).
 static constexpr int maxRank{15};
 }
 #endif  // FORTRAN_COMMON_FORTRAN_H_
--- a/lib/evaluate/constant.cc
+++ b/lib/evaluate/constant.cc
@@ -50,18 +50,13 @@ bool IncrementSubscripts(ConstantSubscripts &indices,
   return false;  // all done
 }
 
-std::optional<std::vector<int>> IsValidDimensionOrder(
-    int rank, const std::optional<std::vector<int>> &order) {
+std::optional<std::vector<int>> ValidateDimensionOrder(
+    int rank, const std::vector<int> &order) {
   std::vector<int> dimOrder(rank);
-  if (!order.has_value()) {
-    for (int j{0}; j < rank; ++j) {
-      dimOrder[j] = j;
-    }
-    return dimOrder;
-  } else if (static_cast<int>(order.value().size()) == rank) {
+  if (static_cast<int>(order.size()) == rank) {
     std::bitset<common::maxRank> seenDimensions;
     for (int j{0}; j < rank; ++j) {
-      int dim{order.value()[j]};
+      int dim{order[j]};
       if (dim < 1 || dim > rank || seenDimensions.test(dim - 1)) {
         return std::nullopt;
       }
--- a/lib/evaluate/constant.cc
+++ b/lib/evaluate/constant.cc
@@ -30,33 +30,15 @@ std::size_t TotalElementCount(const ConstantSubscripts &shape) {
 }
 
 bool IncrementSubscripts(ConstantSubscripts &indices,
-    const ConstantSubscripts &shape, const ConstantSubscripts &lbound) {
-  int rank{GetRank(shape)};
-  CHECK(GetRank(indices) == rank);
-  CHECK(GetRank(lbound) == rank);
-  for (int j{0}; j < rank; ++j) {
-    auto lb{lbound[j]};
-    CHECK(indices[j] >= lb);
-    if (++indices[j] < lb + shape[j]) {
-      return true;
-    } else {
-      CHECK(indices[j] == lb + shape[j]);
-      indices[j] = lb;
-    }
-  }
-  return false;  // all done
-}
-
-bool IncrementSubscripts(ConstantSubscripts &indices,
     const ConstantSubscripts &shape, const ConstantSubscripts &lbound,
-    const std::vector<int> &dimOrder) {
+    const std::vector<int> *dimOrder) {
   int rank{GetRank(shape)};
   CHECK(GetRank(indices) == rank);
   CHECK(GetRank(lbound) == rank);
-  CHECK(static_cast<int>(dimOrder.size()) == rank);
+  CHECK(!dimOrder || static_cast<int>(dimOrder->size()) == rank);
   for (int j{0}; j < rank; ++j) {
     auto lb{lbound[j]};
-    ConstantSubscript k{dimOrder[j]};
+    ConstantSubscript k{dimOrder ? (*dimOrder)[j] : j};
     CHECK(indices[k] >= lb);
     if (++indices[k] < lb + shape[k]) {
       return true;
@@ -70,15 +52,13 @@ bool IncrementSubscripts(ConstantSubscripts &indices,
 
 std::optional<std::vector<int>> IsValidDimensionOrder(
     int rank, const std::optional<std::vector<int>> &order) {
+  std::vector<int> dimOrder(rank);
   if (!order.has_value()) {
-    std::vector<int> result;
     for (int j{0}; j < rank; ++j) {
-      result.push_back(j);
+      dimOrder[j] = j;
     }
-    return {std::move(result)};
-  }
-  if (static_cast<int>(order.value().size()) == rank) {
-    std::vector<int> dimOrder(rank);
+    return dimOrder;
+  } else if (static_cast<int>(order.value().size()) == rank) {
     std::bitset<common::maxRank> seenDimensions;
     for (int j{0}; j < rank; ++j) {
       int dim{order.value()[j]};
@@ -88,21 +68,19 @@ std::optional<std::vector<int>> IsValidDimensionOrder(
       dimOrder[dim - 1] = j;
       seenDimensions.set(dim - 1);
     }
-    return {std::move(dimOrder)};
+    return dimOrder;
+  } else {
+    return std::nullopt;
   }
-  return std::nullopt;
 }
 
 bool IsValidShape(const ConstantSubscripts &shape) {
-  if (shape.size() > common::maxRank) {
-    return false;
-  }
   for (ConstantSubscript extent : shape) {
     if (extent < 0) {
       return false;
     }
   }
-  return true;
+  return shape.size() <= common::maxRank;
 }
 
 template<typename RESULT, typename ELEMENT>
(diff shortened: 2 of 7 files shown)


lib/evaluate/fold.cc
```diff
@@ -295,6 +295,72 @@ static std::optional<std::int64_t> GetInt64ArgOr(
...
+              *source, source->size(), subscripts, &dimOrder.value())};
+          if (copied < resultElements) {
+            CHECK(pad);
+            copied += result.CopyFrom(
+                *pad, resultElements - copied, subscripts, &dimOrder.value());
+          }
+          CHECK(copied == resultElements);
+          return result;
+        } else {
+          context.messages().Say(
```
klausler: In cases where you emit warnings, you might want to return *some* result, so that multiple folds of the same expressions don't emit duplicate messages.

lib/evaluate/fold.cc
```diff
@@ -554,6 +620,10 @@ Expr<Type<TypeCategory::Integer, KIND>> FoldOperation(FoldingContext &context,
     } else if (name == "rank") {
       // TODO assumed-rank dummy argument
       return Expr<T>{args[0].value().Rank()};
+    } else if (name == "reshape") {
+      if (auto maybeConstant{Reshape<T>(context, args)}) {
+        return Expr<T>{std::move(maybeConstant.value())};
```
klausler: It's okay to use `*std::move(maybeConstant)` instead of `.value()` immediately after checking that the value is present, as is done here, if you prefer.

lib/evaluate/fold.cc
```diff
@@ -755,10 +825,14 @@ Expr<Type<TypeCategory::Real, KIND>> FoldOperation(FoldingContext &context,
       if (auto *expr{args[0].value().UnwrapExpr()}) {
         return ToReal<KIND>(context, std::move(*expr));
       }
+    } else if (name == "reshape") {
```
klausler: We're going to want to find a way to fold the transformational intrinsic functions without having to duplicate the code for six different type categories for each one.

jeanPerier: Sure, what about a single entry point for `FunctionRef` that catches shared functions and then calls type specializations ? I have done this.
I also quickly checked if we could add some other shared path (e.g `FoldFloatinPointsIntrinsicFunctions`), but I think it would not save a lot of lines and it would makes the whole thing harder to follow.

lib/evaluate/fold.cc
jeanPerier: `Reshape` function was changed so this part is not needed anymore

lib/evaluate/fold.cc
jeanPerier: It is a good point, however, I am not sure what that result should be... Should we have an `InvalidExpr<T>` to do this in a clean way ? The best I can think of right now is to change the intrinsic name to "invalid" while keeping the characteristic. I have applied it in this PR.

But what if the erroneous expression is nested into an inquiry function? e.g `RANK(RESHAPE(source, shape_with_too_many_dimensions)) ` Should this succeed returning size of `shape_with_too_many_dimensions` or should this fail ? With my change, it will succeed but emit a warning that RESHAPE is wrong.

lib/evaluate/fold.cc
```diff
@@ -295,6 +295,72 @@ static std::optional<std::int64_t> GetInt64ArgOr(
   }
 }
 
+template<typename A, typename B>
```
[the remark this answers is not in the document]
jeanPerier: It would, but the function is also used to extract other types than int64, and I would rather have to avoid having to convert std::vector<int64> to other integer types afterwards.

lib/evaluate/constant.cc
```diff
@@ -104,6 +163,27 @@ auto ConstantBase<RESULT, ELEMENT>::Reshape(
...
+    ConstantSubscripts &resultSubscripts, const std::vector<int> *dimOrder) {
+  std::size_t copied{0};
+  ConstantSubscripts sourceSubscripts{source.lbounds_};
+  while (copied < count) {
+    values_.at(SubscriptsToOffset(resultSubscripts, shape_, lbounds_)) =
+        source.values_.at(SubscriptsToOffset(
+            sourceSubscripts, source.shape_, source.lbounds_));
+    copied++;
+    IncrementSubscripts(sourceSubscripts, source.shape_, source.lbounds_);
+    if (dimOrder) {
```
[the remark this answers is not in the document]
jeanPerier: Good point, thanks

lib/evaluate/constant.cc
```diff
@@ -46,6 +47,64 @@ bool IncrementSubscripts(ConstantSubscripts &indices,
...
+    std::bitset<common::maxRank> seenDimensions;
+    for (int j{0}; j < rank; ++j) {
+      int dim{order.value()[j]};
+      if (dim < 1 || dim > rank || seenDimensions.test(dim - 1)) {
+        return std::nullopt;
+      }
+      dimOrder[dim - 1] = j;
+      seenDimensions.set(dim - 1);
+    }
+    return {std::move(dimOrder)};
```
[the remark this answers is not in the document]
jeanPerier: Done

lib/evaluate/constant.cc
```diff
@@ -46,6 +47,64 @@ bool IncrementSubscripts(ConstantSubscripts &indices,
...
+
+std::optional<std::vector<int>> IsValidDimensionOrder(
+    int rank, const std::optional<std::vector<int>> &order) {
+  if (!order.has_value()) {
+    std::vector<int> result;
+    for (int j{0}; j < rank; ++j) {
+      result.push_back(j);
+    }
+    return {std::move(result)};
+  }
```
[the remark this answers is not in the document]
jeanPerier: Done

lib/evaluate/constant.cc
```diff
@@ -46,6 +47,64 @@ bool IncrementSubscripts(ConstantSubscripts &indices,
...
+}
+
+std::optional<std::vector<int>> IsValidDimensionOrder(
+    int rank, const std::optional<std::vector<int>> &order) {
+  if (!order.has_value()) {
+    std::vector<int> result;
+    for (int j{0}; j < rank; ++j) {
+      result.push_back(j);
+    }
+    return {std::move(result)};
```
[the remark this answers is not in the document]
jeanPerier: Done. I was a bit paranoid with this after #527 and #532 and was not willing to take the risk to break Ubuntu 18.04.... After your comment I checked on Ubuntu and you are correct, this is not an issue here. I was curious and could not take leaving in fear anymore, so I investigated a bit more, the bug #532 seems to be a g++ compiler bug (not an stdlib bug). I could reproduce with simple classes and come with the following workaround rule:
 
> One need to explicitly move an lvalue result that should be automatically moved by the compiler when:
    *  it can only be move constructed into the result type (no copy constructor), and
    *  where the result type constructor is defined using perfect forwarding (template<typename A> Constructor(A&& x): x_{std::forward(x)})

`result` is copyable, so it will be copy constructed into an `std::optional<std::vector<int>>` with Ubuntu 18.04 g++ (which is technically wrong, since it should have selected the move constructor, but at least, it does not crash).

lib/evaluate/constant.cc
```diff
@@ -46,6 +47,64 @@ bool IncrementSubscripts(ConstantSubscripts &indices,
   return false;  // all done
 }
 
+bool IncrementSubscripts(ConstantSubscripts &indices,
```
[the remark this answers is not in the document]
jeanPerier: Done

lib/common/Fortran.h
```diff
@@ -53,5 +53,7 @@ ENUM_CLASS(IoSpecKind, Access, Action, Advance, Asynchronous, Blank, Decimal,
     Convert,  // nonstandard
     Dispose,  // nonstandard
 )
+
+static constexpr int maxRank{15};
```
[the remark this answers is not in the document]
jeanPerier: Done

lib/evaluate/constant.cc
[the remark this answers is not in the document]
klausler: Can you please add a note about this bug with copy elision in older GCC compilers to the documentation somewhere so that we don't forget it?  Thank you.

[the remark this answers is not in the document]
jeanPerier: Sure I added a note at the end of the C++ style document in a new section about workaround for C++ compiler bugs.

lib/evaluate/constant.cc
```diff
@@ -46,6 +47,64 @@ bool IncrementSubscripts(ConstantSubscripts &indices,
...
+bool IsValidShape(const ConstantSubscripts &shape) {
+  if (shape.size() > common::maxRank) {
+    return false;
+  }
+  for (ConstantSubscript extent : shape) {
+    if (extent < 0) {
+      return false;
+    }
+  }
+  return true;
```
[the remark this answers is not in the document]
jeanPerier: This suggestion was also applied in previous commit

lib/evaluate/constant.cc
```diff
@@ -30,22 +30,59 @@ std::size_t TotalElementCount(const ConstantSubscripts &shape) {
...
-      CHECK(indices[j] == lb + shape[j]);
-      indices[j] = lb;
+      CHECK(indices[k] == shape[k] + lb);
+      indices[k] = lb;
     }
   }
   return false;  // all done
 }
 
+std::optional<std::vector<int>> IsValidDimensionOrder(
```
vjayathirtha-nv: I feel the name gives a wrong impression about the return value of the function.

Especially since there are other functions of the form `IsValidShape` which returns `bool`.

vjayathirtha-nv: I noticed that you have a comment next to this function's signature. I suppose that's okay

tskeith: `ValidateDimensionOrder` might be a better name.

Also, it's unfortunate that this function uses `std::optional<std::vector<int>>` in its interface while others in this file use `const std::vector<int> *`. Should the others, like `CopyFrom`, use `const std::optional<std::vector<int>> &` for `dimOrder`?

jeanPerier: Thanks, the name was indeed not that good, I took `ValidateDimensionOrder`.
For the interface aspect, it was more straight forward on the call site in that case, but I do not want to have  an optional `optional` argument in `IncrementSubscripts`. 

I actually completely removed the optional aspect on `ValidateDimensionOrder`, it did not fit with the name and it did not save that much (I also reorganized the if imbrication at the call site to use an if/else-if structure that is more straightforward).

lib/evaluate/fold.cc
```diff
@@ -313,296 +316,383 @@ static std::optional<std::int64_t> GetInt64ArgOr(
...
+      (args[3].has_value() && !order)) {
+    return Expr<T>{std::move(funcRef)};  // Non-constant arguments
+  } else if (!IsValidShape(shape.value())) {
+    context.messages().Say("Invalid SHAPE in RESHAPE"_en_US);
+  } else {
+    int rank{GetRank(shape.value())};
+    std::size_t resultElements{TotalElementCount(shape.value())};
+    std::optional<std::vector<int>> dimOrder;
+    if (order.has_value()) {
+      dimOrder = ValidateDimensionOrder(rank, *order);
```
vjayathirtha-nv: Some instances of accessing `std::optional` variable values use `*` operator and some others use the `value()`. I feel we should try to keep it uniform. However, I do know that both kinds of accesses exist in the code base outside of this change.

jeanPerier: It is a fair point. The rational I try to follow is: `value()` will kill the program if the optional is empty `*` does not do this check, so I use `*` if the line just above is `if  (opt.has_value())` because the checks it obviously redundant, everywhere-else, I use `.value()` for safety. Exception to the exception of using `*` is when it would end up `&(*opt)` which I find disturbing...
And yes, both kinds of access already exists outside of this change.

klausler: Please feel free to use `*` when it is obvious that the presence of the value does not have to be checked, as in this situation.  Using `.value()` on the line after checking `.has_value()` is not necessary.

vjayathirtha-nv: That rationale makes sense to me. Thanks.

PR #545, opened 2019-07-03
Fix some crashes

klausler:
Code in the evaluation library to generate expressions to compute character LEN() values needed to be allowed to fail gracefully by returning an `optional` result rather than crashing.


lib/evaluate/descender.h
```diff
@@ -308,9 +308,10 @@ template<typename VISITOR> class Descender {
     auto lx{ss.lower()};
     Visit(lx);
     ss.set_lower(std::move(lx));
-    auto ux{ss.upper()};
-    Visit(ux);
-    ss.set_lower(std::move(ux));
+    if (auto ux{ss.upper()}) {
+      Visit(ux);
+      ss.set_upper(std::move(*ux));
```
ichoyjx: Could you briefly explain this change?

klausler: The member function `upper` used to return an expression.  Now it returns an `optional` expression.  This change ensures that `set_upper` is called correctly with an expression.

ichoyjx: Okay. Was the call to `set_lower` used to be a typo?

klausler: Yes!  I'm glad I had to make a pass through this code, it gave me a chance to notice that.

lib/evaluate/call.cc
```diff
@@ -165,18 +165,14 @@ Expr<SubscriptInteger> ProcedureRef::LEN() const {
...
-          ConvertTo(stringLen, common::Clone(*nCopiesArg));
-    }
-    if (intrinsic->name == "trim") {
-      // LEN(TRIM(ch)) is unknown without execution.
-      CHECK(arguments_.size() == 1);
-      const auto *stringArg{
-          UnwrapExpr<Expr<SomeCharacter>>(arguments_[0].value())};
-      CHECK(stringArg != nullptr);
-      return stringArg->LEN();
+      if (auto stringLen{stringArg->LEN()}) {
```
vjayathirtha-nv: `stringArg` is of type Expr<SomeType> but I can't seem to find a LEN() associated with this type.
Which LEN() is invoked here?

It's not apparent from the name of the variable `stringLen` or otherwise that it is a pointer. Except that you are dereferencing it later.

vjayathirtha-nv: I'm just trying to understand why `stringLen` wasn't de-referenced before and now it is. Was it a typo?

klausler: It is `const Expr<SomeCharacter> *`, which should be clear from the declaration above (not part of this change).

klausler: `stringLen` is now an `optional` so we extract its value with `*` once it is known to be present.

lib/evaluate/intrinsics.cc
```diff
@@ -772,9 +772,9 @@ static const SpecificIntrinsicInterface specificIntrinsicFunction[]{
     {{"idnint", {{"a", DoublePrecision}}, DefaultInt}, "nint"},
     {{"ifix", {{"a", DefaultReal}}, DefaultInt}, "int", true},
     {{"index", {{"string", DefaultChar}, {"substring", DefaultChar}},
-        DefaultInt}},
+        SubscriptInt}},
     {{"isign", {{"a", DefaultInt}, {"b", DefaultInt}}, DefaultInt}, "sign"},
-    {{"len", {{"string", DefaultChar, Rank::anyOrAssumedRank}}, DefaultInt,
+    {{"len", {{"string", DefaultChar, Rank::anyOrAssumedRank}}, SubscriptInt,
```
jeanPerier: Sorry for commenting on a merged PR, but isn't it a deviation from the standard ? 16.9.112 4 :

> Result Characteristics. Integer scalar. If KIND is present, the kind type parameter is that specified by the
> value of KIND; otherwise the kind type parameter is that of default integer type.

(Same for INDEX). I am not saying it is wrong, but maybe a note should be added about the concept of subscriptIntegerKind into Extension.md ? Something like:

> The integer result of LEN, INDEX and LOC intrinsics when KIND is not present has kind 8 (therefore not necessarily the default integer type) . This is to make address calculations 64-bit safe.

klausler: It is an intentional deviation, and not the first for this issue.  Fortran's default INTEGER kind is simply broken as a choice for anything related to sizes (including loop controls).  But Fortran defines the default INTEGER as being the same size in storage as the default REAL, and that's always going to be 32 bits for probably forever.  I'll ensure that it is documented.

PR #549, opened 2019-07-03
Add DerivedTypeSpec::IsKindCompatibleWith

tskeith:
- Share function that checks derived type kind compatibility with `check-allocate.cc`
- Fix a bug in that function
- Move some member functions out of `DerivedTypeDetails`

### DIFF
--- a/lib/common/idioms.h
+++ b/lib/common/idioms.h
@@ -133,6 +133,10 @@ template<typename A> struct ListItemCount {
 // Given a const reference to a value, return a copy of the value.
 template<typename A> A Clone(const A &x) { return x; }
 
+template<typename A> inline bool PointeeComparison(const A *x, const A *y) {
+  return x == y || (x != nullptr && y != nullptr && *x == *y);
+}
+
 // C++ does a weird and dangerous thing when deducing template type parameters
 // from function arguments: lvalue references are allowed to match rvalue
 // reference arguments.  Template function declarations like
--- a/lib/semantics/type.h
+++ b/lib/semantics/type.h
@@ -251,6 +251,10 @@ public:
   bool operator==(const DerivedTypeSpec &that) const {
     return &typeSymbol_ == &that.typeSymbol_ && parameters_ == that.parameters_;
   }
+
+  // Do the kind type parameters of this have the same values as the
+  // corresponding kind type parameters of that?
+  bool IsKindCompatibleWith(const DerivedTypeSpec &) const;
   std::string AsFortran() const;
 
 private:
--- a/lib/evaluate/type.cc
+++ b/lib/evaluate/type.cc
@@ -93,14 +93,10 @@ bool IsDescriptor(const Symbol &symbol) {
 
 namespace Fortran::evaluate {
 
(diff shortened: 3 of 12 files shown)


lib/common/idioms.h
```diff
@@ -133,6 +133,10 @@ template<typename A> struct ListItemCount {
 // Given a const reference to a value, return a copy of the value.
 template<typename A> A Clone(const A &x) { return x; }
 
+template<typename A> inline bool PointeeComparison(const A *x, const A *y) {
+  return x == y || (x != nullptr && y != nullptr && *x == *y);
+}
```
tskeith: I kept the name of this function from `lib/evaluate/type.cc` but I would like something shorter, like just `Equals`. Any opinions on that?

lib/common/idioms.h
```diff
@@ -133,6 +133,10 @@ template<typename A> struct ListItemCount {
 // Given a const reference to a value, return a copy of the value.
 template<typename A> A Clone(const A &x) { return x; }
 
+template<typename A> inline bool PointeeComparison(const A *x, const A *y) {
```
klausler: If this is being promoted to `idioms.h`, maybe the name should be normalized to a verb (which I should have done originally).

lib/common/idioms.h
vjayathirtha-nv: How about `SameValue` or `EqualValue` ?

klausler: Please use a name that indicates that pointers/pointees are being compared.  `Equals` is vague.

`ComparePointees` is more what I had in mind.

psteinfeld: I'd suggest either IsPointerEqual or IsPointeeEqual.  The latter seems more accurate, but it also seems like a typo.

tskeith: The problem is that both pointers and pointees are being compared so either word is slightly misleading. Using "pointer" implies no dereferencing and using "pointee" implies unconditional dereferencing.

schweitzpgi: It appears to be testing if the pointers are referentially transparent, whether those pointers are valid for dereference or not.

lib/semantics/type.h
```diff
@@ -251,6 +251,10 @@ class DerivedTypeSpec {
   bool operator==(const DerivedTypeSpec &that) const {
     return &typeSymbol_ == &that.typeSymbol_ && parameters_ == that.parameters_;
   }
+
+  // Do the kind type parameters of this have the same values as the
+  // corresponding kind type parameters of that?
+  bool IsKindCompatibleWith(const DerivedTypeSpec &) const;
```
schweitzpgi: I'm curious why this functionality was moved into the class rather than exposing the ordered list of parameter decls. Of course, one could have both with a functional style `IsKindCompatible(x, y)` wrapper that just returns `x->IsKindCompat(y)`.

tskeith:
> Didn't you get linking issues ?

The linking issues were solved by adding FortranSemantics to the target_link_libraries of folding-test.

Issue #476, opened 2019-06-04
Distinguish warnings from informational messages


tskeith:
Currently f18 has two kinds of messages, fatal and non-fatal.
Non-fatal messages may be warnings (indicating a potential error)
or informational (e.g. extra information associated with another message).
We should distinguish between those two kinds of non-fatal messages.

Some things that would need to be done to achieve this:
- Change `bool isFatal_` in message texts to have 3 states
- Add another user-defined string literal for warnings
- Change message output to put "warning:" in warning messages
- Find existing non-fatal messages that should be warnings and change them
- Change `test_errors.sh` to check warnings as well as errors and update
  expected results

Questions:
- Should we have any other kinds of messages? For example, a message attached
  to another message to provide context might be distinguished from stand-alone
  messages that provide information about how the program is optimized.
- Should we rename the user-defined literals to remove `_en_US`? Messages are
  implicitly in that locale so we could shorted the suffixes to `_err`, `_wrn`,
  `_inf` or something like that.
- Do we want to add warning identifiers (issue #44) and should that be done
  at the same time?

klausler:
I see this as being a good way to make the `-Werror` flag not fail compilations with messages that aren't warnings.

I'd like to keep `_en_US` to retain the option of having non-English messages in the base sources.

Language extensions are already associated with per-feature flags.

To avoid cluttering source code with message numbers, pointers to URLs and Fortran standard sections, &c., we should have a little database of distinct messages that can include their translations and links to further information.

tskeith:
Regarding `_en_US`: why would we want to allow non-English messages in the source? That would mean when the compiler is built normally it would produce messages in more than one language. To get consistent messages it would have to go through a localization process.

klausler:
If something like Kanji mode were implemented by actual speakers of Japanese, maybe they'd want to be able to emit idiomatic error messages in Japanese right in the source.  That's the use case I was thinking about.

jeanPerier:
> we should have a little database of distinct messages

Do you mean something like flang error message tables [errtxt](https://github.com/flang-compiler/flang/blob/c82148884626fece6a752040c68ab6d3ce148d18/runtime/flang/error.c) and [kanjitxt](https://github.com/flang-compiler/flang/blob/c82148884626fece6a752040c68ab6d3ce148d18/runtime/flang/kanjidf.h) ? Would you have all messages in such database or only the core ones (not sure how "core" would be defined) ?

jeanPerier:
[...]

PR #551, opened 2019-07-04
Enabling automatic declaration of names in SHARED locality-spec's

psteinfeld:
Prior to this change, the compiler issued an error message when a name in a
SHARED locality-spec had not been declared explicitly in an enclosing scope.
Presumably, this was done to enforce constraint C1124.  This constraint states
that "A variable-name in a locality-spec shall be the name of a variable in the
innermost executable construct or scoping unit that includes the DO CONCURRENT
statement".  For LOCAL and LOCAL_INIT locality-spec's, we were automatically
creating a local variable in the situation where one had not been explicitly
declared rather than issuing an error message.

The only compiler I could find that implements the SHARED locality-spec was the
PGI compiler, which automatically creates a variable and does not emit an error
message.  Also, @vdonaldson said that he had consulted with a member of the
Fortran standards committee who said that the correct thing was to create a
variable if necessary.

This change has the compiler creating a variable in the enclosing scope of a DO
CONCURRENT statement if necessary.  I also changed a test to adapt to the new
behavior.  I also consolidated the semantic checking for the constraints
associated with all of the locality-spec's.

### DIFF
--- a/lib/semantics/resolve-names.cc
+++ b/lib/semantics/resolve-names.cc
@@ -1188,12 +1188,11 @@ bool AttrsVisitor::SetPassNameOn(Symbol &symbol) {
   if (!passName_) {
     return false;
   }
-  std::visit(
-      common::visitors{
-          [&](ProcEntityDetails &x) { x.set_passName(*passName_); },
-          [&](ProcBindingDetails &x) { x.set_passName(*passName_); },
-          [](auto &) { common::die("unexpected pass name"); },
-      },
+  std::visit(common::visitors{
+                 [&](ProcEntityDetails &x) { x.set_passName(*passName_); },
+                 [&](ProcBindingDetails &x) { x.set_passName(*passName_); },
+                 [](auto &) { common::die("unexpected pass name"); },
+             },
       symbol.details());
   return true;
 }
@@ -1345,20 +1344,19 @@ void ImplicitRulesVisitor::Post(const parser::ParameterStmt &x) {
 }
 
 bool ImplicitRulesVisitor::Pre(const parser::ImplicitStmt &x) {
-  bool res = std::visit(
-      common::visitors{
-          [&](const std::list<ImplicitNoneNameSpec> &x) {
-            return HandleImplicitNone(x);
-          },
-          [&](const std::list<parser::ImplicitSpec> &x) {
-            if (prevImplicitNoneType_) {
-              Say("IMPLICIT statement after IMPLICIT NONE or "
-                  "IMPLICIT NONE(TYPE) statement"_err_en_US);
-              return false;
-            }
-            return true;
-          },
-      },
+  bool res = std::visit(common::visitors{
+                            [&](const std::list<ImplicitNoneNameSpec> &x) {
+                              return HandleImplicitNone(x);
+                            },
+                            [&](const std::list<parser::ImplicitSpec> &x) {
+                              if (prevImplicitNoneType_) {
+                                Say("IMPLICIT statement after IMPLICIT NONE or "
+                                    "IMPLICIT NONE(TYPE) statement"_err_en_US);
+                                return false;
+                              }
+                              return true;
+                            },
+                        },
       x.u);
   prevImplicit_ = currStmtSource();
   return res;
@@ -1687,18 +1685,17 @@ void ScopeHandler::EraseSymbol(const parser::Name &name) {
 
 static bool NeedsType(const Symbol &symbol) {
   return symbol.GetType() == nullptr &&
-      std::visit(
-          common::visitors{
-              [](const EntityDetails &) { return true; },
-              [](const ObjectEntityDetails &) { return true; },
-              [](const AssocEntityDetails &) { return true; },
-              [&](const ProcEntityDetails &p) {
-                return symbol.test(Symbol::Flag::Function) &&
-                    p.interface().type() == nullptr &&
-                    p.interface().symbol() == nullptr;
-              },
-              [](const auto &) { return false; },
-          },
+      std::visit(common::visitors{
+                     [](const EntityDetails &) { return true; },
+                     [](const ObjectEntityDetails &) { return true; },
+                     [](const AssocEntityDetails &) { return true; },
+                     [&](const ProcEntityDetails &p) {
+                       return symbol.test(Symbol::Flag::Function) &&
+                           p.interface().type() == nullptr &&
+                           p.interface().symbol() == nullptr;
+                     },
+                     [](const auto &) { return false; },
+                 },
           symbol.details());
 }
 void ScopeHandler::ApplyImplicitRules(Symbol &symbol) {
@@ -1833,15 +1830,14 @@ void ModuleVisitor::Post(const parser::UseStmt &x) {
     // then add a use for each public name that was not renamed.
     std::set<SourceName> useNames;
     for (const auto &rename : *list) {
-      std::visit(
-          common::visitors{
-              [&](const parser::Rename::Names &names) {
-                useNames.insert(std::get<1>(names.t).source);
-              },
-              [&](const parser::Rename::Operators &ops) {
-                useNames.insert(std::get<1>(ops.t).v.source);
-              },
-          },
+      std::visit(common::visitors{
+                     [&](const parser::Rename::Names &names) {
+                       useNames.insert(std::get<1>(names.t).source);
+                     },
+                     [&](const parser::Rename::Operators &ops) {
+                       useNames.insert(std::get<1>(ops.t).v.source);
+                     },
+                 },
           rename.u);
     }
     for (const auto &[name, symbol] : *useModuleScope_) {
@@ -4106,9 +4102,7 @@ bool ConstructVisitor::Pre(const parser::LocalitySpec::LocalInit &x) {
 bool ConstructVisitor::Pre(const parser::LocalitySpec::Shared &x) {
   for (const auto &name : x.v) {
     if (!FindSymbol(name)) {
-      Say(name,
-          "Variable '%s' with SHARED locality implicitly"
-          " declared and may be uninitialized"_en_US);
+      Say(name, "Variable '%s' with SHARED locality implicitly declared"_en_US);
     }
     Symbol &prev{FindOrDeclareEnclosingEntity(name)};
     if (PassesSharedLocalityChecks(name, prev)) {
@@ -4150,17 +4144,16 @@ bool ConstructVisitor::Pre(const parser::DataImpliedDo &x) {
 }
 
 bool ConstructVisitor::Pre(const parser::DataStmtObject &x) {
-  std::visit(
-      common::visitors{
-          [&](const common::Indirection<parser::Variable> &y) {
-            Walk(y.value());
-          },
-          [&](const parser::DataImpliedDo &y) {
-            PushScope(Scope::Kind::ImpliedDos, nullptr);
-            Walk(y);
-            PopScope();
-          },
-      },
+  std::visit(common::visitors{
+                 [&](const common::Indirection<parser::Variable> &y) {
+                   Walk(y.value());
+                 },
+                 [&](const parser::DataImpliedDo &y) {
+                   PushScope(Scope::Kind::ImpliedDos, nullptr);
+                   Walk(y);
+                   PopScope();
+                 },
+             },
       x.u);
   return false;
 }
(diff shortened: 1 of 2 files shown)


lib/semantics/resolve-names.cc
```diff
@@ -4071,16 +4086,8 @@ bool ConstructVisitor::Pre(const parser::LocalitySpec::LocalInit &x) {
...
-    } else if (prev->owner() == currScope()) {
-      SayAlreadyDeclared(name, *prev);  // C1125 and C1126
-    } else if (!IsVariableName(*prev)) {
-      SayLocalMustBeVariable(name, *prev);  // C1124
-    } else {
+    Symbol *prev{EnsureEnclosingEntity(name)};
+    if (PassesSharedLocalityChecks(name, *prev)) {
       auto &symbol{MakeSymbol(name, HostAssocDetails{*prev})};
       symbol.set(Symbol::Flag::LocalityShared);
       name.symbol = &symbol;  // override resolution to parent
```
jeanPerier: I know this is not completely part of your change, but why is:

```
if (auto *type{prev->GetType()}) {
    symbol.SetType(*type);
    if (prev->test(Symbol::Flag::Implicit)) {
      symbol.set(Symbol::Flag::Implicit);
    }
  }
```
performed in `DeclareLocalEntity` for LOCAL and LOCAL_INIT variables and not here for SHARED variables ?

lib/semantics/resolve-names.cc
```diff
@@ -3767,35 +3784,33 @@ bool DeclarationVisitor::PassesLocalityChecks(
...
-  if (symbol.owner() == currScope()) {  // C1125 and C1126
-    SayAlreadyDeclared(name, symbol);
-    return false;
-  }
   // TODO: Check to see if the name can appear in a variable definition context
   return true;
 }
 
-Symbol *DeclarationVisitor::DeclareLocalEntity(const parser::Name &name) {
+Symbol *DeclarationVisitor::EnsureEnclosingEntity(const parser::Name &name) {
```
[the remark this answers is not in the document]
klausler: If a pointer result or variable or member will never be null, it should be a reference.  Being a pointer is a signal to the reader that the pointer can be null and must be checked.  (There are a few necessary exceptions to this general principle, but they get hidden; one cannot use a reference in some situations.)

lib/semantics/resolve-names.cc
```diff
@@ -3767,35 +3784,33 @@ bool DeclarationVisitor::PassesLocalityChecks(
...
   if (prev == nullptr) {
     // Declare the name as an object in the enclosing scope so that
     // the name can't be repurposed there later as something else.
     prev = &MakeSymbol(InclusiveScope(), name.source, Attrs{});
     ConvertToObjectEntity(*prev);
     ApplyImplicitRules(*prev);
-    implicit = true;
   }
+  return prev;
+}
```
klausler: Please emit a warning when creating the shared variable.  It's going to be shared state in concurrent code with no initialization, so code after the loop that uses the shared variable probably will have no way to know whether it was ever written to or simply remains uninitialized.

lib/semantics/resolve-names.cc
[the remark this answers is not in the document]
klausler: If a pointer result (or variable or member) can't be null, it must be a reference, and if it can be null, it should be checked.

test/semantics/resolve35.f90
```diff
@@ -88,7 +88,6 @@ subroutine s7
   do concurrent(integer::i=1:5) local(j, i) &
       !ERROR: 'j' is already declared in this scoping unit
       local_init(k, j) &
-      !ERROR: Variable 'a' not found
       shared(a)
     a(i) = j + 1
```
klausler: How will `a` be dimensioned?

lib/semantics/resolve-names.cc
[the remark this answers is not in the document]
psteinfeld: Thanks!  I'll change to references where possible and add null pointer checks to any remaining code that dereferences pointers.

lib/semantics/resolve-names.cc
```diff
@@ -746,6 +746,8 @@ class DeclarationVisitor : public ArraySpecVisitor,
   bool BeginDecl();
   void EndDecl();
   Symbol &DeclareObjectEntity(const parser::Name &, Attrs);
+  // Make sure that there's an entity in an enclosing scope called Name
+  Symbol *EnsureEnclosingEntity(const parser::Name &);
```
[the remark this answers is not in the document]
psteinfeld: Thanks, @jeanPerier .  I'll make that change.

lib/semantics/resolve-names.cc
psteinfeld: Thanks, @klausler .  Do you think that we should also warn for creating variable whose names appear in LOCAL and LOCAL_INIT locality-spec's?

klausler: No, absolutely not.

test/semantics/resolve35.f90
psteinfeld: Currently, the code creates a variable in the enclosing scope that's a scalar and creates another variable whose scope is the block of the ```DO CONCURRENT``` that's "host associated" with it.

klausler: But the test applies subscripts to `a`.

lib/semantics/resolve-names.cc
psteinfeld: Good question.  After consultation with @tskeith, I don't think that this code is needed in either case.  I'll remove it.

tskeith: The `ApplyImplicitRules` part was not needed in `DeclareLocalEntity` but copying the type and implicit flag still are.

The same code isn't needed here because the new symbol is a host-association with the previous one. The type of `symbol` should come from the type of `prev`.

lib/semantics/resolve-names.cc
```diff
@@ -3767,35 +3784,33 @@ bool DeclarationVisitor::PassesLocalityChecks(
...
   name.symbol = nullptr;
   Symbol &symbol{DeclareEntity<ObjectEntityDetails>(name, {})};
   if (auto *type{prev->GetType()}) {
-    if (implicit) {
-      ApplyImplicitRules(symbol);
-    } else {
-      symbol.SetType(*type);
+    symbol.SetType(*type);
+    if (prev->test(Symbol::Flag::Implicit)) {
+      symbol.set(Symbol::Flag::Implicit);
```
tskeith: You can write this as:
`symbol.set(Symbol::Flag::Implicit, prev->test(Symbol::Flag::Implicit))`

lib/semantics/resolve-names.cc
```diff
@@ -4086,9 +4085,14 @@ bool ConstructVisitor::Pre(const parser::LocalitySpec::LocalInit &x) {
 
 bool ConstructVisitor::Pre(const parser::LocalitySpec::Shared &x) {
   for (const auto &name : x.v) {
-    Symbol *prev{EnsureEnclosingEntity(name)};
-    if (PassesSharedLocalityChecks(name, *prev)) {
-      auto &symbol{MakeSymbol(name, HostAssocDetails{*prev})};
+    if (!FindSymbol(name)) {
+      Say(name,
+          "Warning, variable '%s' with SHARED locality allocated"
+          " automatically and may be uninitialized"_en_US);
```
tskeith: We don't start other warnings with "Warning," so we shouldn't here either. When issue #476 is implemented we will have a way to mark warnings and then they should come out with "warning:" before the message.

I would say "implicitly declared" rather than "allocated automatically". It's not allocated in the Fortran sense.

klausler: And it definitely will be uninitialized within the loop, and probably uninitialized afterwards -- and if it's not used after the loop, it shouldn't have been shared anyway.  Maybe this whole feature is a bad idea.

psteinfeld: Thanks, @tskeith.  I'll change the message.

psteinfeld: @vdonaldson, it was based on information from you in an email message that made me think it was a good idea to allow for implicitly declared variables to be named in a SHARED locality spec.  But I agree with @klausler that it's hard to imagine why anyone would want to do this.  Do you have any insight into this?

vdonaldson: Do concurrent locality lists are new in f18, and not necessarily well thought out or well designed.   Constraint C1124 does not distinguish between local, local_init, and shared sub-cases.  If implicit declaration of local and local_init variables is allowed, then implicit declaration of shared variables is as well, at least with regard to C1124.  As Peter notes, this program idiom is probably standard conforming, but useless:

```
    do concurrent (i=1:10) shared(s)
       ! code that doesn't reference s
    end do
```

The following code also doesn't violate C1124.  It is non-conforming from 11.1.7.5p4 because it has assignments to shared variable 's' in multiple iterations.  The compiler is not required to diagnose this violation. It should not mark the 'shared(s)' locality spec as an error, because that is not the error.  It would be ok to flag the assignment as an error, although that may be beyond current compiler capabilities.

```
    do concurrent (i=1:10) shared(s)
      s = 1.0
      ! do something with s
    end do

```

Short of changes to the standard or clarifying references elsewhere (unlikely), it looks like this
implicit declaration should be treated as a usual implicit declaration.  Two secondary responses seem reasonable:

 - Do nothing (else); this is not a constraint violation.
 - Generate a warning, possibly as neutral as "Implicit declaration of '%s' in a SHARED locality spec", although it might be possible to be more informative about why that is questionable.

Note that subscripting an implicitly declared locality spec name will probably generate some error on use.

psteinfeld: Thanks, Val.

I assume that your reference to the standard was to section 11.1.7.5p3, which discusses SHARED locality.  The more I learn and read about this, the crazier it seems.  This paragraph says (among other things), referring to the variable in the SHARED locality-spec: "If it is defined or becomes undefined during any iteration, it shall not be referenced, defined, or become undefined during any other iteration."  So the only valid use that I can think of would be for one of the iterations to define the variable and then for code outside the DO construct to use the defined value.

vdonaldson: Yes, I meant 11.1.7.5p3.  And yes, Peter K is right about the usefulness of allowing an implicit declaration in a shared locality spec, and about the design of the feature.  Do concurrent is more or less a replacement for a broken forall construct/statement, which is now deprecated.  Maybe the committee will need a third attempt to get it more or less right.

psteinfeld: Thanks, Val.

It seems to me that allowing implicit declaration of a name in SHARED locality-spec means that we're allowing people to do something that we don't recommend and can't effectively test.  Thus, even though it seems like the standard allows this, I recommend that we leave things as they are.  That is, that we emit an error message when an undeclared name appears in a SHARED locality-spec.

I'll pursue this problem with the standard outside of this forum.

If no one objects to this plan, I'll withdraw this pull request.

klausler: I actually don't think that FORALL is all that badly broken, it was just never well implemented.  DO CONCURRENT's default automatic localization rules are what really deserve the term; and the locality specs in 2018 were meant to fix them, and didn't.  They may have been the "third attempt", actually!  Sorry, this is getting totally off-topic for this code review.  Anyway, can you please emit a warning when implicitly declaring a SHARED name?

klausler: Whether we admit implicit declarations of SHARED scalars or not, we also have the question of whether we should emit a warning about implicit declarations of scalars in DO CONCURRENT that have default locality.

I think that you (Pete) should retain this PR -- one could write a valid Fortran program that would depend on implicit declaration, however artificial.  But also look into coughing up a warning in name resolution when there's an implicit declaration in a DO CONCURRENT without a locality spec.

vdonaldson: I'm ok with a warning.  I vote against an error.

psteinfeld: This pull request currently causes the following warning to be issued for undeclared names appearing in a SHARED locality-spec:
```Variable 'a' with SHARED locality implicitly declared and may be uninitialized```

I've added @gklimowicz as a reviewer to see if he has any insight into this.

psteinfeld: > 
> 
> Whether we admit implicit declarations of SHARED scalars or not, we also have the question of whether we should emit a warning about implicit declarations of scalars in DO CONCURRENT that have default locality.
> 
Thanks, Peter.  I've added the requirement to warn for implicit declaration of variables with unspecified locality to my list on Google Drive.

psteinfeld:
All of the feedback I've received is that we should allow for implicit declaration of a name that appears in a SHARED locality-spec.  And, when that happens, we should emit a warning.  That's the current state of this pull request.  It issues the warning: 
```Variable 'a' with SHARED locality implicitly declared and may be uninitialized```

klausler:
> All of the feedback I've received is that we should allow for implicit declaration of a name that appears in a SHARED locality-spec. And, when that happens, we should emit a warning. That's the current state of this pull request. It issues the warning:
> `Variable 'a' with SHARED locality implicitly declared and may be uninitialized`

I would either clarify "may be uninitialized" or remove the phrase from the message.  The variable will certainly be uninitialized within the loop; it will most likely be uninitialized or invalid for use after the loop.

psteinfeld:
> 
> 
> > All of the feedback I've received is that we should allow for implicit declaration of a name that appears in a SHARED locality-spec. And, when that happens, we should emit a warning. That's the current state of this pull request. It issues the warning:
> > `Variable 'a' with SHARED locality implicitly declared and may be uninitialized`
> 
> I would either clarify "may be uninitialized" or remove the phrase from the message. The variable will certainly be uninitialized within the loop; it will most likely be uninitialized or invalid for use after the loop.

Thanks.  I'll remove the phrase before merging.

Issue #546, opened 2019-07-03
Bogus error on labeled DO WHILE


klausler:
```
      j=1
      do40while(j.ne.3)
      j=j+1
      doi=j,j+1
40    enddo
      end
```

elicits "label '40' is not in scope" and "Internal: no symbol found for 'j'" when compiled with semantics enabled.

jeanPerier:
Looks like the same issue as #518

jeanPerier:
@klausler do you confirm it is solved ?

klausler:
Yes, I think that it is.

Issue #518, opened 2019-06-21
Internal error on DO loop index


klausler:
```
  do 1 j=1,2
    if (.true.) then
1   end if
  end
```

elicits a legitimate error message about the label `1` but also emits `error: Internal: no symbol found for 'j'` when semantics is enabled.

jeanPerier:
**First analysis:**
`LabelDoStmt` are supposed to be rewritten to `NonLabelDoStmt` by label resolution before name resolution.
Name resolution forbids name resolution inside `LabelDoStmt` on purpose (see  [resolve-names.cc L918](https://github.com/flang-compiler/f18/blob/248600031d8fc500edc3e97e8921a0bad64681de/lib/semantics/resolve-names.cc#L918)) (because it does not expect them ?).

The `label '1' is not in scope` is not emitted as an error but warning. This is in line with other compilers (e.g pgfortran) that accepts the construct. However, label resolution failed to rewrite the `LabelDoStmt` to `NonLabelDoStmt` in this situation. Hence, `j` is not resolved and error message emitted.

**Planned follow-up:**
- See why label resolution failed to rewrite the `LabelDoStmt` to `NonLabelDoStmt`
- either fix it if possible, or stop guaranteeing the absence of `LabelDoStmt` in later stages

jeanPerier:
@klausler do you confirm it is solved ?

jeanPerier:
Fixed by #553: Code from issue compiles without errors.
Note that `Document issue #518 workaround` commit 9bd5ad9 linked in this issue has a wrong title, it is about issue #527 workaround. No workaround were needed for the issue at hand.

PR #552, opened 2019-07-08
Fix issues #517, #518, and 546

### LINKED ISSUE #517: Label missing in unparsed output after semantics
```
  do 1 i=1,2
    goto 2
1 continue
2 do 3 i=1,2
3 continue
  end
```

loses the label `2` in the unparsed output if semantics is enabled.

jeanPerier:
Fixes #517, #518, and #546:
- Only allow out-of-scope label target for label-do-stmt when the label is on an end-stmt of a scope directly nested into the one of the label-do-stmt.
- Also transform label-do-stmt into non-label-do constructs in such out-of-scope cases so that later phases do not have to handle label-do-stmt.

This PR makes out-of-scope labels more restrictive, e.g, the following is not allowed anymore by label resolution:
```
do 1 i=1,10
 if (i>5) then
1  print*, i
 end if
end
```
but the following remains OK with a warning:
```
do 1 i=1,10
  if (i>5) then
    print*, i
1 end if
end
```
and is now transformed into
```
do  i=1,10
  if (i>5) then
    print*, i
1 end if
end do
end
```
This is done with if, do and select end statements.
This seems to be in line with what pgfortran and NAG are accepting (select type and rank end statements are also accepted, this is to avoid making resolve-labels.cc much more complex). 

Edit: fix for #517 added in later commit because it was small enough


lib/semantics/resolve-labels.cc
```diff
@@ -889,9 +907,16 @@ void CheckLabelDoConstraints(const SourceStmtList &dos,
...
     } else if (!InInclusiveScope(scopes, scope, doTarget.proxyForScope)) {
       // C1133
-      errorHandler.Say(position,
-          parser::MessageFormattedText{
-              "label '%u' is not in scope"_en_US, SayLabel(label)});
+      if (doTarget.legacyAcceptedInDirectSubscope &&
+          ParentScope(scopes, doTarget.proxyForScope) == scope) {
+        errorHandler.Say(doTarget.parserCharBlock,
+            parser::MessageFormattedText{
+                "A DO loop should terminate with an END DO or CONTINUE inside its scope"_en_US});
```
schweitzpgi: Can this case be merged with the succeeding `else if` case rather than replicating the message?

lib/semantics/resolve-labels.cc
```diff
@@ -209,6 +210,14 @@ const parser::CharBlock *GetStmtName(const parser::Statement<A> &stmt) {
   return nullptr;
 }
 
+template<typename A>
+static constexpr bool LabelDoTargetAcceptedInDirectSubscopeOn(
+    const parser::Statement<A> &statement) {
+  return (std::is_same_v<A, parser::EndIfStmt> ||
```
klausler: Needless parentheses; could also be written by defining a `std::tuple<>` of these types and using `common::HasMember`.

lib/semantics/canonicalize-do.cc
```diff
@@ -33,6 +33,30 @@ class CanonicalizationOfDoLoops {
...
+                          caseConstruct.value().t));
+                },
+                [&](common::Indirection<SelectRankConstruct> &selectRank) {
+                  CanonicalizeIfMatch(block, stack, i,
+                      std::get<Statement<EndSelectStmt>>(selectRank.value().t));
+                },
+                [&](common::Indirection<SelectTypeConstruct> &selectType) {
+                  CanonicalizeIfMatch(block, stack, i,
+                      std::get<Statement<EndSelectStmt>>(selectType.value().t));
+                },
```
klausler: Also `associate`, `block`, `forall`, `where`. and `critical`, maybe more.

lib/semantics/resolve-labels.cc
jeanPerier: Thanks, it saves some lines and the merged conditions remain readable.

lib/semantics/resolve-labels.cc
jeanPerier: Thanks, that makes even more sens now that there are more `End..Stmt` to handle.

lib/semantics/canonicalize-do.cc
jeanPerier: OK, I only took the f95 executable constructs (+ `select type` and `select rank`, because they were "dragged" by `select case` support for simplicity), the rest was rejected by resolve-labels.cc. However, it seems more consistent to accept it on all executable construct end statements (+ forall and where end statements) (pgfortran does so for all the ones it supports).

I have added the suggested plus `change team` and renamed the functions/variables accordingly.

I also added a non-standard feature to control warning on label do loops terminating at statements different than `end do` or `continue`. I can revert this last commit if you think it is not justified.

test/semantics/label14.f90
```diff
@@ -13,22 +13,121 @@
...
+
+subroutine foo0()
   do 1 j=1,2
     if (.true.) then
+! CHECK: A DO loop should terminate with an END DO or CONTINUE
 1   end if
   do 2 k=1,2
     do i=3,4
       print*, i+k
+! CHECK: A DO loop should terminate with an END DO or CONTINUE
```
tskeith: I think this test will pass if the message comes out once anywhere in the source. It's not checking that the error is emitted for each place the CHECK comment appears.

It would be a better test if it were split into separate source files, each with one error. Or better yet, switch to using test_errors.sh.

lib/semantics/canonicalize-do.cc
```diff
@@ -34,16 +34,34 @@ class CanonicalizationOfDoLoops {
             common::visitors{
                 [](auto &) {},
                 // Labels on end-stmt of constructs are accepted by f18 for
```
tskeith: The "for" at the end of this line is left over from the previous version of the comment and should be removed.

jeanPerier: Thanks, fixed.

test/semantics/label14.f90
jeanPerier: Thanks, it was indeed not testing as I intended.
I tried using `test_errors.sh` with `-Mstandard -Werror` but that did not work because `-Werror` does not add the `error:` prefix to warnings (which I think is correct, but annoying in this case).
So I went the other way you suggested and split the tests into different sources. To make it more valuable, I also added checks to test if the label-do were rewritten as non-label-do.

PR #554, opened 2019-07-09
2.8.1 SIMD structural checks

ichoyjx:
Straightforward implementation but not including:
* list-item on aligned clause
* nesting check with the ordered construct (parse tree changes needed!)
* list-item attributes check on aligned clause

### DIFF
--- a/lib/semantics/check-omp-structure.cc
+++ b/lib/semantics/check-omp-structure.cc
@@ -216,9 +216,8 @@ void OmpStructureChecker::Leave(const parser::OmpClauseList &) {
           if (const auto safelenValue{GetIntValue(safelenClause.v)}) {
             if (*safelenValue > 0 && *simdlenValue > *safelenValue) {
               context_.Say(clause->source,
-                  "The parameter of the SIMDLEN clause must be "
-                  "less than or equal to "
-                  "the parameter of the SAFELEN clause"_err_en_US);
+                  "The parameter of the SIMDLEN clause must be less than or "
+                  "equal to the parameter of the SAFELEN clause"_err_en_US);
             }
           }
         }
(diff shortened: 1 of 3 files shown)


lib/semantics/check-omp-structure.cc
```diff
@@ -177,7 +201,32 @@ void OmpStructureChecker::Leave(const parser::OmpClauseList &) {
...
+        const auto &safelenClause{
+            std::get<parser::OmpClause::Safelen>(clause2->u)};
+        // simdlen and safelen both have parameters
+        if (const auto simdlenValue{GetIntValue(simdlenClause.v)}) {
+          if (const auto safelenValue{GetIntValue(safelenClause.v)}) {
+            if (*safelenValue > 0 && *simdlenValue > *safelenValue) {
+              context_.Say(clause->source,
+                  "The parameter of the SIMDLEN clause must be "
+                  "less than or equal to "
+                  "the parameter of the SAFELEN clause"_err_en_US);
```
tskeith: You can probably format this message into two source lines.

ichoyjx: Updated. Thanks.

lib/semantics/check-omp-structure.cc
```diff
@@ -177,7 +201,31 @@ void OmpStructureChecker::Leave(const parser::OmpClauseList &) {
 
       // TODO: ordered region binding check (requires nesting implementation)
     }
-  }
+  }  // DO
+
+  // 2.8.1 Simd Construct Restriction
+  if (GetContext().directive == OmpDirective::SIMD) {
+    if (auto *clause{FindClause(OmpClause::SIMDLEN)}) {
```
sscalpone: Here you are checking to see if you have two clauses and fetching their integer values.  Is this going to be a common pattern?  If so, you might want to create function that does the same & returns a tuple of two optionals (or whatever).

ichoyjx: > Here you are checking to see if you have two clauses and fetching their integer values. Is this going to be a common pattern? 
For this pattern, I think we've dealt with all the cases based on the spec.

PR #558, opened 2019-07-10
Examples of Composite and Combined constructs

ichoyjx:
2.8.3 Loop SIMD Construct
2.11.1 Parallel Loop Construct
2.11.3 Parallel Loop SIMD Construct

Not in the implementation:
2.8.3 DO SIMD: A list item may appear in a linear or firstprivate clause but not both

### DIFF
--- a/lib/semantics/check-omp-structure.cc
+++ b/lib/semantics/check-omp-structure.cc
@@ -18,6 +18,12 @@
 
 namespace Fortran::semantics {
 
+std::string OmpStructureChecker::ContextDirectiveAsFortran() {
+  auto dir{EnumToString(GetContext().directive)};
+  std::replace(dir.begin(), dir.end(), '_', ' ');
+  return dir;
+}
+
 bool OmpStructureChecker::HasInvalidWorksharingNesting(
     const parser::CharBlock &source, const OmpDirectiveSet &set) {
   // set contains all the invalid closely nested directives
@@ -199,10 +205,7 @@ void OmpStructureChecker::Enter(
 
 void OmpStructureChecker::Leave(const parser::OmpClauseList &) {
   // 2.7 Loop Construct Restriction
-  if (GetContext().directive == OmpDirective::DO ||
-      GetContext().directive == OmpDirective::PARALLEL_DO ||
-      GetContext().directive == OmpDirective::DO_SIMD ||
-      GetContext().directive == OmpDirective::PARALLEL_DO_SIMD) {
+  if (doSet.test(GetContext().directive)) {
     if (auto *clause{FindClause(OmpClause::SCHEDULE)}) {
       // only one schedule clause is allowed
       const auto &schedClause{std::get<parser::OmpScheduleClause>(clause->u)};
@@ -253,12 +256,10 @@ void OmpStructureChecker::Leave(const parser::OmpClauseList &) {
 
       // TODO: ordered region binding check (requires nesting implementation)
     }
-  }  // DO
+  }  // doSet
 
   // 2.8.1 Simd Construct Restriction
-  if (GetContext().directive == OmpDirective::SIMD ||
-      GetContext().directive == OmpDirective::DO_SIMD ||
-      GetContext().directive == OmpDirective::PARALLEL_DO_SIMD) {
+  if (simdSet.test(GetContext().directive)) {
     if (auto *clause{FindClause(OmpClause::SIMDLEN)}) {
       if (auto *clause2{FindClause(OmpClause::SAFELEN)}) {
         const auto &simdlenClause{
@@ -375,14 +376,11 @@ void OmpStructureChecker::Enter(const parser::OmpClause::Ordered &x) {
     }
 
     // 2.8.3 Loop SIMD Construct Restriction
-    if (GetContext().directive == OmpDirective::DO_SIMD ||
-        GetContext().directive == OmpDirective::PARALLEL_DO_SIMD) {
-      auto dir{EnumToString(GetContext().directive)};
-      std::replace(dir.begin(), dir.end(), '_', ' ');
+    if (doSimdSet.test(GetContext().directive)) {
       context_.Say(GetContext().clauseSource,
           "No ORDERED clause with a parameter can be specified "
           "on the %s directive"_err_en_US,
-          dir);
+          ContextDirectiveAsFortran());
     }
   }
 }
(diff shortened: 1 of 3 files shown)


lib/semantics/check-omp-structure.cc
```diff
@@ -148,9 +164,45 @@ void OmpStructureChecker::Enter(const parser::OmpLoopDirective::Simd &) {
...
+  OmpClauseSet allowedOnce{OmpClause::IF, OmpClause::NUM_THREADS,
+      OmpClause::PROC_BIND, OmpClause::SCHEDULE, OmpClause::COLLAPSE,
+      OmpClause::ORDERED, OmpClause::SAFELEN, OmpClause::SIMDLEN};
+  SetContextAllowedOnce(allowedOnce);
+}
+
 void OmpStructureChecker::Leave(const parser::OmpClauseList &) {
   // 2.7 Loop Construct Restriction
-  if (GetContext().directive == OmpDirective::DO) {
+  if (GetContext().directive == OmpDirective::DO ||
```
vjayathirtha-nv: I wonder if it would be better (cleaner to read) to use a `switch` statement here instead of `if` here?
Even I'm not sure, anyone else have opinions on this?
```suggestion
  switch(GetContext().directve) {
  case OmpDirective::DO:
  case OmpDirective::PARALLEL_DO:
  case OmpDirective::DO_SIMD:
  case OmpDirective::PARALLEL_DO_SIMD:
    <code block>
    break;

  case OmpDirective::SIMD:
  case OmpDirective::DO_SIMD:
  case OmpDirective::PARALLEL_DO_SIMD:
    <code block>
    break;
  }
```

lib/semantics/check-omp-structure.cc
```diff
@@ -204,7 +256,9 @@ void OmpStructureChecker::Leave(const parser::OmpClauseList &) {
   }  // DO
```
vjayathirtha-nv: If you end up not taking the `switch` route, update this comment.

lib/semantics/check-omp-structure.cc
```diff
@@ -148,9 +164,45 @@ void OmpStructureChecker::Enter(const parser::OmpLoopDirective::Simd &) {
...
+  SetContextAllowedOnce(allowedOnce);
+}
+
 void OmpStructureChecker::Leave(const parser::OmpClauseList &) {
   // 2.7 Loop Construct Restriction
-  if (GetContext().directive == OmpDirective::DO) {
+  if (GetContext().directive == OmpDirective::DO ||
+      GetContext().directive == OmpDirective::PARALLEL_DO ||
+      GetContext().directive == OmpDirective::DO_SIMD ||
+      GetContext().directive == OmpDirective::PARALLEL_DO_SIMD) {
```
tskeith: You need to do this a lot (i.e. check if the current directive is one of a set), so it might be worth writing a function for that.  Something like this:
```
bool IsContextDirective(const std::initializer_list<OmpDirective> &directives) {
  auto it{std::find(directives.begin(), directives.end(), GetContext().directive)};
  return it != directives.end();
}
```
Then use it like this:
```
if (IsContextDirective({OmpDirective::DO, OmpDirective::PARALLEL_DO,
        OmpDirective::DO_SIMD, OmpDirective::PARALLEL_DO_SIMD})) {
```

Maybe name it `IsContextDirectiveOneOf` if you want to be more explicit.

klausler: There's also `lib/common/enum-set.h` which provides APIs like `set` and `bitset` to sets of enums.

lib/semantics/check-omp-structure.cc
```diff
@@ -319,6 +373,17 @@ void OmpStructureChecker::Enter(const parser::OmpClause::Ordered &x) {
             "a constant positive integer expression"_err_en_US);
       }
     }
+
+    // 2.8.3 Loop SIMD Construct Restriction
+    if (GetContext().directive == OmpDirective::DO_SIMD ||
+        GetContext().directive == OmpDirective::PARALLEL_DO_SIMD) {
+      auto dir{EnumToString(GetContext().directive)};
+      std::replace(dir.begin(), dir.end(), '_', ' ');
```
tskeith: You should extract this into a function, in part so that the reader can tell why you're doing this string manipulation (either from the name of the function or a comment on it).

We use `AsFortran` for functions that return a representation of something as it can appear in Fortran source, so that seems appropriate here.  Maybe `std::string AsFortran(OmpDirective)` or `std::string ContextDirectiveAsFortran()`.

lib/semantics/check-omp-structure.cc
klausler: I think that the `switch` is more clear.

lib/semantics/check-omp-structure.cc
tskeith: The `EnumSet` approach is especially good if you want to give the sets names so that you can use them in more than one place. For example, the set of DO, PARALLEL_DO, DO_SIMD, and PARALLEL_DO_SIMD is being used twice currectly. If there is a meaningful name you can give this set, you can define it once and use it in both places.

Also, there is already a name for this enum set type, so a set can be defined as:
`static constexpr OmpDirectiveSet thisSet{...};`

ichoyjx: Thank you all for the very detailed suggestions!

lib/semantics/check-omp-structure.cc
ichoyjx: Addressed in update commit. Thanks.

lib/semantics/check-omp-structure.cc
ichoyjx: Addressed in update commit.

lib/semantics/check-omp-structure.cc
ichoyjx: Updated with `EnumSet` instead.

lib/semantics/check-omp-structure.h
```diff
@@ -42,6 +42,14 @@ ENUM_CLASS(OmpDirective, PARALLEL, DO, SECTIONS, SECTION, SINGLE, WORKSHARE,
...
 using OmpDirectiveSet = common::EnumSet<OmpDirective, OmpDirective_enumSize>;
 
+static constexpr OmpDirectiveSet doSet{OmpDirective::DO,
+    OmpDirective::PARALLEL_DO, OmpDirective::DO_SIMD,
+    OmpDirective::PARALLEL_DO_SIMD};
+static constexpr OmpDirectiveSet simdSet{
+    OmpDirective::SIMD, OmpDirective::DO_SIMD, OmpDirective::PARALLEL_DO_SIMD};
+static constexpr OmpDirectiveSet doSimdSet{
+    OmpDirective::DO_SIMD, OmpDirective::PARALLEL_DO_SIMD};
+
```
tskeith: These could probably go in the .cc file.

ichoyjx: Okay.

PR #559, opened 2019-07-11
Move some functions to semantics/tools.h

tskeith:
Move some member functions out of symbols.h etc. and into tools.h. This simplifies things a little and reduces dependencies of the symbol table on the rest of semantics.


lib/semantics/resolve-names.cc
```diff
@@ -1344,19 +1345,20 @@ void ImplicitRulesVisitor::Post(const parser::ParameterStmt &x) {
...
-                            [&](const std::list<parser::ImplicitSpec> &x) {
-                              if (prevImplicitNoneType_) {
-                                Say("IMPLICIT statement after IMPLICIT NONE or "
-                                    "IMPLICIT NONE(TYPE) statement"_err_en_US);
-                                return false;
-                              }
-                              return true;
-                            },
-                        },
+  bool res = std::visit(
```
klausler: Could use {braces}.

PR #561, opened 2019-07-11
Better name resolution for intrinsics

klausler:
Eliminate `semantics::MiscDetails::Kind::SpecificIntrinsic` in favor of representing intrinsic procedures in the symbol table as `ProcEntity`(ie)s with the `INTRINSIC` attr set.  Ensure that `EXTERNAL abs` masks the intrinsic.


klausler:
Rebased to current master to eliminate all the formatting differences, which obscured the real changes here.

PR #565, opened 2019-07-11
When creating symbols for intrinsics, do not put them in the global s…

klausler:
…cope, and do not emit them to module files.

### DIFF
--- a/lib/semantics/resolve-names.cc
+++ b/lib/semantics/resolve-names.cc
@@ -4767,11 +4767,12 @@ void ResolveNamesVisitor::HandleProcedureName(
   CHECK(flag == Symbol::Flag::Function || flag == Symbol::Flag::Subroutine);
   auto *symbol{FindSymbol(name)};
   if (symbol == nullptr) {
-    Attrs attrs;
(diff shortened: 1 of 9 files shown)


klausler:
Please take another look, I've added a fix for bug #571.

lib/semantics/resolve-names.cc
```diff
@@ -829,7 +829,7 @@ class DeclarationVisitor : public ArraySpecVisitor,
   Attrs HandleSaveName(const SourceName &, Attrs);
   void AddSaveName(std::set<SourceName> &, const SourceName &);
   void SetSaveAttr(Symbol &);
-  bool HandleUnrestrictedSpecificIntrinsicFunction(const parser::Name &);
+  Symbol *HandleUnrestrictedSpecificIntrinsicFunction(const parser::Name &);
```
tskeith: I don't understand changing the return type. It looks like `NameIsKnownOrIntrinsic` and `HandleUnrestrictedSpecificIntrinsicFunction` are still only used in a boolean context.

klausler: I'll change it back then.

PR #567, opened 2019-07-12
Use passed-object dummy in distinguishability checks

tskeith:
Complete the checks for distinguishable specifics procedure in a generic
by considering any passed-object dummy arguments.

C1514 rule 3 is implemented and the checks for the other rules are
extended to consider the PASS attribute, including the concept of the
"effective" position of an argument in an argument list, computed by
ignoring passed-object arguments.

Add `pass` to `characteristics::DummyArgument` to mark each
passed-object dummy argument.

Change symbols to store the index of the passed-object dummy argument
rather than its symbol.

Check that specifics of a type-bound generic are distinguishable only
after all of the procedure bindings have been processed. They don't have
to be before the generic.

### DIFF
--- a/lib/evaluate/characteristics.h
+++ b/lib/evaluate/characteristics.h
@@ -145,9 +145,10 @@ struct DummyArgument {
   bool IsOptional() const;
   void SetOptional(bool = true);
   std::ostream &Dump(std::ostream &) const;
-  // name is not a characteristic and so does not participate in operator==
-  // but it is needed to determine if procedures are distinguishable
+  // name and pass are not a characteristics and so does not participate in
+  // operator== but are needed to determine if procedures are distinguishable
   std::string name;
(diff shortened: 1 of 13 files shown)


lib/evaluate/characteristics.h
```diff
@@ -145,9 +145,10 @@ struct DummyArgument {
   bool IsOptional() const;
   void SetOptional(bool = true);
   std::ostream &Dump(std::ostream &) const;
-  // name is not a characteristic and so does not participate in operator==
-  // but it is needed to determine if procedures are distinguishable
+  // name and pass are not a characteristics and so does not participate in
```
vjayathirtha-nv: are not ~a~ characteristics

tskeith: Sorry, I missed this comment. I'll fix this in a future change.

PR #578, opened 2019-07-16
 Loop association between Loop Constructs and do-loops

ichoyjx:
```
!$omp parallel do
do i=1, N
  ...
enddo
[!$omp end parallel do]
```
Here LoopDirective and EndLoopDirective are not wrapped in the same class.
Ideally, it would be `{LoopDirective, ClauseList, DoConstruct, EndLoopDirective}`.
However, to allow different forms of do-loops, OpenMPLoopConstruct is actually
`{LoopDirective, ClauseList}` and the loop association checks are deferred to
Semantics.

In the example, do-loop needs to immediately follow the LoopDirective and
EndLoopDirective needs to immediately follow the ENDDO (`DoConstruct`). Here
the EndLoopdirective is optional because the scope of PARALLEL region is
determined by the scope of the associated do-loop. Unlike BlockDirective,
we pop the context upon the LEAVE of DoConstruct or on the point of failing
to find the associated do-loop.

Note that a new `checkState_` struct was created to hold temporary status.

Msg examples:
```
omp-clause-validity01.f90:220:15: error: do-loop is expected after the PARALLEL DO directive
          !$omp parallel do shared(b)
                ^^^^^^^^^^^
omp-clause-validity01.f90:224:12: error: The END PARALLEL DO must follow the do-loop associated with the loop construct
       !$omp end parallel do
             ^^^^^^^^^^^^^^^
```
Also added simd/ordered constructs example to test the implementation.

### DIFF
--- a/lib/semantics/check-omp-structure.cc
+++ b/lib/semantics/check-omp-structure.cc
@@ -66,7 +66,28 @@ void OmpStructureChecker::CheckAllowed(const OmpClause &type) {
 }
 
 void OmpStructureChecker::Enter(const parser::OpenMPConstruct &x) {
-  // 2.8.1 TODO: Simd Construct with Ordered Construct Nesting check
+  // 2.8.1 Simd Construct Restriction
+  if (CurrentDirectiveIsNested() &&
+      GetContext().directive == OmpDirective::SIMD) {
+    bool hasError{true};
+    if (const auto *block{std::get_if<parser::OpenMPBlockConstruct>(&x.u)}) {
+      const auto &dir{std::get<parser::OmpBlockDirective>(block->t)};
+      const auto &clauses{std::get<parser::OmpClauseList>(block->t)};
+      const auto &list{clauses.v};
+      if (dir.v == parser::OmpBlockDirective::Directive::Ordered &&
+          list.size() == 1 &&
+          std::holds_alternative<parser::OmpClause::Simd>(list.front().u)) {
+        // Simd and Threads clauses cannot be both present
+        hasError = false;
+      }
+    }
+    if (hasError) {
+      context_.Say(GetContext().directiveSource,
+          "An ORDERED construct with the SIMD clause is "
+          "the only OpenMP construct that can be "
+          "encountered during execution of a SIMD region"_err_en_US);
+    }
+  }  // SIMD
 }
 
 void OmpStructureChecker::Enter(const parser::OpenMPLoopConstruct &x) {
--- a/lib/semantics/check-omp-structure.cc
+++ b/lib/semantics/check-omp-structure.cc
@@ -82,10 +82,55 @@ void OmpStructureChecker::Enter(const parser::OpenMPLoopConstruct &x) {
   // push a context even in the error case
   OmpContext ct{dir.source};
   ompContext_.push_back(ct);
+  checkState_.justSawLoopConstruct = true;
 }
 
-void OmpStructureChecker::Leave(const parser::OpenMPLoopConstruct &) {
-  ompContext_.pop_back();
+void OmpStructureChecker::Enter(const parser::ExecutableConstruct &x) {
+  if (checkState_.justSawLoopConstruct) {
+    if (const auto *doDir{
+            std::get_if<common::Indirection<parser::DoConstruct>>(&x.u)}) {
+      // save associated DoConstruct pointer
+      SetContextDo(doDir->value());
+      checkState_.inOmpLoopRegion = true;
+    } else {
+      context_.Say(GetContext().directiveSource,
+          "do-loop is expected after the %s directive"_err_en_US,
+          ContextDirectiveAsFortran());
+      ompContext_.pop_back();
+    }
+    checkState_.justSawLoopConstruct = false;
+  }
+
+  const auto *endDoDir{
+      std::get_if<common::Indirection<parser::OpenMPEndLoopDirective>>(&x.u)};
+  if (!endDoDir && checkState_.allowOptionalEndDirective) {
+    // END directive is not present, loop construct is complete
+    checkState_.allowOptionalEndDirective = false;
+  }
+}
+
+void OmpStructureChecker::Leave(const parser::ExecutableConstruct &x) {
+  if (const auto *doDir{
+          std::get_if<common::Indirection<parser::DoConstruct>>(&x.u)}) {
+    if (checkState_.inOmpLoopRegion) {
+      if (&doDir->value() == GetContext().doDirective) {
+        checkState_.inOmpLoopRegion = false;
+        checkState_.allowOptionalEndDirective = true;
+        // pop at the end of associated do-loop
+        ompContext_.pop_back();
+      }
+    }
+  }
+}
+
+void OmpStructureChecker::Enter(const parser::OpenMPEndLoopDirective &x) {
+  checkState_.inEndDirective = true;
+  if (!checkState_.allowOptionalEndDirective) {
+    context_.Say(x.source,
+        "The %s must follow the do-loop associated with the "
+        "loop construct"_err_en_US,
+        parser::ToUpperCaseLetters(x.source.ToString()));
+  }
 }
 
 void OmpStructureChecker::Enter(const parser::OpenMPBlockConstruct &x) {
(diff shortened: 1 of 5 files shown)


lib/semantics/check-omp-structure.cc
```diff
@@ -82,10 +82,55 @@ void OmpStructureChecker::Enter(const parser::OpenMPLoopConstruct &x) {
...
+      std::get_if<common::Indirection<parser::OpenMPEndLoopDirective>>(&x.u)};
+  if (!endDoDir && checkState_.allowOptionalEndDirective) {
+    // END directive is not present, loop construct is complete
+    checkState_.allowOptionalEndDirective = false;
+  }
 }
 
-void OmpStructureChecker::Leave(const parser::OpenMPLoopConstruct &) {
-  ompContext_.pop_back();
+void OmpStructureChecker::Leave(const parser::ExecutableConstruct &x) {
```
tskeith: Does this work if you see an `ExecutionPartConstruct` that is not an `ExecutableConstruct`?
For example, what if a loop directive is followed by a FORMAT statement?

lib/semantics/check-omp-structure.cc
```diff
@@ -82,10 +82,55 @@ void OmpStructureChecker::Enter(const parser::OpenMPLoopConstruct &x) {
...
+    }
+  }
+}
+
+void OmpStructureChecker::Enter(const parser::OpenMPEndLoopDirective &x) {
+  checkState_.inEndDirective = true;
+  if (!checkState_.allowOptionalEndDirective) {
+    context_.Say(x.source,
+        "The %s must follow the do-loop associated with the "
+        "loop construct"_err_en_US,
```
tskeith: There are some messages in `resolve-labels.cc` that say "DO loop" rather than "do-loop". We should be consistent.

sscalpone:
Is there an advantage to treating these cases like do-stmt and do-construct, where do-stmt is re-written to be like do-construct.  Note that multiple do-loops are allowed in a open mp do:

!$omp do [clause[[,]clause]...]
do-loops
[!$omp end do [nowait]]

I did not check to see if the grammar is already written to handle multiple do-loops.

test/semantics/omp-clause-validity01.f90
```diff
@@ -206,6 +206,26 @@
...
+        !ERROR: do-loop is expected after the PARALLEL DO directive
+        !$omp parallel do shared(b)
+        a = 3.14
+     enddo
+     !ERROR: The END PARALLEL DO must follow the do-loop associated with the loop construct
+     !$omp end parallel do
+  enddo
+  a = 1.414
+  !ERROR: The END PARALLEL DO must follow the do-loop associated with the loop construct
+  !$omp end parallel do
```
tskeith: PARALLEL do should work on a labeled DO loop right? You should have a test for that.

Also, what happens if the directive is on a DO CONCURRENT?

lib/semantics/check-omp-structure.cc
```diff
@@ -82,10 +103,55 @@ void OmpStructureChecker::Enter(const parser::OpenMPLoopConstruct &x) {
...
+    if (checkState_.inOmpLoopRegion) {
+      if (&doDir->value() == GetContext().doDirective) {
+        checkState_.inOmpLoopRegion = false;
+        checkState_.allowOptionalEndDirective = true;
+        // pop at the end of associated do-loop
+        ompContext_.pop_back();
+      }
+    }
+  }
+}
```
jeanPerier: You may want have a return after `ompContext_.pop_back();` and add a `checkState_.allowOptionalEndDirective=false` in the fallback path to avoid accepting the following program that have leave other constructs after the parallel do:

```
  N = 1024
  do i = 1, 10
    !$omp parallel do copyin(a)
    do j = 2*i*N, (2*i+1)*N
    enddo
  enddo
  !$omp end parallel do
end
```
Note that the outer `do` could be an `if` or whatever `ExecutableConstruct` and the `!$omp end parallel do` would also be currently accepted.

lib/semantics/check-omp-structure.cc
```diff
@@ -82,10 +103,55 @@ void OmpStructureChecker::Enter(const parser::OpenMPLoopConstruct &x) {
...
+
+void OmpStructureChecker::Enter(const parser::OpenMPEndLoopDirective &x) {
```
jeanPerier: Should you also check that `OpenMPEndLoopDirective` macthes the `OmpLoopDirective` or will this be part of a future PR ?
Currently, the following program is accepted which seems wrong:
```
  N = 1024
  !$omp parallel do
  do i = 1, N
  enddo
  !$omp end simd
end
```

ichoyjx: Yes! I forgot to mention, directive matching will be in a separate PR. Thanks!

lib/semantics/check-omp-structure.cc
ichoyjx: Sorry, I should've noticed this, I will fix it.

test/semantics/omp-clause-validity01.f90
ichoyjx: Good question. I thought DO CONCURRENT was regular DoConstruct. Let me try.

ichoyjx: Yep, DO CONCURRENT is DoConstruct in the parse tree. It works with OpenMP.

sscalpone: Is DO CONCURRENT documented in the OpenMP spec?

vdonaldson: You might want to be selective about allowing a DO CONCURRENT loop as a OpenMP loop construct.  The Fortran standard knows nothing about OpenMP.  OpenMP knows/says nothing about DO CONCURRENT.  That is because no one has figured out what to do with this combination.  One option is to disallow OpenMP directives on a DO CONCURRENT loop.  Another is to selectively allow some simple cases.  Another is to take the time to allow as many cases as possible.

vdonaldson: pgi fortran currently implements the 2nd of these three choices.

ichoyjx: OpenMP 4.5 spec doesn't mention anything about DO CONCURRENT, yet 5.0 spec says:

> This OpenMP API specification refers to ISO/IEC 1539-1:2010 as Fortran 2008. While future
> versions of the OpenMP specification are expected to address the following features, currently
> their use may result in unspecified behavior.
>   – Submodules
>   – Coarrays
>   – DO CONCURRENT
>   ...

GNU doesn't allow OpenMP + DO CONCURRENT but our PGI compiler does.

ichoyjx: > pgi fortran currently implements the 2nd of these three choices.

Good to know.

vdonaldson: To be more specific,  flang / pgi fortran has code that does DO CONCURRENT LOCAL_INIT assignments and mask expressions, regardless of OpenMP.  In the absence of compiler parallelization of a DO CONCURRENT, OpenMP processing will otherwise ignore locality-spec information.  If there are no locality-spec clauses,  a DO CONCURRENT loop will look like a loop nest to OpenMP, and OpenMP clauses like COLLAPSE should work.  There is no attempt to treat locality-spec clause as PRIVATE, FIRSTPRIVATE, or SHARED clauses.

lib/semantics/check-omp-structure.cc
ichoyjx: This is a nice catch, I forget that the `Leave` of the outer `DoConstrcut` follows the `Leave` of the inner `DoConstruct`.

lib/semantics/check-omp-structure.cc
```diff
@@ -82,10 +103,55 @@ void OmpStructureChecker::Enter(const parser::OpenMPLoopConstruct &x) {
...
 
-void OmpStructureChecker::Leave(const parser::OpenMPLoopConstruct &) {
-  ompContext_.pop_back();
+void OmpStructureChecker::Leave(const parser::ExecutableConstruct &x) {
+  if (const auto *doDir{
+          std::get_if<common::Indirection<parser::DoConstruct>>(&x.u)}) {
+    if (checkState_.inOmpLoopRegion) {
+      if (&doDir->value() == GetContext().doDirective) {
+        checkState_.inOmpLoopRegion = false;
+        checkState_.allowOptionalEndDirective = true;
```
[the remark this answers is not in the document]
ichoyjx: This requires a major change (the way to do the loop-association), I am working on that.

Issue #579, opened 2019-07-16
Bogus "derived type not found" error


klausler:
```
module m1
  private
  public :: t1
  type :: t1
  end type t1
end module m1
module m2
  use m1
  type, extends(t1) :: t2
  end type t2
end module m2
program main
  use m2
  type(t2) :: x
end
```

elicits an error about the derived type `t2` not being found at the declaration of `x`.

klausler:
It would seem that we don't reset the default accessibility back to `PUBLIC` when beginning semantics on a new module; if so, this will be an easy fix.

Issue #571, opened 2019-07-15
Forward references to interfaces are not accepted by f18


klausler:
```
module m
  procedure(iface), pointer :: pp
  abstract interface
    subroutine iface
    end subroutine
  end interface
end module
```

PR #580, opened 2019-07-16
Allow forward references to procedure interfaces in derived types (fi…

klausler:
…xing #571 more)

An earlier change fixed up forward references in `procedure()` interfaces outside derived types.  This change completes the job so that forward references can also be used within derived types.

### DIFF
--- a/lib/semantics/resolve-names.cc
+++ b/lib/semantics/resolve-names.cc
@@ -774,6 +774,7 @@ protected:
   const parser::Name *ResolveVariable(const parser::Variable &);
   const parser::Name *ResolveName(const parser::Name &);
   bool PassesSharedLocalityChecks(const parser::Name &name, Symbol &symbol);
+  Symbol *NoteInterfaceName(const parser::Name &);
   void CheckExplicitInterface(Symbol &);
 
 private:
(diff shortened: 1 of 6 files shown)


lib/semantics/resolve-names.cc
```diff
@@ -3907,32 +3902,27 @@ bool DeclarationVisitor::CanBeTypeBoundProc(const Symbol &symbol) {
...
-        " explicit interface"_err_en_US);
-    symbol = nullptr;
+void DeclarationVisitor::CheckExplicitInterface(Symbol &symbol) {
+  if (const Symbol * interface{FindInterface(symbol)}) {
+    const Symbol *subp{FindSubprogram(*interface)};
+    if (subp == nullptr || !subp->HasExplicitInterface()) {
+      Say(symbol.name(),
+          "The interface of '%s' (%s) is not an abstract interface or a "
+          "procedure with an explicit interface"_err_en_US,
+          symbol.name(), interface->name());
```
tskeith: I think the interface name should be in quotes too, since it's a name from the user program.

klausler:
Also fixes #579.

Issue #582, opened 2019-07-16
Bogus error about bad kind type parameters on sourced allocation w/ array constructor


klausler:
```
integer :: i, n
integer, allocatable :: a(:)
n = 2
allocate(a(n),source=[(i,i=0,n-1)])
end
```

klausler:
Works if `a` is `INTEGER*8`; probably due to automatic use of `INTEGER*8` for loop controls.

jeanPerier:
**Expected behavior from the standard** :
`[(i,i=0,n-1)]` expression is of kind 8 inside f18: `integer, parameter :: k= kind([(i,i=0,n-1)])` is folded to 8. Hence, allocate semantics reports an error as per C946.
> If source-expr appears, the kind type parameters of each allocate-object shall have the same values as the corresponding type parameters of source-expr.

From 7.8 point 2, `[(i,i=0,n-1)]` should have the same kind as expression `i` which is the one of the ac-do-variable. As per 19.4 point 5, this variable should in this case be of default integer type (which is not kind 8 in this case):
> it has the type and type parameters that it would have if it were the name of a variable in the innermost executable construct or scoping unit that includes the DATA statement or array constructor, and this type shall be integer type.

**Root cause**:
In lib/semantics/expression.cc, the kind of implicit `ac-do-variable` is set to compile-time constant `ImpliedDoIndex::Result` which is `SubscriptInteger` (8). This is not what is mandated by 19.4 point 5.

See: https://github.com/flang-compiler/f18/blob/38789942a74f83d38c1132e96708158738ae0784/lib/semantics/expression.cc#L1082

Also, this conflicts with the symbol table where the ac-do-variable was previously added with kind 4 ( -fdebug-dump-symbols output is `ImpliedDos scope: i (Implicit): ObjectEntity type: INTEGER(4)`). See: https://github.com/flang-compiler/f18/blob/8c63aa94972fb3f7b88f23fa1b78f8b67ffa8732/lib/semantics/resolve-names.cc#L4153

**Different proposed solutions:**

1. Follow 19.4 point 5 and make implicit ac-do-variable follow the implicit rules.
2. Relax C946 and allow type conversions in source allocation in certain scenarios (e.g. same rule as in assignment ?).
3. Leave as it is and forbid such usage (user has to write `allocate(a(n),source=[(i,integer(4)::i=0,n-1)]`)).

[...]

PR #583, opened 2019-07-17
Fix issue #582: implicit ac-do-variable kind in expressions

jeanPerier:
Apply first proposed solution for issue #582: use previously resolved type of `ac-do-variable` from the symbol table when analyzing array constructor with `AcImpliedDo`. The symbol type resolution is compliant with 19.4 point 5 requirements.

### DIFF
--- a/lib/semantics/expression.cc
+++ b/lib/semantics/expression.cc
@@ -1079,10 +1079,10 @@ void ArrayConstructorContext::Add(const parser::AcValue &x) {
                 std::get<parser::AcImpliedDoControl::Bounds>(control.t)};
             Analyze(bounds.name);
             parser::CharBlock name{bounds.name.thing.thing.source};
+            const Symbol *symbol{bounds.name.thing.thing.symbol};
             int kind{IntType::kind};
-            if (auto &its{std::get<std::optional<parser::IntegerTypeSpec>>(
-                    control.t)}) {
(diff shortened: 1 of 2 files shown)


lib/semantics/expression.cc
```diff
@@ -1079,10 +1079,10 @@ void ArrayConstructorContext::Add(const parser::AcValue &x) {
                 std::get<parser::AcImpliedDoControl::Bounds>(control.t)};
             Analyze(bounds.name);
             parser::CharBlock name{bounds.name.thing.thing.source};
+            const Symbol *symbol{bounds.name.thing.thing.symbol};
             int kind{IntType::kind};
-            if (auto &its{std::get<std::optional<parser::IntegerTypeSpec>>(
```
klausler: If an explicit integer type appears in the parse tree, it should be used rather than applying implicit typing to the symbol, yes?

jeanPerier: Yes, but the symbol type resolution in:
https://github.com/flang-compiler/f18/blob/8c63aa94972fb3f7b88f23fa1b78f8b67ffa8732/lib/semantics/resolve-names.cc#L4153
is already doing all this, so I do not see a good reason to re-do it here.

The last line I added in the test file is actually checking that explicit integer types are still applied in here.

PR #585, opened 2019-07-17
[OpenMP] Add Sections and Single Construct check

ichoyjx:
Parse tree for OmpEndSingle needs to be modified to save the provenance of END SINGLE directive and check its own clauses

### DIFF
--- a/lib/semantics/check-omp-structure.cc
+++ b/lib/semantics/check-omp-structure.cc
@@ -98,6 +98,72 @@ void OmpStructureChecker::Leave(const parser::OpenMPBlockConstruct &) {
   ompContext_.pop_back();
 }
 
+void OmpStructureChecker::Enter(const parser::OpenMPSectionsConstruct &x) {
+  const auto &dir{std::get<parser::Verbatim>(x.t)};
+  OmpContext ct{dir.source};
+  ompContext_.push_back(ct);
+  SetContextDirectiveEnum(OmpDirective::SECTIONS);
+  OmpClauseSet allowed{OmpClause::PRIVATE, OmpClause::FIRSTPRIVATE,
+      OmpClause::LASTPRIVATE, OmpClause::REDUCTION};
+  SetContextAllowed(allowed);
+}
+
+void OmpStructureChecker::Leave(const parser::OpenMPSectionsConstruct &x) {
+  ompContext_.pop_back();
+}
+
+void OmpStructureChecker::Enter(const parser::OmpSection &x) {
+  const auto &dir{x.v};  // Verbatim
+  OmpContext ct{dir.source};
+  ompContext_.push_back(ct);
+  SetContextDirectiveEnum(OmpDirective::SECTION);
+
+  if (ompContext_.size() == 1) {
+    context_.Say(
+        dir.source, "Orphaned SECTION directive is prohibited"_err_en_US);
+  } else {
+    if (GetPrevContext().directive != OmpDirective::SECTIONS) {
+      context_.Say(dir.source,
+          "SECTION directive must appear within "
+          "the SECTIONS construct"_err_en_US);
+    }
+  }
+  // TODO: Section-Block association
+}
+
+void OmpStructureChecker::Leave(const parser::OmpSection &x) {
+  ompContext_.pop_back();
+}
+
+void OmpStructureChecker::Enter(const parser::OpenMPSingleConstruct &x) {
+  const auto &dir{std::get<parser::Verbatim>(x.t)};
+  OmpContext ct{dir.source};
+  ompContext_.push_back(ct);
+  SetContextDirectiveEnum(OmpDirective::SINGLE);
+  OmpClauseSet allowed{OmpClause::PRIVATE, OmpClause::FIRSTPRIVATE};
+  SetContextAllowed(allowed);
+}
+
+void OmpStructureChecker::Leave(const parser::OpenMPSingleConstruct &x) {
+  ompContext_.pop_back();
+}
+
+void OmpStructureChecker::Enter(const parser::OmpEndSingle &x) {
+  // EndSingle is in SingleConstruct context
+  const auto &dir{std::get<parser::Verbatim>(x.t)};
+  OmpContext ct{dir.source};
+  ompContext_.push_back(ct);
+  SetContextDirectiveEnum(OmpDirective::END_SINGLE);
+  OmpClauseSet allowed{OmpClause::COPYPRIVATE};
+  SetContextAllowed(allowed);
+  OmpClauseSet allowedOnce{OmpClause::NOWAIT};
+  SetContextAllowedOnce(allowedOnce);
+}
+
+void OmpStructureChecker::Leave(const parser::OmpEndSingle &x) {
+  ompContext_.pop_back();
+}
+
 void OmpStructureChecker::Enter(const parser::OmpBlockDirective &x) {
   switch (x.v) {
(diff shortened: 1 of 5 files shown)


lib/semantics/check-omp-structure.cc
```diff
@@ -98,6 +98,72 @@ void OmpStructureChecker::Leave(const parser::OpenMPBlockConstruct &) {
...
+void OmpStructureChecker::Enter(const parser::OmpSection &x) {
+  const auto &dir{x.v};  // Verbatim
+  OmpContext ct{dir.source};
+  ompContext_.push_back(ct);
+  SetContextDirectiveEnum(OmpDirective::SECTION);
+
+  if (ompContext_.size() == 1) {
+    context_.Say(
+        dir.source, "Orphaned SECTION directive is prohibited"_err_en_US);
+  } else {
```
vjayathirtha-nv: Better to use If-else-if chain I think.

lib/semantics/check-omp-structure.cc
```diff
@@ -98,6 +98,72 @@ void OmpStructureChecker::Leave(const parser::OpenMPBlockConstruct &) {
   ompContext_.pop_back();
 }
 
+void OmpStructureChecker::Enter(const parser::OpenMPSectionsConstruct &x) {
+  const auto &dir{std::get<parser::Verbatim>(x.t)};
+  OmpContext ct{dir.source};
+  ompContext_.push_back(ct);
+  SetContextDirectiveEnum(OmpDirective::SECTIONS);
```
tskeith: I suggest adding a `PushContext` function that allows you to replace these three lines with:
`PushContext(dir.source, OmpDirective::SECTIONS)`

For the places where you push without setting the directive, it would be just:
`PushContext(dir.source);`

lib/semantics/check-omp-structure.cc
```diff
@@ -98,6 +98,72 @@ void OmpStructureChecker::Leave(const parser::OpenMPBlockConstruct &) {
...
+
+void OmpStructureChecker::Enter(const parser::OmpSection &x) {
+  const auto &dir{x.v};  // Verbatim
+  OmpContext ct{dir.source};
+  ompContext_.push_back(ct);
+  SetContextDirectiveEnum(OmpDirective::SECTION);
+
+  if (ompContext_.size() == 1) {
+    context_.Say(
+        dir.source, "Orphaned SECTION directive is prohibited"_err_en_US);
```
tskeith: What does "orphaned" mean? Why isn't this just the same message as below, i.e. "SECTION directive must appear within the SECTIONS construct"?

lib/semantics/check-omp-structure.h
```diff
@@ -121,6 +132,7 @@ class OmpStructureChecker : public virtual BaseChecker {
   };
   // back() is the top of the stack
   const OmpContext &GetContext() const { return ompContext_.back(); }
+  const OmpContext &GetPrevContext() const { return ompContext_.rbegin()[1]; }
```
tskeith: There have to be at least 2 elements on the stack for this to be valid, right? Is that something the caller is supposed to ensure? If so, you should add a check that they are keeping their promise.

Same thing with `GetContext()`, I suppose. It may not be called when the stack is empty, right?

lib/semantics/check-omp-structure.cc
ichoyjx:
```
 !$omp section  <- Orphaned
  a = 0.0
  !$omp sections
  !$omp section
  a = 3.14
  !$omp end sections
```
I think you're right, the error msg applies to this case too. I will remove that.

lib/semantics/check-omp-structure.cc
```diff
@@ -98,6 +98,72 @@ void OmpStructureChecker::Leave(const parser::OpenMPBlockConstruct &) {
...
+
+void OmpStructureChecker::Leave(const parser::OpenMPSectionsConstruct &x) {
+  ompContext_.pop_back();
+}
+
+void OmpStructureChecker::Enter(const parser::OmpSection &x) {
+  const auto &dir{x.v};  // Verbatim
+  OmpContext ct{dir.source};
+  ompContext_.push_back(ct);
+  SetContextDirectiveEnum(OmpDirective::SECTION);
```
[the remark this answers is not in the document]
ichoyjx: Yes, `Block` will be the child of `OmpSection` and it is in the plan. And I agree with you on the pushing context part, I will do that the same as the nesting check, which is to check before pushing context. Thanks.

lib/semantics/check-omp-structure.cc
```diff
@@ -291,12 +357,26 @@ void OmpStructureChecker::Leave(const parser::OmpClauseList &) {
...
+    }
+  }
 }
 
 void OmpStructureChecker::Enter(const parser::OmpClause &x) {
   SetContextClause(x);
 }
 
+void OmpStructureChecker::Enter(const parser::OmpNowait &) {
+  CheckAllowed(OmpClause::NOWAIT);
```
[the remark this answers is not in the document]
ichoyjx: I added the restrictions in the `Enter` of `OmpNowait`. This helped to find a parser bug, thanks!

lib/semantics/check-omp-structure.cc
ichoyjx: Addressed by overloading functions.

lib/semantics/check-omp-structure.cc
ichoyjx: `GetPrevContext` is removed based on Jean's comment.

lib/semantics/check-omp-structure.h
ichoyjx: `CHECK` was added to enforce the empty check.

lib/semantics/check-omp-structure.h
```diff
@@ -131,8 +133,11 @@ class OmpStructureChecker : public virtual BaseChecker {
     std::multimap<OmpClause, const parser::OmpClause *> clauseInfo;
   };
   // back() is the top of the stack
-  const OmpContext &GetContext() const { return ompContext_.back(); }
-  const OmpContext &GetPrevContext() const { return ompContext_.rbegin()[1]; }
+  // TODO: remove CHECK after all directives/clauses are checked
```
tskeith: I think you should keep the CHECK (and remove this comment).

lib/semantics/check-omp-structure.h
```diff
@@ -131,8 +133,11 @@ class OmpStructureChecker : public virtual BaseChecker {
...
+  const OmpContext &GetContext() const {
+    CHECK(!ompContext_.empty());
+    return ompContext_.back();
+  }
   void SetContextDirectiveSource(const parser::CharBlock &directive) {
     ompContext_.back().directiveSource = directive;
```
tskeith: Use `GetContext()` here instead of `ompContext_.back()` so that it is checked here too.

ichoyjx: But here `GetContext` is returning a constant reference... Probably better to add `CHECK` for all the setters?

tskeith: It would be better to have a non-const overload of GetContext(), like this:
```
  OmpContext &GetContext() {
    CHECK(!ompContext_.empty());
    return ompContext_.back();
  }
  const OmpContext &GetContext() const {
    return const_cast<OmpStructureChecker *>(this)->GetContext();
  }
```

Or, if you never call GetContext from a const member function, you could omit the second one.

PR #596, opened 2019-07-19
Changes for constraint C1128

psteinfeld:
Specifically, these changes enforce the last sentence of the constraint, which
prohibits names that cannot appear in a variable definition context from
appearing in a locality-spec.  Here are the details.

 - Created the function "IsModifiableName" to return "true" when its parameter
   is the name of a variable that can appear in a variable definition context.
 - Created the function "GetAssociationRoot" to follow construct associations
   to potentially get to an underlying variable.  This function is similar to
   the existing "GetUltimate" function that follows use associations and host
   associations.  One difference is that "GetAssociationRoot" requires access
   to the types "MaybeExpr" and "SomeExpr", which makes is inappropriate to put
   into symbol.cc, which is where "GetUltimate" lives.  Perhaps we should move
   "GetUltimate" to tools.[h,cc].
 - Generalized the functions "IsPureFunction" to "IsPureProcedure" since either
   a pure function or subroutine can provide a context for variables that
   cannot be modified.  Changed "FindPureFunctionContaining" to
   "FindPureProcedueContaining" to go along with this.
 - Added the function "IsExternalInPureContext" to detect the case where a
   nominally pure procedure potentially modifies a variable.
 - Created the function "IsOrContainsEventOrLockComponent" to detect variables
   that either are of EVENT_TYPE or LOCK_TYPE or contain components of these
   types.  Such variables cannot appear in variable definition contexts.
 - Added the test resolve56.f90 to test most of these conditions.  Note that I
   only tested the new code from the perspective of locality-specs.

### DIFF
--- a/lib/semantics/expression.cc
+++ b/lib/semantics/expression.cc
@@ -1335,7 +1335,7 @@ MaybeExpr ExpressionAnalyzer::Analyze(
                 if (auto *msg{Say(expr.source,
                         "Externally visible object '%s' must not be "
                         "associated with pointer component '%s' in a "
-                        "PURE function"_err_en_US,
+                        "PURE procedure"_err_en_US,
                         object->name(), pointer->name())}) {
                   msg->Attach(object->name(), "Object declaration"_en_US)
                       .Attach(pointer->name(), "Pointer declaration"_en_US);
--- a/lib/semantics/expression.cc
+++ b/lib/semantics/expression.cc
@@ -1325,13 +1325,13 @@ MaybeExpr ExpressionAnalyzer::Analyze(
         } else if (symbol->has<semantics::ObjectEntityDetails>()) {
           // C1594(4)
           const auto &innermost{context_.FindScope(expr.source)};
-          if (const auto *pureFunc{
-                  semantics::FindPureFunctionContaining(&innermost)}) {
+          if (const auto *pureProc{
+                  semantics::FindPureProcedureContaining(&innermost)}) {
             if (const Symbol *
                 pointer{semantics::FindPointerComponent(*symbol)}) {
               if (const Symbol *
                   object{semantics::FindExternallyVisibleObject(
-                      *value, *pureFunc)}) {
+                      *value, *pureProc)}) {
                 if (auto *msg{Say(expr.source,
                         "Externally visible object '%s' must not be "
                         "associated with pointer component '%s' in a "
--- a/lib/semantics/resolve-names.cc
+++ b/lib/semantics/resolve-names.cc
@@ -3804,9 +3804,6 @@ bool DeclarationVisitor::PassesSharedLocalityChecks(
 // Checks for locality-specs LOCAL and LOCAL_INIT
 bool DeclarationVisitor::PassesLocalityChecks(
     const parser::Name &name, Symbol &symbol) {
-  if (!PassesSharedLocalityChecks(name, symbol)) {
-    return false;
-  }
   if (IsAllocatable(symbol)) {  // C1128
     SayWithDecl(name, symbol,
         "ALLOCATABLE variable '%s' not allowed in a locality-spec"_err_en_US);
@@ -3846,7 +3843,15 @@ bool DeclarationVisitor::PassesLocalityChecks(
         "Assumed size array '%s' not allowed in a locality-spec"_err_en_US);
     return false;
   }
-  // TODO: Check to see if the name can appear in a variable definition context
+  if (!IsModifiable(symbol, currScope())) {  // C1128
+    SayWithDecl(name, symbol,
+        "'%s' must be able to appear in a variable definition context to "
+        "appear in a locality-spec"_err_en_US);
+    return false;
+  }
+  if (!PassesSharedLocalityChecks(name, symbol)) {
+    return false;
+  }
   return true;
 }
 
--- a/lib/semantics/resolve-names.cc
+++ b/lib/semantics/resolve-names.cc
@@ -3853,8 +3853,8 @@ bool DeclarationVisitor::PassesLocalityChecks(
   if (std::optional<MessageFixedText> msg{
           WhyNotModifiable(symbol, currScope())}) {
     SayWithReason(name, symbol,
-        "'%s' must be able to appear in a variable definition context to "
-        "appear in a locality-spec"_err_en_US,
+        "'%s' may not appear in a locality-spec because it is not "
+        "definable"_err_en_US,
         std::move(*msg));
     return false;
   }
--- a/lib/semantics/resolve-names.cc
+++ b/lib/semantics/resolve-names.cc
@@ -432,6 +432,8 @@ public:
   // about a type; two names & locations
   void SayAlreadyDeclared(const SourceName &, Symbol &);
   void SayAlreadyDeclared(const parser::Name &, Symbol &);
+  void SayWithReason(
+      const parser::Name &, Symbol &, MessageFixedText &&, MessageFixedText &&);
   void SayWithDecl(const parser::Name &, Symbol &, MessageFixedText &&);
   void SayLocalMustBeVariable(const parser::Name &, Symbol &);
   void SayDerivedType(const SourceName &, MessageFixedText &&, const Scope &);
@@ -1527,12 +1529,17 @@ void ScopeHandler::SayAlreadyDeclared(const SourceName &name, Symbol &prev) {
   context().SetError(prev);
 }
 
+void ScopeHandler::SayWithReason(const parser::Name &name, Symbol &symbol,
+    MessageFixedText &&msg1, MessageFixedText &&msg2) {
+  Say2(name, std::move(msg1), symbol, std::move(msg2));
+  context().SetError(symbol, msg1.isFatal());
+}
+
 void ScopeHandler::SayWithDecl(
     const parser::Name &name, Symbol &symbol, MessageFixedText &&msg) {
-  Say2(name, std::move(msg), symbol,
+  SayWithReason(name, symbol, std::move(msg),
       symbol.test(Symbol::Flag::Implicit) ? "Implicit declaration of '%s'"_en_US
                                           : "Declaration of '%s'"_en_US);
-  context().SetError(symbol, msg.isFatal());
 }
 
 void ScopeHandler::SayLocalMustBeVariable(
@@ -3843,16 +3850,15 @@ bool DeclarationVisitor::PassesLocalityChecks(
         "Assumed size array '%s' not allowed in a locality-spec"_err_en_US);
     return false;
   }
-  if (!IsModifiable(symbol, currScope())) {  // C1128
-    SayWithDecl(name, symbol,
+  if (std::optional<MessageFixedText> msg{
+          WhyNotModifiable(symbol, currScope())}) {
+    SayWithReason(name, symbol,
         "'%s' must be able to appear in a variable definition context to "
-        "appear in a locality-spec"_err_en_US);
+        "appear in a locality-spec"_err_en_US,
+        std::move(*msg));
     return false;
   }
-  if (!PassesSharedLocalityChecks(name, symbol)) {
-    return false;
-  }
-  return true;
+  return PassesSharedLocalityChecks(name, symbol);
 }
 
 Symbol &DeclarationVisitor::FindOrDeclareEnclosingEntity(
--- a/lib/semantics/tools.cc
+++ b/lib/semantics/tools.cc
@@ -69,10 +69,10 @@ const Scope *FindProgramUnitContaining(const Symbol &symbol) {
   return FindProgramUnitContaining(symbol.owner());
 }
 
-const Scope *FindPureFunctionContaining(const Scope *scope) {
+const Scope *FindPureProcedureContaining(const Scope *scope) {
   scope = FindProgramUnitContaining(*scope);
   while (scope != nullptr) {
-    if (IsPureFunction(*scope)) {
+    if (IsPureProcedure(*scope)) {
       return scope;
     }
     scope = FindProgramUnitContaining(scope->parent());
@@ -124,14 +124,22 @@ bool IsDummy(const Symbol &symbol) {
   }
 }
 
+bool IsValueDummy(const Symbol &symbol) {
+  return IsDummy(symbol) && symbol.attrs().test(Attr::VALUE);
+}
+
 bool IsPointerDummy(const Symbol &symbol) {
   return IsPointer(symbol) && IsDummy(symbol);
 }
 
 // variable-name
 bool IsVariableName(const Symbol &symbol) {
-  const Symbol &ultimate{symbol.GetUltimate()};
-  return ultimate.has<ObjectEntityDetails>() && !IsParameter(ultimate);
+  if (const Symbol * root{GetAssociationRoot(symbol)}) {
+    return IsValueDummy(symbol) ||
+        (root->has<ObjectEntityDetails>() && !IsParameter(*root));
+  } else {
+    return false;
+  }
 }
 
 // proc-name
@@ -157,13 +165,13 @@ bool IsFunction(const Symbol &symbol) {
       symbol.details());
 }
 
-bool IsPureFunction(const Symbol &symbol) {
-  return symbol.attrs().test(Attr::PURE) && IsFunction(symbol);
+bool IsPureProcedure(const Symbol &symbol) {
+  return symbol.attrs().test(Attr::PURE) && IsProcedure(symbol);
 }
 
-bool IsPureFunction(const Scope &scope) {
+bool IsPureProcedure(const Scope &scope) {
   if (const Symbol * symbol{scope.GetSymbol()}) {
-    return IsPureFunction(*symbol);
+    return IsPureProcedure(*symbol);
   } else {
     return false;
   }
@@ -256,7 +264,7 @@ const Symbol *FindExternallyVisibleObject(
   // TODO: Storage association with any object for which this predicate holds,
   // once EQUIVALENCE is supported.
   if (IsUseAssociated(object, scope) || IsHostAssociated(object, scope) ||
-      (IsPureFunction(scope) && IsPointerDummy(object)) ||
+      (IsPureProcedure(scope) && IsPointerDummy(object)) ||
       (object.attrs().test(Attr::INTENT_IN) && IsDummy(object))) {
     return &object;
   } else if (const Symbol * block{FindCommonBlockContaining(object)}) {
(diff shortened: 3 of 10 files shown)


lib/semantics/resolve-names.cc
```diff
@@ -3796,7 +3793,15 @@ bool DeclarationVisitor::PassesLocalityChecks(
...
     return false;
   }
-  // TODO: Check to see if the name can appear in a variable definition context
+  if (!IsModifiable(symbol, currScope())) {  // C1128
+    SayWithDecl(name, symbol,
+        "'%s' must be able to appear in a variable definition context to "
+        "appear in a locality-spec"_err_en_US);
+    return false;
+  }
+  if (!PassesSharedLocalityChecks(name, symbol)) {
```
vjayathirtha-nv: `PassesSharedLocalityChecks()` has a `IsVariableName()` check in it. There may be a case where the `IsVariableName()` fails, and (because of the reordering) control first goes to one of `IsAllocatable()`, `IsAssumedSizeArray` etc which may not be equipped to handle such a symbol. It may be a risky assumption to make.

psteinfeld: Currently, there are tests for all of the conditions.  Thus, there are tests where ```IsAllocatable()``` gets called and returns false, and then later ```IsVariableName()``` also gets called and returns false.  For example, in the test resolve56.f90, the test case for ```subroutine s5()``` creates the name ```a``` which is construct associated with an expression.  The ```Symbol``` for ```a``` is neither allocatable or a variable name.  Also note that some names might fail to qualify for being in a locality-spec for multiple reasons.  None of this seems like a problem to me.

I ordered these tests the way that currently appears in the code so that we test for the conditions which produce the most specific error messages first.  For example, a variable which is a coarray of type ```event_type``` gets checked first to see if it's a coarray, which produces a more specific error message than the one caused by being ```event_type``` since the latter fails because with a message saying that the variable name must be able to appear in a variable definition context.

lib/semantics/tools.cc
```diff
@@ -327,6 +335,28 @@ const Symbol *FindFunctionResult(const Symbol &symbol) {
   return nullptr;
 }
 
+// Return the variable of a construct association, if associated
+// Return nullptr if the name is associated with an expression
+const Symbol *GetAssociationRoot(const Symbol &symbol) {
+  const Symbol &ultimate{symbol.GetUltimate()};
+  if (ultimate.has<AssocEntityDetails>()) {  // construct association
+    auto &details{ultimate.get<AssocEntityDetails>()};
+    if (const MaybeExpr * pexpr{&details.expr()}) {
```
tskeith: If pexpr is false, `ultimate` will be returned. I think it should be nullptr in that case.

lib/semantics/tools.cc
```diff
@@ -442,6 +488,31 @@ bool IsAssumedSizeArray(const Symbol &symbol) {
...
+
+bool InProtectedContext(const Symbol &symbol, const Scope &currentScope) {
+  return IsProtected(symbol) && !IsHostAssociated(symbol, currentScope);
+}
+
+bool IsModifiable(const Symbol &symbol, const Scope &scope) {
+  if (const Symbol * root{GetAssociationRoot(symbol)}) {
+    return IsVariableName(*root) && !InProtectedContext(*root, scope) &&
+        !IsExternalInPureContext(*root, scope) &&
+        !IsOrContainsEventOrLockComponent(*root);
```
tskeith: Here we know why the symbol can't appear in a variable definition context, so it would be nice to share that with the user. Presumably we will be doing the same check in lots of places.

Maybe return an optional message that says what's wrong with symbol, like "%s is not a variable". The main error message would be "%s may not appear in a locality spec" and then attach the returned message to it.

lib/semantics/resolve-names.cc
vjayathirtha-nv: My theory was that there could be symbols which are not handled in std::visits that exist in one of the other checks, which would have been caught by `IsVariableName()`. However I was unsuccessful in coming up with a case to test out this theory.

lib/semantics/resolve-names.cc
```diff
@@ -3796,7 +3793,15 @@ bool DeclarationVisitor::PassesLocalityChecks(
...
+    return false;
+  }
```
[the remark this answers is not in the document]
psteinfeld: Thanks, Jean.  I'll make this change.

lib/semantics/expression.cc
```diff
@@ -1320,13 +1320,13 @@ MaybeExpr ExpressionAnalyzer::Analyze(
...
-                  semantics::FindPureFunctionContaining(&innermost)}) {
+          if (const auto *pureProc{
+                  semantics::FindPureProcedureContaining(&innermost)}) {
             if (const Symbol *
                 pointer{semantics::FindPointerComponent(*symbol)}) {
               if (const Symbol *
                   object{semantics::FindExternallyVisibleObject(
-                      *value, *pureFunc)}) {
+                      *value, *pureProc)}) {
                 if (auto *msg{Say(expr.source,
```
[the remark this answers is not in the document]
psteinfeld: Good catch!  I'll change the message.

lib/semantics/tools.cc
```diff
@@ -327,6 +335,28 @@ const Symbol *FindFunctionResult(const Symbol &symbol) {
   return nullptr;
 }
 
+// Return the variable of a construct association, if associated
+// Return nullptr if the name is associated with an expression
+const Symbol *GetAssociationRoot(const Symbol &symbol) {
+  const Symbol &ultimate{symbol.GetUltimate()};
+  if (ultimate.has<AssocEntityDetails>()) {  // construct association
+    auto &details{ultimate.get<AssocEntityDetails>()};
```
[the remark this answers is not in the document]
psteinfeld: Good suggestion.  I'll clean up this code to be more concise.

lib/semantics/tools.cc
```diff
@@ -327,6 +335,28 @@ const Symbol *FindFunctionResult(const Symbol &symbol) {
...
+    if (const MaybeExpr * pexpr{&details.expr()}) {
+      if (pexpr->has_value()) {
+        const SomeExpr &expr{**pexpr};
+        if (evaluate::IsVariable(expr)) {
+          if (const Symbol * varSymbol{evaluate::GetLastSymbol(expr)}) {
+            return GetAssociationRoot(*varSymbol);
+          }
```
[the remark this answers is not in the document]
psteinfeld: Good point.  I'll restructure this.

lib/semantics/tools.cc
```diff
@@ -327,6 +335,28 @@ const Symbol *FindFunctionResult(const Symbol &symbol) {
...
 
+// Return the variable of a construct association, if associated
+// Return nullptr if the name is associated with an expression
+const Symbol *GetAssociationRoot(const Symbol &symbol) {
+  const Symbol &ultimate{symbol.GetUltimate()};
+  if (ultimate.has<AssocEntityDetails>()) {  // construct association
+    auto &details{ultimate.get<AssocEntityDetails>()};
+    if (const MaybeExpr * pexpr{&details.expr()}) {
+      if (pexpr->has_value()) {
+        const SomeExpr &expr{**pexpr};
```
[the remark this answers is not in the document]
psteinfeld: Yep.  I'll fix this up.

lib/semantics/tools.cc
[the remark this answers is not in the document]
psteinfeld: Actually, that case should more properly be covered by ```IsVariableName()```.  Tim had a related suggestion that I'm working on.  I'll push another version for your review.

lib/semantics/tools.cc
psteinfeld: Yes.  I'll restructure this to be correct and more obviously so.

lib/semantics/tools.cc
psteinfeld: Will do.

lib/semantics/tools.cc
```diff
@@ -328,6 +337,28 @@ const Symbol *FindFunctionResult(const Symbol &symbol) {
   return nullptr;
 }
 
+const Symbol *GetAssociatedVariable(const AssocEntityDetails &details) {
```
tskeith: This function could be static.

lib/semantics/tools.cc
```diff
@@ -328,6 +337,28 @@ const Symbol *FindFunctionResult(const Symbol &symbol) {
...
+  }
+  return nullptr;
+}
+
+// Return the Symbol of the variable of a construct association, if it exists
+// Return nullptr if the name is associated with an expression
+const Symbol *GetAssociationRoot(const Symbol &symbol) {
+  const Symbol &ultimate{symbol.GetUltimate()};
+  if (ultimate.has<AssocEntityDetails>()) {  // construct association
+    return GetAssociatedVariable(ultimate.get<AssocEntityDetails>());
```
tskeith: This could be simplified slightly:
```
if (const auto *details{ultimate.detailsIf<AssocEntityDetails>()}) {
  return GetAssociatedVariable(*details);
```

lib/semantics/tools.cc
```diff
@@ -443,6 +490,45 @@ bool IsAssumedSizeArray(const Symbol &symbol) {
...
+  if (!root) {
+    return "'%s' is construct associated with an expression"_en_US;
+  } else if (InProtectedContext(*root, scope)) {
+    return "'%s' is protected in this scope"_en_US;
+  } else if (IsExternalInPureContext(*root, scope)) {
+    return "'%s' is an external that is referenced in a PURE procedure"_en_US;
+  } else if (IsOrContainsEventOrLockComponent(*root)) {
+    return "'%s' is an entity with either an EVENT_TYPE or LOCK_TYPE"_en_US;
+  } else if (IsIntentIn(*root) && !IsValueDummy(*root)) {
+    return "'%s' is an INTENT(IN) dummy argument"_en_US;
```
tskeith: C844 prohibits nonpointer INTENT(IN) dummy arguments. That's not the same as VALUE arguments.

lib/semantics/tools.cc
```diff
@@ -443,6 +490,45 @@ bool IsAssumedSizeArray(const Symbol &symbol) {
...
+
+std::optional<parser::MessageFixedText> WhyNotModifiable(
+    const Symbol &symbol, const Scope &scope) {
+  const Symbol *root{GetAssociationRoot(symbol)};
+  if (!root) {
+    return "'%s' is construct associated with an expression"_en_US;
+  } else if (InProtectedContext(*root, scope)) {
+    return "'%s' is protected in this scope"_en_US;
+  } else if (IsExternalInPureContext(*root, scope)) {
+    return "'%s' is an external that is referenced in a PURE procedure"_en_US;
```
tskeith: "External" usually refers to an external procedure so maybe this would be better:
"'%s' is externally visible and referenced in a PURE procedure"

lib/semantics/tools.cc
```diff
@@ -443,6 +490,45 @@ bool IsAssumedSizeArray(const Symbol &symbol) {
   return details && details->IsAssumedSize();
 }
 
+bool IsExternalInPureContext(const Symbol &symbol, const Scope &scope) {
+  if (const auto *pureProc{semantics::FindPureProcedureContaining(&scope)}) {
+    if (const Symbol * root{GetAssociationRoot(symbol)}) {
+      if (semantics::FindExternallyVisibleObject(*root, *pureProc)) {
```
tskeith: `semantics::` is unnecessary here.

lib/semantics/tools.cc
tskeith: C1594 has some other restrictions on what can be modified in a pure procedure, e.g. something in a common block or a co-indexed object.

lib/semantics/tools.cc
```diff
@@ -443,6 +490,45 @@ bool IsAssumedSizeArray(const Symbol &symbol) {
...
+  } else if (IsIntentIn(*root) && !IsValueDummy(*root)) {
+    return "'%s' is an INTENT(IN) dummy argument"_en_US;
+  } else if (!IsVariableName(*root)) {
+    return "'%s' is not a variable"_en_US;
+  } else {
+    return std::nullopt;
+  }
+}
+
+bool IsModifiable(const Symbol &symbol, const Scope &scope) {
```
tskeith: Is this ever used? It seems like we'll always want the reason.

lib/semantics/resolve-names.cc
```diff
@@ -3796,8 +3800,15 @@ bool DeclarationVisitor::PassesLocalityChecks(
         "Assumed size array '%s' not allowed in a locality-spec"_err_en_US);
     return false;
   }
-  // TODO: Check to see if the name can appear in a variable definition context
-  return true;
+  if (std::optional<MessageFixedText> msg{
+          WhyNotModifiable(symbol, currScope())}) {
+    SayWithReason(name, symbol,
+        "'%s' must be able to appear in a variable definition context to "
+        "appear in a locality-spec"_err_en_US,
```
tskeith: This message is precise, but maybe too technical.
What do you think of this:
"'%s' may not appear in a locality-spec because it is not a definable variable"

lib/semantics/tools.cc
```diff
@@ -399,6 +430,22 @@ const Symbol *HasEventOrLockPotentialComponent(
   return nullptr;
 }
 
+bool IsOrContainsEventOrLockComponent(const Symbol &symbol) {
+  if (const Symbol * root{GetAssociationRoot(symbol)}) {
+    if (const auto *details{root->detailsIf<ObjectEntityDetails>()}) {
+      if (const DeclTypeSpec * type{details->type()}) {
+        if (const DerivedTypeSpec * derived{type->AsDerived()}) {
+          if (IsEventTypeOrLockType(derived) ||
```
vjayathirtha-nv: You could have `return IsEventTypeOrLockType(derived) || HasEventOrLockPotentialComponent(*derived);`
Since we're using this form in other places too.

lib/semantics/tools.cc
```diff
@@ -443,6 +490,45 @@ bool IsAssumedSizeArray(const Symbol &symbol) {
   return details && details->IsAssumedSize();
 }
 
+bool IsExternalInPureContext(const Symbol &symbol, const Scope &scope) {
+  if (const auto *pureProc{semantics::FindPureProcedureContaining(&scope)}) {
+    if (const Symbol * root{GetAssociationRoot(symbol)}) {
```
vjayathirtha-nv: Did you run clang format on this? 
I think there should be no space between `*`  and `root`

tskeith: There should be no space, but it is a problem with clang-format. It gets it right with `const auto` but not if there is a type name there.

vjayathirtha-nv: I see. You can ignore my comment Pete.

lib/semantics/tools.cc
psteinfeld: Thanks, I'll change that.

lib/semantics/tools.cc
psteinfeld: Thanks, I'll change this.

lib/semantics/tools.cc
psteinfeld: Yikes!  I misread the standard on this.  I'll fix things.

lib/semantics/tools.cc
psteinfeld: Thanks, I'll change this.

lib/semantics/tools.cc
psteinfeld: I used it in earlier code, but since it's not currently used, I'll remove it.

lib/semantics/resolve-names.cc
psteinfeld: Thanks.  I'll make this change.

lib/semantics/tools.cc
psteinfeld: Good catch!  I'll change this.

lib/semantics/tools.cc
psteinfeld: Thanks for checking, though.

lib/semantics/resolve-names.cc
psteinfeld: Actually, I think I'll use this message here:
"'%s' may not appear in a locality-spec because it is not definable"

The potential problem is that we augment this message with another message that gives the reason that it's not definable.  Sometimes, the reason states that it's not a variable.  If both of these messages mention not being a variable it reads awkwardly.

lib/semantics/tools.cc
psteinfeld: I like your rewording of the message.

With regards to C1594, it looks like I just spaced on this.  I have a list on Google docs of all of the places where various constructs are prohibited from appearing in variable definition contexts, and this document already lists C1594.  It also lists C1101 and C1158 neither of which I currently check.  I have more work to do ...

psteinfeld: More on C1594 ...  It looks to me like all of the cases listed in the constraint are handled by ```FindExternallyVisibleObject()``` except for a coindexed object.  I added a TODO for that case.

lib/semantics/tools.cc
```diff
@@ -328,6 +331,29 @@ const Symbol *FindFunctionResult(const Symbol &symbol) {
...
+      if (const Symbol * varSymbol{evaluate::GetLastSymbol(*expr)}) {
+        return GetAssociationRoot(*varSymbol);
+      }
+    }
+  }
+  return nullptr;
+}
+
+// Return the Symbol of the variable of a construct association, if it exists
+// Return nullptr if the name is associated with an expression
```
vjayathirtha-nv: I think this comment needs to be updated.

psteinfeld: Could you be more specific about the change you want?  Is the comment not understandable, or is it incorrect in some respect?

vjayathirtha-nv: `GetAssociationRoot()` [used to have](https://github.com/flang-compiler/f18/pull/596/commits/c9d2507b74da881dda2eb11805a0394a415db2e4#diff-1f02100eb6511d1af252e0a76b91ecafR353) a `return nullptr;` in your original commit of this PR. With your follow up commits, the structure of this function changed and so the comment about `return nullptr` doesn't seem to fit.

psteinfeld: Right.  The code that used to return a ```nullptr``` was pulled out and put into the static function ```GetAssociatedVariable()```.  But the effect of this is that ```GetAssociationRoot()```, which is the only externally visible function, still either returns a ```nullptr``` in the case where its parameter is construct associated with an expression.

vjayathirtha-nv: That makes sense, thanks for the explanation.

PR #600, opened 2019-07-23
Avoid interference between different construct association in name resolutions

### LINKED ISSUE #598: Bogus error from sourced allocation
```
type t1
end type
type, extends(t1) :: t2
end type
class(t2), allocatable :: x
class(*), allocatable :: a
allocate(t2::a)
select type(a)
  type is(t2)
     associate(y=>a%t1)
     end associate
end select
select type(a)
  class is (t2)
    allocate(x,source=a)
end select
end
```

produces `bug.f90:15:14: error: Allocatable object in ALLOCATE must be type compatible with source expression from MOLD or SOURCE` on the sourced allocation.

The previous `select type` and its nested `associate` constructs need to be present to demonstrate the bug.  `valgrind`, however, does not complain about uninitialized data usage.

jeanPerier:
Fix #598 and similar issues.
Transform `resolve-names.cc` `ConstructVisitor` `association_` member into a stack so that different association construct (select-type, select-rank, change-team, associate) imbrication/succession do not interfere with each other leading to erroneous symbol resolution.

### DIFF
--- a/lib/semantics/resolve-names.cc
+++ b/lib/semantics/resolve-names.cc
@@ -909,6 +909,8 @@ public:
   void Post(const parser::EndAssociateStmt &);
   void Post(const parser::Association &);
   void Post(const parser::SelectTypeStmt &);
+  bool Pre(const parser::SelectTypeConstruct &);
+  void Post(const parser::SelectTypeConstruct &);
   bool Pre(const parser::SelectTypeConstruct::TypeCase &);
   void Post(const parser::SelectTypeConstruct::TypeCase &);
   void Post(const parser::TypeGuardStmt::Guard &);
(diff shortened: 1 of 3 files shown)


lib/semantics/resolve-names.cc
```diff
@@ -4407,6 +4440,20 @@ ConstructVisitor::Selector ConstructVisitor::ResolveSelector(
       x.u);
 }
 
+ConstructVisitor::Association &ConstructVisitor::GetCurrentAssociation() {
+  CHECK(associationStack_.size());
```
klausler: `!associationStack_.empty()` makes the condition clearer.  Also below.

PR #601, opened 2019-07-23
Fix legal pointer target test

klausler:
When an object is assigned as the target of a pointer, perhaps as a component of a structure constructor, the object must be either a pointer or have the target attribute.  The test is a little tricky, and I had botched it in the case of something like `p => a(1)%b(2)%c(3)` when (say) `b` was a pointer or target but neither `a` nor `c` were.  The old code just found the base object `a` and tested it for pointer/target attributes.  It turns out that as long as `c` is not allocatable, it effectively inherits the right to be a pointer target from `b`, and so the assignment is legal.  So the code really needs to traverse the target's designator upwards until it hits a symbol that determines whether it can be use as a target or not.

Solving this is just a few actual lines of code to implement and call a `GetLastTarget()` function in `tools.h` to find `b` (in the example above) in a designator; but I implemented it with the expression traversal framework, and while I was doing so, I re-implemented `IsVariable()` and `GetLastSymbol()` with the traversal framework as well, and beefed up that framework so that visitors and mutators could be written with member function templates as callbacks, not just member functions.  I think that tools implemented with the expression traversal framework might be less immediately readable than using mutually recursive functions, but may be less buggy in the long run (debatable).  This patch is thus mostly infrastructure work in the traversal framework, plus the new call that motivated that work.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,6 @@
+Debug
+Release
+MinSizeRel
 build
 root
 tags
--- a/.gitignore
+++ b/.gitignore
@@ -1,7 +1,5 @@
-Debug
-Release
-MinSizeRel
 build
+root
 tags
 TAGS
 *.o
--- a/lib/semantics/assignment.cc
+++ b/lib/semantics/assignment.cc
@@ -113,21 +113,19 @@ template<typename T>
 void CheckPointerAssignment(parser::ContextualMessages &messages,
     const IntrinsicProcTable &intrinsics, const Symbol &lhs,
     const Designator<T> &d) {
-  if (const auto *symbol{d.GetBaseObject().symbol()}) {
-    const Symbol &ultimate{symbol->GetUltimate()};
+  const Symbol *last{d.GetLastSymbol()}, *base{d.GetBaseObject().symbol()};
+  if (last != nullptr && base != nullptr) {
     std::optional<parser::MessageFixedText> error;
     if (IsProcedurePointer(lhs)) {
       // Shouldn't be here in this function unless lhs is an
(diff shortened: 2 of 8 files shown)


lib/semantics/assignment.cc
```diff
@@ -113,21 +113,19 @@ template<typename T>
 void CheckPointerAssignment(parser::ContextualMessages &messages,
     const IntrinsicProcTable &intrinsics, const Symbol &lhs,
     const Designator<T> &d) {
-  if (const auto *symbol{d.GetBaseObject().symbol()}) {
-    const Symbol &ultimate{symbol->GetUltimate()};
+  const Symbol *last{d.GetLastSymbol()}, *base{d.GetBaseObject().symbol()};
```
tskeith: It's just a matter of opinion, but I prefer one declaration per line, especially when they are initialized.

.gitignore
```diff
@@ -1,7 +1,5 @@
-Debug
-Release
-MinSizeRel
```
tskeith: Why remove these?

klausler: I didn't see how anybody could still be using them; am I wrong?

lib/semantics/assignment.cc
klausler: As you wish.

.gitignore
tskeith: I build in subdirectories named Debug and Release, though I see I have them in my global .gitignore so I guess it doesn't affect me.

klausler: I'll restore them; thanks.

PR #602, opened 2019-07-24
Fold more model inquiry intrinsics

klausler:
EPSILON, HUGE, TINY, &c.  Correct &/or add intrinsic table entries, handle folding, and test.


lib/evaluate/intrinsics.cc
```diff
@@ -296,6 +296,8 @@ static const IntrinsicInterface genericIntrinsicFunction[]{
         {{"i", AnyInt, Rank::elementalOrBOZ},
             {"j", AnyInt, Rank::elementalOrBOZ}},
         DefaultLogical},
+    {"bit_size", {{"i", AnyInt, Rank::anyOrAssumedRank}}, DefaultInt,
```
jeanPerier: From a standard point of view, the result has the same kind as the argument in `BIT_SIZE`. 

Both gfortran and pgfortran seem to follow this and print 1,2,4 for the program below:
```
  integer k1, k2, k4
  k1 = kind(bit_size(1_1))
  k2 = kind(bit_size(1_2))
  k4 = kind(bit_size(1_4))
  print *, k1, k2, k4
end
```
It would also remove the argument visit in folding to follow this.

PR #603, opened 2019-07-24
Fix IsVariable for Pete

klausler:
Does this fix your problem, Pete?

### DIFF
--- a/lib/evaluate/tools.cc
+++ b/lib/evaluate/tools.cc
@@ -24,13 +24,6 @@ using namespace Fortran::parser::literals;
 namespace Fortran::evaluate {
 
 // IsVariable()
-IsVariableVisitor::IsVariableVisitor(std::nullptr_t) {}
-void IsVariableVisitor::Handle(const StaticDataObject &) { Return(false); }
-void IsVariableVisitor::Post(const Substring &) { Return(true); }
-void IsVariableVisitor::Pre(const Component &) { Return(true); }
-void IsVariableVisitor::Pre(const ArrayRef &) { Return(true); }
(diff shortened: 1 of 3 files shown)


psteinfeld:
I'm still investigating, but here's what I've found so far.

Here's my test program;
subroutine s()
  real :: a

  associate (b => a)
    do concurrent (i = 1:5) local (b)
    end do
  end associate
end subroutine s

Here's the function that calls ```IsVariable()```
```
const Symbol *GetAssociationRoot(const Symbol &symbol) {
  const Symbol &ultimate{symbol.GetUltimate()};
  if (ultimate.has<AssocEntityDetails>()) {  // construct association
    auto &details{ultimate.get<AssocEntityDetails>()};
    if (const MaybeExpr * pexpr{&details.expr()}) {
      if (pexpr->has_value()) {
        const SomeExpr &expr{**pexpr};
        if (evaluate::IsVariable(expr)) {
          if (const Symbol * varSymbol{evaluate::GetLastSymbol(expr)}) {
            return GetAssociationRoot(*varSymbol);
          }
        } else {
          return nullptr;
        }
      }
    }
  }
  return &ultimate;
}
```

I've added debug statements to this code, and here's what happens.  The code above gets called for the reference to ```b```.  Here's the symbol dump for ```b```: ```AssocEntity type: REAL(4) expr:a```.  The code proceeds down the cascade of "if" statements until it gets to the call to ```evaluate::IsVariable(expr)```.  This call returns ```false```.

klausler:
Should work now; please confirm whenever convenient.

Issue #594, opened 2019-07-19
Crash with CHECK in check-allocate.cc(199)


klausler:
```
MODULE m
  TYPE :: t
    class(t), pointer :: p
   CONTAINS
    PROCEDURE :: f
  END TYPE
 CONTAINS
  FUNCTION f(this) RESULT (x)
    CLASS(t) :: this
    CLASS(t), allocatable :: x
  END FUNCTION
END MODULE

use m
type(t),target :: o
class(t), allocatable :: p
allocate(p, source=o%f())
end
```

jeanPerier:
The `CHECK` is failing after a typeless expression was found in a context where it is not expected. In the bug report case, the internal error is justified because this o%f() has type "t" so something is probably wrong in `GetType()`.
However, investigating further, the `CHECK` is also too strict because it will crash with actual type-less expressions when it should only emit a user error.

I am working on a fix.

jeanPerier:
Fixed with PR #607

Issue #573, opened 2019-07-15
Forward references to derived types are not supported


klausler:
```
type, abstract :: base
  class(extended), pointer :: p
end type
type, extends(base) :: extended
end type
```

jeanPerier:
What about allocatable components ? I know gfortran, NAG and pgfortran accepts this with pointers and not allocatables, but I did not find the exact rule here that would differentiate allocatable and pointer components on this matter.
The only restrcition to use previously defined derived-types I found are:

7.3.2.3 point 2 says:
> A derived-type-spec in a CLASS type specifier in a type declaration statement shall specify a previously defined derived type.

Which I understand as making not yet defined derived type usage illegal regardless of the attributes in type-declaration-stmt but not in data-component-def-stmt (so it is out of scope here)

C744, specific to data-component-def-stmt, says:
> If neither the POINTER nor the ALLOCATABLE attribute is specified, the declaration-type-spec in the component-def-stmt shall specify an intrinsic type or a previously defined derived type.

Therefore, I do not fully understand why ALLOCATABLE component with forward-derived type are rejected by other compilers, except that it would make checks related to potential components more annoying to perform.

I am not asking, because I want to advocate to support allocatable here too, but because I wondered when implementing `HasEventOrLockPotentialComponent` used in  allocate checks, and if we need to support this for allocatable too, then it will currently end-up in an infinite loop.
See: https://github.com/flang-compiler/f18/blob/8c63aa94972fb3f7b88f23fa1b78f8b67ffa8732/lib/semantics/tools.cc#L383

tskeith:
Here is another example of a forward reference in a data-component-def-stmt. It is from 7.5.4.6(8) Note 5.
```
TYPE NODE
  INTEGER :: VALUE = 0
  TYPE (NODE), POINTER :: NEXT_NODE => SENTINEL
END TYPE
TYPE(NODE), SAVE, TARGET :: SENTINEL
```

klausler:
Jean, can this bug report be closed?

klausler:
Forward references to derived types still block some legacy correctness tests from passing; do not close yet.

(Specifically, `class(t), pointer :: p` as a component fails when `t` is defined later.)

klausler:
Fixed (implemented) in PR #847.

PR #607, opened 2019-07-26
Fix issue 594 and add a derived type component visitor framework

jeanPerier:
**EDIT 07/31:** Derived type components visitor replaced  by iterators (description in comments above `ComponentIterator` declaration in `lib/semantics/tools.h`). Iterator framework now also replaces `DerivedTypeDetails::OrderComponents`.

- Fix issue #594 
- Also do not crash on typeless expressions in source-allocation that could come from the user. 
- Add an derived type component ~~visitor~~ iterator framework to verify component properties
- Uses the framework to replace `HasCoarrayUltimateComponent` and `HasEventOrLockPotentialComponent` and give more useful feedback in allocate checks errors.

**Fix 594:**
This PR fixes issue 594 by adding and handler for 'ProcBindingDetails' in `Symbol::GetType`.
`HostAssocDetails` handling was added at the same time because it did not seem right to handle
`UseDetails` but not this.
  
**Do not crash on typeless source expressions:**
If the `CHECK` failure related to #594 was indeed an internal failure, it lead me to think that a user could very well put a Boz here and an error message should be given rather than crashing.

**~~New derived type component visiting framework:~~ (Replaced by iterators, see tools.h)**
The fix unveiled an issue in `FindUltimateComponent` (type bound procedure were also searched).
Rather than just fixing this, this PR implements a new derived type component visitor framework that allows testing a predicate on ultimate/potential or direct component and also allows extracting the full component path to the matching component for better error messages.

The visitor is not using the template and virtual inheritance pattern of semantic and expression visitors because only `Symbol&` are visited and I did not want to expose the implementation in header file for a feature that is much less than. It is using a more classic callback pattern.

**Better feedback in allocate checks:**
Given the program:

```
  type t1
    real, allocatable :: my_coarray[*]
  end type
  type t2
    type(t1) :: a
  end type
  type t3
    type(t2) :: b
  end type
  type(t3) :: var, test
  allocatable :: test
  allocate(test, source=var)
end
```
The error message is now:
```
error: SOURCE or MOLD expression must not have a type with a coarray ultimate component
    allocate(test, source=var)
                          ^^^
Type 't3' has coarray ultimate component '%b%a%my_coarray' declared here
      real, allocatable :: my_coarray[*]
```

The attached part was previously only: `Coarray ultimate component  declared here` which I found was not helpful enough in case of nested derived types like this.

### DIFF
--- a/lib/semantics/tools.cc
+++ b/lib/semantics/tools.cc
@@ -380,18 +380,6 @@ bool IsTeamType(const DerivedTypeSpec *derived) {
   return IsDerivedTypeFromModule(derived, "iso_fortran_env", "team_type");
 }
 
-const Symbol *FindUltimateComponent(const DerivedTypeSpec &derivedTypeSpec,
-    std::function<bool(const Symbol &)> predicate) {
-  return ComponentVisitor{std::move(predicate)}
-      .VisitUltimateComponents(derivedTypeSpec)
-      .Result();
-}
-
-const Symbol *HasCoarrayUltimateComponent(
-    const DerivedTypeSpec &derivedTypeSpec) {
-  return FindUltimateComponent(derivedTypeSpec, IsCoarray);
-}
-
 const bool IsEventTypeOrLockType(const DerivedTypeSpec *derivedTypeSpec) {
   return IsDerivedTypeFromModule(
              derivedTypeSpec, "iso_fortran_env", "event_type") ||
@@ -412,19 +400,12 @@ bool IsOrContainsEventOrLockComponent(const Symbol &symbol) {
       if (const DeclTypeSpec * type{details->type()}) {
         if (const DerivedTypeSpec * derived{type->AsDerived()}) {
           return IsEventTypeOrLockType(derived) ||
-              HasEventOrLockPotentialComponent(*derived);
+              ComponentVisitor::HasEventOrLockPotential(*derived);
         }
       }
     }
   }
   return false;
-}
-
-const Symbol *HasEventOrLockPotentialComponent(
-    const DerivedTypeSpec &derivedTypeSpec) {
-  return ComponentVisitor{IsEventTypeOrLockTypeObjectEntity}
-      .VisitPotentialComponents(derivedTypeSpec)
-      .Result();
 }
 
 bool IsFinalizable(const Symbol &symbol) {
@@ -888,12 +869,25 @@ ComponentVisitor &ComponentVisitor::VisitDirectComponents(
   return *this;
 }
 
+ComponentVisitor ComponentVisitor::HasEventOrLockPotential(
+    const DerivedTypeSpec &derived) {
+  return ComponentVisitor{IsEventTypeOrLockTypeObjectEntity}
+      .VisitPotentialComponents(derived);
+}
+
 std::string ComponentVisitor::BuildResultDesignatorName() const {
   std::string designator{""};
   for (const Symbol *component : componentStack_) {
     designator += "%" + component->name().ToString();
   }
   return designator;
+}
+
+const Symbol *FindUltimateComponent(const DerivedTypeSpec &derived,
+    std::function<bool(const Symbol &)> predicate) {
+  return ComponentVisitor{std::move(predicate)}
+      .VisitUltimateComponents(derived)
+      .Result();
 }
 
 }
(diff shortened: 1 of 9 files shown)


jeanPerier:
Such a long description and I forgot to say... the main motivation for the component visitor framework is to make it safe against derived type loops that could arise from forward usage of derived type (legal or not). (see issue #573).

lib/semantics/tools.cc
```diff
@@ -794,4 +761,139 @@ static Symbol &InstantiateSymbol(
...
+private:
+  const Symbol *TraverseDerivedType(const DerivedTypeSpec &derived) {
+    const Symbol &derivedTypeSymbol{derived.typeSymbol()};
+    // Avoid infinite loop if the type is already part of the types
+    // being visited. It is possible to have "loops in type" because
+    // C744 does not forbid to use not yet declared type for
+    // ALLOCATABLE or POINTER components.
+    // When looking for potential components, the search could fall
+    // in an infinite loop. Avoid these in other search for safety.
+    for (const Symbol *typeUnderVisit : typeStack_) {
```
tskeith: Wouldn't it be better for `typeStack_` to be a set? You wouldn't have to iterate to check membership.

lib/semantics/tools.cc
```diff
@@ -794,4 +761,139 @@ static Symbol &InstantiateSymbol(
...
+        return nullptr;
+      }
+    }
+    typeStack_.push_back(&derivedTypeSymbol);
+    const Scope *scope{derivedTypeSymbol.scope()};
+    CHECK(scope);  // derived type symbol must have a scope
+    const Symbol *result{nullptr};
+    // Note: parent subcomponents are visited first, then the parent component
+    // (if any), then other components by order of declaration
+    SymbolVector orederedComponents{
```
tskeith: `orederedComponents` -> `orderedComponents`

klausler:
An iterator for visiting components in order would also work as an alternative design pattern for your visitation framework, and could be used to replace `DerivedTypeDetails::OrderComponents()`, which today returns a vector of components.

lib/semantics/tools.cc
```diff
@@ -794,4 +761,139 @@ static Symbol &InstantiateSymbol(
...
+        derivedTypeSymbol.get<DerivedTypeDetails>().OrderComponents(*scope)};
+    for (const Symbol *component : orederedComponents) {
+      if (component->has<ProcEntityDetails>()) {
+        result = VisitProcedureComponent(*component);
+      } else if (component->has<ObjectEntityDetails>()) {
+        result = VisitDataComponent(*component);
+      }
+      if (result) {
+        return result;
```
tskeith: It's surprising to push the type but not always pop it. Maybe add a comment about why the pop is not necessary (assuming it's not).

jeanPerier: > An iterator for visiting components in order would also work as an alternative design pattern for your visitation framework, and could be used to replace `DerivedTypeDetails::OrderComponents()`, which today returns a vector of components.

That is an interesting idea, thanks.
So far, the current visiting framework is actually using `DerivedTypeDetails::OrderComponents()` to traverse derived type sub-components. But, if I transform the ultimate/direct/potential visitor into iterators, they would probably live inside the `DerivedTypeDetails` or at least be `friend` and would not require `OrderComponents`.

I guess I can transform the visitors into iterators by transforming the `componentStack_` into some kind of stack of iterators over `DerivedTypeDetails::componentNames_` and by removing the predicate testing part (and probably a bit more).
I will explore the idea.

lib/semantics/tools.cc
jeanPerier: Thanks, variable removed in iterators framework.

lib/semantics/tools.cc
jeanPerier: Oops, the reply above did not belong here. The pop was not necessary, but yes a comment would have been nice.

lib/semantics/tools.cc
jeanPerier: Thanks, a stack would have been better for the visitor indeed.
In the iterators, the types used for this check are are kept as `DerivedTypeSpec`, but now the order matter because they are also used to get the scopes, so I could not use a `set` there.

test/semantics/resolve05.f90
```diff
@@ -23,6 +23,7 @@ subroutine s
 function f() result(res)
   integer :: res
   !ERROR: 'f' is already declared in this scoping unit
+  !ERROR: The type of 'f' has already been declared
```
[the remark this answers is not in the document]
jeanPerier: Thanks for catching this, now updated.

PR #610, opened 2019-07-30
Create symbols for generics at the start of the scope

tskeith:
Generic interfaces can have the same name as contained subprograms
or derived types. This means a subprogram and derived type can have
the same name **if** there is a generic interface with that name.
Change the order of processing declarations to do generics first
so that we don't report a conflict between the subprogram and type.
(See `module m4b` in `test/semantics/resolve18.f90` for an example.)

`PreSpecificationConstruct` is called on each `SpecificationConstruct`
of a specification part before we continue walking it. We use that
to call `CreateGeneric` for each generic interface so that they already
exist when we do the normal walk of the declarations.

Improve the error message issued when a procedure and generic interface
have the same name but the procedure is not a specific of the generic.

Change `SayAlreadyDeclared` to report the error on the second occurence
of the name when possible. This can arise for declarations the are
processed out of order, e.g. contained subprograms and generic
interfaces.

Avoid multiple "already declared" errors for the case when a contained
subprogram has the same name as a declared entity. We first create the
symbol with SubprogramNameDetails, then replace it with the entity
(and report the error), then replace it with the real subprogram
(and get the error again). By setting and checking the error flag
we avoid the second error.

### DIFF
--- a/lib/semantics/resolve-names.cc
+++ b/lib/semantics/resolve-names.cc
@@ -430,8 +430,9 @@ public:
 
   // Special messages: already declared; referencing symbol's declaration;
   // about a type; two names & locations
-  void SayAlreadyDeclared(const SourceName &, Symbol &);
   void SayAlreadyDeclared(const parser::Name &, Symbol &);
+  void SayAlreadyDeclared(const SourceName &, Symbol &);
+  void SayAlreadyDeclared(const SourceName &, const SourceName &);
   void SayWithReason(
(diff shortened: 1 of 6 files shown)


klausler:
Only PGI and XLF seem to allow a derived type to have the same name as a contained procedure.  Are you sure that the F'2018 standard requires this?  Can the names coexist if they *don't* appear in a generic interface?

tskeith:
Based on Peter's input I'm rethinking whether it is okay to use the same name for a generic interface, derived type, and procedure all in the same scope.

PR #612, opened 2019-07-31
Changes to enforce constraint C1130

psteinfeld:
The constraint states that "If the locality-spec DEFAULT ( NONE ) appears in a DO CONCURRENT statement; a variable that is a local or construct entity of a scope containing the DO CONCURRENT construct; and that appears in the block of the construct; shall have its locality explicitly specified by that statement."

 - In check-do.cc: Implemented the function
   CheckDefaultNoneImpliesExplicitLocality() to do the checking.  This involved
   adding the class DoConcurrentVariableEnforce to walk the DO loop's block
   looking for variable names.  I also cleaned up the code a little in
   CheckDoExpression() and EnforceConcurrentLoopControl().
 - Added the test dosemantics05.f90

### DIFF
--- a/lib/semantics/check-do.cc
+++ b/lib/semantics/check-do.cc
@@ -324,8 +324,9 @@ class DoConcurrentVariableEnforce {
 public:
   DoConcurrentVariableEnforce(
       SemanticsContext &context, parser::CharBlock doConcurrentSourcePosition)
-    : context_{context}, doConcurrentSourcePosition_{
-                             doConcurrentSourcePosition} {}
+    : context_{context},
+      doConcurrentSourcePosition_{doConcurrentSourcePosition},
+      blockScope_{context.FindScope(doConcurrentSourcePosition_)} {}
 
   template<typename T> bool Pre(const T &) { return true; }
   template<typename T> void Post(const T &) {}
@@ -335,10 +336,11 @@ public:
     if (const Symbol * symbol{name.symbol}) {
       if (IsVariableName(*symbol)) {
         const Scope &variableScope{symbol->owner()};
-        if (DoesScopeContain(&variableScope, GetBlockScope())) {
+        if (DoesScopeContain(&variableScope, blockScope_)) {
           context_.Say(name.source,
-              "Variable '%s' from enclosing scope in a DEFAULT(NONE) DO "
-              "CONCURRENT, must appear in a locality-spec"_err_en_US,
+              "Variable '%s' from an enclosing scope referenced in a DO "
+              "CONCURRENT with DEFAULT(NONE) must appear in a "
+              "locality-spec"_err_en_US,
               name.source);
         }
       }
@@ -346,21 +348,19 @@ public:
   }
 
 private:
-  const Scope &GetBlockScope() {
-    return context_.FindScope(doConcurrentSourcePosition_);
-  }
-
   SemanticsContext &context_;
   parser::CharBlock doConcurrentSourcePosition_;
+  const Scope &blockScope_;
 };  // class DoConcurrentVariableEnforce
 
-using CS = std::vector<const Symbol *>;
+using SymbolContainer = std::set<const Symbol *>;
 
 enum GatherWhichVariables { All, NotShared, Local };
 
-static CS GatherVariables(const std::list<parser::LocalitySpec> &localitySpecs,
+static SymbolContainer GatherVariables(
+    const std::list<parser::LocalitySpec> &localitySpecs,
     GatherWhichVariables which) {
-  CS symbols;
+  SymbolContainer symbols;
   for (auto &ls : localitySpecs) {
     auto names{std::visit(
         [=](const auto &x) {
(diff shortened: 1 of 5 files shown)


lib/semantics/check-do.cc
```diff
@@ -319,14 +319,42 @@ class DoConcurrentLabelEnforce {
...
+              "Variable '%s' from enclosing scope in a DEFAULT(NONE) DO "
+              "CONCURRENT, must appear in a locality-spec"_err_en_US,
+              name.source);
+        }
+      }
+    }
+  }
+
+private:
+  const Scope &GetBlockScope() {
```
klausler: This function is going to be called once for every variable name in the loop and it's going to return the same thing every time.

test/semantics/dosemantics05.f90
```diff
@@ -0,0 +1,67 @@
...
+    jvar = 5 ! OK, jvar appears in a locality spec
+    kvar = 5 ! OK, kvar appears in a locality spec
+
+!ERROR: Variable 'mvar' from enclosing scope in a DEFAULT(NONE) DO CONCURRENT, must appear in a locality-spec
+    mvar = 3.5
+  end do
+
+  x = 5.0  ! OK, we're not in a DO CONCURRENT
+  
+end subroutine s1
```
klausler: What about construct entities from ASSOCIATE & al.?  Can they appear in a nested DO CONCURRENT with a DEFAULT(NONE)?

lib/semantics/check-do.cc
```diff
@@ -319,14 +319,42 @@ class DoConcurrentLabelEnforce {
...
+  template<typename T> void Post(const T &) {}
+
+  // Check to see if the name is a variable from an enclosing scope
+  void Post(const parser::Name &name) {
+    if (const Symbol * symbol{name.symbol}) {
+      if (IsVariableName(*symbol)) {
+        const Scope &variableScope{symbol->owner()};
+        if (DoesScopeContain(&variableScope, GetBlockScope())) {
+          context_.Say(name.source,
+              "Variable '%s' from enclosing scope in a DEFAULT(NONE) DO "
```
klausler: "DEFAULT(NONE) DO CONCURRENT" is awkward, and the comma really shouldn't be there.

lib/semantics/check-do.cc
```diff
@@ -319,14 +319,42 @@ class DoConcurrentLabelEnforce {
...
+private:
+  const Scope &GetBlockScope() {
+    return context_.FindScope(doConcurrentSourcePosition_);
+  }
+
+  SemanticsContext &context_;
+  parser::CharBlock doConcurrentSourcePosition_;
+};  // class DoConcurrentVariableEnforce
+
+using CS = std::vector<const Symbol *>;
```
klausler: Why a `vector` instead of a `set`?

lib/semantics/check-do.cc
tskeith: How about this?
"Variable '%s' from an enclosing scope referenced in a DO CONCURRENT with DEFAULT(NONE) must appear in a locality-spec"

lib/semantics/check-do.cc
```diff
@@ -319,14 +319,42 @@ class DoConcurrentLabelEnforce {
...
+                             doConcurrentSourcePosition} {}
 
-struct GatherSymbols {
-  CS symbols;
-  template<typename T> constexpr bool Pre(const T &) { return true; }
-  template<typename T> constexpr void Post(const T &) {}
-  void Post(const parser::Name &name) { symbols.push_back(name.symbol); }
-};
+  template<typename T> bool Pre(const T &) { return true; }
+  template<typename T> void Post(const T &) {}
```
tskeith: It seems like overkill creating a whole new visitor just for this check. It adds 700 new Walk functions to the generated code.

Did you consider adding a Checker (or adding to an existing one)? You would need Enter/Leave functions to manage a stack of active DO CONCURRENTs with DEFAULT(NONE) and Enter on parser::Name to perform the check.

lib/semantics/check-do.cc
psteinfeld: Thanks, Peter.  I'll fix that.

test/semantics/dosemantics05.f90
psteinfeld: The standard states that the variable can be a "local or construct entity".  If it's from the enclosing scope, it must appear in a locality-spec.  I'll add tests for that.

lib/semantics/check-do.cc
psteinfeld: Thanks, Tim.  I'll make this change.

lib/semantics/check-do.cc
psteinfeld: I'm not sure.  That code was written before my time.  It doesn't look like the uses of ```CS``` require ordering.  I'll look into changing it.

lib/semantics/check-do.cc
psteinfeld: I'll look into adding another Checker.  There are additional semantic checks that we need to do for DO loops, which makes another checker more desirable.

psteinfeld: I created issue #624 for this.

lib/semantics/check-do.cc
```diff
@@ -335,32 +336,31 @@ class DoConcurrentVariableEnforce {
...
-    return context_.FindScope(doConcurrentSourcePosition_);
-  }
-
   SemanticsContext &context_;
   parser::CharBlock doConcurrentSourcePosition_;
+  const Scope &blockScope_;
 };  // class DoConcurrentVariableEnforce
 
-using CS = std::vector<const Symbol *>;
+using SymbolContainer = std::set<const Symbol *>;
```
tskeith: I think `SymbolSet` is a better name for this type.

psteinfeld: Thanks.  I'll make that change.

lib/semantics/check-do.cc
psteinfeld: @klausler , can you please look at my latest commit and see if it satisfies your concerns?  Note that this comment is blocking my ability to merge.

PR #614, opened 2019-07-31
 Improve handling of generics with same name as type or procedure

tskeith:
Create symbols for generics in a pre-pass over the specification part so
it is easier to handle cases when they have the same name as a derived
type or subprogram. This is done by calling `PreSpecificationConstruct`
on each `SpecificationConstruct` of a specification part before we
continue walking it. The generics symbols are created there and the same
mechanism will be used to handle forward references to derived types.

Report an error when the same name is used for a generic interface,
derived type, and subprogram.

Improve the error message issued when a procedure and generic interface
have the same name but the procedure is not a specific of the generic.

Change `SayAlreadyDeclared` to report the error on the second occurence
of the name when possible. This can arise for declarations the are
processed out of order, e.g. contained subprograms and generic interfaces.

Avoid multiple "already declared" errors for the case when a contained
subprogram has the same name as a declared entity. We first create the
symbol with SubprogramNameDetails, then replace it with the entity (and
report the error), then replace it with the real subprogram (and get the
error again). By setting and checking the error flag we avoid the second
error.

### DIFF
--- a/lib/semantics/resolve-names.cc
+++ b/lib/semantics/resolve-names.cc
@@ -430,8 +430,9 @@ public:
 
   // Special messages: already declared; referencing symbol's declaration;
   // about a type; two names & locations
-  void SayAlreadyDeclared(const SourceName &, Symbol &);
   void SayAlreadyDeclared(const parser::Name &, Symbol &);
+  void SayAlreadyDeclared(const SourceName &, Symbol &);
+  void SayAlreadyDeclared(const SourceName &, const SourceName &);
   void SayWithReason(
       const parser::Name &, Symbol &, MessageFixedText &&, MessageFixedText &&);
   void SayWithDecl(const parser::Name &, Symbol &, MessageFixedText &&);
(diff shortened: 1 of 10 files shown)


lib/semantics/resolve-names.cc
```diff
@@ -2237,7 +2199,14 @@ void InterfaceVisitor::CheckGenericProcedures(Symbol &generic) {
   ResolveSpecificsInGeneric(generic);
   auto &details{generic.get<GenericDetails>()};
   if (auto *proc{details.CheckSpecific()}) {
-    SayAlreadyDeclared(generic.name(), *proc);
+    auto msg{
```
klausler: Is this check deferred until it is certain that all of the specific procedures for the generic interface have been seen, including the cases where multiple `GENERIC` statements &/or generic interface blocks with the same generic name are used?

tskeith: Yes, this is called from `FinishSpecificationPart` at the end of the specification part, so they should all have been seen by then.

lib/semantics/resolve-names.cc
```diff
@@ -5027,7 +4996,77 @@ static bool NeedsExplicitType(const Symbol &symbol) {
...
+  }
+  GenericDetails genericDetails;
+  if (Symbol * existing{currScope().FindSymbol(symbolName)}) {
+    if (existing->has<GenericDetails>()) {
+      info.Resolve(existing);
+      return;  // already have generic, add to it
+    }
+    Symbol &ultimate{existing->GetUltimate()};
+    if (ultimate.has<GenericDetails>()) {
+      genericDetails.AddSpecificProcsFrom(ultimate);
```
jeanPerier: Should potential derived type/subroutine with the same name also be copied ?
Currently the following program gives error `error: 'foo' is not a derived type` for `type(foo) something`

```
module m1
  interface foo
    procedure :: f
  end interface
  type foo
    integer mbar
  end type
  contains
    function f()
    end function
end module

module m2
  use m1
  interface foo
    procedure :: f2
  end interface
  contains
    function f2(x)
    end function
end module

use m2
type(foo) something
end
```
It compiles fine if `foo` is not a generic in `m1`. NAG and gfortran and pgfortran compile this ok.

jeanPerier: The comment may also apply to other places that are "copying" genericDetails, where it seems to me it would make sens to also bring the type/subroutine e.g:
https://github.com/flang-compiler/f18/blob/1c32a289b53d787c0033ed45033db86afde33458/lib/semantics/resolve-names.cc#L1987
https://github.com/flang-compiler/f18/blob/1c32a289b53d787c0033ed45033db86afde33458/lib/semantics/resolve-names.cc#L1999

tskeith: Thanks, Jean, you are right, that should work. I've pushed a fix for it.

Issue #581, opened 2019-07-16
Bogus error about indistinguishable interfaces with extended PDTs


klausler:
```
module m
  type :: t(k)
    integer, kind :: k
  end type t
  type, extends(t) :: t2
  end type t2
  interface gen
    module procedure s1, s2
  end interface
 contains
  subroutine s1(x)
    type(t2(1)), intent(in) :: x
  end subroutine
  subroutine s2(x)
    type(t2(2)), intent(in) :: x
  end subroutine
end module
```

jeanPerier:
Closed by PR #615.

PR #615, opened 2019-08-02
Fix issue 581: extended derived type kind parameter comparison 

jeanPerier:
**Fix extended derived type kind compatibility check (#581):**

`HaveCompatibleKindParameters` was not considering the kind parameters of the parent types leading to false answers. This change fixes this by directly looking into the map of `ParamValue` of the `DeclTypeSpec` instead of going through the symbols of the derived type scope. This map contains all the parent and implicit type parameters.

**Fix ParamValue attribute (kind/len) of implicit type parameters:**
While testing fix for issue 581 it appeared that 'ParamValue`
implicit len parameters had incorrect attribute kind.

- Set correct attribute when creating `ParamValue` for implicit type parameter.

- Also set the correct attribute for character length `ParamValue` though this attribute is currently not used anywhere.

-  Change some `std::int64_t` to `common::ConstantSubscript` on the way.

### DIFF
--- a/lib/evaluate/type.cc
+++ b/lib/evaluate/type.cc
@@ -151,14 +151,11 @@ bool DynamicType::IsTypeCompatibleWith(const DynamicType &that) const {
 // corresponding kind type parameters of the type2?
 static bool IsKindCompatible(const semantics::DerivedTypeSpec &type1,
     const semantics::DerivedTypeSpec &type2) {
-  for (const auto &[name, symbol] : *type1.scope()) {
-    if (const auto *details{symbol->detailsIf<semantics::TypeParamDetails>()}) {
-      if (details->attr() == common::TypeParamAttr::Kind) {
-        const semantics::ParamValue *param1{type1.FindParameter(name)};
-        const semantics::ParamValue *param2{type2.FindParameter(name)};
-        if (!PointeeComparison(param1, param2)) {
-          return false;
-        }
+  for (const auto &[name, param1] : type1.parameters()) {
+    if (param1.isKind()) {
+      const semantics::ParamValue *param2{type2.FindParameter(name)};
+      if (!PointeeComparison(&param1, param2)) {
+        return false;
       }
     }
   }
--- a/lib/semantics/resolve-names.cc
+++ b/lib/semantics/resolve-names.cc
@@ -834,6 +834,8 @@ private:
   Symbol *MakeTypeSymbol(const parser::Name &, Details &&);
   bool OkToAddComponent(const parser::Name &, const Symbol * = nullptr);
   ParamValue GetParamValue(const parser::TypeParamValue &);
+  ParamValue GetLenParamValue(const parser::TypeParamValue &);
+  ParamValue GetLenParamValue(common::ConstantSubscript);
   Symbol &MakeCommonBlockSymbol(const parser::Name &);
   void CheckCommonBlockDerivedType(const SourceName &, const Symbol &);
   std::optional<MessageFixedText> CheckSaveAttr(const Symbol &);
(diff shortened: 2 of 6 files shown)


lib/semantics/resolve-names.cc
```diff
@@ -4058,6 +4060,19 @@ ParamValue DeclarationVisitor::GetParamValue(const parser::TypeParamValue &x) {
...
 
+ParamValue DeclarationVisitor::GetLenParamValue(
+    const parser::TypeParamValue &x) {
+  ParamValue param{GetParamValue(x)};
+  param.set_attr(common::TypeParamAttr::Len);
+  return param;
+}
+
+ParamValue DeclarationVisitor::GetLenParamValue(common::ConstantSubscript l) {
+  ParamValue param{l};
```
schweitzpgi: A lowercase 'l' is not the most readable variable name.

lib/evaluate/type.cc
```diff
@@ -150,14 +150,11 @@ bool DynamicType::IsTypeCompatibleWith(const DynamicType &that) const {
...
     const semantics::DerivedTypeSpec &type2) {
-  for (const auto &[name, symbol] : *type1.scope()) {
-    if (const auto *details{symbol->detailsIf<semantics::TypeParamDetails>()}) {
-      if (details->attr() == common::TypeParamAttr::Kind) {
-        const semantics::ParamValue *param1{type1.FindParameter(name)};
-        const semantics::ParamValue *param2{type2.FindParameter(name)};
-        if (!PointeeComparison(param1, param2)) {
-          return false;
-        }
+  for (const auto &[name, param1] : type1.parameters()) {
```
klausler: Please ensure that both PDTs have had their defaulted type parameters made explicit in their `ParamValue` lists.  I think that they do, but the representation has changed a few times during development.

lib/semantics/tools.cc
```diff
@@ -705,7 +705,10 @@ void ProcessParameterExpressions(
       if (paramValue != nullptr) {
         paramValue->SetExplicit(std::move(*expr));
       } else {
-        spec.AddParamValue(symbol->name(), ParamValue{std::move(*expr)});
+        ParamValue implicitParam{std::move(*expr)};
+        implicitParam.set_attr(details.attr());
+        spec.AddParamValue(
+            symbol->name(), ParamValue{std::move(implicitParam)});
```
tskeith: It would probably be worthwhile to add a constructor for `ParamValue` that takes an expr and `TypeParamAttr` (maybe defaulting to `Kind`). Then this would just be:
```
spec.AddParamValue(symbol->name(), ParamValue{std::move(*expr), details.attr()});
```
And you may decide you don't even need to add `GetLenParamValue()`.

lib/semantics/resolve-names.cc
jeanPerier: Yes, that was a bad name. I removed the function anyway based on Tim's comment.

lib/evaluate/type.cc
jeanPerier: Looking at the code, I think it is happening here:
https://github.com/flang-compiler/f18/blob/46d77ab006fd5a633ea9ab18b208ac9a7baee4f2/lib/semantics/tools.cc#L695
I have added tests that involved implicit kind parameters to ensure this.

lib/semantics/tools.cc
jeanPerier: Thanks, I have actually made this the only kind of public constructor accepting expression so that the attribute cannot be forgotten to be safer. I removed `GetLenParamValue`.

jeanPerier:
> The changes look good, and the tests pass.
> 
> It would be nice to have a test for this change, especially since issue #581 includes sample code.

It is a fair point, I added more tests playing with PDT instance distinguish-ability.

PR #617, opened 2019-08-02
Fix bug with use-associated base of component reference

### LINKED ISSUE #616: Incorrect message about invalid base for a component reference
% f18 -fdebug-semantics cl-test.f90
cl-test.f90:16:3: error: 'g' is an invalid base for a component reference
    g%x=  0.0_8
    ^

MODULE dm
  TYPE gt
    REAL(KIND=8)       :: x
  END TYPE gt
  TYPE(gt)        :: g
END

MODULE cm
  USE dm
  IMPLICIT NONE
CONTAINS
END

SUBROUTINE r()
  USE cm
  g%x=  0.0_8
END

tskeith:
When the base of a component reference (`a` in `a%b`) was use
associated, we were failing to follow it to the real symbol.

Fixes #616.

### DIFF
--- a/lib/semantics/resolve-names.cc
+++ b/lib/semantics/resolve-names.cc
@@ -4675,7 +4675,7 @@ const parser::Name *DeclarationVisitor::FindComponent(
   if (!base || !base->symbol) {
     return nullptr;
   }
-  auto &symbol{*base->symbol};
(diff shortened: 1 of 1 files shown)


tskeith:
> Should we add a test?

I thought it was too minor to bother with a test.

PR #620, opened 2019-08-04
[OpenMP] parse tree changes for declarative directives

ichoyjx:
1. Changes are to save provenance for directive names
2. check-omp-structure.* is updated to avoid assertion errors
3. Tests added now are only for the basic usages for the declarative directives,
   more complete examples will be added once we start implementing the semantics
   checks for declarative directives

Note that there is a potential issue for saving the provenance starting from
"!$OMP", which will be investigated separately.


lib/parser/parse-tree.h
```diff
@@ -3529,6 +3529,12 @@ struct OpenMPDeclareTargetConstruct {
   std::variant<WithClause, WithExtendedList, Implicit> u;
 };
 
+struct OpenMPDeclareTargetConstruct {
+  TUPLE_CLASS_BOILERPLATE(OpenMPDeclareTargetConstruct);
+  CharBlock source;
+  std::tuple<Verbatim, OpenMPDeclareTargetSpecifier> t;
```
tskeith: What is the difference between the source range in `source` and the one in `Verbatim`?

ichoyjx: For example, `!$omp declare target to(Q)`. The `Verbatim` is the directive `declare target` itself; the `source` is `declare target to(Q)`.

PR #630, opened 2019-08-06
Perform more checks on array-specs

tskeith:
The commit titled "Perform more checks on array-specs" is the real change.


lib/parser/grammar.h
```diff
@@ -1088,9 +1088,9 @@ TYPE_PARSER(construct<ExplicitCoshapeSpec>(
 // a need for forced look-ahead.
 TYPE_PARSER(
     construct<ArraySpec>(parenthesized(nonemptyList(explicitShapeSpec))) ||
+    construct<ArraySpec>(parenthesized(deferredShapeSpecList)) ||
```
klausler: A comment here about the need to attempt `deferredShapeSpecList` first to resolve a potential ambiguity with assumed shape specs would be great.

lib/semantics/type.cc
```diff
@@ -106,6 +106,30 @@ std::ostream &operator<<(std::ostream &o, const ShapeSpec &x) {
...
+  return CheckAll([](const ShapeSpec &x) {
+    return x.lbound().isDeferred() && x.ubound().isDeferred();
+  });
+}
+bool ArraySpec::IsImpliedShape() const {
+  return CheckAll([](const ShapeSpec &x) { return x.ubound().isAssumed(); });
+}
+bool ArraySpec::IsAssumedSize() const {
+  return !empty() &&
+      std::all_of(begin(), end() - 1,
```
klausler: `end() - 1` is the same as `back()`, at least when the list isn't empty.  You probably want `end() - 2` here and a length check, or a different implementation.

module/ieee_exceptions.f90
```diff
@@ -28,9 +28,9 @@ module ieee_exceptions
     ieee_inexact = ieee_flag_type(16)
 
   type(ieee_flag_type), parameter :: &
-    ieee_usual(:) = [ &
+    ieee_usual(*) = [ &
```
klausler: Were the colon specifications incorrect, bad style, or what?

psteinfeld:
I haven't looked at the changes yet, but I'm having trouble building.  

Here's my CMake command:
```cmake -D LLVM_DIR=/home/sw/thirdparty/clang/llvm8+gcc-8.2.0/install/lib/cmake/llvm -D CMAKE_BUILD_TYPE=Release ..```

When I build, I get a bunch of undefined references:

```
[ 92%] Linking CXX executable intrinsics-test
../../lib/evaluate/libFortranEvaluate.a(characteristics.cc.o): In function `Fortran::evaluate::characteristics::TypeAndShape::AcquireShape(Fortran::semantics::ObjectEntityDetails const&)':
characteristics.cc:(.text+0x47c0): undefined reference to `Fortran::semantics::ArraySpec::IsAssumedRank() const'
../../lib/evaluate/libFortranEvaluate.a(shape.cc.o): In function `Fortran::evaluate::GetExtent(Fortran::evaluate::FoldingContext&, Fortran::evaluate::NamedEntity const&, int)':
shape.cc:(.text+0x106c6): undefined reference to `Fortran::semantics::ArraySpec::IsAssumedSize() const'
../../lib/evaluate/libFortranEvaluate.a(shape.cc.o): In function `Fortran::evaluate::GetUpperBound(Fortran::evaluate::FoldingContext&, Fortran::evaluate::NamedEntity const&, int)':
shape.cc:(.text+0x10cb4): undefined reference to `Fortran::semantics::ArraySpec::IsAssumedSize() const'
```

File: test/semantics/modfile25.f90 [hunk 4]
```diff
@@ -15,51 +15,51 @@
...
-! integer(8),parameter::a1n(1_8:)=[Integer(8)::125_8,5_8,5_8]
-! integer(8),parameter::a1sn(1_8:)=[Integer(8)::3_8,1_8,1_8]
-! integer(8),parameter::ac1s(1_8:)=[Integer(8)::1_8]
-! integer(8),parameter::ac2s(1_8:)=[Integer(8)::3_8]
-! integer(8),parameter::ac3s(1_8:)=[Integer(8)::4_8]
-! integer(8),parameter::ac3bs(1_8:)=[Integer(8)::4_8]
-! integer(8),parameter::ac4s(1_8:)=[Integer(8)::36_8]
-! integer(8),parameter::ac5s(1_8:)=[Integer(8)::9_8]
-! integer(8),parameter::rss(1_8:)=[Integer(8)::10_8,9_8]
+! integer(8),parameter::a1s(1_8:*)=[Integer(8)::5_8,5_8,5_8]
```
klausler: Implied-shape array parameters can be explicit-shape in module files.  And everywhere else, once their initializers have been folded, yes?

lib/semantics/resolve-names.cc
```diff
@@ -2849,14 +2857,97 @@ Symbol &DeclarationVisitor::DeclareObjectEntity(
...
+    if (!isExplicit && !isImplied) {
+      Say(name,
+          "Named constant '%s' array must have explicit or implied shape"_err_en_US);
+      return false;
+    }
+  } else if (!IsAllocatableOrPointer(symbol) && !isExplicit) {
+    if (symbol.owner().IsDerivedType()) {  // C749
+      Say(name,
+          "Component array '%s' without ALLOCATABLE or POINTER attribute must"
+          " have explicit shape"_err_en_US);
```
psteinfeld: Should there be a ```return false;``` here?

lib/semantics/resolve-names.cc
```diff
@@ -2849,14 +2857,97 @@ Symbol &DeclarationVisitor::DeclareObjectEntity(
...
+    } else {  // C816
+      Say(name,
+          "Array '%s' without ALLOCATABLE or POINTER attribute must have"
+          " explicit shape"_err_en_US);
+    }
```
psteinfeld: Should there be a return false; here?

test/semantics/stop01.f90
```diff
@@ -18,11 +18,11 @@ program main
   integer, pointer :: p_i
   integer(kind = 1) :: invalid = 0
   integer, dimension(1:100) :: iarray
-  integer, dimension(1:100), pointer :: p_iarray
-  integer, allocatable, dimension(1:100) :: aiarray
+  integer, dimension(:), pointer :: p_iarray
```
psteinfeld: I notice that this file has an ARM copyright.  Should we be adding NVIDIA to it?

lib/semantics/type.cc
tskeith: `std::all_of(first, last, predicate)` tests the predicate on `[first, last)`. So this is checking the predicate on all but `back()` and then checking the upper bound of `back()` is assumed. (And `back()` is known to exist because of `!empty()`.)

lib/parser/grammar.h
tskeith: OK

lib/semantics/resolve-names.cc
tskeith: Yes

test/semantics/modfile25.f90
tskeith: Yes, once they are checked and folded that seems like a good idea.

test/semantics/stop01.f90
tskeith: Good question for @sscalpone. In this case there is not a lot of new IP going in so I'm not going to worry about it.

tskeith:
> I haven't looked at the changes yet, but I'm having trouble building.

It links on Linux now.

test/semantics/stop01.f90
sscalpone: NVIDIA policy is to add copyright if the change is substantial, a contribution worthy of recognition of ownership.

psteinfeld:
Thanks, Tim.  All looks good.

Issue #589, opened 2019-07-18
Bad "relational operands do not have comparable types" with TBP call


klausler:
```
module m
type,abstract :: t
contains
   procedure, nopass :: f
end type
contains
recursive integer function f(i)
 integer i
 if( i .eq. 1 ) then
   f = -2
  else
   f = 1 + f(i-1)
 endif
end function
end module

use m
type(t) :: x
print *, x%f(10) .eq. 7
end
```

Oddly, the error does not appear if the executable part of the function 'f' is simplified to just `f = i - 3`.

jeanPerier:
Starting investigation on this one.
Valgrind is whispering about some conditional jump or move depends on uninitialised value:
```
==6434==    by 0x5A6AFD: Fortran::common::Interval<char const*>::ExtendToCover(Fortran::common::Interval<char const*> const&) (interval.h:76)
==6434==    by 0x5873DC: Fortran::parser::CharBlock::ExtendToCover(Fortran::parser::CharBlock const&) (char-block.h:56)
==6434==    by 0x1E37F2B: Fortran::semantics::Scope::AddSourceRange(Fortran::parser::CharBlock const&) (scope.cc:252)
==6434==    by 0x276250C: bool Fortran::semantics::ScopeHandler::Pre<Fortran::parser::ActionStmt>(Fortran::parser::Statement<Fortran::parser::ActionStmt> const&) (resolve-names.cc:424)
```

I will start by following that lead.

jeanPerier:
Valgrind issue is a fake lead for this issue. It seams to be another issue with anonymous program.
Adding `program test` before `use m` in the example clears Valgrind errors.
I will treat it the Valgrind issue later in an other issue.

jeanPerier:
**Smallest reproducer:**
```
function f()
  real x,f
  x = f()
  f = 1.
end function
print*, f().eq.1.
end
```
Notes:

- the issue disappear if `x = f()` is removed.

- similar issues appear if `f` is called in the main program in situations where the type is checked (e.g: `print*, acos(f())`)

**First Analysis:**
The duality of recursive function names seems not yet handled in f18. Here is the chain of events that leads to the bug:
- During name resolution, when encountering the `parser::ProgramUnit` for `f`, the specifications of `f`, including its result type are handled by `ResolveNamesVisitor::ResolveSpecificationParts(ProgramTree &node)`. See:
[...]

PR #631, opened 2019-08-07
Prevent result symbol to be converted into function when they are called

jeanPerier:
This PR:
+ Fixes issue #589
+ Adds an error for empty subscript list in array reference

In name resolution, when skimming through the execution statement of a function, calls to the result symbol should not trigger the conversion of this symbol to a function symbol. The result is a data object and cannot be called unless it was explictly declared to be a procedure pointer.
Notably, recursive function calls cannot be made if RESULT was not used.

The symbol is prevented from being transformed into a function symbol by transforming it into an object before skimming through the executable statement. This is done after processing all the specifications so that if the result actually is a procedure pointer, the call to `ConvertToObjectEntity` introduced by this commit will not convert it to an object by mistake.

Also introduce a check when fixing mis-parsed function reference into array reference to verify the array reference has array subscripts. Currently this went uncaught. It is not possible to complain later in expressions because the subscript list of expression might be empty for unrelated error recovery reasons (e.g. if an entity of the wrong type appeared as subscripts).

### DIFF
--- a/lib/semantics/expression.cc
+++ b/lib/semantics/expression.cc
@@ -1830,6 +1830,35 @@ MaybeExpr ExpressionAnalyzer::Analyze(const parser::Expr::DefinedBinary &) {
   return std::nullopt;
 }
 
+static void CheckFuncRefToArrayElementRefHasSubscripts(
+    semantics::SemanticsContext &context,
+    const parser::FunctionReference &funcRef) {
+  // Emit message if the function reference fix will end-up an array element
+  // reference with no subscripts because it will not be possible to later tell
+  // the difference in expressions between empty subscript list due to bad
+  // subscripts error recovery or because the user did not put any.
+  if (std::get<std::list<parser::ActualArgSpec>>(funcRef.v.t).empty()) {
+    auto &proc{std::get<parser::ProcedureDesignator>(funcRef.v.t)};
+    const auto *name{std::get_if<parser::Name>(&proc.u)};
+    name = {name
+            ? name
+            : &std::get<parser::ProcComponentRef>(proc.u).v.thing.component};
+    auto &msg{context.Say(funcRef.v.source,
+        "Reference to array '%s' with empty subscript list"_err_en_US,
+        name->source)};
+    if (name->symbol) {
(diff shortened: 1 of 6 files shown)


lib/semantics/expression.cc
```diff
@@ -1831,6 +1831,35 @@ MaybeExpr ExpressionAnalyzer::Analyze(const parser::Expr::DefinedBinary &) {
...
+    semantics::SemanticsContext &context,
+    const parser::FunctionReference &funcRef) {
+  // Emit message if the function reference fix will end-up an array element
+  // reference with no subscripts because it will not be possible to later tell
+  // the difference in expressions between empty subscript list due to bad
+  // subscripts error recovery or because the user did not put any.
+  if (std::get<std::list<parser::ActualArgSpec>>(funcRef.v.t).empty()) {
+    auto &proc{std::get<parser::ProcedureDesignator>(funcRef.v.t)};
+    const auto *name{std::get_if<parser::Name>(&proc.u)};
+    name = {name
```
klausler: I am surprised by these braces.  They may be affecting clang-format.  The statement might be more clear as an `if` (or with GNU's `?:` extension that we probably should not use).

lib/semantics/tools.cc
```diff
@@ -790,4 +790,16 @@ static Symbol &InstantiateSymbol(
   return result;
 }
 
+bool IsFunctionResultWithSameNameAsFunction(const Symbol &symbol) {
+  if ((symbol.has<semantics::ObjectEntityDetails>() &&
```
klausler: Consider defining a function in tools.h (`IsFunctionResult(const Symbol &)`) and calling it here.  It might be a useful thing elsewhere and would hide some of the internal details.

psteinfeld: I agree with Peter.

lib/semantics/expression.cc
jeanPerier: It was weird and clang-format made it even weirder indeed... I changed it to a simple `if`.

lib/semantics/tools.cc
jeanPerier: Done

PR #632, opened 2019-08-07
[OpenMP] parse tree changes for `OpenMPBlockConstruct`

ichoyjx:
1. merge `Workshare` and `Single` into `OpenMPBlockConstruct` because
   they both accept structured-block and syntax is similar to other block
   directives.

2. `OpenMPBlockConstruct` changes to structure like `{Begin, Block, End}`,
   where `Begin` and `End` are tuple of `{Directive, ClauseList}`.

3. Updated the check-omp-structure.* for necessary parts. Added all the END
   directive enumeration types that may have clauses.

More tests will be added during Semantics.

TODO: investigate `Sections` related directives, see whether they can be merged


lib/semantics/check-omp-structure.h
```diff
@@ -26,19 +26,20 @@
...
 
-ENUM_CLASS(OmpDirective, PARALLEL, DO, SECTIONS, SECTION, SINGLE, END_SINGLE,
-    WORKSHARE, SIMD, DECLARE_SIMD, DO_SIMD, TASK, TASKLOOP, TASKLOOP_SIMD,
-    TASKYIELD, TARGET_DATA, TARGET_ENTER_DATA, TARGET_EXIT_DATA, TARGET,
-    TARGET_UPDATE, DECLARE_TARGET, TEAMS, DISTRIBUTE, DISTRIBUTE_SIMD,
-    DISTRIBUTE_PARALLEL_DO, DISTRIBUTE_PARALLEL_DO_SIMD, PARALLEL_DO,
-    PARALLEL_SECTIONS, PARALLEL_WORKSHARE, PARALLEL_DO_SIMD, TARGET_PARALLEL,
-    TARGET_PARALLEL_DO, TARGET_PARALLEL_DO_SIMD, TARGET_SIMD, TARGET_TEAMS,
-    TEAMS_DISTRIBUTE, TEAMS_DISTRIBUTE_SIMD, TARGET_TEAMS_DISTRIBUTE,
+ENUM_CLASS(OmpDirective, PARALLEL, DO, SECTIONS, SECTION, SINGLE, WORKSHARE,
```
klausler: Consider applying alphabetical order.

ichoyjx: Updated, thanks!

lib/semantics/check-omp-structure.cc
```diff
@@ -238,12 +206,51 @@ void OmpStructureChecker::Enter(const parser::OmpBlockDirective &x) {
         OmpClause::IF, OmpClause::NUM_THREADS, OmpClause::PROC_BIND};
     SetContextAllowedOnce(allowedOnce);
   } break;
+  // 2.7.3 single-clause -> private-clause |
+  //                        firstprivate-clause
+  case parser::OmpBlockDirective::Directive::Single: {
+    SetContextDirectiveEnum(OmpDirective::SINGLE);
+    OmpClauseSet allowed{OmpClause::PRIVATE, OmpClause::FIRSTPRIVATE};
+    SetContextAllowed(allowed);
```
tskeith: If you replaced these two lines with `SetContextAllowed({OmpClause::PRIVATE, OmpClause::FIRSTPRIVATE});` you wouldn't need the block around the body of the `case`.

lib/semantics/check-omp-structure.cc
```diff
@@ -88,7 +88,8 @@ void OmpStructureChecker::Leave(const parser::OpenMPLoopConstruct &) {
 }
 
 void OmpStructureChecker::Enter(const parser::OpenMPBlockConstruct &x) {
-  const auto &dir{std::get<parser::OmpBlockDirective>(x.t)};
+  const auto &beginDir{std::get<parser::OmpBeginBlockDirective>(x.t)};
+  const auto &dir{std::get<parser::OmpBlockDirective>(beginDir.t)};
```
jeanPerier: You may want to check here or in `Leave(const parser::OpenMPBlockConstruct &)` that the `OmpBeginBlockDirective` and `OmpEndBlockDirective` contain the same `Directive`.
The following program is now accepted after the parse-tree changes:

```
real :: a
!$omp parallel
a =1.
!$omp end single
end
```
It maybe part of the tests you intended to do in a later PR, which is fine by me.

lib/semantics/check-omp-structure.cc
```diff
@@ -238,12 +206,51 @@ void OmpStructureChecker::Enter(const parser::OmpBlockDirective &x) {
...
   default:
     // TODO others
     break;
   }
 }
 
+void OmpStructureChecker::Enter(const parser::OmpEndBlockDirective &x) {
+  const auto &dir{std::get<parser::OmpBlockDirective>(x.t)};
+  SaveContextDirectiveSource();
+  SetContextDirectiveSource(dir.source);
```
jeanPerier: You may want to simply pop and push a new `OmpContext` here.
Rational:
-   the `allowedClauses`/ `allowedOnceClausesand` seen clauses from the `OmpEndBlockDirective` are still set here, and this can lead to weird thing in the `default` like accepting:
```
real :: a
!$omp parallel
a =1.
!$omp end parallel copyin(a)
end
```
-  I have not looked in detail, but I am a bit surprised that there is a need to save/restore the source location of the `OmpBeginBlockDirective` here given the `OmpContext` stack is basically pop just after restoring it because `Leave(const parser::OmpEndBlockDirective &)` is directly followed by `Leave(const parser::OpenMPBlockConstruct &)`.

lib/semantics/check-omp-structure.cc
ichoyjx: Thanks! It is already listed in the plan doc (4a). I will do the matching checks along with `LoopConstruct`.

lib/semantics/check-omp-structure.cc
ichoyjx: Thanks... For the first item, I removed the pushing-new-`OmpContext` in the latest commit on Tim's review because there is no "real" context that needs to be pushed (although it was handy). For the second item, I thought about it too, I wasn't sure whether the source location info might be useful in the `Leave` of `OpenMPBlockConstruct`. I will fix both of them.

ichoyjx: @tskeith I have committed the change for clearing the context before processing `OmpEndBlockDirective`. Thanks for the advice.

lib/semantics/check-omp-structure.h
```diff
@@ -146,16 +145,13 @@ class OmpStructureChecker : public virtual BaseChecker {
     CHECK(!ompContext_.empty());
     return ompContext_.back();
   }
+  void ClearContext() {
+    CHECK(!ompContext_.empty());
+    ompContext_.back() = {};
```
tskeith: I think this will set the `directive` data member to `OmpDirective::PARALLEL` because it will be zero-initialized and that is the zero value of `OmpDirective`. That doesn't seem like a good default value.

It would be better to make `directive` have type `std::optional<OmpDirective>`. Then access it as `GetContext().directive.value()` so if it hasn't been set you'll get a runtime error and not `OmpDirective::PARALLEL`. Then maybe put that into a member function so you can get the directive with `GetContextDirective()`.

tskeith: By the way, this problem already exists. The version of `PushContext()` that doesn't set the directive also leaves it in a questionable state until it is set.

Issue #604, opened 2019-07-25
CHECK failure in scope.h(94)


psteinfeld:
Here's the code that produces the failure:
```
subroutine s8()

  block
    import, all
  end block

end subroutine s8
```

jeanPerier:
Fixed by PR #634.

PR #634, opened 2019-08-07
Fix a crash with IMPORT in BLOCK (issue 604)

jeanPerier:
Fixes issue #604.
In `CheckImports`, do not call `scope.name()` without checking that  the scope is associated to a symbol (and therefore has a name). It is legal to put imports in scopes that are not linked to symbols, such as BLOCK.

Add a comment in `scope.h` to not call `name()` without knowing it is OK.

### DIFF
--- a/lib/semantics/mod-file.cc
+++ b/lib/semantics/mod-file.cc
@@ -36,14 +36,15 @@ using namespace parser::literals;
 // are using another encoding.
 static constexpr auto magic{"\xef\xbb\xbf!mod$ v1 sum:"};
 
-static const SourceName *GetSubmoduleParent(const parser::Program &);
+static const std::optional<SourceName> GetSubmoduleParent(
+    const parser::Program &);
 static std::string ModFilePath(const std::string &dir, const SourceName &,
     const std::string &ancestor, const std::string &suffix);
 static std::vector<const Symbol *> CollectSymbols(const Scope &);
 static void PutEntity(std::ostream &, const Symbol &);
 static void PutObjectEntity(std::ostream &, const Symbol &);
 static void PutProcEntity(std::ostream &, const Symbol &);
-static void PutPassName(std::ostream &, const SourceName *);
+static void PutPassName(std::ostream &, const std::optional<SourceName> &);
 static void PutTypeParam(std::ostream &, const Symbol &);
 static void PutEntity(std::ostream &, const Symbol &, std::function<void()>);
 static void PutInit(std::ostream &, const Symbol &, const MaybeExpr &);
@@ -488,7 +489,7 @@ void PutProcEntity(std::ostream &os, const Symbol &symbol) {
   os << '\n';
 }
 
-void PutPassName(std::ostream &os, const SourceName *passName) {
+void PutPassName(std::ostream &os, const std::optional<SourceName> &passName) {
   if (passName) {
     PutLower(os << ",pass(", passName->ToString()) << ')';
   }
@@ -713,7 +714,7 @@ Scope *ModFileReader::Read(const SourceName &name, Scope *ancestor) {
   Scope *parentScope;  // the scope this module/submodule goes into
   if (!ancestor) {
     parentScope = &context_.globalScope();
-  } else if (auto *parent{GetSubmoduleParent(*parseTree)}) {
+  } else if (std::optional<SourceName> parent{GetSubmoduleParent(*parseTree)}) {
     parentScope = Read(*parent, ancestor);
   } else {
     parentScope = ancestor;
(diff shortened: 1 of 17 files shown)


lib/semantics/resolve-names.cc
```diff
@@ -5118,7 +5118,7 @@ void ResolveNamesVisitor::CheckImports() {
     // C8102: all entities in host must not be hidden
     for (const auto &pair : scope.parent()) {
       auto &name{pair.first};
-      if (name != scope.name()) {
+      if (!scope.GetSymbol() || name != scope.name()) {
```
klausler: This is a little too obscure.  It would be better to use the Symbol pointer returned from `GetSymbol()` and, when not null, test its name.  Or change `Scope::name()` to be a real accessor, actually -- it is named like one, and accessors should never crash; right now, it's dangerous.  Maybe it should be renamed and modified to return an optional result -- that seems best to me.

lib/semantics/scope.h
```diff
@@ -89,6 +89,8 @@ class Scope {
   const Symbol *GetSymbol() const;
   const Scope *GetDerivedTypeParent() const;
 
+  // It is only safe to call name() for kind of scopes for which GetSymbol
```
klausler: See above.  The real problem is that `name()` looks like an accessor and clients shouldn't have to worry about crashing.

`std::optional<SourceName> GetName() const { ... }` would be better.

tskeith: I agree, though I think the return type should be `const SourceName *` because other member functions of Scope with optional return values return pointers (e.g. `symbol()`, `GetDerivedTypeParent()`).

Clients can get the old behavior with `DEREF(scope.GetName())` when they know the scope has a name.

klausler: In this particular case, a `SourceName` is small, and only value semantics are needed, I think.

jeanPerier: OK, I think the most important here is to be consistent throughout  f18 regarding how "maybe" `SourceName` are returned. Here is a quick survey of the existing code base:

- 4 functions are returning `const SourceName *` (8 bytes, but it is a pointer)

- 1 is returning `std::optional<SourceName>` (24 bytes)

Should I transform everything to `std::optional<SourceName>` or go with the existing majority and use `const SourceName *` everywhere ?


Detailed Survey:


`const SourceName *GetSubmoduleParent(const parser::Program &);`
https://github.com/flang-compiler/f18/blob/a1e26df148c4d480178aa8b765aad4ceaaa5eae2/lib/semantics/mod-file.cc#L39
`const SourceName *MessageHandler::currStmtSource();`
https://github.com/flang-compiler/f18/blob/a1e26df148c4d480178aa8b765aad4ceaaa5eae2/lib/semantics/resolve-names.cc#L100
`const SourceName *BaseVisitor::currStmtSource();`
https://github.com/flang-compiler/f18/blob/a1e26df148c4d480178aa8b765aad4ceaaa5eae2/lib/semantics/resolve-names.cc#L140
`const SourceName *WithPassArg::passName();`
https://github.com/flang-compiler/f18/blob/009db27bfaca2a84b0795c0d977c68cd8b57ce88/lib/semantics/symbol.h#L186


`std::optional<SourceName> DerivedTypeDetails::GetParentComponentName();`
https://github.com/flang-compiler/f18/blob/009db27bfaca2a84b0795c0d977c68cd8b57ce88/lib/semantics/symbol.h#L233

klausler: Are you sure that all of those APIs have value semantics?  Some might be returning optional references encoded as pointers.  It might be hard to tell.

It's okay if you don't change all of these APIs, just this one.  We can change other APIs to value semantics later when visiting that code.

tskeith: Any of these could return `std::optional<SourceName>` or `const SourceName *`. I think it is good for them to be consistent, and also consistent with `symbol()`, which returns `const Symbol *` because it can't return by value.

I don't see any benefit to using `std::optional<SourceName>`.

klausler: `symbol()` is returning an optional reference to an object, not a value.  `GetName()` is returning an optional value, not a reference.  The benefit of making a distinction between references and values is that references to objects raise questions about ownership and expiration; values results don't.

jeanPerier: Comment treated in latest changes:
I have changed all `SourceName` to use value semantics by replacing all `SourceName *` by `std::optional<SourceName>`. This includes both some function return types and some class member types. I also changed argument types to be consistent.

`SourceName Scope::name()` is changed to `std::optional<SourceName> Scope::GetName()` and related call sites modified.

lib/semantics/mod-file.cc
```diff
@@ -36,14 +36,15 @@ using namespace parser::literals;
 // are using another encoding.
 static constexpr auto magic{"\xef\xbb\xbf!mod$ v1 sum:"};
 
-static const SourceName *GetSubmoduleParent(const parser::Program &);
+static const std::optional<SourceName> GetSubmoduleParent(
```
tskeith: `const` doesn't add anything here.

jeanPerier: Thanks, I have removed the leftover `const`

Issue #547, opened 2019-07-03
IMPLICIT in module subroutine doesn't work with IMPLICIT NONE in module


klausler:
```
module m
  implicit none
contains
  subroutine s
    implicit real (a-h)
    h = 1
  end subroutine
end module
```

fail semantics checking with "No explicit type declared for 'h'".

jeanPerier:
Fixed by PR #635.

PR #635, opened 2019-08-08
Fix IMPLICIT issue 547

jeanPerier:
Fix issue #547 and refactor `ImplicitRules` a bit.
In name resolution, the `IMPLICIT NONE` (for type) is represented by `isImplicitNoneType_` in the `ImplicitRules` class.
So far, it was a ternary with (I guess) the following intention:
- unset -> nothing about implicit in this scope, look into parent scope `ImplicitRules`
- set true -> There is an IMPLICIT NONE for types in this scope.
- set to false -> There is an IMPLICIT statement mapping types in this scope.

However, it was never set to false, so the IMPORT NONE of parent scopes was "leaking" when it should not.
Rather than just setting this to false, this PR actually change `ImplicitRules` a little based on the observation that the default mappings were actually not entered in the type `map_` as the comment suggested but hard coded  in `GetType`. So the `set to false` can be deduced from `!(set to true) && !map_.empty()`. Also, I do not think a ternary for `isImplicitNoneExternal_` is necessary as it can only be set to true and always "leak" to hosted scopes.

So both ternary are transformed to `bool` and related functions modified.

Last, `GetType` is modified to check `isImplicitNoneType_` because IMPLCIT NONE of parents should be considered when recursively calling `ImplicitRules` parents `GetType`.
This make the `isImplicitNone` redundant in `ApplyImplicitRules`. Instead, `GetImplicitType` is modified to not emit an error message and `ApplyImplicitRules` emits all error messages when it gets a nullptr from  `GetImplicitType`.

### DIFF
--- a/lib/semantics/resolve-names.cc
+++ b/lib/semantics/resolve-names.cc
@@ -73,8 +73,10 @@ public:
   void set_inheritFromParent(bool x) { inheritFromParent_ = x; }
   // Get the implicit type for identifiers starting with ch. May be null.
   const DeclTypeSpec *GetType(char ch) const;
-  // Record the implicit type for this range of characters.
-  void SetType(const DeclTypeSpec &type, parser::Location lo, parser::Location);
+  // Record the implicit type for the range of characters [fromLetter,
+  // toLetter].
+  void SetTypeMapping(const DeclTypeSpec &type, parser::Location fromLetter,
+      parser::Location toLetter);
 
 private:
   static char Incr(char ch);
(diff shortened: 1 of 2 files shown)


lib/semantics/resolve-names.cc
```diff
@@ -1757,7 +1760,10 @@ static bool NeedsType(const Symbol &symbol) {
 }
 void ScopeHandler::ApplyImplicitRules(Symbol &symbol) {
   if (NeedsType(symbol)) {
-    if (isImplicitNoneType()) {
+    if (const auto *type{GetImplicitType(symbol)}) {
```
klausler: `DeclTypeSpec` might be more clear than `auto` here.

lib/semantics/resolve-names.cc
```diff
@@ -1757,7 +1760,10 @@ static bool NeedsType(const Symbol &symbol) {
...
+      symbol.set(Symbol::Flag::Implicit);
+      symbol.SetType(*type);
+    } else {
```
vjayathirtha-nv: You could make this an `if-else-if` chain instead.

lib/semantics/resolve-names.cc
```diff
@@ -74,18 +74,19 @@ class ImplicitRules {
   // Get the implicit type for identifiers starting with ch. May be null.
   const DeclTypeSpec *GetType(char ch) const;
   // Record the implicit type for this range of characters.
-  void SetType(const DeclTypeSpec &type, parser::Location lo, parser::Location,
-      bool isDefault = false);
+  void SetType(const DeclTypeSpec &type, parser::Location lo, parser::Location);
```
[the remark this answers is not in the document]
jeanPerier: I have changed the function name to `SetTypeMapping` and the `parser::Location` arguments to `fromLetter` and  `toLetter`. Is this more helpful ?

lib/semantics/resolve-names.cc
jeanPerier: Done.

lib/semantics/resolve-names.cc
jeanPerier: Thanks, done.

PR #637, opened 2019-08-08
Tests for DO loop semantics

psteinfeld:
These are tests for checks that are already implemented, but for which we
did not have tests.

### DIFF
--- a/test/semantics/doconcurrent01.f90
+++ b/test/semantics/doconcurrent01.f90
@@ -14,7 +14,7 @@
 !
 ! C1141
 ! A reference to the procedure IEEE_GET_FLAG, IEEE_SET_HALTING_MODE, or
-! IEEE_GET_- HALTING_MODE from the intrinsic module IEEE_EXCEPTIONS, shall not
+! IEEE_GET_HALTING_MODE from the intrinsic module IEEE_EXCEPTIONS, shall not
 ! appear within a DO CONCURRENT construct.
 !
 ! C1137
@@ -25,27 +25,6 @@
 !
 ! (11.1.7.5), paragraph 4
 ! In a DO CONCURRENT, can't have an i/o statement with an ADVANCE= specifier
-
-module ieee_exceptions
-  interface
-     subroutine ieee_get_flag(i, j)
-       integer :: i, j
-     end subroutine ieee_get_flag
-     subroutine ieee_get_halting_mode(i, j)
-       integer :: i
-       logical :: j
-     end subroutine ieee_get_halting_mode
-     subroutine ieee_set_halting_mode(i, j)
-       integer :: i
-       logical :: j
-     end subroutine ieee_set_halting_mode
-  end interface
-end module ieee_exceptions
-
-module iso_fortran_env
-  type :: team_type
-  end type
-end
 
 subroutine do_concurrent_test1(i,n)
   implicit none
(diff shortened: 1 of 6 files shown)


test/semantics/doconcurrent01.f90
```diff
@@ -49,11 +65,22 @@ subroutine do_concurrent_test2(i,j,n,flag)
   use iso_fortran_env, only: team_type
   implicit none
   integer :: i, n, flag, flag2
+  logical :: halting
   type(team_type) :: j
   do concurrent (i = 1:n)
     change team (j)
+!ERROR: call to impure subroutine in DO CONCURRENT not allowed
+!ERROR: IEEE_GET_FLAG not allowed in DO CONCURRENT
```
klausler: These symbols aren't coming from the intrinsic module, so they shouldn't  trigger errors that are particular to intrinsic modules.

test/semantics/dosemantics08.f90
```diff
@@ -0,0 +1,27 @@
...
+!
+! Unless required by applicable law or agreed to in writing, software
+! distributed under the License is distributed on an "AS IS" BASIS,
+! WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+! See the License for the specific language governing permissions and
+! limitations under the License.
+
+! C1138 -- 
+! A branch (11.2) within a DO CONCURRENT construct shall not have a branch
+! target that is outside the construct.
```
klausler: You can also branch with `EXIT` and `CYCLE`.

test/semantics/doconcurrent01.f90
psteinfeld: Will there be a way to distinguish between the actual intrinsic modules and use defined modules with the same names?

klausler: Yes, we have to support `USE, INTRINSIC` and `USE, NON_INTRINSIC` too.

test/semantics/dosemantics08.f90
psteinfeld: Yep.  There were already tests for those.

test/semantics/doconcurrent01.f90
```diff
@@ -11,21 +11,34 @@
...
-! CHECK: RETURN not allowed in DO CONCURRENT
-! CHECK: call to impure subroutine in DO CONCURRENT not allowed
-! CHECK: IEEE_GET_FLAG not allowed in DO CONCURRENT
-! CHECK: ADVANCE specifier not allowed in DO CONCURRENT
-! CHECK: SYNC ALL
-! CHECK: SYNC IMAGES
+!
+! C1141
+! A reference to the procedure IEEE_GET_FLAG, IEEE_SET_HALTING_MODE, or
+! IEEE_GET_- HALTING_MODE from the intrinsic module IEEE_EXCEPTIONS, shall not
```
vjayathirtha-nv: this looks like a typo. `IEEE_GET_HALTING_MODE`

test/semantics/doconcurrent01.f90
```diff
@@ -11,21 +11,34 @@
...
+     subroutine ieee_get_halting_mode(i, j)
+       integer :: i
+       logical :: j
+     end subroutine ieee_get_halting_mode
+     subroutine ieee_set_halting_mode(i, j)
+       integer :: i
+       logical :: j
+     end subroutine ieee_set_halting_mode
   end interface
 end module ieee_exceptions
```
tskeith: It would be better to add `ieee_get_halting_mode` and `ieee_set_halting_mode` to `module/ieee_exceptions.f90` and use that in this test. Eventually that will happen and we'll have to remember to remove it here.

test/semantics/doconcurrent01.f90
psteinfeld: Thanks for catching this!

PR #638, opened 2019-08-08
Handle forward references to procedures and interfaces

klausler:
Add a pass to name resolution to allow for forward references to procedures and interfaces to appear in pointer initializations, procedure component initializations, and type-bound procedure bindings.

(The new `test/semantics/symbol15.f90` walks through the cases and might be worth reading first.)

Also fix `Scope::FindSymbol()` (actually, `CanImport()`) to ensure that symbols from the global scope don't leak into independent compilation units.

Also add a predicate `IsSaved()` to check for the SAVE attribute, including conditions that imply the SAVE attribute.

### DIFF
--- a/lib/semantics/resolve-names.cc
+++ b/lib/semantics/resolve-names.cc
@@ -858,7 +858,7 @@ private:
   bool HandleUnrestrictedSpecificIntrinsicFunction(const parser::Name &);
   const parser::Name *FindComponent(const parser::Name *, const parser::Name &);
   void CheckInitialDataTarget(const Symbol &, const SomeExpr &, SourceName);
-  void CheckInitialProcTarget(const Symbol &, const parser::Name &);
+  void CheckInitialProcTarget(const Symbol &, const parser::Name &, SourceName);
   void Initialization(const parser::Name &, const parser::Initialization &,
       bool inComponentDecl);
   bool PassesLocalityChecks(const parser::Name &name, Symbol &symbol);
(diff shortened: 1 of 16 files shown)


lib/semantics/resolve-names.cc
```diff
@@ -3410,25 +3418,45 @@ void DeclarationVisitor::Post(
...
     }
     if (auto *s{MakeTypeSymbol(bindingName, ProcBindingDetails{*procedure})}) {
       SetPassNameOn(*s);
     }
   }
 }
 
+void DeclarationVisitor::CheckBindings(
+    const parser::TypeBoundProcedureStmt::WithoutInterface &tbps) {
+  CHECK(currScope().kind() == Scope::Kind::DerivedType);
```
tskeith: This condition could be written `currScope().IsDerivedType()`.

lib/semantics/symbol.h
```diff
@@ -204,8 +204,13 @@ class ProcEntityDetails : public EntityDetails, public WithPassArg {
   void set_interface(const ProcInterface &interface) { interface_ = interface; }
   inline bool HasExplicitInterface() const;
 
+  std::optional<const Symbol *> init() const { return init_; }
+  void set_init(const Symbol &symbol) { init_ = &symbol; }
+  void set_init(std::nullptr_t) { init_ = nullptr; }
+
 private:
   ProcInterface interface_;
+  std::optional<const Symbol *> init_;  // if present but null => NULL()
```
tskeith: I think the comment should go with `init()` and `set_init()`. The distinction between nullopt and an optional holding nullptr is pretty subtle.

lib/semantics/resolve-names.cc
klausler: Sure.  And it's probably preferable, in case some future change decides to discriminate between derived types, PDT instantiations, and some future standard feature for generics.  Should `kind()` become private?

lib/semantics/symbol.h
klausler: Sure.

PR #641, opened 2019-08-09
[OpenMP] parse tree changes for Critical Construct

ichoyjx:
Very straightforward changes: add source provenance for directive itself and renaming.

The test added is only for basic correctness check.


ichoyjx:
@Peter, could you please take a quick look? Thanks.

lib/parser/openmp-grammar.h
```diff
@@ -439,18 +439,18 @@ TYPE_PARSER(construct<OpenMPAtomicConstruct>(Parser<OmpAtomicRead>{}) ||
...
-    construct<OmpEndCritical>(maybe(parenthesized(name))) / endOmpLine)
-
-TYPE_PARSER(
-    sourced(construct<OpenMPCriticalConstructDirective>(
-        "CRITICAL" >> maybe(parenthesized(name)),
-        maybe("HINT" >> construct<OpenMPCriticalConstructDirective::Hint>(
-                            parenthesized(constantExpr))))) /
+TYPE_PARSER(startOmpLine >>
+    sourced(construct<OmpEndCriticalDirective>(
+        verbatim("END CRITICAL"_tok), maybe(parenthesized(name)))) /
```
klausler: If the space is required in `END CRITICAL`, use `_sptok`, not `_tok`.

ichoyjx: Thanks! Based on the OpenMP spec, here the space is optional.

Issue #643, opened 2019-08-11
Unexpected error when using PRESENT


sscalpone:
% f18 -fparse-only -fdebug-semantics /tmp/b.f90
b.f90:9:8: error: Typeless item not allowed for 'a=' argument
    if ( present(cmp) ) continue
         ^^^^^^^

subroutine h(cmp)
  interface
     function cmp(a) result(r)
       logical :: r
       real, intent(in) :: a
     end
  end interface
  optional cmp
  if ( present(cmp) ) continue
end

sscalpone:
Fixed. Thanks

PR #646, opened 2019-08-12
Fix bug #643

klausler:
Allow arguments that don't have value types to a few intrinsics (PRESENT, LOC, ASSOCIATED) as long as they're not BOZ literals.


klausler:
> My test case worked. Are you going to add a test? Will we need the same for %ref, %val, and %loc?

Those three extensions don't go through the intrinsics table.  I wasn't going to add a specific test.

PR #650, opened 2019-08-12
Fix bugs writing .mod files

tskeith:
See individual commits.

When a `.mod` file is compiled as Fortran (with `-fbackslash`) it should produce exactly the same `.mod` file again. There are no more known cases where that does not happen.

### DIFF
--- a/lib/semantics/symbol.cc
+++ b/lib/semantics/symbol.cc
@@ -247,6 +247,13 @@ bool Symbol::CanReplaceDetails(const Details &details) const {
   }
 }
 
+// Usually a symbol's name is the first occurrence in the source, but sometimes
+// we want to replace it with one at a different location (but same characters).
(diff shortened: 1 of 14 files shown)


lib/semantics/symbol.cc
```diff
@@ -247,6 +247,15 @@ bool Symbol::CanReplaceDetails(const Details &details) const {
   }
 }
 
+// Usually a symbol's name is the first occurrence in the source, but sometimes
+// we want to replace it with one at a different location (but same characters).
+void Symbol::ReplaceName(const SourceName &name) {
+  CHECK(name == name_);
+  if (name.begin() != name_.begin()) {
```
klausler: You can just assign `name_ = name;` unconditionally.

tskeith: Yes, good point.

PR #651, opened 2019-08-12
Create a throwaway wrapper for the throwaway driver.

sscalpone:
This change makes the throwaway driver more useful as a driver for other compilers because it changes the suffix used for module files to .fmf, including the module files that are generated as part of the build, thus avoiding potential naming collisions with .mod files.  The wrapper is called flang. It also sets the f18 -I option to point to the directory containing the fmf files for the predefined modules.

### DIFF
--- a/tools/f18/CMakeLists.txt
+++ b/tools/f18/CMakeLists.txt
@@ -50,15 +50,26 @@ set(MODULES
 
 # Create module files directly from the top-level module source directory
 foreach(filename ${MODULES})
-  add_custom_command(OUTPUT ${CMAKE_CURRENT_BINARY_DIR}/include/${filename}.mod
-    COMMAND f18 -fparse-only -fdebug-semantics ${PROJECT_SOURCE_DIR}/module/${filename}.f90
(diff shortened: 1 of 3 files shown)


tools/f18/CMakeLists.txt
```diff
@@ -50,15 +50,26 @@ set(MODULES
...
 install(TARGETS f18 f18-parse-demo DESTINATION bin)
+
+file(COPY flang.sh
+  DESTINATION "${CMAKE_CURRENT_BINARY_DIR}/bin"
+  FILE_PERMISSIONS
+    OWNER_READ OWNER_WRITE OWNER_EXECUTE
+    GROUP_READ GROUP_EXECUTE
+    WORLD_READ WORLD_EXECUTE
+)
+file(RENAME "${CMAKE_CURRENT_BINARY_DIR}/bin/flang.sh" "${CMAKE_CURRENT_BINARY_DIR}/bin/flang")
```
sscalpone: I wanted the check-in source file to be called flang.sh instead of just flang.  CMake does not have a COPY command that can rename files or a RENAME command that can set file permissions.

PR #653, opened 2019-08-13
Create a throwaway wrapper for the throwaway driver.

sscalpone:
Add a script called flang to the bin directory that calls f18 with the -I option pointing to the module files for the predefined modules. The wrapper also sets the module suffix to .fmf to avoid conflicts with module file names generated by other compilers. Also, now the build creates and installs both
.mod and .fmf files for the predefined module files, the latter to accommodate the flang script.

### DIFF
--- a/tools/f18/flang.sh
+++ b/tools/f18/flang.sh
@@ -19,4 +19,4 @@ function abspath() {
 
 wd=`abspath $(dirname "$0")/..`
 
-$wd/bin/f18 -fdebug-semantics -module-suffix .fmf -I $wd/include $*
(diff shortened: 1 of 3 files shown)


tools/f18/flang.sh
```diff
@@ -0,0 +1,22 @@
...
+    if [ "$cur_dir" == "/" ]; then
+      echo "$cur_dir`basename \"$1\"`";
+    else
+      echo "$cur_dir/`basename \"$1\"`";
+    fi;
+  fi;
+  popd > /dev/null;
+}
+
+wd=`abspath $(dirname "$0")/..`
```
tskeith: I think this can just be:
`wd="$(cd "$(dirname "$0")"/..; pwd)"`

PR #663, opened 2019-08-14
Checks for constraint C1129

psteinfeld:
"A variable that is referenced by the scalar-mask-expr of a concurrent-header or by any concurrent-limit or concurrent-step in that concurrent-header shall not appear in a LOCAL locality-spec in the same DO CONCURRENT statement."

In the process of implementing these checks, I found and fixed some other problems.  I also cleaned up some of the code in check-do.cc.  I ran into two notable difficulties in implementing these checks.  First, the symbols associated with the names in a locality spec get created when the locality specs are process during name resolution.  Thus, they're different from the symbols associated with names that appear in the control expressions.  At Tim's suggestion, I dealt with this by looking up the symbols from the names in the locality spec starting with the closest enclosing scope containing the DO construct.  Second, the symbols can be hidden behind host- use- and construct-associations.

### DIFF
--- a/lib/semantics/check-do.cc
+++ b/lib/semantics/check-do.cc
@@ -29,9 +29,9 @@ namespace Fortran::semantics {
 using namespace parser::literals;
 
 // 11.1.7.5 - enforce semantics constraints on a DO CONCURRENT loop body
-class DoConcurrentEnforcement {
+class DoConcurrentBodyEnforce {
 public:
-  DoConcurrentEnforcement(SemanticsContext &context) : context_{context} {}
+  DoConcurrentBodyEnforce(SemanticsContext &context) : context_{context} {}
   std::set<parser::Label> labels() { return labels_; }
   std::set<parser::CharBlock> names() { return names_; }
   template<typename T> bool Pre(const T &) { return true; }
@@ -222,7 +222,7 @@ private:
   std::set<parser::Label> labels_;
   parser::CharBlock currentStatementSourcePosition_;
   SemanticsContext &context_;
-};  // class DoConcurrentEnforcement
+};  // class DoConcurrentBodyEnforce
 
 class DoConcurrentLabelEnforce {
 public:
@@ -372,8 +372,8 @@ private:
     context_.Say(sourceLocation, "DO controls should be INTEGER"_err_en_US);
   }
 
-  void CheckDoControl(parser::CharBlock sourceLocation, bool isReal) {
-    bool warn{context_.warnOnNonstandardUsage() ||
+  void CheckDoControl(const parser::CharBlock &sourceLocation, bool isReal) {
+    const bool warn{context_.warnOnNonstandardUsage() ||
         context_.ShouldWarn(parser::LanguageFeature::RealDoControls)};
     if (isReal && !warn) {
       // No messages for the default case
@@ -387,8 +387,7 @@ private:
 
   void CheckDoVariable(const parser::ScalarName &scalarName) {
     const parser::CharBlock &sourceLocation{scalarName.thing.source};
-    const Symbol *symbol{scalarName.thing.symbol};
-    if (symbol) {
+    if (const Symbol * symbol{scalarName.thing.symbol}) {
       if (!IsVariableName(*symbol)) {
         context_.Say(
             sourceLocation, "DO control must be an INTEGER variable"_err_en_US);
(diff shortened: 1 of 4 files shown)


test/semantics/dosemantics09.f90
```diff
@@ -0,0 +1,125 @@
...
+
+subroutine s2()
+!ERROR: 'i' is already declared in this scoping unit
+  do concurrent (i=1:10) local_init(i)
+  end do
+end subroutine s2
+
+subroutine s3()
+!ERROR: 'i' is already declared in this scoping unit
+  do concurrent (i=i:10) shared(i)
```
ichoyjx: You meant `i=i:10` or `i=1:10`?

lib/semantics/check-do.cc
```diff
@@ -583,48 +574,68 @@ class DoContext {
...
-    auto &mask{std::get<std::optional<parser::ScalarLogicalExpr>>(header.t)};
-    if (mask.has_value()) {
-      CheckMaskIsPure(*mask);
-    }
-    auto &localitySpecs{
+  void CheckLocalitySpecs(const parser::LoopControl::Concurrent &concurrent,
+      const parser::Block &block) const {
+    const auto &header{std::get<parser::ConcurrentHeader>(concurrent.t)};
+    const auto &controls{
+        std::get<std::list<parser::ConcurrentControl>>(header.t)};
```
ichoyjx: Is it safe to assume that `controls` will not be empty?

test/semantics/dosemantics09.f90
psteinfeld: Good catch!  This is just a bogus extra test.  I'll get rid of it.

lib/semantics/check-do.cc
```diff
@@ -493,87 +434,137 @@ class DoContext {
...
-    int count{0};
-    for (auto &ls : localitySpecs) {
-      if (std::holds_alternative<parser::LocalitySpec::DefaultNone>(ls.u)) {
-        ++count;
-        if (count > 1) {
-          context_.Say(currentStatementSourcePosition_,
-              "only one DEFAULT(NONE) may appear"_en_US);
-          return;
+    SymbolSet symbols;
+    const Scope &scope{
```
vjayathirtha-nv: This is a nitpick, but naming this variable something like `parentScope` would make the intentions of this code slightly more clear. However, you can choose to ignore this suggestion since you already have a comment that clearly tells all that needs to be told.

lib/semantics/check-do.cc
psteinfeld: Yes.  The associated element of the grammar is a ```concurrent-control-list```.  Section 4.1.3 of the standard in R401 states ```xyz-list is xyz [ , xyz ] ...```.  So if there's not at least one ```concurrent-control```, there will be a parsing error.

lib/semantics/check-do.cc
psteinfeld: Good point.  I'll make that change.

lib/semantics/check-do.cc
```diff
@@ -493,87 +434,137 @@ class DoContext {
...
+            .Attach(name, "Declaration of impure procedure '%s'"_en_US, name);
         return;
       }
     }
   }
 
-  void CheckNoCollisions(const SymbolContainer &refs,
-      const SymbolContainer &defs,
-      const parser::MessageFixedText &errorMessage) const {
+  void CheckNoCollisions(const SymbolSet &refs, const SymbolSet &set,
```
vjayathirtha-nv: why change name from `defs` to `set` ? 

`refs` and `defs` makes sense to me.

psteinfeld: The function ```CheckNoCollisions()``` is used to compare names in limit and step expressions against index names and to compare names in the ```scalar-mask-expr``` against names in ```locality-spec```s.  In the first case, the index-names can be considered to be definitions or defs.  But in the second case, there really aren't any definitions.  In both cases, though, the names in the expressions can be considered references or refs.

vjayathirtha-nv: Okay, then maybe you could name it `uses` or something that conveys what it is. `set` doesn't say anything really.

psteinfeld: Good suggestion.  I'll do that.  Thanks!

test/semantics/dosemantics09.f90
```diff
@@ -0,0 +1,125 @@
...
+  end do
+end subroutine s3
+
+subroutine s4()
+!ERROR: concurrent-header expression references variable 'i' in locality-spec
+  do concurrent (j=i:10) local(i)
```
tskeith: This message points to the expression where `i` is referenced but not the locality spec that causes the error. I think it would be easier to understand if it did, i.e. have the attached message point to the `i` in `local(i)`.

Or have the error point to the second `i` and the attached message point to the first one. That would require rewording the message, something like this:
`'i' may not appear in LOCAL locality-spec because it is referenced in the concurrent-header`

Also, this constraint only applies to the LOCAL locality-spec so I think the message should say that.

psteinfeld: I like the suggestion of specifying that the problem is specific to LOCAL locality-specs.  But I think it's better for the error message to refer to the variable name's use in the expression since I'd guess that this location would be more difficult to find than the name in the locality-spec.  Also, referring to the name's declaration is a good thing since the actual declaration might require traversing a construct association which changes the name of the variable.

I'll change the text of the error message.

PR #675, opened 2019-08-22
Fix bugs in .mod file generation

tskeith:


lib/semantics/mod-file.cc
```diff
@@ -283,17 +288,22 @@ void ModFileWriter::PutSubprogram(const Symbol &symbol) {
     bindAttrs.set(Attr::BIND_C, true);
     attrs.set(Attr::BIND_C, false);
   }
-  if (attrs.test(Attr::PRIVATE)) {
-    PutAttr(decls_, Attr::PRIVATE) << "::";
+  Attrs prefixAttrs{subprogramPrefixAttrs};
+  prefixAttrs &= attrs;
```
ichoyjx: Is this same as `Attrs prefixAttrs{subprogramPrefixAttrs & attrs};`?

tskeith: That's how I wrote it originally, but it didn't compile because the `operator&` returns `EnumSet<Attr>`. Adding the appropriate constructors to `Attrs` allows it to work and makes the code clearer, so I've done that. Thanks for the suggestion.

Issue #661, opened 2019-08-14
Missing error or warning on dconjg misuse


sscalpone:
```
! gfortran always issues an error
! if INTRINSIC is used, ifort issues an error
! if INTRINSIC is commented out, ifort assumes EXTERNAL
! pgfortran says "Type conversion of expression performed"
! f18 is slient; still a todo
subroutine test(a)
  complex(4) :: a
  complex(4) :: c
  INTRINSIC dconjg
  c = dconjg(a)
end
```

jeanPerier:
Fix by PR #680.
f18 now outputs error:
```
error: Actual argument for 'a=' has bad type or kind 'Complex(4)'
    c = dconjg(a)
        ^^^^^^
```

sscalpone:
It took me a minute to understand the "=" in "a=".

PR #680, opened 2019-08-23
 Fix issue: Enforce specific intrinsic characteristics

jeanPerier:
Fix issue #661.
The issue was that when probing a specific intrinsic, the constraints of the related generic intrinsic were tested instead of the more restrictive constraints of the specific intrinsic.

For issue #661 program, f18 now raises error: 
```
error: Actual argument for 'a=' has bad type or kind 'Complex(4)'
    c = dconjg(a)
        ^^^^^^
```

During the fix, found and fix an issue with `DOUBLE COMPLEX` in intrinsic matching. The intrinsic matching was only expecting REAL for kind code `KindCode::doublePrecision` that is also used for `DoublePrecisionComplex` (This caused an internal error when using `dconjg` with an complex(8) arg).

### DIFF
--- a/lib/evaluate/intrinsics.cc
+++ b/lib/evaluate/intrinsics.cc
@@ -1206,10 +1206,9 @@ std::optional<SpecificCall> IntrinsicInterface::Match(
           DynamicType{*category, defaults.GetDefaultKind(TypeCategory::Real)};
       break;
     case KindCode::doublePrecision:
-      CHECK(result.categorySet == RealType);
-      CHECK(*category == TypeCategory::Real);
-      resultType =
-          DynamicType{TypeCategory::Real, defaults.doublePrecisionKind()};
+      CHECK(result.categorySet == CategorySet{*category});
+      CHECK(FloatingType.test(*category));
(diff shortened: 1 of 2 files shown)


lib/evaluate/intrinsics.cc
```diff
@@ -1618,40 +1617,34 @@ std::optional<SpecificCall> IntrinsicProcTable::Implementation::Probe(
        ++specIter) {
     // We only need to check the cases with distinct generic names.
```
klausler: This comment may now be obsolete with the new approach.

jeanPerier: I think it still holds to justify the `if (const char *genericName{specIter->second->generic})` just below. 
My understanding is:
If `specIter->second->generic` is `nullptr`, then the specific name is also a generic name and the related generic that has already been attempted few lines before (and did not match if we got here). Hence there is no need to attempt the specific version that will fail because it is more restrictive than the generic version that already failed.

jeanPerier:
> blems by running f90_correct with semantics checking enabled after this change is merged; it looks okay to me, but maybe there's a reason that the original code did what it did that I'm not remembering now.

Issue #664, opened 2019-08-15
clang++ 9 with libc++ fails to compile parsing.cc


DavidTruby:
libc++ rL363692 onwards
This is due to a change in the standard meaning that narrowing conversions are no longer considered when selecting a constructor for variant. See http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2018/p0608r3.html
Unfortunately since this is a 'bug fix' in the standard it's likely this will also break in the same way with libstdc++ in future.
```
clang++ -stdlib=libc++  -I/home/davtru01/f18/build/../include -I/home/davtru01/f18/build/include -I/home/davtru01/llvm/include -std=c++17 -Wall -pedantic -fno-rtti -fno-exceptions -O2   -D_GNU_SOURCE -D__STDC_CONSTANT_MACROS -D__STDC_FORMAT_MACROS -D__STDC_LIMIT_MACROS -fsyntax-only ../lib/parser/parsing.cc
In file included from ../lib/parser/parsing.cc:15:
In file included from ../lib/parser/parsing.h:20:
In file included from ../lib/parser/instrumented-parser.h:19:
In file included from ../lib/parser/parse-state.h:28:
In file included from ../lib/parser/user-state.h:25:
../lib/parser/parse-tree.h:615:3: error: no matching constructor for initialization of 'std::variant<ScalarIntExpr, Star, Deferred>' (aka 'variant<Scalar<Integer<common::Indirection<Expr> > >, Fortran::parser::Star, Fortran::parser::TypeParamValue::Deferred>')
  UNION_CLASS_BOILERPLATE(TypeParamValue);
  ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
../lib/parser/parse-tree.h:105:22: note: expanded from macro 'UNION_CLASS_BOILERPLATE'
  classname(A &&x) : u(std::move(x)) {} \
                     ^ ~~~~~~~~~~~~
/home/davtru01/llvm/bin/../include/c++/v1/variant:620:9: note: in instantiation of function template specialization 'Fortran::parser::TypeParamValue::TypeParamValue<unsigned long, void>' requested here
      : __value(_VSTD::forward<_Args>(__args)...) {}
        ^
../lib/parser/basic-parsers.h:724:18: note: in instantiation of function template specialization 'Fortran::parser::CharLength::CharLength<unsigned long, void>' requested here
          return RESULT{std::move(*arg)};
                 ^
[...]

PR #681, opened 2019-08-23
Fix for builds with libc++9 and libstdc++10

DavidTruby:
The behaviour of std::variant has been changed in the standard to forbid narrowing conversions. This behaviour is applied to variant in every standard version in both libc++9 and libstdc++10. This patch simply adds casts so that the narrowing converions don't occur anymore.

Clang (or libstdc++) also has a bug whereby the compiler greedily instantiates too far down a template stack and fails to find a valid constructor even though one exists. This patch also adds a workaround in the one place where this manifests in the f18 code, forcing clang to find the correct constructor.

Fixes issue #664

### DIFF
--- a/lib/parser/parse-tree.h
+++ b/lib/parser/parse-tree.h
@@ -632,7 +632,7 @@ WRAPPER_CLASS(IntegerTypeSpec, std::optional<KindSelector>);
 // R723 char-length -> ( type-param-value ) | digit-string
 struct CharLength {
   UNION_CLASS_BOILERPLATE(CharLength);
-  std::variant<TypeParamValue, std::int64_t> u;
+  std::variant<TypeParamValue, std::uint64_t> u;
 };
 
 // R722 length-selector -> ( [LEN =] type-param-value ) | * char-length [,]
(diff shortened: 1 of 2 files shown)


lib/parser/parse-tree.h
```diff
@@ -2931,6 +2931,9 @@ struct GenericStmt {
 // R1503 interface-stmt -> INTERFACE [generic-spec] | ABSTRACT INTERFACE
 struct InterfaceStmt {
   UNION_CLASS_BOILERPLATE(InterfaceStmt);
+  // Workaround for clang with libstc++10 bug
+  InterfaceStmt(Abstract x) : u(x) {}
```
klausler: We typically use curly braces for member initialization in constructors.

The parse tree node types all have move constructors, and intentionally omit copy semantics; does this WAR still succeed if written `InterfaceStmt(Abstract &&x) : u{std::move(x)} {}`?

DavidTruby: Unfortunately it doesn't as this still triggers the defective behaviour. My first thought was to implement it this way for consistency. However the copy should be cheap/free as Abstract is an empty class.
I can possibly use braces for the member initialisation instead to follow the style guide, I'll check if that still works.

lib/semantics/resolve-names.cc
```diff
@@ -2998,8 +2998,9 @@ void DeclarationVisitor::Post(const parser::CharSelector::LengthAndKind &x) {
   }
 }
 void DeclarationVisitor::Post(const parser::CharLength &x) {
-  if (const auto *length{std::get_if<std::int64_t>(&x.u)}) {
-    charInfo_.length = ParamValue{*length, common::TypeParamAttr::Len};
+  if (const auto *length{std::get_if<std::uint64_t>(&x.u)}) {
+    charInfo_.length = ParamValue{static_cast<ConstantSubscript>(*length),
+                                  common::TypeParamAttr::Len};
```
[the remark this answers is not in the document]
DavidTruby: I've corrected this, will remember to run clang-format in future

PR #683, opened 2019-08-23
Requirements and first draft design for procedure call protocol

klausler:
Here's an incomplete draft of my understanding of the requirements imposed by the standard on f18's non-intrinsic procedure reference call/entry/exit/return protocol, specifically concentrating on the situations where we'll have to deal with temporary storage.  It's going to get fleshed out further next week but maybe you'll find it already of interest at this point.  Please comment on any point that seems wrong, unjustified, or unclear.

### DIFF
--- a/documentation/Calls.md
+++ b/documentation/Calls.md
@@ -125,7 +125,7 @@ some design alternatives that are explored further below.
    package them into single address (for calls to internal procedures &
    for calls that pass internal procedures as arguments).
 1. Resolve the target procedure's polymorphic binding, if any.
-1. Marshal actual argument addresses/values into registers.
+1. Marshal actual argument addresses (or values for `VALUE`/`%VAL()`) into registers.
 1. Marshal an extra argument for the `CHARACTER` result
    length (assumed or explicit) or the function result descriptor.
 1. Set the "host instance" (static link) register when calling an internal
@@ -187,15 +187,16 @@ argument.
 First, actual arguments that are expressions, not designators, obviously
 need to be computed and captured into memory in order to be passed
 by reference.
-This includes parenthesized designators like `(X)` as an important
-special case, which are expressions in Fortran.
-(This case also technically includes constants, but those are better
-implemented by passing addresses in read-only memory when the interface
-is explicit.)
+This includes parenthesized designators like `(X)`, which are
+expressions in Fortran, as an important special case.
+(This case also technically includes unparenthesized constants,
+but those are better implemented by passing addresses in read-only
+memory.)
 The dummy argument cannot be known to have `INTENT(OUT)` or
 `INTENT(IN OUT)`.
 
-Small scalar or elemental `VALUE` arguments may be passed in registers.
+Small scalar or elemental `VALUE` arguments may be passed in registers,
+as should arguments wrapped in the legacy VMS `%VAL()` notation.
 Multiple elemental `VALUE` arguments might be packed into SIMD registers.
 
 Actual arguments that are designators, not expressions, must also
@@ -506,9 +507,9 @@ For examples: `Fa.mod.foo`, `Fa.mod.submod.foo`.
 ## Summary of checks to be enforced in semantics analysis
 
 15.5.1 procedure references:
-* C1533 (can't pass non-intrinsic `ELEMENTAL` as argument)
-* C1536 alternate return labels must be in the inclusive scope
-* C1537 coindexed argument cannot have a `POINTER` ultimate component
+* (C1533) can't pass non-intrinsic `ELEMENTAL` as argument
+* (C1536) alternate return labels must be in the inclusive scope
+* (C1537) coindexed argument cannot have a `POINTER` ultimate component
 
 15.5.2.4 requirements for non-`POINTER` non-`ALLOCATABLE` dummies:
 * (2) dummy must be monomorphic for coindexed polymorphic actual
@@ -524,7 +525,7 @@ For examples: `Fa.mod.foo`, `Fa.mod.submod.foo`.
 * (21) array sections with vector subscripts can't be passed to definable dummies
        (`INTENT(OUT)`, `INTENT(IN OUT)`, `ASYNCHRONOUS`, `VOLATILE`)
 * (22) `VOLATILE` attributes must match when dummy has a coarray ultimate component
-* C1538 - C1540: checks for `ASYNCHRONOUS` and `VOLATILE`
+* (C1538 - C1540) checks for `ASYNCHRONOUS` and `VOLATILE`
 
 15.5.2.5 requirements for `ALLOCATABLE` & `POINTER` arguments when both
 the dummy and actual arguments have the same attributes:
@@ -540,8 +541,8 @@ the dummy and actual arguments have the same attributes:
 * (7) `INTENT(OUT)` & `INTENT(IN OUT)` dummies require definable actuals
 
 15.5.2.7 `POINTER` dummy arguments:
-* C1541: `CONTIGUOUS` dummy requires simply contiguous actual
-* C1542: actual argument cannot be coindexed unless procedure is intrinsic
+* (C1541) `CONTIGUOUS` dummy requires simply contiguous actual
+* (C1542) actual argument cannot be coindexed unless procedure is intrinsic
 * (2) actual argument must be `POINTER` unless dummy is `INTENT(IN)` and
   actual could be the right-hand side of a pointer assignment statement
 
@@ -575,7 +576,7 @@ For interoperable procedures and interfaces (18.3.6):
 ## Further topics to document
 
 * Alternate return specifiers
-* `%VAL()` and `%REF()`
+* `%VAL()`, `%REF()`, and `%DESCR()` legacy VMS interoperability extensions
 * Unrestricted specific intrinsic functions as actual arguments
 * SIMD variants of `ELEMENTAL` procedures (& unrestricted specific intrinsics)
 * Elemental subroutine calls with array arguments
--- a/documentation/Calls.md
+++ b/documentation/Calls.md
@@ -117,7 +117,7 @@ some design alternatives that are explored further below.
 1. Shuffle `ENTRY` dummy arguments & jump to common entry point.
 1. Complete `VALUE` copying if this step will not always be done
    by the caller (as I think it should be).
-1. Finalize &/or re-initialize `INTENT(OUT)` non-pointer non-allocatable
+1. Finalize &/or re-initialize `INTENT(OUT)` non-pointer
    actual arguments (see below).
 1. Optionally compact assumed-shape arguments for contiguity on one
    or more leading dimensions to improve SIMD vectorization, if not
@@ -147,7 +147,8 @@ the subroutine's alternate return.
 1. Copy actual argument array designator data that was copied into
    a temporary back into its original storage (see below).
 1. Complete deallocation of actual argument temporaries (not `VALUE`).
-1. Reload host-escaping local objects from memory.
+1. Reload definable host-escaping local objects from memory, if they
+   were saved to memory by the host before the call.
 1. `GO TO` alternate return, if any.
 1. Use the function result in an expression.
 1. Eventually, finalize &/or deallocate the function result.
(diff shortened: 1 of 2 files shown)


documentation/Calls.md
```diff
@@ -0,0 +1,443 @@
...
+reentrancy;
+PGI passes host variable links in descriptors in additional arguments
+that are not always successfully forwarded across implicit interfaces,
+sometimes leading to crashes when they turn out to be needed.
+
+F18 will use descriptors in most cases to pass procedures as arguments,
+but also maintain a pool of trampolines in its runtime support library
+to implement calls to routines that need to support an
+implicit interface.
+
```
kiranchandramohan: Will this design allow for cases like the one mentioned in the following ticket? Here the C program calls a Fortran subroutine 'c', passing it a pointer to the subroutine 'b'. Subroutine 'c' calls 'b'. (Not supported by Flang but works with gfortran and Intel. We are told it is a common pattern) 
https://github.com/flang-compiler/flang/issues/631

documentation/Calls.md
```diff
@@ -0,0 +1,636 @@
...
+   result if the function is F77ish.
+1. Marshal an extra argument for the function result's descriptor,
+   if it needs one.
+1. Set the "host instance" (static link) register when calling an internal
+   procedure from its host or another internal procedure, a procedure pointer,
+   or dummy procedure (when it has a descriptor).
+1. Jump.
+
+### On entry:
+1. Shuffle `ENTRY` dummy arguments & jump to common entry point.
```
jeanPerier: It is not really clear to me what is happening here I guess "`ENTRY` dummy arguments" means that the callee is a procedure with multiple `ENTRY` points. Is it already clear if we are going to have one compiled function per entry point or a single function with some kind of dispatcher at the beginning ?
In the first case, I do not really understand why we would need the set described here.
In the second case, wouldn't we need to pass some kind of index as arguments to indicate where the control flow should be dispatched ?

documentation/Calls.md
```diff
@@ -0,0 +1,636 @@
...
+Effective arguments that are designators, not expressions, must also
+be copied into temporaries in the following situations.
+
+1. Coindexed objects need to be copied into the local image.
+   This can get very involved if they contain `ALLOCATABLE`
+   components, which also need to be copied, along with their
+   `ALLOCATABLE` components, and may be best implemented with a runtime
+   library routine working off a description of the type.
+1. Effective arguments associated with dummies with the `VALUE`
+   attribute need to copied; this could be done on either
```
jeanPerier: "need to **be** copied"

documentation/Calls.md
```diff
@@ -0,0 +1,636 @@
...
+   side of the call, but there are optimization opportunities
+   on the caller's side.
+1. In non-elemental calls, the values of array sections with
+   vector-valued subscripts need to be gathered into temporaries.
+   These effective arguments are not definable, and they are not allowed to
+   be associated with non-`VALUE` dummy arguments with the attributes
+   `INTENT(IN)`, `INTENT(IN OUT)`, `ASYNCHRONOUS`, or `VOLATILE`
```
jeanPerier: `INTENT(IN)` -> `INTENT(OUT)`

documentation/Calls.md
```diff
@@ -0,0 +1,636 @@
...
+   (15.4.2.4(21)); `INTENT()` can't always be checked.
```
jeanPerier: "(15.4.2.4(21))" -> "(15.**5**.2.4(21))" (at least in N2162).

documentation/Calls.md
```diff
@@ -0,0 +1,636 @@
...
+minimal initialization of any `ALLOCATABLE` components so that
+the runtime doesn't crash when the callee finalizes and deallocates
+them.
+Note that calls to implicit interfaces must conservatively allow
+for the use of `INTENT(OUT)` by the callee.
+
+Except for `VALUE` and known `INTENT(IN)` dummy arguments, the original
+contents of local designators that have been compacted into temporaries
+could optionally have their `ALLOCATABLE` components invalidated
+across the call as an aid to debugging.
```
jeanPerier: I do not understand what is meant by "invalidated" here. Does this mean to de-allocate the `ALLOCATABLE` components of the original content that was compacted ? Is it to ensure they are not accessed inside the callee ?

documentation/Calls.md
```diff
@@ -0,0 +1,636 @@
...
+Derived type components that are `ALLOCATABLE` are finalized
+and deallocated.
+Components with initializers are (re)initialized.
+
+The preparation of effective arguments for `INTENT(OUT)` could be
+done on either side of the call.  If the preparation is
+done by the caller, there is an optimization opportunity
+in situations where unmodified incoming `INTENT(OUT)` dummy
+arguments are being passed onward as outgoing `INTENT(OUT)`
+arguments.
```
jeanPerier: Would it be needed to prove that finalization of the argument has no side effects to do such optimization ?

documentation/Calls.md
```diff
@@ -0,0 +1,636 @@
...
+Dummy arguments are represented with the addresses of descriptors
+when they have any of the following characteristics:
+
+1. assumed-shape array (`DIMENSION(:)`)
+1. assumed-rank array (`DIMENSION(..)`)
+1. parameterized derived type with assumed `LEN` parameters
+1. polymorphic (`CLASS(T)`, `CLASS(*)`)
+1. assumed-type (`TYPE(*)`)
+1. coarray dummy argument
+1. `INTENT(IN) POINTER` argument (15.5.2.7, C.10.4)
```
jeanPerier: Doesn't all `POINTER` or `ALLOCATABLE` dummy arguments also need to be represented as descriptors ?
Or is this list just listing cases where the dummy needs a descriptor when the actual argument may not already have had one ?

documentation/Calls.md
```diff
@@ -0,0 +1,636 @@
...
+callee, not the caller.
+(A Fortran pitfall: declaring `A(0:9)`, passing it to a dummy
+array `D(:)`, and assuming that `LBOUND(D,1)` will be zero
+in the callee.)
+If the declaration of an assumed-shape dummy argument array
+contains an explicit lower bound expression (R819), its value
+needs to be computed by the callee;
+it may be captured and saved in the incoming descriptor
+as long as we assume that argument descriptors can be modified
+by callees.
```
jeanPerier: Assuming so would prevent forwarding descriptors when possible, right ?
I would have stored local lower bounds locally to avoid leaking them and having weird bugs, on the other hand if we already have storage space for them, that may be a waste.

documentation/Calls.md
```diff
@@ -0,0 +1,636 @@
...
+
+Except for `VALUE` and known `INTENT(IN)` dummy arguments and array sections
+with vector-valued subscripts (15.5.2.4(21)), temporary storage into
+which effective argument data were compacted for contiguity before the call
+must be redistributed back to its original storage by the caller after
+the return.
+
+In conjunction with saved cosubscript values, a standard descriptor
+suffices to represent a pointer to the original storage into which the
+temporary data should be redistributed.
```
jeanPerier: Do you mean we would actually create two descriptors for such case inside the caller ? One for the actual argument and one for the dummy and we would do inside the caller :
```
Copy("dummy descriptor", "actual descriptor") // maybe with runtime help for PDTS
Call("dummy descriptor")
Copy("actual descriptor", "dummy descriptor") 
```

documentation/Calls.md
```diff
@@ -0,0 +1,636 @@
...
+XLF manages trampolines in its runtime support library, which adds some overhead
+to their construction and a reclamation obligation;
+NAG Fortran manages a static fixed-sized stack of trampolines
+per call site, imposing a hidden limit on recursion and foregoing
+reentrancy;
+PGI passes host instance links in descriptors in additional arguments
+that are not always successfully forwarded across implicit interfaces,
+sometimes leading to crashes when they turn out to be needed.
+
+F18 will manage a pool of trampolines in its runtime support library
```
jeanPerier: I am not sure I fully understand the concept of trampoline.
Would that be something like:
in the host function
```func_pointer => SetTrampoline(proc, host_link)```

In SetTrampoline:
```
N = GetFreeTrampoline()
host_links[N] = host_link
proc[N] = proc
return trampolines[N] // trampoline<N>
```

In the trampoline function `trampoline<N>`
```
host_register = host_links[N]
return proc[N]()
```

File: documentation/Calls.md [hunk 12]
```diff
@@ -0,0 +1,636 @@
...
+Pure `ELEMENTAL` functions (15.8) must use distinct names for any alternate
+entry points used for packed SIMD arguments of various widths if we support
+calls to these functions in SIMD parallel contexts.
+There are already conventions for these names in `libpgmath`.
+
+The names of non-F77ish external procedures
+should be distinguished as such so that incorrect attempts to call or pass
+them with an implicit interface will fail to resolve at link time.
+Fortran 2018 explicitly enables us to do this with a correction to Fortran
+2003 in 4.3.4(2).
```
jeanPerier: Also 4.3.4(5) ?

documentation/Calls.md
jeanPerier: OK, so it would not be a pre-compiled pool, more like a JIT compiled one.
I did not know it was safer than stack execution, but after reading on the topic a bit, I guess you want to use something like W^X protection (http://man7.org/linux/man-pages/man2/mprotect.2.html). Sounds great to allow deeper recursions.

documentation/Calls.md
jeanPerier: It's sounds worth testing. I guess we may want to come with a clever "merry-go-round" of allocatable component storages anyway to save some memory allocation and de-allocation around calls. Instead of copying allocatable components we could "move" them into the temp and move them back into the original variable after the call.

documentation/Calls.md
jeanPerier: OK, you already replied to this comment in another comment (in "Before the call" l158. )
You only mention assumed-shape, but for explicit-shape/assumed-sized PDT or polymorphic argument we would also have a descriptor, right ? Will descriptors also be used to store the lower/upper bounds that maybe be non-constants in such cases ?

klausler: Yes, there are multiple reasons for having descriptors, and if they're around, we might as well use their storage for bounds if necessary.

documentation/Calls.md
```diff
@@ -0,0 +1,664 @@
...
+  and their contents are contiguous.
+* An *assumed-size* array is an explicit-shape array with `*` as its
+  final dimension, which is the most-significant one in Fortran
+  and whose value does not affect indexed address calculations.
+* A *deferred-shape* array (`DIMENSION::A(:)`) is a `POINTER` or `ALLOCATABLE`.
+  `POINTER` target data might not be contiguous.
+* An *assumed-shape* (not size!) array is a dummy argument whose
+  upper bounds take their values from the *shape* of the effective
+  argument; lower bounds can be set by the procedure that receives
+  them and default to 1.
```
tskeith: Maybe add a definition of *shape* to this list.

documentation/Calls.md
```diff
@@ -0,0 +1,664 @@
...
+is acceptable for use in LLVM IR and assembly language but not in user
+Fortran or C code, or by switching case
+(so long as there's a way to cope with extension names that don't begin
+with letters).
+
+In particular, the period (`.`) seems safe to use as a separator character,
+so a `Fa.` prefix can serve to isolate these discretionary names from
+other uses and to identify the earliest link-compatible version.
+For examples: `Fa.mod.foo`, `Fa.mod.submod.foo`, and (for an external
+subprogram that requires an explicit interface) `Fa.foo`.
```
tskeith: I would like to be clear on when the `Fa.` prefix would change. If we change f18 in a way that is link-incompatible, does that mean we would also change the prefix to `Fb.`? So that code compiled with a compiler from before that change won't link with code from the new version?

klausler: Yes, that's right.  I'll add text.

vdonaldson: A minor consideration: numbers are easier to increment indefinitely than letters, and might make the purpose for their inclusion in the scheme slightly more obvious: F1 ... F10 ... vs. Fa ... Fz ...

PR #685, opened 2019-08-26
Add the OpenMP module

sscalpone:
Add omp_lib as two files: a Fortran file used to create the omp_lib module and a .h file that can be used directly, which apparently some codes do. Because of the split, and wanting to avoid forcing the user to use isc_c_binding, use int_ptr_kind()  instead of c_intptr_t.


module/omp_lib.h
```diff
@@ -0,0 +1,548 @@
+! Copyright (c) 2019, NVIDIA CORPORATION.  All rights reserved.
```
kiranchandramohan: 1) Can we use the headers provided by the OpenMP library?
2) Can this Copyright be removed? I am guessing that for the fixed-form case this header will be directly placed included the source. Alternatively, can this be preprocessed and included?
3)Which OpenMP version does this header correspond to?

PR #686, opened 2019-08-26
Changes to add an executable construct stack

psteinfeld:
I added a stack of ExecutableConstruct's to SemanticsContext along with
functions to push and pop constructs.  I added code to the SemanticsVisitor
to use these new functions.  I also added functions Pre and Post functions
for UnlabeledStatement's so that we could isolate the source positions for
statements embedded in "if" statements to improve error messages.

I also added code to check-do.[h,cc] to use this new infrastructure to check
for CYCLE and EXIT statements that are not contained within DO constructs
along with a test.

### DIFF
--- a/lib/semantics/check-do.cc
+++ b/lib/semantics/check-do.cc
@@ -647,21 +647,20 @@ void DoChecker::Leave(const parser::DoConstruct &x) {
   doContext.Check(x);
 }
 
-void DoChecker::CheckLoneStmt(const char *stmtString) const {
-  for (auto &executable : context_.executables()) {
-    if (std::holds_alternative<common::Indirection<parser::DoConstruct>>(
-            executable->u)) {
-      return;
-    }
+// C1134
+void DoChecker::Enter(const parser::CycleStmt &) {
+  if (!context_.InsideDoConstruct()) {
+    context_.Say(
+        *context_.location(), "CYCLE must be within a DO construct"_err_en_US);
   }
-  context_.Say(*context_.location(),
-      "%s must be within a DO construct"_err_en_US, stmtString);
 }
 
-// C1134
-void DoChecker::Enter(const parser::CycleStmt &) { CheckLoneStmt("CYCLE"); }
-
 // C1166
-void DoChecker::Enter(const parser::ExitStmt &) { CheckLoneStmt("EXIT"); }
+void DoChecker::Enter(const parser::ExitStmt &) {
+  if (!context_.InsideDoConstruct()) {
+    context_.Say(
+        *context_.location(), "EXIT must be within a DO construct"_err_en_US);
+  }
+}
 
 }  // namespace Fortran::semantics
--- a/lib/semantics/check-do.cc
+++ b/lib/semantics/check-do.cc
@@ -647,4 +647,21 @@ void DoChecker::Leave(const parser::DoConstruct &x) {
   doContext.Check(x);
 }
 
+void DoChecker::CheckLoneStmt(const char *stmtString) const {
+  for (auto &executable : context_.executables()) {
+    if (std::holds_alternative<common::Indirection<parser::DoConstruct>>(
+            executable->u)) {
+      return;
+    }
+  }
+  context_.Say(*context_.location(),
+      "%s must be within a DO construct"_err_en_US, stmtString);
+}
+
+// C1134
+void DoChecker::Enter(const parser::CycleStmt &) { CheckLoneStmt("CYCLE"); }
+
+// C1166
+void DoChecker::Enter(const parser::ExitStmt &) { CheckLoneStmt("EXIT"); }
+
 }  // namespace Fortran::semantics
--- a/lib/semantics/check-do.h
+++ b/lib/semantics/check-do.h
@@ -27,8 +27,12 @@ class DoChecker : public virtual BaseChecker {
 public:
   explicit DoChecker(SemanticsContext &context) : context_{context} {}
   void Leave(const parser::DoConstruct &);
+  void Enter(const parser::CycleStmt &);
+  void Enter(const parser::ExitStmt &);
 
 private:
+  void CheckLoneStmt(const char *const) const;
+
   SemanticsContext &context_;
 };
 }
--- a/lib/semantics/check-do.h
+++ b/lib/semantics/check-do.h
@@ -19,6 +19,8 @@
 
 namespace Fortran::parser {
 struct DoConstruct;
+struct CycleStmt;
+struct ExitStmt;
 }
 
 namespace Fortran::semantics {
@@ -31,8 +33,6 @@ public:
   void Enter(const parser::ExitStmt &);
 
 private:
-  void CheckLoneStmt(const char *const) const;
-
   SemanticsContext &context_;
 };
 }
--- a/lib/semantics/semantics.cc
+++ b/lib/semantics/semantics.cc
@@ -57,10 +57,18 @@ public:
     : C{context}..., context_{context} {}
 
   template<typename N> bool Pre(const N &node) {
+    if constexpr (common::HasMember<const N *, ConstructNode>) {
+      context_.PushConstruct(node);
+    }
     Enter(node);
     return true;
   }
-  template<typename N> void Post(const N &node) { Leave(node); }
+  template<typename N> void Post(const N &node) {
+    Leave(node);
+    if constexpr (common::HasMember<const N *, ConstructNode>) {
+      context_.PopConstruct();
+    }
+  }
 
   template<typename T> bool Pre(const parser::Statement<T> &node) {
     context_.set_location(node.source);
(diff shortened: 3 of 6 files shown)


lib/semantics/semantics.cc
```diff
@@ -168,6 +298,24 @@ Scope &SemanticsContext::FindScope(parser::CharBlock source) {
   }
 }
 
+void SemanticsContext::PushConstruct(const ConstructNode &construct) {
```
klausler: I'm actually surprised that C++ implicitly converts the pointers at the various call sites into a `ConstructNode` that can be passed by reference.

This function might be more clear if it used move semantics.

lib/semantics/semantics.cc
```diff
@@ -168,6 +298,24 @@ Scope &SemanticsContext::FindScope(parser::CharBlock source) {
...
+  constructStack_.emplace_back(construct);
+}
+
+void SemanticsContext::PopConstruct() {
+  CHECK(!constructStack_.empty());
```
klausler: This `CHECK` may not add any value; the `pop_back()` will crash if the stack is empty.

lib/semantics/semantics.cc
```diff
@@ -168,6 +298,24 @@ Scope &SemanticsContext::FindScope(parser::CharBlock source) {
...
+  constructStack_.pop_back();
+}
+
+bool SemanticsContext::InsideDoConstruct() const {
+  for (const ConstructNode construct : constructStack_) {
```
klausler: Rather than looping over the entries in the stack and looking for `DO` loops, you could maintain a count of active `DO` loops.

lib/semantics/semantics.cc
```diff
@@ -67,10 +67,30 @@ template<typename... C> class SemanticsVisitor : public virtual C... {
...
+  template<typename T> bool Pre(const parser::UnlabeledStatement<T> &node) {
+    context_.set_location(node.source);
+    Enter(node);
+    return true;
+  }
   template<typename T> void Post(const parser::Statement<T> &node) {
     Leave(node);
     context_.set_location(std::nullopt);
   }
+  template<typename T> void Post(const parser::UnlabeledStatement<T> &node) {
```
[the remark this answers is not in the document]
psteinfeld: Apparently so.  When I remove them, there are times when the location is uninitialized.

lib/semantics/semantics.cc
psteinfeld: Actually, that was my original implementation.  When I reviewed that with Tim, he suggested using a stack.  The stack is heavier weight, but it has the promise of being useful in other kinds of semantic analysis.

lib/semantics/semantics.cc
psteinfeld: ```CHECK``` has the advantage that the message coming from the compiler contains that name of the file and source line of the failing check.  When ```pop_back()``` fails, the compiler just produces a segmentation fault with no indication of the location of the failure.

lib/semantics/semantics.cc
psteinfeld: I want to make sure that I understand your suggestion.  Are you saying that, for example, the code for the ```Pre``` function for a ```DoConstruct``` should look like this?
```
  bool Pre(const parser::DoConstruct &doConstruct) {
    ConstructNode doNode{std::move(&doConstruct)};
    context_.PushConstruct(doNode);
    Enter(doConstruct);
    return true;
  }
```

klausler: No, that doesn't work.  I meant something like `context_.PushConstruct(ConstructNode{doConstruct});` in `Pre()` with the definition of `PushConstruct` accepting a `ConstructNode &&construct` argument.

lib/semantics/semantics.cc
klausler: You can still maintain a counter.

vdonaldson: Or maybe a cumulative ancestor set for all ancestor types, not just do loops.  That is, for top level constructs, this set is empty.  For nested constructs, the set is the parent's set plus the parent's type.  That way, queries of the form "this construct has at least one kind X ancestor" can be answered without searching.  And queries of the form "this construct is immediately nested in a kind X parent' can also be directly answered by looking one up the stack.  Those two queries probably cover a large majority of all potential queries.

vjayathirtha-nv: Val suggestion is more efficient compared to the stack approach. However one would still have to traverse up get a count of active `DO` loops. So a cumulative ancestor set + depth record for `DO` (and any other construct that needs its depth saved) seems like the best idea.

lib/semantics/semantics.cc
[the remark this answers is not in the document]
psteinfeld: Yes.  When I remove them, the call to ```set_location()``` is not made for some nodes and several tests fail.

lib/semantics/check-do.cc
```diff
@@ -647,4 +647,21 @@ void DoChecker::Leave(const parser::DoConstruct &x) {
   doContext.Check(x);
 }
 
+void DoChecker::CheckLoneStmt(const char *stmtString) const {
```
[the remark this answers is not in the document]
psteinfeld: Note that, in my latest version of the code, I added the function ```InsideDoConstruct()``` to the class```SemanticsContext``` which supersedes this code.

lib/semantics/semantics.h
```diff
@@ -136,6 +143,7 @@ class SemanticsContext {
   evaluate::FoldingContext foldingContext_{defaultKinds_};
 
   bool CheckError(bool);
+  std::vector<const parser::ExecutableConstruct *> executables_;
```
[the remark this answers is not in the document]
psteinfeld: I'm pretty sure that we'll need the more general capabilities of ```std::vector``` as we add more semantic checks.

lib/semantics/check-do.h
```diff
@@ -27,8 +27,12 @@ class DoChecker : public virtual BaseChecker {
 public:
   explicit DoChecker(SemanticsContext &context) : context_{context} {}
   void Leave(const parser::DoConstruct &);
+  void Enter(const parser::CycleStmt &);
+  void Enter(const parser::ExitStmt &);
 
 private:
+  void CheckLoneStmt(const char *const) const;
```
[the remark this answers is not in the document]
psteinfeld: Note that this function has been eliminated in the latest version of the code.

lib/semantics/semantics.cc
psteinfeld: Thanks, Peter.  I'll change the ```Pre()``` functions.

lib/semantics/semantics.cc
psteinfeld: Thanks for the suggestions.  One of the checks I plan to implement (although not in this pull request) is to semantic checking for ```CYCLE do-construct-name```.  For that case, I'll need to walk up the currently active set of nested ```DO``` constructs looking for one with ```do-construct-name```.  For just checking a lone ```CYCLE``` statement, a counter is sufficient and more efficient.  Since I don't think I can get rid of the stack, and since I believe that it's sufficient for the cases of ```CYCLE``` statements with and without ```do-construct-name```s, I plan to just go with a stack.  I'll keep the notion of counters in mind as I add the rest of the checks, though.

lib/semantics/semantics.cc
tskeith: Take a look at the version in PR #708

psteinfeld: In Brian's code, ```PushConstruct()``` is a template that updates the data member ```constructStack_```.  But since ```constructStack_``` is of a non-template type, I don't see the advantage of implementing ```PushConstruct()``` as a template.

Why make it a template?

vjayathirtha-nv: The template suggestion doesn't have anything to do with `PushConstruct` or `constructStack_` being templated. This PR has roughly 10 `Pre()` and `Post()` routines manually defined, functionally they all do exactly the same thing. Instead you could use a templated routine each for `Pre()` and `Post()` .

vjayathirtha-nv: Please disregard my comment. I misunderstood things.

psteinfeld: Thanks, Tim.  I'll change my code to match Brian's.

psteinfeld:
I changed the implementation of ```PushConstruct()``` to match Brian's -- to be
a template that accepts any kind of node and do all of the work of creating the
```ConstructNode``` and putting it on the stack.

File: lib/semantics/semantics.cc [hunk 8]
```diff
@@ -67,10 +67,140 @@ template<typename... C> class SemanticsVisitor : public virtual C... {
...
+    Leave(node);
+    context_.set_location(std::nullopt);
+  }
+
+  bool Pre(const parser::AssociateConstruct &associateConstruct) {
```
klausler: It would be great if you could combine all of these repetitive `Pre` overloads.  It's hard to see the differences between them, if any, and it's hard to tell which ones are paired up with `Post` overloads and which are not.

klausler: Specifically, you could augment the default `Pre()` and `Post()` member function templates above (ca. line #60) with something like

'''
if constexpr (common::HasMember<const N *, ConstructNode>) {
  context_.PushConstruct(node);
}
```

psteinfeld: Peter, you're a genius!  I wanted to combine those calls, but I couldn't figure out how to do it.

psteinfeld:
I combined all of the functions that called ```PushConstruct()``` and ```PopConstruct()``` based on a suggestion from Peter.

PR #692, opened 2019-08-27
Create OpenMP-semantics.md

sscalpone:
Initial markdown document.  The author is @ichoyjx.

### DIFF
--- a/documentation/OpenMP-semantics.md
+++ b/documentation/OpenMP-semantics.md
@@ -1,6 +1,6 @@
 # OpenMP Semantic Analysis
 
-## F18 for OpenMP
+## OpenMP for F18
 
 1. Define and document the parse tree representation for
     * Directives (listed below)
@@ -26,7 +26,7 @@ The directives that are in the same categories share some characteristics.
 
 #### Declarative directives
 
-An OpenMP directive that may only be placed in a declarative context.
+An OpenMP directive may only be placed in a declarative context.
 A declarative directive results in one or more declarations only;
 it is not associated with the immediate execution of any user code.
 
@@ -50,14 +50,14 @@ holds all four of the node types as discriminated unions
 along with the source provenance for the entire directive
 starting from `!$OMP`.
 
-In the `parser-tree.h`,
+In `parser-tree.h`,
 `OpenMPDeclarativeConstruct` is part
 of the `SpecificationConstruct` and `SpecificationPart`
 in F18 because
 a declarative directive can only be placed in the specification part
 of a Fortran program.
 
-All the `_Names` or `_Designators` associated
+All the `Names` or `Designators` associated
 with the declarative directive will be resolved in later phases.
 
 #### Executable directives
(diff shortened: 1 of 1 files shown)


documentation/OpenMP-semantics.md
```diff
@@ -0,0 +1,662 @@
...
+The stand-alone directive is described in the next section.
+
+The constructs associated with code blocks listed below
+share a similar structure:
+_Begin Directive_, _Clause List_, _Code Block_, _End Directive_.
+The _End Directive_ is optional for constructs
+like Loop-associated constructs.
+
+* Block-associated constructs (`OpenMPBlockConstruct`)
+* Loop-associated constructs (`OpenMPLoopConstruct`)
```
kiranchandramohan: Is there any special handling for OpenMP loop constructs like collapse(n) which affects not just the immediate loop but the outermost n loops?

documentation/OpenMP-semantics.md
```diff
@@ -0,0 +1,662 @@
...
+the data-mapping attributes of any variable referenced in a target region
+will be determined and represented as `Flag` in the `Symbol` object
+of the variable.
+No `Symbol` or `Scope` will be created.
+
+The basic steps to determine the data-mapping attribute are:
+
+1. If _map_ clause is present,
+the data-mapping attribute is determined by the _map-type_
+on the clause and its corresponding `Flag` are listed below:
```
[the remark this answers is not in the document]
ichoyjx: Thanks, I agree.

documentation/OpenMP-semantics.md
ichoyjx: Would you mind being more specific on the "special handling"? If you are talking about `collapse(n)` on the parse tree level, then no. Parse tree node is not designed for this. For `collapse(n)` or `ordered(n)`, during Semantics, we will do the `associated loop(s)` check based on the parameter `n` and the information of `associated loop(s)` (# of nesting levels, etc.) will be stored in the context of `OpenMPLoopConstruct`. More details are still TBD, for example, OpenMP 5.0 does not require perfectly nested loops any more if `collapse(n)` is present.

kiranchandramohan: Thanks for the reply.
I was asking because with constructs like collapse there is a relation between the OpenMP construct and n nested loops (not just the next loop). 
When this loop will be lowered to MLIR we might want to tag those n loops with some information so that an optimization pass can collapse them.

sscalpone: The nested loops will be children of the directive/clause so the lowering code ought to be able to apply the collapse clause.  Makes sense that the collapse could be written as an MLIR pass.

sscalpone:
> Also, Is there a plan to support the -fopenmp-simd flag?

I'll add to the project page.

documentation/OpenMP-semantics.md
schweitzpgi: The OpenMP `collapse` clause ought to be associated with the OpenMP construct in the parse tree.  The lowering of that to the OpenMP dialect of MLIR needs some design, obviously.  One possible approach could be defining a collapse attribute to some OpenMP operation that applies to its Region, which in turn contains the loop nest.

PR #697, opened 2019-08-29
Name resolution for Cray pointers

tskeith:
Resolve the pointer and pointee names in a `BasedPointerStmt` and
enforce some of the constraints on them. There are still some
constraints to be implemented, mainly about what kind of attributes
the pointers and pointees can have.

The rules for these are a little vague. I mostly followed:
- Cray Fortran Reference Manual section 9.3.2
- https://gcc.gnu.org/onlinedocs/gfortran/Cray-pointers.html
- VSI Fortran for OpenVMS Language Reference Manual section B.11

Note that the first two use the term "Cray pointer" but the last does
not. That is confusing because you have to know from context whether
it is referring to Cray pointers or Fortran pointers, so I used
"Cray pointer" and "Cray pointee" in error messages to refer to the
two names in the pointer statement.

### DIFF
--- a/lib/semantics/resolve-names.cc
+++ b/lib/semantics/resolve-names.cc
@@ -383,7 +383,7 @@ private:
 // 3. ALLOCATABLE :: x(:)
 // 4. DIMENSION :: x(10)
 // 5. COMMON x(10)
-// 6. TODO: BasedPointerStmt
+// 6. BasedPointerStmt
 class ArraySpecVisitor : public virtual BaseVisitor {
 public:
   void Post(const parser::ArraySpec &);
@@ -768,6 +768,7 @@ public:
   void Post(const parser::CommonBlockObject &);
   bool Pre(const parser::EquivalenceStmt &);
   bool Pre(const parser::SaveStmt &);
+  bool Pre(const parser::BasedPointerStmt &);
 
   void PointerInitialization(
       const parser::Name &, const parser::InitialDataTarget &);
@@ -2696,7 +2697,9 @@ bool DeclarationVisitor::Pre(const parser::BindEntity &x) {
 bool DeclarationVisitor::Pre(const parser::NamedConstantDef &x) {
   auto &name{std::get<parser::NamedConstant>(x.t).v};
   auto &symbol{HandleAttributeStmt(Attr::PARAMETER, name)};
-  if (!ConvertToObjectEntity(symbol)) {
+  if (!ConvertToObjectEntity(symbol) ||
+      symbol.test(Symbol::Flag::CrayPointer) ||
+      symbol.test(Symbol::Flag::CrayPointee)) {
     SayWithDecl(
         name, symbol, "PARAMETER attribute not allowed on '%s'"_err_en_US);
     return false;
@@ -2969,13 +2972,20 @@ Symbol &DeclarationVisitor::DeclareObjectEntity(
 // - A single "*" or "lb:*" might be assumed-size or implied-shape-list
 bool DeclarationVisitor::CheckArraySpec(const parser::Name &name,
     const Symbol &symbol, const ArraySpec &arraySpec) {
-  CHECK(arraySpec.Rank() > 0);
+  if (arraySpec.Rank() == 0) {
+    return true;
+  }
   bool isExplicit{arraySpec.IsExplicitShape()};
   bool isDeferred{arraySpec.IsDeferredShape()};
   bool isImplied{arraySpec.IsImpliedShape()};
   bool isAssumedShape{arraySpec.IsAssumedShape()};
   bool isAssumedSize{arraySpec.IsAssumedSize()};
   bool isAssumedRank{arraySpec.IsAssumedRank()};
+  if (symbol.test(Symbol::Flag::CrayPointee) && !isExplicit && !isAssumedSize) {
+    Say(name,
+        "Cray pointee '%s' must have must have explicit shape or assumed size"_err_en_US);
+    return false;
+  }
   if (IsAllocatableOrPointer(symbol) && !isDeferred && !isAssumedRank) {
     if (symbol.owner().IsDerivedType()) {  // C745
       if (IsAllocatable(symbol)) {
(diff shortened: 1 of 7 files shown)


lib/semantics/resolve-names.cc
```diff
@@ -3568,6 +3571,78 @@ bool DeclarationVisitor::Pre(const parser::StructureConstructor &x) {
...
+      Say(pointerName.source, "Cray pointer '%s' must have type %s"_err_en_US,
+          pointerName.source, pointerType.AsFortran());
+    }
+    if (ResolveName(pointeeName)) {
+      Symbol &pointee{*pointeeName.symbol};
+      if (pointee.has<UseDetails>()) {
+        Say(pointeeName,
+            "'%s' cannot be a Cray pointee as it is use-associated"_err_en_US);
+      } else if (!ConvertToObjectEntity(pointee) || IsParameter(pointee)) {
+        Say(pointeeName, "'%s' is not a variable"_err_en_US);
```
jeanPerier: Must pointees always have an `ObjectEntityDetails` ? Gfortran seems to allow pointee procedures, from the linked manual :
> A Cray pointer may also point to a function or a subroutine.

lib/semantics/resolve-names.cc
```diff
@@ -3568,6 +3571,78 @@ bool DeclarationVisitor::Pre(const parser::StructureConstructor &x) {
...
+      }
+      if (const auto *pointeeType{pointee.GetType()}) {
+        if (const auto *derived{pointeeType->AsDerived()}) {
+          if (!derived->typeSymbol().get<DerivedTypeDetails>().sequence()) {
+            Say(pointeeName,
+                "Type of Cray pointee '%s' is a non-sequence derived type"_err_en_US);
+          }
+        }
+      }
+      // process the pointee array-spec, if present
```
jeanPerier: Does this fully implement the following TODO regarding array spec ?
https://github.com/flang-compiler/f18/blob/a1af134ab004736010eacef018855cb9ddf139c6/lib/semantics/resolve-names.cc#L386

lib/semantics/resolve-names.cc
```diff
@@ -3568,6 +3571,78 @@ bool DeclarationVisitor::Pre(const parser::StructureConstructor &x) {
...
+      if (spec.empty()) {
+        // No array spec
+      } else if (pointee.Rank() > 0) {
+        SayWithDecl(pointeeName, pointee,
+            "Array spec was already declared for '%s'"_err_en_US);
+      } else if (!spec.IsExplicitShape() && !spec.IsAssumedSize()) {
+        Say(pointeeName,
+            "'%s' must have must have explicit shape or assumed size"_err_en_US);
+      }
+      ClearArraySpec();
```
jeanPerier: Shouldn't the computed shape be added to the pointee details ?

lib/semantics/resolve-names.cc
```diff
@@ -3568,6 +3571,78 @@ bool DeclarationVisitor::Pre(const parser::StructureConstructor &x) {
...
+      Walk(std::get<std::optional<parser::ArraySpec>>(bp.t));
+      const auto &spec{arraySpec()};
+      if (spec.empty()) {
+        // No array spec
+      } else if (pointee.Rank() > 0) {
+        SayWithDecl(pointeeName, pointee,
+            "Array spec was already declared for '%s'"_err_en_US);
+      } else if (!spec.IsExplicitShape() && !spec.IsAssumedSize()) {
+        Say(pointeeName,
+            "'%s' must have must have explicit shape or assumed size"_err_en_US);
```
jeanPerier: This check should probably also be done if the pointee shape is declare outside of the BasedPointeeStmt else the following will be accepted:
```
  function foo(x)
    real :: x(:)
    pointer(i, x)
  end function
```

lib/semantics/resolve-names.cc
```diff
@@ -3568,6 +3571,78 @@ bool DeclarationVisitor::Pre(const parser::StructureConstructor &x) {
...
+  for (const parser::BasedPointer &bp : x.v) {
+    const parser::ObjectName &pointerName{std::get<0>(bp.t)};
+    const parser::ObjectName &pointeeName{std::get<1>(bp.t)};
+    auto *pointer{FindSymbol(pointerName)};
+    if (!pointer) {
+      pointer = &MakeSymbol(pointerName, ObjectEntityDetails{});
+    } else if (pointer->has<UseDetails>()) {
+      Say(pointerName,
+          "'%s' cannot be a Cray pointer as it is use-associated"_err_en_US);
+    } else if (!ConvertToObjectEntity(*pointer) || IsParameter(*pointer)) {
```
jeanPerier: Just a heads up that I renamed `IsParameter` to `IsNamedConstant` so you'll have a logical conflict once you rebase/merge with master.

tskeith:
> Will Cray pointee/pointers link be stored in the pointer/pointee symbol details somehow ? I am not sure for the pointers, but that seems like a useful info in the pointee symbol.

The symbol has the `CrayPointer` or `CrayPointee` flag set. That should be good enough.

lib/semantics/resolve-names.cc
```diff
@@ -3568,6 +3571,78 @@ bool DeclarationVisitor::Pre(const parser::StructureConstructor &x) {
...
+      SayWithDecl(pointerName, *pointer, "'%s' is not a variable"_err_en_US);
+    } else if (pointer->Rank() > 0) {
+      SayWithDecl(pointerName, *pointer,
+          "Cray pointer '%s' must be a scalar"_err_en_US);
+    } else if (pointer->test(Symbol::Flag::CrayPointee)) {
+      Say(pointerName,
```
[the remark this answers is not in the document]
tskeith: It is allowed by PGI, but no other compiler I could find. I'm okay with allowing it if that's the consensus.

lib/semantics/resolve-names.cc
```diff
@@ -3568,6 +3571,78 @@ bool DeclarationVisitor::Pre(const parser::StructureConstructor &x) {
...
+bool DeclarationVisitor::Pre(const parser::BasedPointerStmt &x) {
+  for (const parser::BasedPointer &bp : x.v) {
+    const parser::ObjectName &pointerName{std::get<0>(bp.t)};
+    const parser::ObjectName &pointeeName{std::get<1>(bp.t)};
+    auto *pointer{FindSymbol(pointerName)};
+    if (!pointer) {
+      pointer = &MakeSymbol(pointerName, ObjectEntityDetails{});
+    } else if (pointer->has<UseDetails>()) {
+      Say(pointerName,
+          "'%s' cannot be a Cray pointer as it is use-associated"_err_en_US);
```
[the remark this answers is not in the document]
tskeith: You're right, we can and should allow it.

It's confusing because the pointer statement is both a declaration and an association. If the pointer is already declared and of the right type we should be able to associate a pointee with it, even if it is use-associated.

lib/semantics/resolve-names.cc
```diff
@@ -3568,6 +3571,78 @@ bool DeclarationVisitor::Pre(const parser::StructureConstructor &x) {
...
+        SayWithDecl(pointeeName, pointee,
+            "Array spec was already declared for '%s'"_err_en_US);
+      } else if (!spec.IsExplicitShape() && !spec.IsAssumedSize()) {
+        Say(pointeeName,
+            "'%s' must have must have explicit shape or assumed size"_err_en_US);
+      }
+      ClearArraySpec();
+    }
+  }
+  return false;
```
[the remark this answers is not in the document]
tskeith: Right, I'll add something.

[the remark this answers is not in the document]
tskeith: I added a map to the Scope to record this information.

Issue #220, opened 2018-10-29
Document potential runtime checks


sscalpone:
Create a document to track runtime checks that flang might implement.

- Check that assignments to entities with the contiguous attribute are indeed contiguous.
- Check that a pointer is not null or uninitialized before it is used.
- Check for stack overflow.

kiranchandramohan:
Please consider the following also,
-> Array bounds check

PR #698, opened 2019-08-29
Added -fget-definition and -fget-symbols-sources.

tinlun:
-fget-definition with specified line and two column numbers will return the line and column numbers of the symbol definition.
-fget-symbols-sources will output all symbol names inside of the file along with their line and column numbers. For symbols that exists in another module, it will print the name of the source module instead.

A few notes:
I added ReverseMap as a way to go from Provenance to CookedSource. I was only able to implement it as a sequential search because a binary search would not work when there are macros or when the source is fixed-form, as the provenance within provenance map is no longer sorted. Please let me know if my approach is wrong.
I'm unsure about some of these method names, so feedback would be appreciated.
I added a way to use a temporary file (%t) to test_any.sh.
I will squash these commits before these changes get merged.

Thanks!

### DIFF
--- a/lib/parser/provenance.cc
+++ b/lib/parser/provenance.cc
@@ -239,18 +239,18 @@ void AllSources::EmitMessage(std::ostream &o,
           [&](const Inclusion &inc) {
             o << inc.source.path();
             std::size_t offset{origin.covers.MemberOffset(range->start())};
-            std::pair<int, int> pos{inc.source.FindOffsetLineAndColumn(offset)};
-            o << ':' << pos.first << ':' << pos.second;
+            SourcePosition pos{inc.source.FindOffsetLineAndColumn(offset)};
+            o << ':' << pos.line << ':' << pos.column;
             o << ": " << message << '\n';
             if (echoSourceLine) {
               const char *text{inc.source.content() +
-                  inc.source.GetLineStartOffset(pos.first)};
+                  inc.source.GetLineStartOffset(pos.line)};
               o << "  ";
               for (const char *p{text}; *p != '\n'; ++p) {
                 o << *p;
               }
               o << "\n  ";
-              for (int j{1}; j < pos.second; ++j) {
+              for (int j{1}; j < pos.column; ++j) {
                 char ch{text[j - 1]};
                 o << (ch == '\t' ? '\t' : ' ');
               }
@@ -260,8 +260,8 @@ void AllSources::EmitMessage(std::ostream &o,
                 if (&MapToOrigin(last) == &origin) {
                   auto endOffset{origin.covers.MemberOffset(last)};
                   auto endPos{inc.source.FindOffsetLineAndColumn(endOffset)};
-                  if (pos.first == endPos.first) {
-                    for (int j{pos.second}; j < endPos.second; ++j) {
+                  if (pos.line == endPos.line) {
+                    for (int j{pos.column}; j < endPos.column; ++j) {
                       o << '^';
                     }
                   }
@@ -322,7 +322,7 @@ std::optional<SourcePosition> AllSources::GetSourcePosition(
   const Origin &origin{MapToOrigin(prov)};
   if (const auto *inc{std::get_if<Inclusion>(&origin.u)}) {
     std::size_t offset{origin.covers.MemberOffset(prov)};
-    return SourcePosition{inc->source, offset};
+    return inc->source.FindOffsetLineAndColumn(offset);
   } else {
     return std::nullopt;
   }
@@ -345,7 +345,7 @@ std::string AllSources::GetPath(Provenance at) const {
 int AllSources::GetLineNumber(Provenance at) const {
   std::size_t offset{0};
   const SourceFile *source{GetSourceFile(at, &offset)};
-  return source ? source->FindOffsetLineAndColumn(offset).first : 0;
+  return source ? source->FindOffsetLineAndColumn(offset).line : 0;
 }
 
 Provenance AllSources::CompilerInsertionProvenance(char ch) {
--- a/lib/parser/provenance.cc
+++ b/lib/parser/provenance.cc
@@ -111,17 +111,6 @@ ProvenanceRange OffsetToProvenanceMappings::Map(std::size_t at) const {
   return provenanceMap_[low].range.Suffix(offset);
 }
 
-std::optional<std::size_t> OffsetToProvenanceMappings::ReverseMap(
-    Provenance at) const {
-  for (const auto &[start, range] : provenanceMap_) {
-    if (range.Contains(at)) {
-      std::size_t offset{at.offset() - range.start().offset()};
-      return start + offset;
-    }
-  }
-  return std::nullopt;
-}
-
 void OffsetToProvenanceMappings::RemoveLastBytes(std::size_t bytes) {
   for (; bytes > 0; provenanceMap_.pop_back()) {
     CHECK(!provenanceMap_.empty());
@@ -432,38 +421,19 @@ std::optional<ProvenanceRange> CookedSource::GetProvenanceRange(
   return {ProvenanceRange{first.start(), last.start() - first.start()}};
 }
 
-std::optional<CharBlock> CookedSource::GetCharBlock(
-    ProvenanceRange range) const {
-  CHECK(!invertedMap_.empty() &&
-      "CompileProvenanceRangeToOffsetMappings not called");
-  if (auto to{invertedMap_.Map(range)}) {
-    return CharBlock{data_.c_str() + *to, range.size()};
-  } else {
-    return std::nullopt;
-  }
-}
-
 std::optional<CharBlock> CookedSource::GetCharBlockFromLineAndColumns(
     int line, int startColumn, int endColumn) const {
   // 2nd column is exclusive, meaning it is target column + 1.
   CHECK(line > 0 && startColumn > 0 && endColumn > 0);
+  CHECK(startColumn < endColumn);
   auto provenanceStart{allSources_.GetFirstFileProvenance().value().start()};
   if (auto sourceFile{allSources_.GetSourceFile(provenanceStart)}) {
     CHECK(line <= static_cast<int>(sourceFile->lines()));
-    if (auto firstOffset{
-            provenanceMap_.ReverseMap(sourceFile->GetLineStartOffset(line) +
-                provenanceStart.offset() + startColumn - 1)}) {
-      if (auto secondOffset{
-              provenanceMap_.ReverseMap(sourceFile->GetLineStartOffset(line) +
-                  provenanceStart.offset() + endColumn - 2)}) {
-        if (*secondOffset >= *firstOffset) {
-          // Returned 2nd column is also exclusive.
-          return CharBlock(
-              &data_[*firstOffset], *secondOffset - *firstOffset + 1);
-        }
-      }
-    }
+    return GetCharBlock(ProvenanceRange(sourceFile->GetLineStartOffset(line) +
+            provenanceStart.offset() + startColumn - 1,
+        endColumn - startColumn));
   }
+  return std::nullopt;
 }
 
 std::optional<std::pair<SourcePosition, SourcePosition>>
@@ -477,6 +447,17 @@ CookedSource::GetSourcePositionRange(CharBlock cookedRange) const {
     }
   }
   return std::nullopt;
+}
+
+std::optional<CharBlock> CookedSource::GetCharBlock(
+    ProvenanceRange range) const {
+  CHECK(!invertedMap_.empty() &&
+      "CompileProvenanceRangeToOffsetMappings not called");
+  if (auto to{invertedMap_.Map(range)}) {
+    return CharBlock{data_.c_str() + *to, range.size()};
+  } else {
+    return std::nullopt;
+  }
 }
 
 void CookedSource::Marshal() {
--- a/lib/parser/provenance.cc
+++ b/lib/parser/provenance.cc
@@ -111,6 +111,17 @@ ProvenanceRange OffsetToProvenanceMappings::Map(std::size_t at) const {
   return provenanceMap_[low].range.Suffix(offset);
 }
 
+std::optional<std::size_t> OffsetToProvenanceMappings::ReverseMap(
+    Provenance at) const {
+  for (const auto &[start, range] : provenanceMap_) {
+    if (range.Contains(at)) {
+      std::size_t offset{at.offset() - range.start().offset()};
+      return start + offset;
+    }
+  }
+  return std::nullopt;
+}
+
 void OffsetToProvenanceMappings::RemoveLastBytes(std::size_t bytes) {
   for (; bytes > 0; provenanceMap_.pop_back()) {
     CHECK(!provenanceMap_.empty());
@@ -306,6 +317,26 @@ const SourceFile *AllSources::GetSourceFile(
       origin.u);
 }
 
+std::optional<SourcePosition> AllSources::GetSourcePosition(
+    Provenance prov) const {
+  const Origin &origin{MapToOrigin(prov)};
+  if (const auto *inc{std::get_if<Inclusion>(&origin.u)}) {
+    std::size_t offset{origin.covers.MemberOffset(prov)};
+    return SourcePosition{inc->source, offset};
+  } else {
+    return std::nullopt;
+  }
+}
+
+std::optional<ProvenanceRange> AllSources::GetFirstFileProvenance() const {
+  for (const auto &origin : origin_) {
+    if (std::holds_alternative<Inclusion>(origin.u)) {
+      return origin.covers;
+    }
+  }
+  return std::nullopt;
+}
+
 std::string AllSources::GetPath(Provenance at) const {
   const SourceFile *source{GetSourceFile(at)};
   return source ? source->path() : ""s;
@@ -410,6 +441,42 @@ std::optional<CharBlock> CookedSource::GetCharBlock(
   } else {
     return std::nullopt;
   }
+}
+
+std::optional<CharBlock> CookedSource::GetCharBlockFromLineAndColumns(
+    int line, int startColumn, int endColumn) const {
+  // 2nd column is exclusive, meaning it is target column + 1.
+  CHECK(line > 0 && startColumn > 0 && endColumn > 0);
+  auto provenanceStart{allSources_.GetFirstFileProvenance().value().start()};
+  if (auto sourceFile{allSources_.GetSourceFile(provenanceStart)}) {
+    CHECK(line <= static_cast<int>(sourceFile->lines()));
+    if (auto firstOffset{
+            provenanceMap_.ReverseMap(sourceFile->GetLineStartOffset(line) +
+                provenanceStart.offset() + startColumn - 1)}) {
+      if (auto secondOffset{
+              provenanceMap_.ReverseMap(sourceFile->GetLineStartOffset(line) +
+                  provenanceStart.offset() + endColumn - 2)}) {
+        if (*secondOffset >= *firstOffset) {
+          // Returned 2nd column is also exclusive.
+          return CharBlock(
+              &data_[*firstOffset], *secondOffset - *firstOffset + 1);
+        }
+      }
+    }
+  }
+}
+
+std::optional<std::pair<SourcePosition, SourcePosition>>
+CookedSource::GetSourcePositionRange(CharBlock cookedRange) const {
+  if (auto range{GetProvenanceRange(cookedRange)}) {
+    if (auto firstOffset{allSources_.GetSourcePosition(range->start())}) {
+      if (auto secondOffset{
+              allSources_.GetSourcePosition(range->start() + range->size())}) {
+        return std::pair{*firstOffset, *secondOffset};
+      }
+    }
+  }
+  return std::nullopt;
 }
 
 void CookedSource::Marshal() {
--- a/lib/parser/provenance.h
+++ b/lib/parser/provenance.h
@@ -52,8 +52,6 @@ namespace Fortran::parser {
 // by the upper bits of an offset, but that does not appear to be
 // necessary.)
 
-class AllSources;
-
 class Provenance {
 public:
   Provenance() {}
@@ -124,6 +122,7 @@ public:
   void Put(ProvenanceRange);
   void Put(const OffsetToProvenanceMappings &);
   ProvenanceRange Map(std::size_t at) const;
+  std::optional<std::size_t> ReverseMap(Provenance) const;
   void RemoveLastBytes(std::size_t);
   ProvenanceRangeToOffsetMappings Invert(const AllSources &) const;
   std::ostream &Dump(std::ostream &) const;
@@ -137,6 +136,17 @@ private:
   // Elements appear in ascending order of distinct .start values;
   // their .range values are disjoint and not necessarily adjacent.
   std::vector<ContiguousProvenanceMapping> provenanceMap_;
+};
+
+struct SourcePosition {
+  SourcePosition(const SourceFile &file, int line, int column)
+    : file{file}, line{line}, column{column} {}
+  SourcePosition(const SourceFile &file, std::pair<int, int> pos)
+    : file{file}, line{pos.first}, column{pos.second} {}
+  SourcePosition(const SourceFile &, std::size_t);
+
+  const SourceFile &file;
+  int line, column;
 };
 
 // A singleton AllSources instance for the whole compilation
@@ -173,6 +183,8 @@ public:
       const std::string &message, bool echoSourceLine = false) const;
   const SourceFile *GetSourceFile(
       Provenance, std::size_t *offset = nullptr) const;
+  std::optional<SourcePosition> GetSourcePosition(Provenance) const;
+  std::optional<ProvenanceRange> GetFirstFileProvenance();
   std::string GetPath(Provenance) const;  // __FILE__
   int GetLineNumber(Provenance) const;  // __LINE__
   Provenance CompilerInsertionProvenance(char ch);
@@ -237,6 +249,10 @@ public:
 
   std::optional<ProvenanceRange> GetProvenanceRange(CharBlock) const;
   std::optional<CharBlock> GetCharBlock(ProvenanceRange) const;
+  std::optional<CharBlock> GetCharBlockFromLineAndColumns(
+      int line, int startColumn, int endColumn) const;
+  std::optional<std::pair<SourcePosition, SourcePosition>>
+      GetSourcePositionRange(CharBlock) const;
 
   // The result of a Put() is the offset that the new data
   // will have in the eventually marshaled contiguous buffer.
--- a/lib/parser/provenance.h
+++ b/lib/parser/provenance.h
@@ -138,17 +138,6 @@ private:
   std::vector<ContiguousProvenanceMapping> provenanceMap_;
 };
 
-struct SourcePosition {
-  SourcePosition(const SourceFile &file, int line, int column)
-    : file{file}, line{line}, column{column} {}
-  SourcePosition(const SourceFile &file, std::pair<int, int> pos)
-    : file{file}, line{pos.first}, column{pos.second} {}
-  SourcePosition(const SourceFile &, std::size_t);
-
-  const SourceFile &file;
-  int line, column;
-};
-
 // A singleton AllSources instance for the whole compilation
 // is shared by reference.
 class AllSources {
--- a/lib/parser/provenance.h
+++ b/lib/parser/provenance.h
@@ -52,6 +52,8 @@ namespace Fortran::parser {
 // by the upper bits of an offset, but that does not appear to be
 // necessary.)
 
+class AllSources;
+
 class Provenance {
 public:
   Provenance() {}
@@ -122,7 +124,6 @@ public:
   void Put(ProvenanceRange);
   void Put(const OffsetToProvenanceMappings &);
   ProvenanceRange Map(std::size_t at) const;
-  std::optional<std::size_t> ReverseMap(Provenance) const;
   void RemoveLastBytes(std::size_t);
   ProvenanceRangeToOffsetMappings Invert(const AllSources &) const;
   std::ostream &Dump(std::ostream &) const;
@@ -173,7 +174,7 @@ public:
   const SourceFile *GetSourceFile(
       Provenance, std::size_t *offset = nullptr) const;
   std::optional<SourcePosition> GetSourcePosition(Provenance) const;
-  std::optional<ProvenanceRange> GetFirstFileProvenance();
+  std::optional<ProvenanceRange> GetFirstFileProvenance() const;
   std::string GetPath(Provenance) const;  // __FILE__
   int GetLineNumber(Provenance) const;  // __LINE__
   Provenance CompilerInsertionProvenance(char ch);
@@ -237,11 +238,11 @@ public:
   bool IsValid(ProvenanceRange r) const { return allSources_.IsValid(r); }
 
   std::optional<ProvenanceRange> GetProvenanceRange(CharBlock) const;
-  std::optional<CharBlock> GetCharBlock(ProvenanceRange) const;
   std::optional<CharBlock> GetCharBlockFromLineAndColumns(
       int line, int startColumn, int endColumn) const;
   std::optional<std::pair<SourcePosition, SourcePosition>>
       GetSourcePositionRange(CharBlock) const;
+  std::optional<CharBlock> GetCharBlock(ProvenanceRange) const;
 
   // The result of a Put() is the offset that the new data
   // will have in the eventually marshaled contiguous buffer.
--- a/lib/parser/source.h
+++ b/lib/parser/source.h
@@ -33,6 +33,18 @@ std::string DirectoryName(std::string path);
 std::string LocateSourceFile(
     std::string name, const std::vector<std::string> &searchPath);
 
+class SourceFile;
+
+struct SourcePosition {
+  SourcePosition(const SourceFile &file, int line, int column)
+    : file{file}, line{line}, column{column} {}
+  SourcePosition(const SourceFile &file, std::pair<int, int> pos)
+    : file{file}, line{pos.first}, column{pos.second} {}
+
+  const SourceFile &file;
+  int line, column;
+};
+
 class SourceFile {
 public:
   explicit SourceFile(Encoding e) : encoding_{e} {}
@@ -46,7 +58,7 @@ public:
   bool Open(std::string path, std::stringstream *error);
   bool ReadStandardInput(std::stringstream *error);
   void Close();
-  std::pair<int, int> FindOffsetLineAndColumn(std::size_t) const;
+  SourcePosition FindOffsetLineAndColumn(std::size_t) const;
   std::size_t GetLineStartOffset(int lineNumber) const {
     return lineStart_.at(lineNumber - 1);
   }
--- a/lib/parser/source.h
+++ b/lib/parser/source.h
@@ -36,11 +36,6 @@ std::string LocateSourceFile(
 class SourceFile;
 
 struct SourcePosition {
-  SourcePosition(const SourceFile &file, int line, int column)
-    : file{file}, line{line}, column{column} {}
-  SourcePosition(const SourceFile &file, std::pair<int, int> pos)
-    : file{file}, line{pos.first}, column{pos.second} {}
-
   const SourceFile &file;
   int line, column;
 };
--- a/lib/semantics/semantics.cc
+++ b/lib/semantics/semantics.cc
@@ -40,7 +40,9 @@
 
 namespace Fortran::semantics {
 
+using NameToSymbolMap = std::map<const char *, const Symbol *>;
 static void DoDumpSymbols(std::ostream &, const Scope &, int indent = 0);
+static void GetSymbolNames(const Scope &, NameToSymbolMap &);
 static void PutIndent(std::ostream &, int indent);
 
 // A parse tree visitor that calls Enter/Leave functions from each checker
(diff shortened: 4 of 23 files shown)


lib/parser/provenance.cc
```diff
@@ -69,6 +69,18 @@ ProvenanceRange OffsetToProvenanceMappings::Map(std::size_t at) const {
   return provenanceMap_[low].range.Suffix(offset);
 }
 
+std::optional<std::size_t> OffsetToProvenanceMappings::ReverseMap(
+    Provenance at) const {
+  for (unsigned int i = 0; i < provenanceMap_.size(); i++) {
```
klausler: `std::size_t i{0};`

klausler: `++i` is preferred usage; it doesn't matter here, but with more general iterators it can save creation of a copy of the iterator.

lib/parser/provenance.cc
```diff
@@ -69,6 +69,18 @@ ProvenanceRange OffsetToProvenanceMappings::Map(std::size_t at) const {
...
+    if (provenanceMap_[i].range.Contains(at)) {
+      std::size_t offset{
+          at.offset() - provenanceMap_[i].range.start().offset()};
+      return {provenanceMap_[i].start + offset};
```
klausler: The braces are probably not necessary.

lib/parser/provenance.cc
```diff
@@ -241,6 +253,26 @@ const SourceFile *AllSources::GetSourceFile(
       origin.u);
 }
 
+std::optional<std::pair<int, int>> AllSources::GetSourcePosition(
+    Provenance prov) const {
+  const Origin &origin{MapToOrigin(prov)};
+  if (auto inc = std::get_if<Inclusion>(&origin.u)) {
```
klausler: `const auto *inc{std::get_if<Inclusion>(&origin.u)}`

lib/parser/provenance.cc
```diff
@@ -241,6 +253,26 @@ const SourceFile *AllSources::GetSourceFile(
...
+    std::size_t offset{origin.covers.MemberOffset(prov)};
+    return {std::pair<int, int>(inc->source.FindOffsetLineAndColumn(offset))};
```
klausler: You probably don't need the braces or the `pair` constructor; `FindOffsetLineAndColumn()` returns a `pair`, yes?

lib/parser/provenance.cc
```diff
@@ -241,6 +253,26 @@ const SourceFile *AllSources::GetSourceFile(
...
+  } else {
+    return std::nullopt;
+  }
+}
+
+std::optional<ProvenanceRange> AllSources::GetFirstFileProvenance() {
+  for (unsigned int i = 0; i < origin_.size(); i++) {
```
klausler: `std::size_t i{0};` and `++i`

lib/parser/provenance.cc
```diff
@@ -241,6 +253,26 @@ const SourceFile *AllSources::GetSourceFile(
...
+    if (std::get_if<Inclusion>(&origin_[i].u)) {
```
klausler: Since you don't use the result of `get_if`, `std::holds_alternative<Inclusion>(origin_[i].u)` suffices.

lib/parser/provenance.cc
```diff
@@ -328,6 +360,43 @@ std::optional<ProvenanceRange> CookedSource::GetProvenanceRange(
   return {ProvenanceRange{first.start(), last.start() - first.start()}};
 }
 
+std::optional<CharBlock> CookedSource::GetCharBlockFromLineAndColumns(
+    int line, int column1, int column2) const {
+  auto provenanceStart = allSources_.GetFirstFileProvenance().value().start();
```
klausler: We prefer the use of braced initializers; they're safer from dangerous conversions when the type is not `auto`, and consistent with other initializers when the type is `auto`.

lib/parser/provenance.cc
```diff
@@ -328,6 +360,43 @@ std::optional<ProvenanceRange> CookedSource::GetProvenanceRange(
...
+  auto sourceFile = allSources_.GetSourceFile(provenanceStart);
```
klausler: Consider using `if (auto sourceFile{allSources_.GetSourceFile(provenanceStart)}) { ... `.

lib/parser/provenance.cc
```diff
@@ -328,6 +360,43 @@ std::optional<ProvenanceRange> CookedSource::GetProvenanceRange(
...
+  if (!sourceFile) {
+    return std::nullopt;
+  }
+  std::optional<std::size_t> firstOffset =
```
klausler: A nest of `if (auto firstOffset{...}) { if (auto secondOffset{...}) { ...` would be more clear.  You're setting up a sequence of hurdles that all have to be true in order to get to a non-empty `return`.

lib/parser/provenance.cc
tinlun: I see some examples where braces are used in this way, such as CookedSource::GetProvenanceRange. Can you explain to me when it should/shouldn't be used?

lib/parser/provenance.cc
tinlun: Is this what you had in mind? I figured that too much if nesting would clog up the code, but I understand that this is easier to understand.

tinlun:
I will be incorporating Peter's current feedback with the rest of the changes.

tskeith:
One of the tests fails for me:
```
$ ctest -V -R getdefinition01
```
gets this output:
```
220: Test command: /Users/tsk/f18/tinlun-master/test/semantics/test_any.sh "getdefinition01.f90" "/Users/tsk/f18/tinlun-master/build/Release-gcc-8.3.0/tools/f18/bin/f18"
220: Test timeout computed to be: 10000000
220: Symbol not found.
220: String range: >yyy<
220: Found symbol name: yyy
220: Line: 25. Columns: 24, 27
220: Symbol not found.
220: Not found: 25 21 25 22
220: Not found: 39 24 39 25
220: FAIL  /Users/tsk/f18/tinlun-master/test/semantics/getdefinition01.f90
220: PASS  /Users/tsk/f18/tinlun-master/build/Release-gcc-8.3.0/tools/f18/bin/f18
1/1 Test #220: getdefinition01.f90 ..............***Failed    0.60 sec
```

lib/parser/provenance.cc
```diff
@@ -69,6 +69,18 @@ ProvenanceRange OffsetToProvenanceMappings::Map(std::size_t at) const {
   return provenanceMap_[low].range.Suffix(offset);
 }
 
+std::optional<std::size_t> OffsetToProvenanceMappings::ReverseMap(
+    Provenance at) const {
+  for (std::size_t i = 0; i < provenanceMap_.size(); ++i) {
```
tskeith: You can use a ranged-based for loop here:
`for (const ContiguousProvenanceMapping &mapping : provenanceMap_) {`

In fact, you can simplify it further with structured bindings:
`for (const auto &[start, range] : provenanceMap_) {`

lib/semantics/semantics.cc
```diff
@@ -228,6 +230,45 @@ void DoDumpSymbols(std::ostream &os, const Scope &scope, int indent) {
   --indent;
 }
 
+void Semantics::DumpSymbolsSources(std::ostream &os) {
+  std::list<Symbol> symbols;
+  auto less = [&](Symbol a, Symbol b) {
+    return a.name().begin() < b.name().begin();
+  };
+  std::set<Symbol, decltype(less)> visited(less);
```
tskeith: `std::set<const Symbol *>` is probably better. You shouldn't need to copy symbols into the set. Then you probably won't need `less` as well.

lib/semantics/semantics.cc
```diff
@@ -228,6 +230,45 @@ void DoDumpSymbols(std::ostream &os, const Scope &scope, int indent) {
...
+                   symbol.detailsIf<semantics::UseDetails>()}) {
+      os << symbol.name().ToString() << ": "
+         << useDetails->rootModule().name().ToString() << "\n";
+      visited.emplace(symbol);
+    }
+  }
+}
+
+void GetAllSymbols(std::ostream &os, const Scope &scope,
+    const parser::CookedSource &cooked, std::list<Symbol> &symbols) {
```
tskeith: What are `os` and `cooked` for?

tskeith: If you use `std::set` instead of `std::list` you won't need `visited` up above, right?

lib/semantics/semantics.cc
```diff
@@ -228,6 +230,45 @@ void DoDumpSymbols(std::ostream &os, const Scope &scope, int indent) {
...
+    }
+    if (sourceInfo) {
+      os << symbol.name().ToString() << ": " << std::get<0>(*sourceInfo) << ","
+         << std::get<1>(*sourceInfo) << "," << std::get<0>(*sourceInfo) << ","
+         << std::get<2>(*sourceInfo) << "\n";
+      visited.emplace(symbol);
+    } else if (const auto useDetails{
+                   symbol.detailsIf<semantics::UseDetails>()}) {
+      os << symbol.name().ToString() << ": "
+         << useDetails->rootModule().name().ToString() << "\n";
```
tskeith: I suggest using `symbol.GetUltimate().owner().symbol()->name()` to get the module name. Then you don't need `rootModule()`.

test/semantics/getdefinition01.f90
```diff
@@ -0,0 +1,51 @@
...
+!
+! Unless required by applicable law or agreed to in writing, software
+! distributed under the License is distributed on an "AS IS" BASIS,
+! WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+! See the License for the specific language governing permissions and
+! limitations under the License.
+
+! Tests -fget-definition returning source position of symbol definition.
+
+!DEF: /m Module
```
tskeith: What's the purpose of the "DEF:" and "REF:" comments.

lib/parser/provenance.cc
```diff
@@ -241,6 +253,26 @@ const SourceFile *AllSources::GetSourceFile(
...
+  if (const auto inc = std::get_if<Inclusion>(&origin.u)) {
+    std::size_t offset{origin.covers.MemberOffset(prov)};
+    return inc->source.FindOffsetLineAndColumn(offset);
+  } else {
+    return std::nullopt;
+  }
+}
+
+std::optional<ProvenanceRange> AllSources::GetFirstFileProvenance() {
+  for (std::size_t i = 0; i < origin_.size(); ++i) {
```
vjayathirtha-nv: range based for loop here as well.

tools/f18/f18.cc
```diff
@@ -478,6 +524,14 @@ int main(int argc, char *const argv[]) {
       driver.encoding = Fortran::parser::Encoding::UTF_8;
     } else if (arg == "-flatin") {
       driver.encoding = Fortran::parser::Encoding::LATIN_1;
+    } else if (arg == "-fget-definition") {
+      driver.getDefinition = true;
+      for (int i = 0; i < 3; i++) {
+        driver.sourcePosition[i] = std::stoi(args.front());
+        args.pop_front();
+      }
```
tskeith: What happens if the user provides incorrect input? (I.e. anything that isn't three numbers in the appropriate range.)

lib/parser/provenance.h
```diff
@@ -143,6 +144,8 @@ class AllSources {
       const std::string &message, bool echoSourceLine = false) const;
   const SourceFile *GetSourceFile(
       Provenance, std::size_t *offset = nullptr) const;
+  std::optional<std::pair<int, int>> GetSourcePosition(Provenance) const;
```
tskeith: I think you should replace the `std::pair` with a struct so that the members have names. It makes it easier to figure out what these two `int`s represent.

lib/parser/provenance.h
```diff
@@ -208,6 +211,9 @@ class CookedSource {
   bool IsValid(ProvenanceRange r) const { return allSources_.IsValid(r); }
 
   std::optional<ProvenanceRange> GetProvenanceRange(CharBlock) const;
+  std::optional<CharBlock> GetCharBlockFromLineAndColumns(int, int, int) const;
+  std::optional<std::tuple<int, int, int>> GetLineAndColumnsFromCooked(
```
tskeith: Same thing here.

test/semantics/getdefinition01.f90
tinlun: These comments were copied over from another test. They act as sanity check while I was writing the checks, and they also act as sporadic comments for testing purposes.

tinlun:
> One of the tests fails for me:
> 
> ```
> $ ctest -V -R getdefinition01
> ```
> 
> gets this output:
> 
> ```
> 220: Test command: /Users/tsk/f18/tinlun-master/test/semantics/test_any.sh "getdefinition01.f90" "/Users/tsk/f18/tinlun-master/build/Release-gcc-8.3.0/tools/f18/bin/f18"
> 220: Test timeout computed to be: 10000000
> 220: Symbol not found.
> 220: String range: >yyy<
> 220: Found symbol name: yyy
> 220: Line: 25. Columns: 24, 27
> 220: Symbol not found.
> 220: Not found: 25 21 25 22
> 220: Not found: 39 24 39 25
> 220: FAIL  /Users/tsk/f18/tinlun-master/test/semantics/getdefinition01.f90
> 220: PASS  /Users/tsk/f18/tinlun-master/build/Release-gcc-8.3.0/tools/f18/bin/f18
> 1/1 Test #220: getdefinition01.f90 ..............***Failed    0.60 sec
> ```

Sorry, I stealth pushed a fix, should pass now.

lib/semantics/semantics.cc
tinlun: For the purpose of not printing duplicate definitions, I can't use `Symbol *` because the duplicates aren't the same instances of each other, they just have the same SourceName. I'm not sure of a way to compare their SourceName without using a custom comparator. Maybe I can add an operator < for Symbol but I don't think that really makes sense.

tskeith: Can you give an example where there are two different `Symbol *` that have the same `SourceName`? I'm not sure that should ever happen.

lib/semantics/semantics.cc
```diff
@@ -228,6 +230,45 @@ void DoDumpSymbols(std::ostream &os, const Scope &scope, int indent) {
...
+  GetAllSymbols(os, context_.globalScope(), cooked_, symbols);
+  for (auto symbol : symbols) {
+    auto sourceInfo = cooked_.GetLineAndColumnsFromCooked(symbol.name());
+    if (visited.find(symbol) != visited.end()) {
+      continue;
+    }
+    if (sourceInfo) {
+      os << symbol.name().ToString() << ": " << std::get<0>(*sourceInfo) << ","
+         << std::get<1>(*sourceInfo) << "," << std::get<0>(*sourceInfo) << ","
+         << std::get<2>(*sourceInfo) << "\n";
```
tskeith: Why print element 0 of the tuple (the line number?) twice?

It might be easier to understand the output if the format were:
name: line, col1-col2

lib/parser/provenance.cc
```diff
@@ -241,6 +253,26 @@ const SourceFile *AllSources::GetSourceFile(
       origin.u);
 }
 
+std::optional<std::pair<int, int>> AllSources::GetSourcePosition(
+    Provenance prov) const {
+  const Origin &origin{MapToOrigin(prov)};
+  if (const auto inc = std::get_if<Inclusion>(&origin.u)) {
```
tskeith: When you use `const auto` for something that is a pointer, we always include the `*`. So:
`if (const auto *inc{...}) {`

lib/parser/provenance.cc
```diff
@@ -328,6 +360,45 @@ std::optional<ProvenanceRange> CookedSource::GetProvenanceRange(
   return {ProvenanceRange{first.start(), last.start() - first.start()}};
 }
 
+std::optional<CharBlock> CookedSource::GetCharBlockFromLineAndColumns(
+    int line, int column1, int column2) const {
+  // 2nd column is exclusive, meaning it is target column + 1.
+  auto provenanceStart{allSources_.GetFirstFileProvenance().value().start()};
```
vjayathirtha-nv: `GetFirstFileProvenance()` could return `std::nullopt`

lib/semantics/semantics.cc
tinlun: You're right, that format is easier to understand. line1,col1,line2,col2 format was just easier to parse so I used it, but your suggestion is better.

lib/parser/provenance.cc
```diff
@@ -328,6 +360,45 @@ std::optional<ProvenanceRange> CookedSource::GetProvenanceRange(
...
+    if (auto firstOffset{
+            provenanceMap_.ReverseMap(sourceFile->GetLineStartOffset(line) +
+                provenanceStart.offset() + column1 - 1)}) {
+      if (auto secondOffset{
+              provenanceMap_.ReverseMap(sourceFile->GetLineStartOffset(line) +
+                  provenanceStart.offset() + column2 - 2)}) {
+        if (*secondOffset >= *firstOffset) {
+          // Returned 2nd column is also exclusive.
+          return CharBlock(
+              &data_[0] + *firstOffset, *secondOffset - *firstOffset + 1);
```
vjayathirtha-nv: I feel `&data[*firstOffset]` would be easier to read.

lib/parser/provenance.cc
tinlun: I figured we would never come across a situation where we don't find a file while running this function so I didn't add a guard. Maybe it's a good idea to guard it anyways. What do you think?

vjayathirtha-nv: I think you're okay. Unless someone else can think of a case where it could be a problem.

lib/semantics/semantics.cc
tinlun: For example, if I print out each symbol's address and information in getdefinition01.f90, and then each symbol's source position (without excluding duplicates), I get:
```
0x7fffbd3d1760
m: Module
0x7fffbd3d1d90
f, PRIVATE, PURE, RECURSIVE (Function): Subprogram () result(REAL(4) x)
0x7fffbd3d1970
m, PUBLIC: Misc ScopeName
0x7fffbd3d1b80
s, BIND(C), PUBLIC, PURE (Subroutine): Subprogram (REAL(4) x, REAL(4) yyy)
0x7fffbd3d1fa0
s (Subroutine): HostAssoc
0x7fffbd3d21b0
ss, PURE (Subroutine): Subprogram ()
0x7fffbd3d23c0
x, INTENT(IN) (Implicit): ObjectEntity dummy type: REAL(4)
0x7fffbd3d25d0
yyy, INTENT(INOUT) (Implicit): ObjectEntity dummy type: REAL(4)
0x7fffbd3d27e0
ss (Subroutine): HostAssoc
0x7fffbd3d29f0
f (Function): HostAssoc
0x7fffbd3d2c00
x, ALLOCATABLE: ObjectEntity funcResult type: REAL(4)

m: 18,8,18,9
m: 18,8,18,9
s: 25,18,25,19
f: 37,26,37,27
s: 25,18,25,19
ss: 32,19,32,21
x: 25,21,25,22
yyy: 25,24,25,27
ss: 32,19,32,21
f: 37,26,37,27
x: 39,24,39,25
```

With excluding duplicates using visited, the source position list looks like: 
```
m: 18,8,18,9
f: 37,26,37,27
s: 25,18,25,19
ss: 32,19,32,21
x: 25,21,25,22
yyy: 25,24,25,27
x: 39,24,39,25
```

It seems like all of these symbols found are different from each other, but they can resolve into the same SourceName.

tools/f18/f18.cc
tinlun: Currently stoi throws an error if the argument provided is not an int, which happens when -fget-definitions is provided with less than 3 arguments. Should I catch the exception or just let it error?

lib/parser/provenance.h
tinlun: I replaced the tuple with std::pair of the SourcePosition struct. Still verbose, but hopefully is much more clear what it represents.

lib/semantics/semantics.cc
tinlun: Change DumpSymbolsSources to use a map between Symbol's SourceName and pointer, thanks to Tim's suggestions and patch.

lib/parser/provenance.cc
klausler: They're not needed for `std::optional<>` results.  Some old versions of C++ compilers &/or STL header files cause bogus errors to be emitted for constructors of C++17 template classes and the braces can be used as workarounds.

lib/parser/provenance.cc
klausler: Why?

vjayathirtha-nv: Well, ideally for an array x[] `x+offset` would be the most readable way. But since `data` is an std::string which only allows array access syntax, it is easier to read it as address of the char at `firstOffset` position rather than, address of first element + `firstOffest`. I suppose this is a subjective preference.

klausler: Agree; thanks.

lib/parser/provenance.cc
tinlun: Thank you.

lib/parser/provenance.cc
```diff
@@ -241,6 +252,26 @@ const SourceFile *AllSources::GetSourceFile(
       origin.u);
 }
 
+std::optional<SourcePosition> AllSources::GetSourcePosition(
+    Provenance prov) const {
+  const Origin &origin{MapToOrigin(prov)};
+  if (const auto *inc{std::get_if<Inclusion>(&origin.u)}) {
+    std::size_t offset{origin.covers.MemberOffset(prov)};
+    return SourcePosition{inc->source.FindOffsetLineAndColumn(offset)};
```
tskeith: It seems like `FindOffsetLineAndColumn` should return `SourcePosition` too.

lib/parser/provenance.cc
```diff
@@ -328,6 +359,44 @@ std::optional<ProvenanceRange> CookedSource::GetProvenanceRange(
...
+        }
+      }
+    }
+  }
+  return std::nullopt;
+}
+
+std::optional<std::pair<SourcePosition, SourcePosition>>
+CookedSource::GetSourcePositionRange(CharBlock cookedRange) const {
+  // Returns tuple of (line, column1, column2).
```
tskeith: This comment is no longer accurate.

lib/parser/provenance.h
```diff
@@ -143,6 +151,8 @@ class AllSources {
       const std::string &message, bool echoSourceLine = false) const;
   const SourceFile *GetSourceFile(
       Provenance, std::size_t *offset = nullptr) const;
+  std::optional<SourcePosition> GetSourcePosition(Provenance) const;
+  std::optional<ProvenanceRange> GetFirstFileProvenance();
```
tskeith: Should be a `const` function?

lib/semantics/semantics.cc
```diff
@@ -228,6 +230,35 @@ void DoDumpSymbols(std::ostream &os, const Scope &scope, int indent) {
   --indent;
 }
 
+void Semantics::DumpSymbolsSources(std::ostream &os) {
+  NameToSymbolMap symbols;
+  GetAllSymbols(context_.globalScope(), symbols);
+  for (const auto pair : symbols) {
```
tskeith: If you add `const Symbol &symbol{*pair.second};` here you can use `symbol` below and it will be easier to understand.

lib/semantics/symbol.h
```diff
@@ -679,4 +679,4 @@ inline bool ProcEntityDetails::HasExplicitInterface() const {
   return false;
 }
 }
-#endif  // FORTRAN_SEMANTICS_SYMBOL_H_
+#endif  // FORTRAN_SEMANTICS_SYMBOL_H_
```
tskeith: It looks like you accidentally remove the newline from the end of this file.

tinlun: Oops. VS Code has been doing really weird things to this file specifically.

tools/f18/f18.cc
tskeith: An unhandled exception is not a good way to report a user error. You can't catch the exception because we build with `-fno-exceptions`. Probably `std::strtol` is the best way to do this conversion.

lib/parser/provenance.cc
```diff
@@ -328,6 +359,44 @@ std::optional<ProvenanceRange> CookedSource::GetProvenanceRange(
   return {ProvenanceRange{first.start(), last.start() - first.start()}};
 }
 
+std::optional<CharBlock> CookedSource::GetCharBlockFromLineAndColumns(
+    int line, int column1, int column2) const {
```
klausler: There are better names than `column1` and `column2` that could be used.

lib/parser/provenance.h
```diff
@@ -97,6 +104,7 @@ class OffsetToProvenanceMappings {
   void Put(ProvenanceRange);
   void Put(const OffsetToProvenanceMappings &);
   ProvenanceRange Map(std::size_t at) const;
+  std::optional<std::size_t> ReverseMap(Provenance at) const;
```
klausler: You can omit the name `at` and we typically do so unless it has documentary value.

lib/parser/provenance.h
```diff
@@ -208,6 +218,9 @@ class CookedSource {
   bool IsValid(ProvenanceRange r) const { return allSources_.IsValid(r); }
 
   std::optional<ProvenanceRange> GetProvenanceRange(CharBlock) const;
+  std::optional<CharBlock> GetCharBlockFromLineAndColumns(int, int, int) const;
```
klausler: Here's a case where argument names would reduce the mystery of the three `int` arguments.

lib/semantics/semantics.cc
```diff
@@ -39,7 +39,9 @@
 
 namespace Fortran::semantics {
 
+using NameToSymbolMap = std::map<const char *, const Symbol *>;
 static void DoDumpSymbols(std::ostream &, const Scope &, int indent = 0);
+static void GetAllSymbols(const Scope &, NameToSymbolMap &);
```
klausler: Why is a prototype needed for this static function?

lib/semantics/semantics.cc
```diff
@@ -228,6 +230,35 @@ void DoDumpSymbols(std::ostream &os, const Scope &scope, int indent) {
   --indent;
 }
 
+void Semantics::DumpSymbolsSources(std::ostream &os) {
```
klausler: Can this function be `const`?

lib/semantics/semantics.cc
```diff
@@ -228,6 +230,35 @@ void DoDumpSymbols(std::ostream &os, const Scope &scope, int indent) {
...
+         << "\n";
+    } else if (pair.second->has<semantics::UseDetails>()) {
+      os << pair.second->name().ToString() << ": "
+         << pair.second->GetUltimate().owner().symbol()->name().ToString()
+         << "\n";
+    }
+  }
+}
+
+void GetAllSymbols(const Scope &scope, NameToSymbolMap &symbols) {
```
klausler: Might be better as a `const` member function.

lib/semantics/semantics.cc
```diff
@@ -228,6 +230,35 @@ void DoDumpSymbols(std::ostream &os, const Scope &scope, int indent) {
...
+  for (const auto &pair : scope) {
```
klausler: You're taking all of the names in a scope and its children and flattening them into one `map` in such a way that conflicts are resolved in favor of `COMMON` blocks and the most deeply nested child scopes.

tools/f18/f18.cc
```diff
@@ -101,6 +101,9 @@ struct DriverOptions {
   bool unparseTypedExprsToPGF90{false};
   std::vector<std::string> pgf90Args;
   const char *prefix{nullptr};
+  bool getDefinition{false};
+  bool getSymbolsSources{false};
+  int sourcePosition[3];
```
klausler: Can you please document or rename this magic three-`int` encoding?  Maybe you could create a struct for it and pass that struct as an argument below rather than marshaling the three `int`s as individual arguments.

lib/parser/provenance.h
```diff
@@ -52,6 +52,13 @@ namespace Fortran::parser {
 // by the upper bits of an offset, but that does not appear to be
 // necessary.)
 
+struct SourcePosition {
```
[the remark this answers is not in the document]
tinlun: Currently I don't think it's necessary to include the file context as well, just line and column number would suffice. What do you think I should name it instead? Maybe just "Position"? And if file context is excluded from the struct we can use it for FindOffsetLineAndColumn as well.

Maybe it's a good idea to include the file context anyways for future cases?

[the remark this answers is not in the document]
klausler: Without file context, how do you distinguish hits in `INCLUDE` lines and `#include` directive files from hits in source files?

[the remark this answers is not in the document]
tinlun: You're right. That is a case I was not testing for.

lib/semantics/semantics.cc
tinlun: I think this is fine because symbols with the same name resolve into the same source position and I am not using any other symbol information.

lib/semantics/semantics.cc
tinlun: I think this function is too specific to be a member function of Semantics, because its purpose is to find all symbol names and not collect duplicates. I will rename it to be more specific.

lib/semantics/semantics.cc
tinlun: I thought that it is good practice to have function prototypes so that functions can be referenced before it's defined. I also see that PutIndent has a function prototype in the same way. Can you explain to me why it would be a bad idea to have a function prototype here?

lib/parser/provenance.cc
```diff
@@ -69,6 +69,17 @@ ProvenanceRange OffsetToProvenanceMappings::Map(std::size_t at) const {
   return provenanceMap_[low].range.Suffix(offset);
 }
 
+std::optional<std::size_t> OffsetToProvenanceMappings::ReverseMap(
+    Provenance at) const {
+  for (const auto &[start, range] : provenanceMap_) {
```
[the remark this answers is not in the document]
tinlun: I have discussed with Peter about this. Even with excluding compiler generated provenances, the problem is that using INCLUDE and #include inserts out of order provenance into the provenance map, so a binary search would not work.

tools/f18/f18.cc
tinlun: I have renamed and commented the variable. I could not find a clean way to add a struct for this.

lib/parser/provenance.cc
tskeith: True, so maybe it's not the right thing to call here. But given the addition of the `SourcePosition` type, `FindOffsetLineAndColumn()` should be modified to make use of it.

lib/semantics/semantics.cc
klausler: If there is a variable `x` and a common block `/x/` in the same subprogram, clicking on a use of `x` will resolve to the common block, and that will always be wrong.  (19.3.2)

If there is a variable `x` in a subprogram and a variable `x` in one or more nested `BLOCK` constructs, whither will a click on a use of `x` resolve -- a declaration in the last `BLOCK`?

lib/semantics/semantics.cc
klausler: In modern C++ usage, prototypes of non-member functions are rare outside of header files unless they're required for mutual recursion.  It's more common to define the static function ahead of its first use, and then you don't have to worry about keeping the prototype and the definition consistent in later changes.

lib/parser/provenance.cc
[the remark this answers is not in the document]
klausler: A reverse map data structure that inverts the one in `CookedSource` could be constructed as part of the same process that builds the current one.  I'll take a whack at it.

tools/f18/f18.cc
klausler: What were the problems that prevented you from defining a `struct`?

tinlun: To add a struct, I have to loop through the 3 arguments to validate them and store them in a temporary int array before putting it into the struct, which seems moot because the result is still an array. Now that it is appropriately named and commented with what the values are, I think it's more clear what the data represents.

lib/semantics/semantics.cc
tinlun: I see. Then do you think PutIndent should be defined earlier and its prototype removed?

klausler: It should probably have been written so, yes.

tools/f18/f18.cc
klausler: It's great that there's a comment there now, and it's definitely "more clear", but it's still a long way from being as obviously readable as it could trivially could be as a structure.

lib/semantics/semantics.cc
tinlun: Looking into the cases that you presented, in a subprogram with a variable `x` and a common block `/x/`, the function collects both the variable `x` and the common block `/x/` because they resolve into different locations. A similar thing happens for variable `x` in subprogram and variable `x` in `BLOCK` construct, both of them are collected into the map.

Example:
```
17 program main
18   integer :: x
19   integer :: y
20   common /x/ y
21   x = y
22 end program

f18  -fparse-only -fdebug-semantics  -fget-symbols-sources getsymbols04.f90
main: getsymbols04.f90, 17, 9-13
x: getsymbols04.f90, 18, 14-15
y: getsymbols04.f90, 19, 14-15
x: getsymbols04.f90, 20, 11-12
```
```
17 program main
18   integer :: x
19   integer :: y
20   block
21     integer :: x
22     x = y
23   end block
24 end program

f18  -fparse-only -fdebug-semantics  -fget-symbols-sources getsymbols05.f90
main: getsymbols05.f90, 17, 9-13
x: getsymbols05.f90, 18, 14-15
y: getsymbols05.f90, 19, 14-15
x: getsymbols05.f90, 21, 16-17
```

tools/f18/f18.cc
tinlun: I've added a structure to represent these arguments.

lib/parser/provenance.cc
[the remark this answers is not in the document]
klausler: Please review https://github.com/flang-compiler/f18/pull/715 when convenient; it implements an inverted mapping that's O(N) to build but logarithmic to probe.  (If github doesn't let you review it, let us know.)

lib/semantics/semantics.cc
klausler: Got it, you're comparing the values of `CharBlock`s rather than the contents of `CharBlock`s.

tinlun:
I've addressed all reviews in this pull request. Let me know if there's anything else I should take a look at.

tools/f18/f18.cc
```diff
@@ -76,6 +76,12 @@ void CleanUpAtExit() {
   }
 }
 
+struct GetDefinitionArgs {
+  GetDefinitionArgs(int line, int startColumn, int endColumn)
+    : line{line}, startColumn{startColumn}, endColumn{endColumn} {}
```
tskeith: This constructor is unnecessary.

test/semantics/CMakeLists.txt
```diff
@@ -242,6 +242,22 @@ set(FORALL_TESTS
   forall*.[Ff]90
 )
 
+set(GETSYMBOLS_TESTS
+  getsymbols01.f90
+  getsymbols02-*.f90
+  getsymbols03-a.f90
```
tskeith: Is this supposed to be `getsymbols03-*.f90`?

tskeith: > Is this supposed to be `getsymbols03-*.f90`?

Nevermind, I see that `getsymbols03-b.f90` is only included, not compiled.

lib/parser/source.h
```diff
@@ -33,6 +33,18 @@ std::string DirectoryName(std::string path);
 std::string LocateSourceFile(
     std::string name, const std::vector<std::string> &searchPath);
 
+class SourceFile;
+
+struct SourcePosition {
+  SourcePosition(const SourceFile &file, int line, int column)
+    : file{file}, line{line}, column{column} {}
+  SourcePosition(const SourceFile &file, std::pair<int, int> pos)
+    : file{file}, line{pos.first}, column{pos.second} {}
```
tskeith: These constructors aren't necessary.

lib/parser/provenance.cc
```diff
@@ -228,18 +228,18 @@ void AllSources::EmitMessage(std::ostream &o,
           [&](const Inclusion &inc) {
             o << inc.source.path();
             std::size_t offset{origin.covers.MemberOffset(range->start())};
-            std::pair<int, int> pos{inc.source.FindOffsetLineAndColumn(offset)};
-            o << ':' << pos.first << ':' << pos.second;
+            SourcePosition pos{inc.source.FindOffsetLineAndColumn(offset)};
```
klausler: Should the `SourceFile` in the `SourcePosition` be emitted, or checked, or both?

lib/parser/provenance.cc
```diff
@@ -306,6 +306,34 @@ const SourceFile *AllSources::GetSourceFile(
...
+std::optional<ProvenanceRange> AllSources::GetFirstFileProvenance() const {
+  for (const auto &origin : origin_) {
+    if (std::holds_alternative<Inclusion>(origin.u)) {
+      return origin.covers;
+    }
+  }
+  return std::nullopt;
+}
+
+ProvenanceRange AllSources::GetContiguousRangeAround(
```
klausler: I don't think that this member function is ever used.

lib/semantics/semantics.cc
```diff
@@ -184,6 +198,23 @@ void Semantics::DumpSymbols(std::ostream &os) {
   DoDumpSymbols(os, context_.globalScope());
 }
 
+void Semantics::DumpSymbolsSources(std::ostream &os) const {
+  NameToSymbolMap symbols;
+  GetSymbolNames(context_.globalScope(), symbols);
+  for (const auto pair : symbols) {
+    const Symbol &symbol{*pair.second};
+    auto sourceInfo{cooked_.GetSourcePositionRange(symbol.name())};
```
klausler: These two lines should be combined.

tools/f18/f18.cc
```diff
@@ -247,6 +255,45 @@ std::string CompileFortran(std::string path, Fortran::parser::Options options,
...
+      auto symbol{scope.FindSymbol(*cb)};
+      if (!symbol) {
+        std::cerr << notFoundText;
+        exitStatus = EXIT_FAILURE;
+        return {};
+      }
+      std::cerr << "Found symbol name: "
+                << std::string(symbol->name().begin(), symbol->name().size())
+                << "\n";
+      auto sourceInfo{parsing.cooked().GetSourcePositionRange(symbol->name())};
```
klausler: This declaration should be the predicate of an `if ()`.

tools/f18/f18.cc
```diff
@@ -247,6 +255,45 @@ std::string CompileFortran(std::string path, Fortran::parser::Options options,
           std::cout, parseTree, driver.encoding);
       return {};
     }
+    if (driver.getSymbolsSources) {
+      semantics.DumpSymbolsSources(std::cout);
+      return {};
+    }
+    if (driver.getDefinition) {
+      std::string notFoundText{"Symbol not found.\n"};
+      auto cb{parsing.cooked().GetCharBlockFromLineAndColumns(
```
klausler: This declaration should be the predicate of an `if ()`.

tools/f18/f18.cc
klausler: This comment remains unresolved.

tinlun: You might be looking at an outdated version. Github didn't mark this review as outdated.

lib/parser/provenance.cc
tinlun: My bad, this was incorrectly merged.

tools/f18/f18.cc
klausler: If this conversation is resolved, please click the "Resolve conversation" button.  The conversation appears on the current sources and is not marked as being "outdated".

lib/parser/provenance.cc
tinlun: What about `SourceFile` do you think should be emitted or checked? Currently the `SourceFile`'s path is already emitted through `o << inc.source.path();`.

klausler: That's fine, then.

lib/parser/provenance.h
```diff
@@ -173,6 +173,9 @@ class AllSources {
       const std::string &message, bool echoSourceLine = false) const;
   const SourceFile *GetSourceFile(
       Provenance, std::size_t *offset = nullptr) const;
+  std::optional<SourcePosition> GetSourcePosition(Provenance) const;
+  std::optional<ProvenanceRange> GetFirstFileProvenance() const;
+  ProvenanceRange GetContiguousRangeAround(ProvenanceRange) const;
```
klausler: I think that this member function prototype is for an unused function that no longer exists.

tinlun:
This PR is ready to be merged. I don't have merge access so I would appreciate it if someone would. Thank you for the reviews!

PR #699, opened 2019-08-30
Implement folding of x**y where y is real or complex

### LINKED ISSUE #695: PARAMETER cannot be computed as a constant value
```
% flang -c foo.f90
foo.f90:3:33: error: Initialization expression for PARAMETER 'tmax_r' (1.e1_8**9._8) cannot be computed as a constant value
    real(8), parameter, public :: tmax_r = 10.0d0**tmax
                                  ^^^^^^
```
```
% flang -c bar.f90
error: Initialization expression for PARAMETER 'tmax_r' (1.e1_8**9._8) cannot be computed as a constant value
```
```
% cat foo.f90
module am
  real(8), parameter, public :: tmax = 9.0d0
  real(8), parameter, public :: tmax_r = 10.0d0**tmax
end module am

subroutine a
use am
  real(8), parameter :: t = tmax_r
end
```
```
% cat bar.f90
subroutine b
use am
  real(8), parameter :: t = tmax_r
end
```

jeanPerier:
Fix #695 which was actually a TODO.
The implementation uses the host run-time function pow, either from libm or libpgmath.


lib/evaluate/intrinsics-library.cc
```diff
@@ -66,9 +66,10 @@ static void AddLibmRealHostProcedures(
       {"exp", F{std::exp}, true}, {"gamma", F{std::tgamma}, true},
       {"hypot", F2{std::hypot}, true}, {"log", F{std::log}, true},
       {"log10", F{std::log10}, true}, {"log_gamma", F{std::lgamma}, true},
-      {"mod", F2{std::fmod}, true}, {"sin", F{std::sin}, true},
-      {"sinh", F{std::sinh}, true}, {"sqrt", F{std::sqrt}, true},
-      {"tan", F{std::tan}, true}, {"tanh", F{std::tanh}, true}};
+      {"mod", F2{std::fmod}, true}, {"pow", F2{std::pow}, true},
+      {"sin", F{std::sin}, true}, {"sinh", F{std::sinh}, true},
+      {"sqrt", F{std::sqrt}, true}, {"tan", F{std::tan}, true},
+      {"tanh", F{std::tanh}, true}};
```
tskeith: Adding a comma after the last element of the array initializer (i.e. `}}` -> `},}`) gets clang-format to format these one per line, which makes diffs easier to read.

lib/evaluate/intrinsics-library.cc
```diff
@@ -92,16 +93,23 @@ template<typename HostT>
...
       {"atan", F{std::atan}, true}, {"atanh", F{std::atanh}, true},
       {"cos", F{std::cos}, true}, {"cosh", F{std::cosh}, true},
       {"exp", F{std::exp}, true}, {"log", F{std::log}, true},
-      {"sin", F{std::sin}, true}, {"sinh", F{std::sinh}, true},
-      {"sqrt", F{std::sqrt}, true}, {"tan", F{std::tan}, true},
-      {"tanh", F{std::tanh}, true}};
+      {"pow", F2{std::pow}, true}, {"pow", F2a{std::pow}, true},
+      {"pow", F2b{std::pow}, true}, {"sin", F{std::sin}, true},
+      {"sinh", F{std::sinh}, true}, {"sqrt", F{std::sqrt}, true},
+      {"tan", F{std::tan}, true}, {"tanh", F{std::tanh}, true}};
```
tskeith: Same here.

lib/evaluate/intrinsics-library.cc
jeanPerier: Thanks, the output is much clearer.

PR #706, opened 2019-09-02
Fold more cases of cmplx and dcmplx.

sscalpone:


sscalpone:
I'll add some tests soon.

PR #711, opened 2019-09-04
Tests for call semantics

klausler:
Here's some new tests for semantics to exercise the constraints and restrictions that we'll need to check on procedures and their references.  The tests are not enabled.  More to come, too.

### DIFF
--- a/test/semantics/call01.f90
+++ b/test/semantics/call01.f90
@@ -1,0 +1,130 @@
+! Copyright (c) 2019, NVIDIA CORPORATION.  All rights reserved.
+!
+! Licensed under the Apache License, Version 2.0 (the "License");
+! you may not use this file except in compliance with the License.
+! You may obtain a copy of the License at
+!
+!     http://www.apache.org/licenses/LICENSE-2.0
+!
+! Unless required by applicable law or agreed to in writing, software
+! distributed under the License is distributed on an "AS IS" BASIS,
+! WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+! See the License for the specific language governing permissions and
+! limitations under the License.
+
+! Confirm enforcement of constraints and restrictions in 15.6.2.1
+
+non_recursive function f01(n) result(res)
+  integer, value :: n
+  integer :: res
+  if (n <= 0) then
+    res = n
+  else
+    !ERROR: non recursive function can't recurse
+    res = n * f01(n-1) ! 15.6.2.1(3)
+  end if
+end function
+
+non_recursive function f02(n) result(res)
+  integer, value :: n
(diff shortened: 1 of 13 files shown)


test/semantics/call01.f90
```diff
@@ -0,0 +1,130 @@
...
+    res = ''
+  else
+    res = nested()
+  end if
+ contains
+  character(1) function nested
+    !ERROR: assumed-length CHARACTER(*) function can't recurse
+    nested = f14(n-1) ! 15.6.2.1(3)
+  end function nested
+end function
```
tskeith: C723 places constraints on a "function name declared with an asterisk type-param-value", not just `CHARACTER(*)` functions. But it seems that other constraints prevent `TYPE(t(*))`, for example. Is that right?

klausler: That's exactly right and the Standard should be more specific.  But note that the relevant text in the Standard is in a smaller font: that means that the text pertains only to a deprecated feature.  So there's no point asking for a clarifying edit.

test/semantics/call03.f90
```diff
@@ -0,0 +1,305 @@
...
+    ! ERROR: effective argument associated with INTENT(OUT) dummy must be definable
+    call intentout(3.14159)
+    ! ERROR: effective argument associated with INTENT(OUT) dummy must be definable
+    call intentout(in + 1.)
+    ! ERROR: effective argument associated with INTENT(IN OUT) dummy must be definable
+    call intentinout(in)
+    ! ERROR: effective argument associated with INTENT(IN OUT) dummy must be definable
+    call intentinout(3.14159)
+    ! ERROR: effective argument associated with INTENT(IN OUT) dummy must be definable
+    call intentinout(in + 1.)
```
tskeith: Maybe add a test for a parenthesized name:
```
call intentout(x)  ! OK
call intentout((x))  ! ERROR
```

klausler: Yes.

test/semantics/call07.f90
```diff
@@ -0,0 +1,54 @@
...
+
+  subroutine test
+    ! ERROR: CONTIGUOUS pointer must be an array
+    real, pointer, contiguous :: a01 ! C830
+    real, pointer :: a02
+    real, target :: a03(10)
+    real :: a04(10) ! not TARGET
+    call s01(a03) ! ok
+    ! ERROR: Effective argument associated with CONTIGUOUS POINTER dummy argument must be simply contiguous
+    call s01(a02)
```
tskeith: Wouldn't this just be an error due to rank mismatch?

klausler: Right, I omitted the deferred bounds on the declaration.

PR #713, opened 2019-09-05
[OpenMP] remaining Task related structural checks

ichoyjx:
Please see each commit for description. Basically it's adding checks for directive `Task` [2.9.1] and `Taskloop Simd` [2.9.3] along with the fixes for issues I found.

### DIFF
--- a/lib/parser/openmp-grammar.h
+++ b/lib/parser/openmp-grammar.h
@@ -181,8 +181,8 @@ TYPE_PARSER("ALIGNED" >>
                     parenthesized(scalarIntExpr))) ||
     "DIST_SCHEDULE" >> construct<OmpClause>(construct<OmpClause::DistSchedule>(
                            parenthesized("STATIC ," >> scalarIntExpr))) ||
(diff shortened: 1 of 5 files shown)


lib/parser/openmp-grammar.h
```diff
@@ -181,8 +181,8 @@ TYPE_PARSER("ALIGNED" >>
                     parenthesized(scalarIntExpr))) ||
     "DIST_SCHEDULE" >> construct<OmpClause>(construct<OmpClause::DistSchedule>(
                            parenthesized("STATIC ," >> scalarIntExpr))) ||
-    "FINAL" >> construct<OmpClause>(
-                   construct<OmpClause::Final>(parenthesized(scalarIntExpr))) ||
```
vjayathirtha-nv: Is this causing a bug? The spec says `scalar-expression`

ichoyjx: You must be looking at the C/C++ syntax...

vjayathirtha-nv: Ya you're right.

PR #715, opened 2019-09-06
Implement a mapping from source file Provenances to cooked stream offsets

klausler:
Implements `CookedSource::GetCharBlock(ProvenanceRange)`, which queries an inverted map that's optionally built when requested by setting `parser::Options::needProvenanceRangeToCharBlockMappings = true`.

Not well tested, but manual inspection of dump output looks good.

This would be useful in a `flangd` server that served multiple requests from one parse.  Building the reverse map is O(N) but then the lookups are O(log(N)).

This code review is requested from Tim because Tin is not in the drop-down menu of reviewers.


lib/parser/provenance.cc
```diff
@@ -328,18 +401,43 @@ std::optional<ProvenanceRange> CookedSource::GetProvenanceRange(
   return {ProvenanceRange{first.start(), last.start() - first.start()}};
 }
 
+std::optional<CharBlock> CookedSource::GetCharBlock(
```
tskeith: Is this ever used?

lib/parser/provenance.cc
```diff
@@ -328,18 +401,43 @@ std::optional<ProvenanceRange> CookedSource::GetProvenanceRange(
...
+    ProvenanceRange range) const {
+  if (auto to{invertedMap_.Map(range)}) {
```
tskeith: Should this assert that `invertedMap_` has been built?

klausler: Maybe.  Or maybe it should be non-`const` and build the inverted map on the first reference.

lib/parser/provenance.cc
```diff
@@ -14,10 +14,51 @@
...
+  }
+  if (fromTo.second != map_.end()) {
+    map_.emplace_hint(fromTo.second, range, offset);
+  } else {
+    map_.emplace(range, offset);
+  }
+}
+
+std::optional<std::size_t> ProvenanceRangeToOffsetMappings::Map(
+    ProvenanceRange range) const {
```
tinlun: Since the function only maps the start of the `ProvenanceRange` to offset, maybe the input should only be `Provenance`? This would make it easier if the caller only cares about finding the offset of the start of the `ProvenanceRange`. And it avoids the situation where the caller accidentally uses a larger than necessary `ProvenanceRange`.

klausler: The `multimap` was configured with a comparator that will cause the call to `equal_range(range)` to return a range of iterators that might overlap the entirety of `range`, not just its first position.  This was intentional.  In the event that a client wants to map only a single byte, a singleton range can be used.

lib/parser/provenance.h
```diff
@@ -83,21 +86,47 @@ class Provenance {
...
+
+private:
+  // A comparison function object for use in std::multimap<Compare=>.
+  // Intersecting intervals will effectively compare equal, not being
+  // either < nor >= each other.
+  struct WhollyPrecedes {
+    bool operator()(ProvenanceRange, ProvenanceRange) const;
+  };
+
+  std::multimap<ProvenanceRange, std::size_t, WhollyPrecedes> map_;
```
tinlun: Does this have to be multimap?

klausler: Perhaps not now, after I've given up on mapping macro definitions.  But it probably doesn't matter either way, all probes use `equal_range()`.

lib/parser/provenance.cc
tinlun: My understanding is that if the caller calls this function with a range that doesn't fit any `ProvenanceRange` inside the map, even if the start of the input range is inside of the ProvenanceRange.
Is there a situation where the caller would prefer to call this function with a full `ProvenanceRange` over calling it with a singleton range? Calling this function with the full range has the pitfall of giving it too large a range and whiffing. I think we should ask for a single Provenance and convert it to a ProvenanceRange for the caller.

klausler: Passing a range of multiple byte positions assures the caller that any returned offset points to a corresponding contiguous range in the cooked character stream -- `FOO` in the editor gets mapped to a `foo` in the CCS.  Passing a single position (`F`) would return a result that just points to the `f`, and that's not what one would be probing the symbol table with later.

(What this code can't do is map fixed-form `F O O` to `foo`, and maybe it should.  Same with identifiers broken across line continuations.)

tinlun: Ah, I understand what you're saying. The caller would be assured that the range passed in corresponds to the range they expect if it hits, even if the function only returns the start of it. Thank you for the explanations.

klausler:
ping

lib/parser/provenance.h
```diff
@@ -22,6 +22,7 @@
 #include "../common/idioms.h"
 #include "../common/interval.h"
 #include <cstddef>
+#include <functional>
```
tskeith: Is this needed?

tinlun:
The change looks very good, thank you! I've also looked at dump output with my test cases for flangd and manually verified that they are correct.

lib/parser/provenance.h
klausler: Thanks, no, not any longer.

PR #718, opened 2019-09-08
[OpenMP] formalize `DEFAULTMAP` clause

ichoyjx:
OpenMP 4.5 only accepts `defaultmap(tofrom:scalar)`. The original implementation
only parses the entire `tofrom:scalar` string and does nothing else. This commit
makes it treat `tofrom` (`ImplicitBehavior`) and `scalar` (`VariableCategory`)
separately, which is clear and extendable for OpenMP 5.0 Spec.


lib/parser/openmp-grammar.h
```diff
@@ -97,18 +104,19 @@ TYPE_PARSER(construct<OmpScheduleClause>(maybe(Parser<OmpScheduleModifier>{}),
...
+                pure(OmpIfClause::DirectiveNameModifier::TargetEnterData) ||
+            "TARGET EXIT DATA" >>
+                pure(OmpIfClause::DirectiveNameModifier::TargetExitData) ||
+            "TARGET DATA" >>
+                pure(OmpIfClause::DirectiveNameModifier::TargetData) ||
+            "TARGET UPDATE" >>
+                pure(OmpIfClause::DirectiveNameModifier::TargetUpdate) ||
+            "TARGET" >> pure(OmpIfClause::DirectiveNameModifier::Target) ||
+            "TASK"_id >> pure(OmpIfClause::DirectiveNameModifier::Task) ||
+            "TASKLOOP" >> pure(OmpIfClause::DirectiveNameModifier::Taskloop)) /
```
DavidTruby: This seems to be an unrelated formatting change? Although the new formatting is what I get if I run clang-format on master too so it makes sense to keep it.

ichoyjx: Yes, it's unrelated formatting change by clang-format.

PR #719, opened 2019-09-09
Semantic checks for target construct

DavidTruby:
This patch implements basic semantic checks for the TARGET construct.
The requirement that only a single device or nowait clause can appear is technically from OpenMP 5.0 not 4.5, however it is just an added clarification in 5.0 because there are no reasonable semantics for multiple clauses of these types. As such I've added that requirement here.

### DIFF
--- a/lib/semantics/check-omp-structure.cc
+++ b/lib/semantics/check-omp-structure.cc
@@ -385,6 +385,24 @@ void OmpStructureChecker::Enter(const parser::OpenMPBlockConstruct &x) {
         OmpClause::IF, OmpClause::FINAL, OmpClause::PRIORITY};
     SetContextAllowedOnce(allowedOnce);
   } break;
+  // 2.10.4 target-clause -> if-clause |
+  //                         device-clause |
+  //                         private-clause |
+  //                         firstprivate-clause |
+  //                         map-clause |
+  //                         is-device-ptr-clause |
+  //                         defaultmap-clause |
+  //                         nowait-clause |
+  //                         depend-clause
+  case parser::OmpBlockDirective::Directive::Target: {
+    PushContext(beginDir.source, OmpDirective::TARGET);
+    OmpClauseSet allowed{OmpClause::IF, OmpClause::PRIVATE, OmpClause::MAP,
+        OmpClause::DEPEND, OmpClause::FIRSTPRIVATE, OmpClause::IS_DEVICE_PTR};
+    SetContextAllowed(allowed);
+    OmpClauseSet allowedOnce{
+        OmpClause::DEVICE, OmpClause::DEFAULTMAP, OmpClause::NOWAIT};
+    SetContextAllowedOnce(allowedOnce);
+  } break;
   default:
     // TODO others
     break;
--- a/lib/semantics/check-omp-structure.cc
+++ b/lib/semantics/check-omp-structure.cc
@@ -396,8 +396,9 @@ void OmpStructureChecker::Enter(const parser::OpenMPBlockConstruct &x) {
   //                         depend-clause
   case parser::OmpBlockDirective::Directive::Target: {
     PushContext(beginDir.source, OmpDirective::TARGET);
-    OmpClauseSet allowed{OmpClause::IF, OmpClause::PRIVATE, OmpClause::MAP,
-        OmpClause::DEPEND, OmpClause::FIRSTPRIVATE, OmpClause::IS_DEVICE_PTR};
+    OmpClauseSet allowed{OmpClause::IF, OmpClause::PRIVATE,
+        OmpClause::FIRSTPRIVATE, OmpClause::MAP, OmpClause::IS_DEVICE_PTR,
+        OmpClause::DEPEND};
     SetContextAllowed(allowed);
     OmpClauseSet allowedOnce{
         OmpClause::DEVICE, OmpClause::DEFAULTMAP, OmpClause::NOWAIT};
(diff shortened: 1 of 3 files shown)


ichoyjx:
> A map-type in a map clause must be to, from, tofrom or alloc.

Should be checked. In `parse-tree.h`, map type is defined as `ENUM_CLASS(Type, To, From, Tofrom, Alloc, Release, Delete)`.

Also, I've updated the `defaultmap` change into master branch, you also need to check that `scalar` is present (`VariableCategory`) on the clause.

DavidTruby:
> > A map-type in a map clause must be to, from, tofrom or alloc.
> 
> Should be checked. In `parse-tree.h`, map type is defined as `ENUM_CLASS(Type, To, From, Tofrom, Alloc, Release, Delete)`.


I was going to do this separately as the map clause is a separate section in the spec that applies to multiple constructs, and as such I wanted it to be a separate PR for consideration. Can just add that to this one though if that's easier

> Also, I've updated the defaultmap change into master branch, you also need to check that scalar is > present (VariableCategory) on the clause.

I'll add this check in

ichoyjx:
> > > A map-type in a map clause must be to, from, tofrom or alloc.
> > 
> > 
> > Should be checked. In `parse-tree.h`, map type is defined as `ENUM_CLASS(Type, To, From, Tofrom, Alloc, Release, Delete)`.
> 
> I was going to do this separately as the map clause is a separate section in the spec that applies to multiple constructs, and as such I wanted it to be a separate PR for consideration. Can just add that to this one though if that's easier

Okay with me if you want to do `map` in a separate PR for multiple directives.

lib/semantics/check-omp-structure.cc
```diff
@@ -385,6 +385,24 @@ void OmpStructureChecker::Enter(const parser::OpenMPBlockConstruct &x) {
...
+  //                         firstprivate-clause |
+  //                         map-clause |
+  //                         is-device-ptr-clause |
+  //                         defaultmap-clause |
+  //                         nowait-clause |
+  //                         depend-clause
+  case parser::OmpBlockDirective::Directive::Target: {
+    PushContext(beginDir.source, OmpDirective::TARGET);
+    OmpClauseSet allowed{OmpClause::IF, OmpClause::PRIVATE, OmpClause::MAP,
+        OmpClause::DEPEND, OmpClause::FIRSTPRIVATE, OmpClause::IS_DEVICE_PTR};
```
kiranchandramohan: Nit. Can you re-order the allowed set to match the order of the clauses above?

test/semantics/omp-device-constructs.f90
```diff
@@ -0,0 +1,61 @@
...
+  enddo
+  !$omp end target
+
+  !$omp target device(0)
+  do i = 1, N
+     a = 3.14
+  enddo
+  !$omp end target
+
+
```
kiranchandramohan: Nit. Extra line?

lib/semantics/check-omp-structure.cc
DavidTruby: Done, they are now the same order as the grammar

lib/semantics/check-omp-structure.cc
```diff
@@ -807,8 +826,14 @@ void OmpStructureChecker::Enter(const parser::OmpAlignedClause &x) {
 void OmpStructureChecker::Enter(const parser::OmpDefaultClause &) {
   CheckAllowed(OmpClause::DEFAULT);
 }
-void OmpStructureChecker::Enter(const parser::OmpDefaultmapClause &) {
+void OmpStructureChecker::Enter(const parser::OmpDefaultmapClause &x) {
   CheckAllowed(OmpClause::DEFAULTMAP);
+  using VariableCategory = parser::OmpDefaultmapClause::VariableCategory;
+  if (!std::get<std::optional<VariableCategory>>(x.t)) {
```
ichoyjx: I never saw the usage of `!` with `std::get<std::optional<...>>(...)` in the F18 code base before. I think it's better to write it as:
```
const auto &vc{std::get<std::optional<VariableCategory>>(x.t)};
if (!vc.has_value()) {
...
```
Adding Tim @tskeith.

lib/semantics/check-omp-structure.cc
```diff
@@ -807,8 +826,14 @@ void OmpStructureChecker::Enter(const parser::OmpAlignedClause &x) {
...
+    context_.Say(GetContext().clauseSource,
+        "The scalar VARIABLECATEGORY must be specified on the DEFAULTMAP "
```
ichoyjx: Users may not be able to understand this, how about `The argument TOFROM:SCALAR must be specified on the DEFAULTMAP clause` ?

lib/semantics/check-omp-structure.cc
klausler: Or `if (!std::get<std::optional<VariableCategory>>(x.t).has_value()) {`.

But the original usage is fine by me too.

ichoyjx: Thanks, good to know!

lib/semantics/check-omp-structure.cc
DavidTruby: Yes, that sounds better. I did think as I wrote it that it didn't read that well but I couldn't decide what to change it to

lib/semantics/check-omp-structure.cc
DavidTruby: I'm happy to change this if there's a style we particularly want to stick to, I was just using optional how I've used it previously in other projects without really thinking. Let me know if you want me to change it or if it's fine as-is

PR #732, opened 2019-09-10
Begin semantic checks for calls

klausler:
Defines a static declaration checker for things that don't have to be caught on the fly during name resolution; use it for for basic constraints on assumed-length `CHARACTER(*)` function definitions.  Rearrange some code in `semantics/expression.cc` to lay groundwork for semantic checking of call sites.  Get test `call01.f90` to pass, activate it.

### DIFF
--- a/lib/semantics/semantics.h
+++ b/lib/semantics/semantics.h
@@ -193,5 +193,8 @@ struct BaseChecker {
   template<typename N> void Enter(const N &) {}
   template<typename N> void Leave(const N &) {}
 };
+
+// Static declaration checks
+void CheckDeclarations(SemanticsContext &);
 }
 #endif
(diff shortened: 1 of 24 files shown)


lib/semantics/semantics.h
```diff
@@ -193,5 +193,8 @@ struct BaseChecker {
   template<typename N> void Enter(const N &) {}
   template<typename N> void Leave(const N &) {}
 };
+
+// Static declaration checks
+void CheckDeclarations(SemanticsContext &);
```
tskeith: In my opinion this should be in `check-declarations.h`, even though it means creating an extra file.

klausler: No problem.

PR #737, opened 2019-09-11
Allow module, submodule, and program names to be used as local identi…

klausler:
…fiers.  Check for USE of m in m.

### DIFF
--- a/lib/semantics/resolve-names.cc
+++ b/lib/semantics/resolve-names.cc
@@ -1655,7 +1655,8 @@ void ScopeHandler::PushScope(Scope &scope) {
   }
   // The name of a module or submodule cannot be "used" in its scope,
   // as we read 19.3.1(2), so we allow the name to be used as a local
-  // identifier in the module or submodule too.  Same with programs.
+  // identifier in the module or submodule too.  Same with programs
+  // (14.1(3)).
   if (!currScope_->IsDerivedType() && kind != Scope::Kind::Module &&
       kind != Scope::Kind::MainProgram) {
     if (auto *symbol{scope.symbol()}) {
@@ -2098,7 +2099,7 @@ Scope *ModuleVisitor::FindModule(const parser::Name &name, Scope *ancestor) {
     Say(name, "'%s' is not a module"_err_en_US);
     return nullptr;
   }
-  if (DoesScopeContain(scope, currScope())) {
+  if (DoesScopeContain(scope, currScope())) {  // 14.2.2(1)
     Say(name, "Module '%s' cannot USE itself."_err_en_US);
   }
   Resolve(name, scope->symbol());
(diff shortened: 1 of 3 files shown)


lib/semantics/resolve-names.cc
```diff
@@ -2086,14 +2090,17 @@ void ModuleVisitor::BeginModule(const parser::Name &name, bool isSubmodule) {
...
+  Scope *scope{reader.Read(name.source, ancestor)};
   if (!scope) {
     return nullptr;
   }
   if (scope->kind() != Scope::Kind::Module) {
     Say(name, "'%s' is not a module"_err_en_US);
     return nullptr;
   }
+  if (DoesScopeContain(scope, currScope())) {
+    Say(name, "Module '%s' cannot USE itself."_err_en_US);
```
tskeith: Should return `nullptr` here too?

lib/semantics/resolve-names.cc
```diff
@@ -1653,7 +1653,11 @@ void ScopeHandler::PushScope(Scope &scope) {
   if (kind != Scope::Kind::Block) {
     ImplicitRulesVisitor::BeginScope(scope);
   }
-  if (!currScope_->IsDerivedType()) {
+  // The name of a module or submodule cannot be "used" in its scope,
+  // as we read 19.3.1(2), so we allow the name to be used as a local
```
vdonaldson: Did you mean C1406 here?

lib/semantics/resolve-names.cc
vdonaldson: Add a C1406 comment here?

klausler: I worried that doing so would elicit needless error messages in the following code.

lib/semantics/resolve-names.cc
klausler: No.

lib/semantics/resolve-names.cc
klausler: I don't think that C1406 applies.  I'll reference the relevant text.

klausler:
Tim, any remaining concern?

lib/semantics/resolve-names.cc
vdonaldson: R1409 use-stmt is USE ...

C1406 (R1409) A scoping unit shall not directly reference an intrinsic module and a nonintrinsic module of the same name.

Seems like this is what the message is checking.  I don't get any valid hits on C1406 in f18 code.

tskeith:
> Tim, any remaining concern?

Nope, just forgot to hit Approve.

lib/semantics/resolve-names.cc
klausler: That constraint prevents
```
USE,INTRINSIC::IEEE_ARITHMETIC
USE,NON_INTRINSIC::IEEE_ARITHMETIC`
```

vdonaldson: I see; the conjunction _and_ vs. _or_ makes the difference from what I was reading.  Sorry for the extra cycles.

PR #740, opened 2019-09-12
Fix bugs with use-association renames

tskeith:
Fix some bugs where we weren't keeping the local-name from a rename in a use-association. They showed up in writing module files but the real problem was saving the correct information.


lib/semantics/expression.cc
```diff
@@ -597,17 +597,25 @@ MaybeExpr ExpressionAnalyzer::Analyze(const parser::Name &n) {
...
       // A bare reference to a derived type parameter (within a parameterized
       // derived type definition)
       return AsMaybeExpr(MakeBareTypeParamInquiry(&ultimate));
     } else {
-      return Designate(DataRef{ultimate});
+      const auto *details{ultimate.detailsIf<semantics::ObjectEntityDetails>()};
+      if (details && semantics::IsNamedConstant(ultimate) && details->type() &&
+          details->type()->AsIntrinsic() && !details->IsArray() &&
+          details->init()) {
+        return details->init();
```
jeanPerier: Wasn't evaluate `GetNamedConstantValue` already doing this replacement while folding expressions ? I guess not completely if you did it, but I am curious what.
The reason I am asking is that if `GetNamedConstantValue` was not yet called on ultimate, then conversion of the init expression to the parameter type may have not occurred yet. 
From a few tests, it looks OK, but it would probably be safer to add `details->initWasValidated()` in the check above.

tskeith: I did it this way because I found that if I just returned a DataRef to the original symbol (i.e. `k2` in `use m, k2 => k1`) the constant value was not propagated. But folding is a better place to take care of that, by changing `GetNamedConstantValue()` to follow the use-association.

@jeanPerier, @klausler, please take another look.

jeanPerier: Thanks for explaining and fixing `GetNamedConstantValue`, I had not even noticed it did not resolve use association...

PR #741, opened 2019-09-13
Fix bug merging generics on USE

### LINKED ISSUE #733: Bogus error on generic interface (specific proc. conflicts with itself)
```
module m1
  type :: t1
  end type
  interface generic
    module procedure s1
  end interface
 contains
  subroutine s1(x)
    type(t1) :: x
  end subroutine
end module

module m2
  use m1
  type, extends(t1) :: t2
  end type
  interface generic
    module procedure s2
  end interface
 contains
  subroutine s2(x)
    type(t2) :: x
  end subroutine
end module

use m1
use m2
end
```

produces (with -fdebug-semantics):

```
/tmp/bug3.f90:4:13: error: Generic 'generic' may not have specific procedures 's1' and 's1' as their interfaces are not distinguishable
    interface generic
              ^^^^^^^
/tmp/bug3.f90:8:14: Declaration of 's1'
    subroutine s1(x)
               ^^
/tmp/bug3.f90:8:14: Declaration of 's1'
    subroutine s1(x)
               ^^
```

tskeith:
When we use-associate the same generic name from two different modules
they are merged together. If the same specific procedure occurs in both
generics it should only occur once in the merged one.

Similarly, it's not an error if they both have a derived type or
specific procedure named the same as the generic, as long as they are
the same symbol.

Fixes #733.

### DIFF
--- a/lib/semantics/symbol.cc
+++ b/lib/semantics/symbol.cc
@@ -187,8 +187,8 @@ void GenericDetails::CopyFrom(const GenericDetails &from) {
     derivedType_ = from.derivedType_;
   }
   for (const Symbol *symbol : from.specificProcs_) {
-    auto it{std::find(specificProcs_.begin(), specificProcs_.end(), symbol)};
-    if (it == specificProcs_.end()) {
+    if (std::find(specificProcs_.begin(), specificProcs_.end(), symbol) ==
+        specificProcs_.end()) {
       specificProcs_.push_back(symbol);
     }
   }
--- a/lib/semantics/symbol.cc
+++ b/lib/semantics/symbol.cc
@@ -179,15 +179,19 @@ Symbol *GenericDetails::CheckSpecific() {
 
 void GenericDetails::CopyFrom(const GenericDetails &from) {
   if (from.specific_) {
-    CHECK(!specific_);
+    CHECK(!specific_ || specific_ == from.specific_);
     specific_ = from.specific_;
   }
   if (from.derivedType_) {
-    CHECK(!derivedType_);
+    CHECK(!derivedType_ || derivedType_ == from.derivedType_);
     derivedType_ = from.derivedType_;
   }
-  auto &procs{from.specificProcs_};
-  specificProcs_.insert(specificProcs_.end(), procs.begin(), procs.end());
+  for (const Symbol *symbol : from.specificProcs_) {
+    auto it{std::find(specificProcs_.begin(), specificProcs_.end(), symbol)};
+    if (it == specificProcs_.end()) {
+      specificProcs_.push_back(symbol);
+    }
+  }
 }
 
 // The name of the kind of details for this symbol.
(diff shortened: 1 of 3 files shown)


lib/semantics/symbol.cc
```diff
@@ -179,15 +179,19 @@ Symbol *GenericDetails::CheckSpecific() {
 
 void GenericDetails::CopyFrom(const GenericDetails &from) {
   if (from.specific_) {
-    CHECK(!specific_);
+    CHECK(!specific_ || specific_ == from.specific_);
```
jeanPerier: It looks like a bit brutal to me to die if `specific_ != from.specific_`. Are we sure this case was handled before ?
The following code, where `g` appears as two different specific proc, crashes f18 for me:
```
module m9a
  interface g
    module procedure s1
    module procedure g
  end interface
contains
  subroutine g()
  end
  subroutine s1(x)
    integer :: x
  end
end module
module m9b
  interface g
    module procedure s2
    module procedure g
  end interface
contains
  subroutine s2(x)
    real :: x
  end
  subroutine g(i)
    integer :: i
  end
end module
! Merge use-associated generics that have the same symbol (s1)
subroutine s9
  use m9a
  use m9b
end
```

lib/semantics/symbol.cc
```diff
@@ -179,15 +179,19 @@ Symbol *GenericDetails::CheckSpecific() {
...
     specific_ = from.specific_;
   }
   if (from.derivedType_) {
-    CHECK(!derivedType_);
+    CHECK(!derivedType_ || derivedType_ == from.derivedType_);
```
jeanPerier: Comment above probably also applies here.

lib/semantics/symbol.cc
tskeith: Yes, you're right. I'll change it to check for these cases before calling CopyFrom.

lib/semantics/symbol.cc
```diff
@@ -179,15 +179,19 @@ Symbol *GenericDetails::CheckSpecific() {
...
     derivedType_ = from.derivedType_;
   }
-  auto &procs{from.specificProcs_};
-  specificProcs_.insert(specificProcs_.end(), procs.begin(), procs.end());
+  for (const Symbol *symbol : from.specificProcs_) {
+    auto it{std::find(specificProcs_.begin(), specificProcs_.end(), symbol)};
```
klausler: I'd bash these two lines together and avoid the needless temporary.  Calling `find` and comparing its result to `end()` is idiomatic.

lib/semantics/symbol.cc
tskeith: Now we report an error in this case.

lib/semantics/symbol.cc
tskeith: OK

PR #742, opened 2019-09-13
Improve flag dumping for -funparse-with-symbols

ichoyjx:
Addressing one of Tim's comments in #739, which is to dump the flags we care about along with the `Unparse` process. This commit gets rid of the hard-coded dump for four flags and replaces them with a separate dump function.

### DIFF
--- a/lib/semantics/symbol.cc
+++ b/lib/semantics/symbol.cc
@@ -17,6 +17,7 @@
 #include "../common/idioms.h"
 #include <ostream>
 #include <string>
+#include <unordered_map>
 
 namespace Fortran::semantics {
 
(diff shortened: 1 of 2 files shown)


lib/semantics/symbol.cc
```diff
@@ -530,6 +531,33 @@ static void DumpUniqueName(std::ostream &os, const Scope &scope) {
   }
 }
 
+// Dump flag list for a symbol in Unparse
+static void DumpFlagsForUnparse(std::ostream &os, const Symbol::Flags &flags) {
+  static std::unordered_map<Symbol::Flag, const std::string> flagMap{
```
jeanPerier: You can probably make it `static const`.

lib/semantics/symbol.cc
```diff
@@ -530,6 +531,33 @@ static void DumpUniqueName(std::ostream &os, const Scope &scope) {
...
+      {Symbol::Flag::Implicit, "implicit"},
+      {Symbol::Flag::LocalityLocal, "local"},
+      {Symbol::Flag::LocalityLocalInit, "local_init"},
+      {Symbol::Flag::LocalityShared, "shared"}};
+  std::size_t seen{0};
+
+  flags.IterateOverMembers([&](Symbol::Flag flag) {
+    if (flags.test(flag)) {
```
jeanPerier: Isn't `flags.IterateOverMembers` only iterating over the flags that are set in `flags`? This would make this test redundant.

lib/semantics/symbol.cc
```diff
@@ -530,6 +531,33 @@ static void DumpUniqueName(std::ostream &os, const Scope &scope) {
...
+// Dump flag list for a symbol in Unparse
+static void DumpFlagsForUnparse(std::ostream &os, const Symbol::Flags &flags) {
+  static std::unordered_map<Symbol::Flag, const std::string> flagMap{
+      {Symbol::Flag::Implicit, "implicit"},
+      {Symbol::Flag::LocalityLocal, "local"},
+      {Symbol::Flag::LocalityLocalInit, "local_init"},
+      {Symbol::Flag::LocalityShared, "shared"}};
+  std::size_t seen{0};
+
+  flags.IterateOverMembers([&](Symbol::Flag flag) {
```
jeanPerier: I think I would reverse the logic here and iterate through `flagMap` instead of `flags`, `IterateOverMembers` looks like it is doing a copy of the set to perform iterations, plus the lambda looks more complex to me than:
```
for (const auto& pair: flagMap) {
  if (flags.test(pair.first)) {
       if (seen++ > 0) {
          os << ", ";
        } else {
          os << " (";
        }
        os << pair.second;
  }
}
```

If you go for this you should probably also make `flagMap` a `std::map` instead of an `std::unordered_map` to guarantee the output order.

ichoyjx: Thanks, Jean, I will change the logic here.

PR #743, opened 2019-09-13
Prevent RESULT name from being the same as the function name

### LINKED ISSUE #726: Regression: internal error when `RESULT()` has same name as function
```
integer function f() result(f)
  f = 0
end
```

elicits `error: Internal: no symbol found for 'f'` with the caret pointing at the `f` in `result(f)`.

jeanPerier:
Fix #726.
Check and emit error message for such scenario. Also set error on the result symbol because it is unclear what it is inside the function.

### DIFF
--- a/lib/semantics/resolve-names.cc
+++ b/lib/semantics/resolve-names.cc
@@ -2473,6 +2473,7 @@ void SubprogramVisitor::Post(const parser::FunctionStmt &stmt) {
   }
   const parser::Name *funcResultName;
   if (funcInfo_.resultName && funcInfo_.resultName->source != name.source) {
+    // Note that RESULT is ignored if it has the same name as the function.
     funcResultName = funcInfo_.resultName;
   } else {
     EraseSymbol(name);  // was added by PushSubprogramScope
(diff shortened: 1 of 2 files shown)


lib/semantics/resolve-names.cc
```diff
@@ -2483,6 +2483,13 @@ void SubprogramVisitor::Post(const parser::FunctionStmt &stmt) {
   funcResultDetails.set_funcResult(true);
   funcInfo_.resultSymbol =
       &MakeSymbol(*funcResultName, std::move(funcResultDetails));
+  if (funcInfo_.resultName && funcInfo_.resultName->source == name.source) {
+    // C1560. TODO also enforce on entry names when entry implemented
+    Say(funcInfo_.resultName->source,
+        "'%s' is already the function name and cannot appear in RESULT"_err_en_US,
```
klausler: I would emit a warning, not an error, and ignore the `RESULT` name.

jeanPerier: OK, pgfortran is accepting this (none of the other tested compilers do) and I do not see issues that could come from relaxing the constraint.
I have transformed the error into a warning that states that references to `f` inside the function scope will be considered references to the result only.

PR #745, opened 2019-09-13
Implement checks for test/semantics/call02.f90

klausler:
More semantic checking for calls.  This PR also implements enough analysis of coindexed references to catch an error condition in the test.

### DIFF
--- a/lib/semantics/expression.cc
+++ b/lib/semantics/expression.cc
@@ -1501,7 +1501,7 @@ auto ExpressionAnalyzer::Procedure(const parser::ProcedureDesignator &pd,
                 if (symbol.attrs().test(
                         semantics::Attr::NON_RECURSIVE)) {  // 15.6.2.1(3)
                   if (auto *msg{Say(
-                          "NON_RECURSIVE procedure '%s' cannot call itself."_err_en_US,
+                          "NON_RECURSIVE procedure '%s' cannot call itself"_err_en_US,
                           n.source)}) {
                     msg->Attach(
                         symbol.name(), "definition of '%s'"_en_US, n.source);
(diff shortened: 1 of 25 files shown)


lib/semantics/expression.cc
```diff
@@ -1515,18 +1564,31 @@ std::optional<ActualArgument> ExpressionAnalyzer::AnalyzeActualArgument(
...
-  if (const Symbol * assumedTypeDummy{AssumedTypeDummy(var)}) {
-    return ActualArgument{ActualArgument::AssumedType{*assumedTypeDummy}};
-  } else if (MaybeExpr argExpr{Analyze(var)}) {
-    return ActualArgument{std::move(*argExpr)};
+    Expr<SomeType> x{Fold(GetFoldingContext(), std::move(*argExpr))};
+    if (const auto *proc{std::get_if<ProcedureDesignator>(&x.u)}) {
+      if (!std::holds_alternative<SpecificIntrinsic>(proc->u) &&
+          proc->IsElemental()) {  // C1533
+        Say(expr.source,
+            "Non-intrinsic ELEMENTAL procedure cannot be passed as argument."_err_en_US);
```
tskeith: We haven't been putting periods at the end of error messages.

lib/semantics/expression.cc
```diff
@@ -1515,18 +1564,31 @@ std::optional<ActualArgument> ExpressionAnalyzer::AnalyzeActualArgument(
...
+      }
+    }
+    if (auto coarrayRef{ExtractCoarrayRef(x)}) {
+      const Symbol &coarray{coarrayRef->GetLastSymbol()};
+      if (const semantics::DeclTypeSpec * type{coarray.GetType()}) {
+        if (const semantics::DerivedTypeSpec * derived{type->AsDerived()}) {
+          if (auto ptr{semantics::FindPointerUltimateComponent(*derived)}) {
+            if (auto *msg{Say(expr.source,
+                    "Coindexed object '%s' with POINTER ultimate component '%s' cannot be passed as argument."_err_en_US,
```
tskeith: Same here.

test/semantics/call02.f90
```diff
@@ -63,22 +63,6 @@ elemental real function elem04(x)
...
-    subroutine altreturn(*)
-    end subroutine
-  end interface
- contains
-  subroutine test
-1   continue
-   contains
-    subroutine internal
-      !ERROR: alternate return label must be in the inclusive scope
-      call altreturn(*1)
```
tskeith: Is this check still a TODO? Should the test be moved to a different call??.f90?

klausler: It was redundant; label validation catches the error, but then none of the other semantic checks are applied.

lib/semantics/expression.cc
klausler: Ok.

PR #748, opened 2019-09-16
Added semantic checks for map clauses

DavidTruby:
This patch adds semantic checks for map clauses, as well as the related checks for target data, target enter data and target exit data.

### DIFF
--- a/lib/semantics/check-omp-structure.cc
+++ b/lib/semantics/check-omp-structure.cc
@@ -550,7 +550,11 @@ void OmpStructureChecker::Enter(
     PushContext(dir.source, OmpDirective::TASKYIELD);
   } break;
   case parser::OmpSimpleStandaloneDirective::Directive::TargetEnterData: {
-    // 2.10.2 target-enter-data
+    // 2.10.2 target-enter-data-clause -> if-clause |
+    //                                    device-clause |
+    //                                    map-clause |
+    //                                    depend-clause |
+    //                                    nowait-clause
     PushContext(dir.source, OmpDirective::TARGET_ENTER_DATA);
     OmpClauseSet allowed{OmpClause::MAP, OmpClause::DEPEND, OmpClause::NOWAIT};
     SetContextAllowed(allowed);
@@ -559,7 +563,11 @@ void OmpStructureChecker::Enter(
     SetContextRequired({OmpClause::MAP});
   } break;
   case parser::OmpSimpleStandaloneDirective::Directive::TargetExitData: {
-    // 2.10.3 target-exit-data
+    // 2.10.3  target-enter-data-clause -> if-clause |
+    //                                     device-clause |
+    //                                     map-clause |
+    //                                     depend-clause |
+    //                                     nowait-clause
     PushContext(dir.source, OmpDirective::TARGET_EXIT_DATA);
     OmpClauseSet allowed{OmpClause::MAP, OmpClause::DEPEND, OmpClause::NOWAIT};
     SetContextAllowed(allowed);
--- a/lib/semantics/check-omp-structure.cc
+++ b/lib/semantics/check-omp-structure.cc
@@ -437,9 +437,8 @@ void OmpStructureChecker::Enter(const parser::OpenMPBlockConstruct &x) {
     OmpClauseSet allowed{
         OmpClause::IF, OmpClause::MAP, OmpClause::USE_DEVICE_PTR};
     SetContextAllowed(allowed);
-    OmpClauseSet allowedOnce{OmpClause::DEVICE};
-    SetContextAllowedOnce(allowedOnce);
-    SetContextRequired(OmpClauseSet{OmpClause::MAP});
+    SetContextAllowedOnce({OmpClause::DEVICE});
+    SetContextRequired({OmpClause::MAP});
   } break;
   default:
     // TODO others
@@ -557,7 +556,7 @@ void OmpStructureChecker::Enter(
     SetContextAllowed(allowed);
     OmpClauseSet allowedOnce{OmpClause::DEVICE, OmpClause::IF};
     SetContextAllowedOnce(allowedOnce);
-    SetContextRequired(OmpClauseSet{OmpClause::MAP});
+    SetContextRequired({OmpClause::MAP});
   } break;
   case parser::OmpSimpleStandaloneDirective::Directive::TargetExitData: {
     // 2.10.3 target-exit-data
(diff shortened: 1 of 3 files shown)


lib/semantics/check-omp-structure.cc
```diff
@@ -130,6 +130,15 @@ void OmpStructureChecker::CheckAllowed(OmpClause type) {
   SetContextClauseInfo(type);
 }
 
+void OmpStructureChecker::CheckRequired(OmpClause c) {
+  auto *clause{FindClause(c)};
+  if (!clause) {
```
tskeith: Why not just `if (!FindClause(c))` ?

lib/semantics/check-omp-structure.cc
```diff
@@ -881,8 +909,41 @@ void OmpStructureChecker::Enter(const parser::OmpLinearClause &x) {
     }
   }
 }
-void OmpStructureChecker::Enter(const parser::OmpMapClause &) {
+void OmpStructureChecker::Enter(const parser::OmpMapClause &x) {
   CheckAllowed(OmpClause::MAP);
+  if (const auto &maptype{std::get<std::optional<parser::OmpMapType>>(x.t)}) {
+    using Type = parser::OmpMapType::Type;
+    const Type &type = std::get<Type>(maptype->t);
```
tskeith: Use braced initialization here.

lib/semantics/check-omp-structure.cc
```diff
@@ -881,8 +909,41 @@ void OmpStructureChecker::Enter(const parser::OmpLinearClause &x) {
...
+    switch (GetContext().directive) {
+    case OmpDirective::TARGET:
+    case OmpDirective::TARGET_DATA: {
+      if (type != Type::To && type != Type::From && type != Type::Tofrom &&
+          type != Type::Alloc) {
+        context_.Say(GetContext().clauseSource,
+            "Only the to, from, tofrom or alloc map types are permitted "
+            "for MAP clauses on the %s directive"_err_en_US,
```
tskeith: I think it would be better to capitalize "to", "from", etc. in these messages because they are keywords, like MAP.

lib/semantics/check-omp-structure.cc
```diff
@@ -404,6 +412,15 @@ void OmpStructureChecker::Enter(const parser::OpenMPBlockConstruct &x) {
         OmpClause::DEVICE, OmpClause::DEFAULTMAP, OmpClause::NOWAIT};
     SetContextAllowedOnce(allowedOnce);
   } break;
+  case parser::OmpBlockDirective::Directive::TargetData: {
+    PushContext(beginDir.source, OmpDirective::TARGET_DATA);
+    OmpClauseSet allowed{
+        OmpClause::IF, OmpClause::MAP, OmpClause::USE_DEVICE_PTR};
+    SetContextAllowed(allowed);
+    OmpClauseSet allowedOnce{OmpClause::DEVICE};
+    SetContextAllowedOnce(allowedOnce);
```
ichoyjx: Could be similar to the next line `SetContextAllowedOnce(OmpClauseSet{OmpClause::DEVICE});`.

lib/semantics/check-omp-structure.cc
```diff
@@ -881,8 +908,41 @@ void OmpStructureChecker::Enter(const parser::OmpLinearClause &x) {
...
+  if (const auto &maptype{std::get<std::optional<parser::OmpMapType>>(x.t)}) {
+    using Type = parser::OmpMapType::Type;
+    const Type &type{std::get<Type>(maptype->t)};
+    switch (GetContext().directive) {
+    case OmpDirective::TARGET:
+    case OmpDirective::TARGET_DATA: {
+      if (type != Type::To && type != Type::From && type != Type::Tofrom &&
+          type != Type::Alloc) {
+        context_.Say(GetContext().clauseSource,
+            "Only the TO, FROM, TOFROM or ALLOC map types are permitted "
```
ichoyjx: `TOFROM, or ALLOC`

lib/semantics/check-omp-structure.cc
```diff
@@ -881,8 +908,41 @@ void OmpStructureChecker::Enter(const parser::OmpLinearClause &x) {
...
+        context_.Say(GetContext().clauseSource,
+            "Only the TO or ALLOC map types are permitted "
+            "for MAP clauses on the %s directive"_err_en_US,
+            ContextDirectiveAsFortran());
+      }
+    } break;
+    case OmpDirective::TARGET_EXIT_DATA: {
+      if (type != Type::Delete && type != Type::Release && type != Type::From) {
+        context_.Say(GetContext().clauseSource,
+            "Only the FROM, RELEASE or DELETE map types are permitted "
```
ichoyjx: `RELEASE, or DELETE`

lib/semantics/check-omp-structure.h
```diff
@@ -151,10 +152,13 @@ class OmpStructureChecker : public virtual BaseChecker {
   // collected information for END directive
   void ResetPartialContext(const parser::CharBlock &source) {
     CHECK(!ompContext_.empty());
+    GetContext().requiredClauses.IterateOverMembers(
+        [this](OmpClause c) { CheckRequired(c); });
```
ichoyjx: Not every construct would call `ResetPartialContext`. Besides, it's little weird to do the checks in the `Reset` function. How about in `Leave(const parser::OmpClauseList &)`?

lib/semantics/check-omp-structure.cc
tskeith: There are many places in this file where there is a named temporary like this. It would be good for them to be consistent.

I think you could just write these like:
`SetContextAllowedOnce({OmpClause::DEVICE});`

ichoyjx: Right.

DavidTruby: I think when there's only one we should consistently not create a temporary, and more than one we should as it's clearer. But I don't have strong feelings either way so I'll change it to whatever we think is best

lib/semantics/check-omp-structure.cc
```diff
@@ -516,10 +532,20 @@ void OmpStructureChecker::Enter(
   case parser::OmpSimpleStandaloneDirective::Directive::TargetEnterData: {
     // 2.10.2 target-enter-data
```
ichoyjx: Please help fill the comments for grammar here.

lib/semantics/check-omp-structure.cc
```diff
@@ -516,10 +532,20 @@ void OmpStructureChecker::Enter(
...
     PushContext(dir.source, OmpDirective::TARGET_ENTER_DATA);
+    OmpClauseSet allowed{OmpClause::MAP, OmpClause::DEPEND, OmpClause::NOWAIT};
+    SetContextAllowed(allowed);
+    OmpClauseSet allowedOnce{OmpClause::DEVICE, OmpClause::IF};
+    SetContextAllowedOnce(allowedOnce);
+    SetContextRequired({OmpClause::MAP});
   } break;
   case parser::OmpSimpleStandaloneDirective::Directive::TargetExitData: {
     // 2.10.3 target-exit-data
```
ichoyjx: Same here.

sscalpone:
@DavidTruby please resolve the conflicts & I will merge.

DavidTruby:
@sscalpone there should be no conflicts now

PR #749, opened 2019-09-16
Allow more argument types in some specific intrinsic

### LINKED ISSUE #724: Calls to specific intrinsics no longer convert arguments
```
integer(2) :: j
real :: x
x = float(j)
```

no longer works, and elicits `error: Actual argument for 'i=' has bad type or kind 'Integer(2)'`.  I think it's because the intrinsic table probing code no longer allows for type conversions on calls to specific intrinsic functions.

jeanPerier:
Fix #724.

- Allow all type kinds for arguments in restricted specific intrinsic conversion functions. The result type remains as specified in the standard. No warning, this is ubiquitous and emitting a warning about type conversion when that is what the user request might be weird. Make `dcmplx` and `dreal` restricted intrinsic to avoid issues with this change. These are non-standard intrinsics so it is unspecified whether or not they are restricted, but all standard specific conversion functions are restricted.

- Allow MAX/MIN restricted intrinsic (AMAX0...) to be replaced by the related generic followed by a type conversion of the result to the type of the specific functions. Emit a warning because xlf and ifort are doing so but pgfortran is converting the arguments instead.

- For all the other specific intrinsic with a names that differ from the related generics, only accept the standard argument types because it is less clear what the behavior should be (arg type conversion vs using the generic).

### DIFF
--- a/lib/evaluate/intrinsics.cc
+++ b/lib/evaluate/intrinsics.cc
@@ -1584,6 +1584,21 @@ static bool ApplySpecificChecks(
   return ok;
 }
 
+static DynamicType GetReturnType(const SpecificIntrinsicInterface &interface,
+    const common::IntrinsicTypeDefaultKinds &defaults) {
+  TypeCategory category{TypeCategory::Integer};
+  switch (interface.result.kindCode) {
+  case KindCode::defaultIntegerKind: break;
+  case KindCode::doublePrecision:
+  case KindCode::defaultRealKind: category = TypeCategory::Real; break;
+  default: CRASH_NO_CASE;
+  }
+  int kind{interface.result.kindCode == KindCode::doublePrecision
+          ? defaults.doublePrecisionKind()
+          : defaults.GetDefaultKind(category)};
+  return DynamicType{category, kind};
+}
+
 // Probe the configured intrinsic procedure pattern tables in search of a
 // match for a given procedure reference.
 std::optional<SpecificCall> IntrinsicProcTable::Implementation::Probe(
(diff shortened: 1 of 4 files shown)


lib/evaluate/intrinsics.cc
```diff
@@ -714,10 +714,24 @@ static const IntrinsicInterface genericIntrinsicFunction[]{
 // Tables 16.2 and 16.3 in Fortran 2018.  The "unrestricted" functions
 // in Table 16.2 can be used as actual arguments, PROCEDURE() interfaces,
 // and procedure pointer targets.
+// Note that the restricted conversion functions dcmplx, dreal, float, idint,
+// ifix and sngl are extended to accept any argument kind because this a common
```
klausler: And it's safe as far as we can tell, and useful.

psteinfeld: This line has more than 80 characters.

lib/evaluate/intrinsics.cc
```diff
@@ -714,10 +714,24 @@ static const IntrinsicInterface genericIntrinsicFunction[]{
...
+// Fortran compilers behavior.
 struct SpecificIntrinsicInterface : public IntrinsicInterface {
   const char *generic{nullptr};
   bool isRestrictedSpecific{false};
-  bool forceResultType{false};
+  // Exact actual/dummy type matching is required by default for specific
+  // intrinsics. If useGenericAndForceResultType is set, then the probing will
```
psteinfeld: This line and several other lines in this comment have more than 80 characters.

lib/evaluate/intrinsics.cc
psteinfeld: "this is a common Fortran behavior" rather than "this a common Fortran behavior" (missing the "is").

lib/evaluate/intrinsics.cc
```diff
@@ -1651,6 +1653,51 @@ std::optional<SpecificCall> IntrinsicProcTable::Implementation::Probe(
       } else {
         specificBuffer.clear();
       }
+      // If there was no exact match with a specific, try to match the related
```
psteinfeld: This line length exceeds 80 characters.

psteinfeld: This ```Probe()``` function is very long, which reduces its readability.

lib/evaluate/intrinsics.cc
jeanPerier: As far as I can see it has 79 col in my editors (vim & notepad++). Plus clang-format is not breaking these line, and it should if they were over 80. Are you saying this because Github viewer is breaking the line on you screen, or is it more than 80 in you editor ?
(For the other lines you are pointing out to I also measure <= 80 col)

jeanPerier: Thanks

psteinfeld: My bad.  I was fooled by Windows 10 automatically resizing my terminal window.  Please ignore these line length comments.  I'll be more careful in the future.

lib/evaluate/intrinsics.cc
jeanPerier: I agree it is not the most readable function after my change. I have pushed a new commit that does not make it much shorter but makes all the logic and paths much more easy to follow (IMHO). It mainly hides the message buffer problem a bit.
@psteinfeld, is it more readable ?
@klausler,  I have removed the usage of a local buffer for the `NULL` handling, it seemed not useful to me since it was always copied into the actual context anyway. Do you see a reason that would justify keeping it ?

psteinfeld: Thanks, Jean.  This is a gigantic improvement.

PR #750, opened 2019-09-17
Semantic checks for OpenMP teams directive

DavidTruby:
Simple semantic checks for omp teams

### DIFF
--- a/lib/semantics/check-omp-structure.cc
+++ b/lib/semantics/check-omp-structure.cc
@@ -404,6 +404,13 @@ void OmpStructureChecker::Enter(const parser::OpenMPBlockConstruct &x) {
         OmpClause::DEVICE, OmpClause::DEFAULTMAP, OmpClause::NOWAIT};
     SetContextAllowedOnce(allowedOnce);
   } break;
+  // 2.10.7 teams-clause -> num-teams-clause |
+  //                        thread-limit-clause |
+  //                        default-clause |
+  //                        private-clause |
+  //                        firstprivate-clause |
+  //                        shared-clause |
+  //                        reduction-clause
   case parser::OmpBlockDirective::Directive::Teams: {
     PushContext(beginDir.source, OmpDirective::TEAMS);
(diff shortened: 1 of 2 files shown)


lib/semantics/check-omp-structure.cc
```diff
@@ -404,6 +404,15 @@ void OmpStructureChecker::Enter(const parser::OpenMPBlockConstruct &x) {
         OmpClause::DEVICE, OmpClause::DEFAULTMAP, OmpClause::NOWAIT};
     SetContextAllowedOnce(allowedOnce);
   } break;
+  case parser::OmpBlockDirective::Directive::Teams: {
```
kiranchandramohan: For other directives, the clauses are listed in comments before the case statement. Could you do the same here?

DavidTruby: Good point. I even made a mental note to do this when I went off to write the tests to check it, then never came back and did it..

ichoyjx:
Do you think whether the following restriction can be done in this PR?
> If specified, a teams construct must be contained within a target construct. That target construct must contain no statements, declarations or directives outside of the teams construct.

DavidTruby:
I don't think this is verifiable in the general case? As in, I believe it's permissible to have a target construct containing a function call and a teams construct inside that function. In which case the teams construct can't actually check that it appears inside a target region.

Possibly we could add some metadata to a function signature in the module file that signifies that a function containing a teams directive can only be called in a target region?

DavidTruby:
I've just realised that I misread the restriction, and the target construct cannot contain other statements (and therefore can't contain a function call as in the example I used before). So yes, this check should be doable, I'll add it

kiranchandramohan:
There is a card for nesting of regions (2.17) in the projects page. It gave the impression that this is a separate block of work.

DavidTruby:
I agree that this should go in a separate PR once all the directives are done, it requires a little more machinery that will be common across any nesting checks so to me it makes sense to do that all at once

PR #751, opened 2019-09-17
Update C++style.md

sscalpone:
Add guidelines for error messages and commas.  Add a version number for clang-format. Add spaces before ### sections to restart numbering.

### DIFF
--- a/documentation/C++style.md
+++ b/documentation/C++style.md
@@ -20,11 +20,15 @@ is clear on usage, follow it.
 
 ## In particular:
 
-Use serial commas in comments, error messages, and documentation.
+Use serial commas in comments, error messages, and documentation
+unless they introduce ambiguity.
 
 ### Error messages
+1. Messages should be a single sentence with few exceptions.
 1. Fortran keywords should appear in upper case.
 1. Names from the program appear in single quotes.
+1. Messages should start with a capital letter.
+1. Messages should not end with a period.
 
 ### Files
 1. File names should use dashes, not underscores.  C++ sources have the
--- a/documentation/C++style.md
+++ b/documentation/C++style.md
@@ -3,9 +3,11 @@ Copyright (c) 2018-2019, NVIDIA CORPORATION.  All rights reserved.
 -->
 
 ## In brief:
-* Use *clang-format* on all C++ source and header files before
-  every merge to master.  All code layout should be determined
-  by means of clang-format.
+* Use *clang-format*
+from llvm 7
+on all C++ source and header files before
+every merge to master.  All code layout should be determined
+by means of clang-format.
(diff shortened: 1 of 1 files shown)


documentation/C++style.md
```diff
@@ -15,7 +17,15 @@ is clear on usage, follow it.
 * Reasonable exceptions to these guidelines can be made.
 * Be aware of some workarounds for known issues in older C++ compilers that should
   still be able to compile f18. They are listed at the end of this document.
+
 ## In particular:
+
+Use serial commas in comments, error messages, and documentation.
```
klausler: ... unless they introduce ambiguity?

I'm not sure that we want to say much about English usage in a C++ style guide, other than to encourage correct spelling and clear syntax.

documentation/C++style.md
```diff
@@ -15,7 +17,15 @@ is clear on usage, follow it.
...
+
+### Error messages
+1. Fortran keywords should appear in upper case.
+1. Names from the program appear in single quotes.
```
klausler: Omit the period if a message comprises a single sentence (complete or not).

tskeith: Messages should be a single sentence with very few exceptions. The only multi-sentence message I could find was:
`TiesAwayFromZero rounding mode is not available not available when folding constants with host runtime. Using TiesToEven instead.`
Even in that case I think a colon would be better.

And I see there's a typo with "not available" duplicated.

documentation/C++style.md
psteinfeld: I'd guess that people in general will not get the reference to "serial commas" and understand what it means.  How about "In a comma separated list, place a comma after the next-to-last item and the 'and' or 'or' that precedes the last item."

sscalpone: That's why I included an example.

documentation/C++style.md
jeanPerier: My mistake, I will fix this.
Question regarding this message and and one that I am adding in #749. Would it be better to put the second part, that states how f18 handles the issue, as a note attached to the main message ?

Similar message in #749 is:
```
"Argument type does not match specific intrinsic '%s' requirements,"
" using '%s' generic instead and converting the result to %s if needed"_en_US,
```

tskeith: I think it's good having it all in one message. I suggest:
`TiesAwayFromZero rounding mode is not available when folding constants with host runtime; using TiesToEven instead`

The other one is probably better with a semi-colon instead of the comma as well.

jeanPerier: Thanks for the feedback. Both messages are fixed in commit https://github.com/flang-compiler/f18/pull/749/commits/621ded353f6179045e8fc4c18ef13a4aa05293b9 in PR #749.

PR #753, opened 2019-09-18
[OpenMP] Simplify `OmpObject` with `std::variant`

ichoyjx:
Instead of using an `ENUM_CLASS Kind` to distinguish a `Designator`
or common-block name, change it to `std::variant`. If the `Name`
is available, then it is common-block name.

This arose from a comment in PR#739.

### DIFF
--- a/lib/parser/unparse.cc
+++ b/lib/parser/unparse.cc
@@ -1757,9 +1757,12 @@ public:
   }
   // OpenMP Clauses & Directives
   void Unparse(const OmpObject &x) {
-    bool isCommon{std::get<OmpObject::Kind>(x.t) == OmpObject::Kind::Common};
-    const char *slash{isCommon ? "/" : ""};
-    Put(slash), Walk(std::get<Designator>(x.t)), Put(slash);
+    std::visit(
+        common::visitors{
+            [&](const Designator &y) { Walk(y); },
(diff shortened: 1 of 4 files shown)


lib/parser/unparse.cc
```diff
@@ -1757,9 +1757,11 @@ class UnparseVisitor {
...
   // OpenMP Clauses & Directives
   void Unparse(const OmpObject &x) {
-    bool isCommon{std::get<OmpObject::Kind>(x.t) == OmpObject::Kind::Common};
-    const char *slash{isCommon ? "/" : ""};
-    Put(slash), Walk(std::get<Designator>(x.t)), Put(slash);
+    if (const auto *designator{std::get_if<Designator>(&x.u)}) {
+      Walk(*designator);
+    } else if (const auto *name{std::get_if<Name>(&x.u)}) {
+      Put("/"), Walk(*name), Put("/");
+    }
```
tskeith: `std::visit` is better here. This makes it looks like there could be other alternatives that are begin ignored. And if a third alternative were ever added, the `std::visit` form would get a compilation error that reminds you to add something here.

ichoyjx: Okay I'll change that. Thanks

ichoyjx: Updated.

PR #755, opened 2019-09-20
Check specification expression constraints (10.1.11)

klausler:
And use a new expression traversal utility to do so (lib/evaluate/traverse.h).

Clients of the previous expression traversal utility have been updated to the new one, and the old one was deleted.

Some specialized constraints on functions that can or cannot be called in specification expressions are not (yet) implemented; we don't yet have any way of distinguishing intrinsic modules like IEEE_ARITHMETIC and making special allowances for their inquiry functions.


lib/evaluate/tools.h
```diff
@@ -678,32 +683,36 @@ struct TypeKindVisitor {
...
-      Return(nullptr);
+  template<typename T> Result operator()(const Expr<T> &x) const {
+    if constexpr (common::HasMember<T, AllIntrinsicTypes> ||
+        std::is_same_v<T, SomeDerived>) {
+      if (const auto *designator{std::get_if<Designator<T>>(&x.u)}) {
+        if (auto known{(*this)(*designator)}) {
+          return known;
+        }
+      }
+      return nullptr;
```
jeanPerier: Shouldn't this be `return {};` ? It looks weird to me to have optional containing `nullptr`, though I agree it should be functionally equivalent given how `GetLastSymbolHelper ` is currently used.
I think you could also directly have `return (*this)(*designator)` above.

lib/evaluate/shape.h
```diff
@@ -86,56 +86,57 @@ MaybeExtentExpr GetSize(Shape &&);
...
+    return Shape{GetArrayConstructorExtent(aconst)};
   }
-  void Handle(const ImpliedDoIndex &) { Scalar(); }
-  void Handle(const DescriptorInquiry &) { Scalar(); }
-  template<int KIND> void Handle(const TypeParamInquiry<KIND> &) { Scalar(); }
-  void Handle(const BOZLiteralConstant &) { Scalar(); }
-  void Handle(const NullPointer &) { Return(); }
   template<typename D, typename R, typename LO, typename RO>
-  void Handle(const Operation<D, R, LO, RO> &operation) {
+  Result operator()(const Operation<D, R, LO, RO> &operation) const {
```
jeanPerier: As mentioned more in detail in another comment, I do not think this overload can be selected.
See:
```
  integer :: x, y(10)
  integer, parameter :: p1(*) = shape(abs(x) + y), p2(*) = shape(y + abs(x))
end
```

lib/evaluate/traverse.h
```diff
@@ -0,0 +1,311 @@
...
+  }
+
+  Visitor &visitor_;
+};
+
+template<typename Visitor, typename Result> class TraverseBase {
+public:
+  explicit TraverseBase(Visitor &v) : traverse_{v} {}
+  static Result Default() { return {}; }
+  template<typename A> Result operator()(const A &x) const {
```
jeanPerier: This default template overload is the issue with `Operation<>` overloads. It has priority over down casts. I cannot find a clever SFINAE solution to prevent that.
The old framework, though more complex, did not have this issue.

jeanPerier: One in-between solution I can think of is to have another CRTP helper class to catch operations.
Something like:
```
template<typename Visitor>
class OperationHandler {
  using Result = typename Visitor::Result;
  // only here all ops must be defined
  template<typename T>
  Result operation()(const Add<T>& add) {
    return Visitor::HandleOperation(add);
  }
  // explicitly define operator for all other ops ....
};

class GetShapeHelper
  : public AnyTraverse<GetShapeHelper, std::optional<Shape>>, OperationHandler<GetShapeHelper> {
public: 
  // ....
  template<typename D, typename R, typename LO, typename RO>
  Result HandleOperation(const Operation<D, R, LO, RO> &operation) const {
    // ...
  }
};
```
Not sure it is a nice design pattern though.

lib/evaluate/check-expression.cc
```diff
@@ -0,0 +1,240 @@
...
+  const semantics::Scope &scope_;
+};
+
+template<typename A>
+void CheckSpecificationExpr(const A &x, parser::ContextualMessages &messages,
+    const semantics::Scope &scope) {
+  if (auto why{CheckSpecificationExprHelper{scope}(x)}) {
+    std::stringstream ss;
+    ss << x;
+    messages.Say("The expression (%s) cannot be used as a "
```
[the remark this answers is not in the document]
klausler: If you like.

lib/evaluate/check-expression.cc
```diff
@@ -0,0 +1,240 @@
...
+        (*this)(x.parent());
+  }
+  bool operator()(const DescriptorInquiry &) const { return false; }
+  template<typename T> bool operator()(const ArrayConstructor<T> &) const {
+    return false;
+  }
+  bool operator()(const StructureConstructor &) const { return false; }
+  template<typename T> bool operator()(const FunctionRef<T> &) { return false; }
+  template<typename D, typename R, typename... O>
+  bool operator()(const Operation<D, R, O...> &) const {
```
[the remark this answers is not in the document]
klausler: Understood.  In the case of `IsInitialDataTarget()` the grammar requires that the target be a designator anyway.

lib/evaluate/tools.h
klausler: `return {};` would be the same as `return std::nullopt;`, meaning that the result of the search remains unknown.  `return nullptr;` means that the result of the search is known and that it has failed.

lib/evaluate/check-expression.cc
```diff
@@ -0,0 +1,240 @@
...
+private:
+  const semantics::Scope &scope_;
+};
+
+template<typename A>
+void CheckSpecificationExpr(const A &x, parser::ContextualMessages &messages,
+    const semantics::Scope &scope) {
+  if (auto why{CheckSpecificationExprHelper{scope}(x)}) {
+    std::stringstream ss;
+    ss << x;
```
tskeith: Is it possible to use `AsFortran()` here? That's more specific about what kind of string representation will come out.

klausler: I could use `x.AsFortran(ss)` but I don't think that the output would be different.

klausler: Also, `operator<<` copes with `std::optional<>` better.

lib/evaluate/traverse.h
klausler: The solution turned out to be simple: just don't have a catch-all template in the visitor.

lib/evaluate/shape.h
klausler: This example works now.

klausler: This case works now.

lib/evaluate/traverse.h
klausler: The default general template turns out to not be necessary.

lib/evaluate/check-expression.cc
klausler: `operator<<` can handle `std::optional<>`.

PR #756, opened 2019-09-22
Implementation of semantic checks C1135, C1167, and C1168

psteinfeld:
These constraints state that CYCLE and EXIT statements should not leave DO
CONCURRENT, CRITICAL, or CHANGE TEAM constructs.

I added checking code to check-do.cc and removed some superseded code from
check-do.cc and semantics.cc.  The new code uses the construct stack
implemented in my previous pull request.

I also added a new test -- dosemantics11.f90 and modified the tests
dosemantics10.f90, doconcurrent05.f90, and doconcurrent06.f90 to adapt to
the new error messages.  I converted these latter two tests to use
test_error.sh since they only reported errors.

### DIFF
--- a/lib/common/template.h
+++ b/lib/common/template.h
@@ -111,6 +111,16 @@ std::optional<A> JoinOptional(std::optional<std::optional<A>> &&x) {
   return std::nullopt;
 }
 
+// Convert an std::optional to an ordinary pointer
+template<typename A>
+const A *GetPtrFromOptional(const std::optional<A> &x) {
+  if (x.has_value()) {
+    return &x.value();
+  } else {
+    return nullptr;
+  }
+}
+
 // Copy a value from one variant type to another.  The types allowed in the
 // source variant must all be allowed in the destination variant type.
 template<typename TOV, typename FROMV> TOV CopyVariant(const FROMV &u) {
--- a/lib/common/template.h
+++ b/lib/common/template.h
@@ -115,7 +115,7 @@ std::optional<A> JoinOptional(std::optional<std::optional<A>> &&x) {
 template<typename A>
 const A *GetPtrFromOptional(const std::optional<A> &x) {
   if (x.has_value()) {
-    return &x.value();
+    return &*x;
   } else {
     return nullptr;
   }
--- a/lib/semantics/check-do.cc
+++ b/lib/semantics/check-do.cc
@@ -658,17 +658,17 @@ static parser::CharBlock GetNodePosition(const ConstructNode &construct) {
       [&](const auto &x) { return GetConstructPosition(*x); }, construct);
 }
 
-void DoChecker::SayBadLeave(const char *stmtName, const char *enclosingStmtName,
+void DoChecker::SayBadLeave(StmtType stmtType, const char *enclosingStmtName,
     const ConstructNode &construct) const {
   context_
-      .Say("%s must not leave a %s statement"_err_en_US, stmtName,
+      .Say("%s must not leave a %s statement"_err_en_US, EnumToString(stmtType),
           enclosingStmtName)
       .Attach(GetNodePosition(construct), "The construct that was left"_en_US);
 }
 
 static const parser::DoConstruct *MaybeGetDoConstruct(
     const ConstructNode &construct) {
-  if (auto *const doNode{
+  if (auto const *doNode{
           std::get_if<const parser::DoConstruct *>(&construct)}) {
     return *doNode;
   }
@@ -683,43 +683,38 @@ static bool ConstructIsDoConcurrent(const ConstructNode &construct) {
 // Check that CYCLE and EXIT statements do not cause flow of control to
 // leave DO CONCURRENT, CRITICAL, or CHANGE TEAM constructs.
 void DoChecker::CheckForBadLeave(
-    const char *stmtTypeName, const ConstructNode &construct) const {
+    StmtType stmtType, const ConstructNode &construct) const {
   std::visit(
       common::visitors{
           [&](const parser::DoConstruct *doConstructPtr) {
             if (doConstructPtr->IsDoConcurrent()) {
               // C1135 and C1167
-              SayBadLeave(stmtTypeName, "DO CONCURRENT", construct);
+              SayBadLeave(stmtType, "DO CONCURRENT", construct);
             }
           },
           [&](const parser::CriticalConstruct *) {
             // C1135 and C1168
-            SayBadLeave(stmtTypeName, "CRITICAL", construct);
+            SayBadLeave(stmtType, "CRITICAL", construct);
           },
           [&](const parser::ChangeTeamConstruct *) {
             // C1135 and C1168
-            SayBadLeave(stmtTypeName, "CHANGE TEAM", construct);
+            SayBadLeave(stmtType, "CHANGE TEAM", construct);
           },
           [](const auto *) {},
       },
       construct);
 }
 
-static const char *GetStmtTypeName(bool isExit) {
-  if (isExit) {
-    return "EXIT";
-  } else {
-    return "CYCLE";
-  }
-}
+static bool isExit(StmtType stmtType) { return stmtType == StmtType::EXIT; }
 
-static bool StmtMatchesConstruct(const parser::Name *stmtName, bool isExit,
-    const parser::Name *constructName, const ConstructNode &construct) {
+static bool StmtMatchesConstruct(const parser::Name *stmtName,
+    StmtType stmtType, const parser::Name *constructName,
+    const ConstructNode &construct) {
   bool inDoConstruct{MaybeGetDoConstruct(construct) != nullptr};
   if (stmtName == nullptr) {
     return inDoConstruct;  // Unlabeled statements match all DO constructs
   } else if (constructName && constructName->source == stmtName->source) {
-    return isExit || inDoConstruct;
+    return isExit(stmtType) || inDoConstruct;
   } else {
     return false;
   }
@@ -727,9 +722,9 @@ static bool StmtMatchesConstruct(const parser::Name *stmtName, bool isExit,
 
 // C1167 Can't EXIT from a DO CONCURRENT
 void DoChecker::CheckDoConcurrentExit(
-    bool isExit, const ConstructNode &construct) const {
-  if (isExit && ConstructIsDoConcurrent(construct)) {
-    SayBadLeave("EXIT", "DO CONCURRENT", construct);
+    StmtType stmtType, const ConstructNode &construct) const {
+  if (isExit(stmtType) && ConstructIsDoConcurrent(construct)) {
+    SayBadLeave(StmtType::EXIT, "DO CONCURRENT", construct);
   }
 }
 
@@ -737,36 +732,35 @@ void DoChecker::CheckDoConcurrentExit(
 // levels looking for a construct that matches the CYCLE or EXIT statment.  At
 // every construct, check for a violation.  If we find a match without finding
 // a violation, the check is complete.
-void DoChecker::CheckNesting(bool isExit, const parser::Name *stmtName) const {
+void DoChecker::CheckNesting(
+    StmtType stmtType, const parser::Name *stmtName) const {
   const ConstructStack &stack{context_.constructStack()};
   for (auto rIter{stack.crbegin()}; rIter != stack.crend(); ++rIter) {
     const ConstructNode &construct{*rIter};
     const parser::Name *constructName{MaybeGetNodeName(construct)};
-    if (StmtMatchesConstruct(stmtName, isExit, constructName, construct)) {
-      CheckDoConcurrentExit(isExit, construct);
+    if (StmtMatchesConstruct(stmtName, stmtType, constructName, construct)) {
+      CheckDoConcurrentExit(stmtType, construct);
       return;  // We got a match, so we're finished checking
     }
-    CheckForBadLeave(GetStmtTypeName(isExit), construct);
+    CheckForBadLeave(stmtType, construct);
   }
 
   // We haven't found a match in the enclosing constructs
-  if (isExit) {
-    context_.Say(*context_.location(),
-        "No matching construct for EXIT statement"_err_en_US);
+  if (isExit(stmtType)) {
+    context_.Say("No matching construct for EXIT statement"_err_en_US);
   } else {
-    context_.Say(*context_.location(),
-        "No matching DO construct for CYCLE statement"_err_en_US);
+    context_.Say("No matching DO construct for CYCLE statement"_err_en_US);
   }
 }
 
 // C1135
 void DoChecker::Enter(const parser::CycleStmt &cycleStmt) {
-  CheckNesting(/*isExit*/ false, GetPtrFromOptional(cycleStmt.v));
+  CheckNesting(StmtType::CYCLE, GetPtrFromOptional(cycleStmt.v));
 }
 
 // C1167 and C1168
 void DoChecker::Enter(const parser::ExitStmt &exitStmt) {
-  CheckNesting(/*isExit*/ true, GetPtrFromOptional(exitStmt.v));
+  CheckNesting(StmtType::EXIT, GetPtrFromOptional(exitStmt.v));
 }
 
 }  // namespace Fortran::semantics
--- a/lib/semantics/check-do.cc
+++ b/lib/semantics/check-do.cc
@@ -28,12 +28,34 @@ namespace Fortran::semantics {
 
 using namespace parser::literals;
 
+static NamePtr GetNamePtr(const std::optional<parser::Name> &name) {
+  if (name.has_value()) {
+    return &(name.value());
+  } else {
+    return nullptr;
+  }
+}
+
+template<typename A> static NamePtr GetConstructName(const A &a) {
+  return GetNamePtr(std::get<0>(std::get<0>(a.t).statement.t));
+}
+
+static NamePtr GetConstructName(const parser::BlockConstruct &blockConstruct) {
+  return GetNamePtr(
+      std::get<parser::Statement<parser::BlockStmt>>(blockConstruct.t)
+          .statement.v);
+}
+
+template<typename A> static NamePtr GetStmtName(const A &a) {
+  return GetNamePtr(std::get<0>(a.t));
+}
+
 // 11.1.7.5 - enforce semantics constraints on a DO CONCURRENT loop body
 class DoConcurrentBodyEnforce {
 public:
   DoConcurrentBodyEnforce(SemanticsContext &context) : context_{context} {}
   std::set<parser::Label> labels() { return labels_; }
-  std::set<parser::CharBlock> names() { return names_; }
+  std::set<SourceName> names() { return names_; }
   template<typename T> bool Pre(const T &) { return true; }
   template<typename T> void Post(const T &) {}
   template<typename T> bool Pre(const parser::Statement<T> &statement) {
@@ -45,53 +67,53 @@ public:
   }
 
   // C1167
-  bool Pre(const parser::WhereConstructStmt &s) {
-    addName(std::get<std::optional<parser::Name>>(s.t));
+  bool Pre(const parser::WhereConstruct &s) {
+    addName(GetConstructName(s));
     return true;
   }
 
-  bool Pre(const parser::ForallConstructStmt &s) {
-    addName(std::get<std::optional<parser::Name>>(s.t));
+  bool Pre(const parser::ForallConstruct &s) {
+    addName(GetConstructName(s));
     return true;
   }
 
-  bool Pre(const parser::ChangeTeamStmt &s) {
-    addName(std::get<std::optional<parser::Name>>(s.t));
+  bool Pre(const parser::ChangeTeamConstruct &s) {
+    addName(GetConstructName(s));
     return true;
   }
 
-  bool Pre(const parser::CriticalStmt &s) {
-    addName(std::get<std::optional<parser::Name>>(s.t));
+  bool Pre(const parser::CriticalConstruct &s) {
+    addName(GetConstructName(s));
     return true;
   }
 
   bool Pre(const parser::LabelDoStmt &s) {
-    addName(std::get<std::optional<parser::Name>>(s.t));
+    addName(GetStmtName(s));
     return true;
   }
 
   bool Pre(const parser::NonLabelDoStmt &s) {
-    addName(std::get<std::optional<parser::Name>>(s.t));
+    addName(GetStmtName(s));
     return true;
   }
 
   bool Pre(const parser::IfThenStmt &s) {
-    addName(std::get<std::optional<parser::Name>>(s.t));
+    addName(GetStmtName(s));
     return true;
   }
 
   bool Pre(const parser::SelectCaseStmt &s) {
-    addName(std::get<std::optional<parser::Name>>(s.t));
+    addName(GetStmtName(s));
     return true;
   }
 
   bool Pre(const parser::SelectRankStmt &s) {
-    addName(std::get<0>(s.t));
+    addName(GetStmtName(s));
     return true;
   }
 
   bool Pre(const parser::SelectTypeStmt &s) {
-    addName(std::get<0>(s.t));
+    addName(GetStmtName(s));
     return true;
   }
 
@@ -213,9 +235,9 @@ private:
     return false;
   }
 
-  void addName(const std::optional<parser::Name> &nm) {
-    if (nm.has_value()) {
-      names_.insert(nm.value().source);
+  void addName(NamePtr nm) {
+    if (nm) {
+      names_.insert(nm->source);
     }
   }
 
@@ -238,15 +260,8 @@ public:
     return true;
   }
 
-  bool Pre(const parser::DoConstruct &) {
-    ++do_depth_;
-    return true;
-  }
-
   template<typename T> void Post(const T &) {}
 
-  // C1138: branch from within a DO CONCURRENT shall not target outside loop
-  void Post(const parser::ExitStmt &exitStmt) { checkName(exitStmt.v); }
   void Post(const parser::GotoStmt &gotoStmt) { checkLabelUse(gotoStmt.v); }
   void Post(const parser::ComputedGotoStmt &computedGotoStmt) {
     for (auto &i : std::get<std::list<parser::Label>>(computedGotoStmt.t)) {
@@ -277,22 +292,6 @@ public:
   void Post(const parser::ErrLabel &errLabel) { checkLabelUse(errLabel.v); }
   void Post(const parser::EndLabel &endLabel) { checkLabelUse(endLabel.v); }
   void Post(const parser::EorLabel &eorLabel) { checkLabelUse(eorLabel.v); }
-  void Post(const parser::DoConstruct &) { --do_depth_; }
-  void checkName(const std::optional<parser::Name> &nm) {
-    if (!nm.has_value()) {
-      if (do_depth_ == 0) {
-        context_.Say(currentStatementSourcePosition_,
-            "exit from DO CONCURRENT construct (%s)"_err_en_US,
-            doConcurrentSourcePosition_);
-      }
-      // nesting of named constructs is assumed to have been previously checked
-      // by the name/label resolution pass
-    } else if (names_.find(nm.value().source) == names_.end()) {
-      context_.Say(currentStatementSourcePosition_,
-          "exit from DO CONCURRENT construct (%s) to construct with name '%s'"_err_en_US,
-          doConcurrentSourcePosition_, nm.value().source);
-    }
-  }
 
   void checkLabelUse(const parser::Label &labelUsed) {
     if (labels_.find(labelUsed) == labels_.end()) {
@@ -305,7 +304,6 @@ private:
   SemanticsContext &context_;
   std::set<parser::Label> labels_;
   std::set<parser::CharBlock> names_;
-  int do_depth_{0};
   parser::CharBlock currentStatementSourcePosition_{nullptr};
   parser::CharBlock doConcurrentSourcePosition_{nullptr};
 };  // class DoConcurrentLabelEnforce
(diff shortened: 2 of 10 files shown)


lib/semantics/check-do.cc
```diff
@@ -647,20 +645,131 @@ void DoChecker::Leave(const parser::DoConstruct &x) {
...
+static parser::CharBlock GetNodePosition(const ConstructNode &construct) {
+  return std::visit(
+      [&](const auto &x) { return GetConstructPosition(*x); }, construct);
+}
+
+void DoChecker::SayBadLeave(const char *stmtChecked,
+    const char *enclosingStmtType, const ConstructNode &construct) const {
+  context_
+      .Say(*context_.location(), "%s must not leave a %s statement"_err_en_US,
+          stmtChecked, enclosingStmtType)
```
tskeith: You should be able to omit the first argument to `Say`. That's what the location defaults to.

lib/semantics/check-do.cc
```diff
@@ -647,20 +645,131 @@ void DoChecker::Leave(const parser::DoConstruct &x) {
...
+      .Attach(GetNodePosition(construct), "The construct that was left"_en_US);
+}
+
+static parser::DoConstruct const *MaybeGetDoConstruct(
```
tskeith: Everywhere else const pointers are written with the const before the type.

lib/semantics/check-do.cc
```diff
@@ -647,20 +645,131 @@ void DoChecker::Leave(const parser::DoConstruct &x) {
...
+  }
+}
+
+static bool ConstructIsDoConcurrent(const ConstructNode &construct) {
+  parser::DoConstruct const *doConstruct{MaybeGetDoConstruct(construct)};
+  return doConstruct && doConstruct->IsDoConcurrent();
+}
+
+static bool isExitStmt(const char *stmtType) {
+  return std::strncmp("EXIT", stmtType, 4) == 0;
```
tskeith: It looks like `stmtType` can only ever be "EXIT" or "CYCLE". If that's the case an enum would be a better representation.

lib/semantics/check-do.cc
```diff
@@ -647,20 +645,131 @@ void DoChecker::Leave(const parser::DoConstruct &x) {
...
+  } else if (constructName->source == stmtName->source) {
+    if (isExitStmt(stmtType)) {
+      return true;  // EXIT name matches any construct name
+    } else {
+      if (inDoConstruct) {
+        return true;  // CYCLE name matches only DO construct name
+      }
+    }
+  }
+  return false;
```
tskeith: This return expression could be a lot simpler:
```
  if (stmtName == nullptr) {
    return inDoConstruct;
  } else if (constructName && constructName->source == stmtName->source) {
    return isExitStmt(stmtType) || inDoConstruct;
  } else {
    return false;
  }
```

lib/semantics/check-do.h
```diff
@@ -25,6 +25,8 @@ struct ExitStmt;
 
 namespace Fortran::semantics {
 
+using NamePtr = parser::Name const *;
```
tskeith: In my opinion this type alias just obscures things and it would be better to use `const parser::Name *` everywhere.

test/semantics/CMakeLists.txt
```diff
@@ -137,16 +137,17 @@ set(ERROR_TESTS
   allocate12.f90
   allocate13.f90
   doconcurrent01.f90
+  dosemantics05.f90
+  dosemantics06.f90
```
tskeith: Why move these out of order?

test/semantics/doconcurrent05.f90
```diff
@@ -11,21 +11,17 @@
...
+! C1167 -- An exit-stmt shall not appear within a DO CONCURRENT construct if 
+! it belongs to that construct or an outer construct.
 
 subroutine do_concurrent_test1(n)
   implicit none
   integer :: n
   integer :: j,k
   mydoc: do concurrent(j=1:n)
   mydo:    do k=1,n
+!ERROR: EXIT must not leave a DO CONCURRENT statement
```
tskeith: You need to change `CMakeLists.txt` so this is executed by `test_errors.sh`.

test/semantics/doconcurrent05.f90
```diff
@@ -58,7 +56,10 @@ subroutine do_concurrent_test4(n)
   mytest4: if (n>0) then
   mydoc:    do concurrent(j=1:n)
               do concurrent(k=1:n)
+!ERROR: EXIT must not leave a DO CONCURRENT statement
                 if (k==5) exit
+!ERROR: EXIT must not leave a DO CONCURRENT statement
+!ERROR: EXIT must not leave a DO CONCURRENT statement
```
tskeith: Why does this error come out twice?

test/semantics/doconcurrent06.f90
```diff
@@ -30,6 +25,9 @@ subroutine do_concurrent_test1(n)
   nc4:             do i4=1,n
   nc5:               do concurrent(i5=1:n)
   nc6:                 do i6=1,n
+!ERROR: EXIT must not leave a DO CONCURRENT statement
+!ERROR: EXIT must not leave a DO CONCURRENT statement
+!ERROR: EXIT must not leave a DO CONCURRENT statement
```
tskeith: Repeated three times?

test/semantics/dosemantics11.f90
```diff
@@ -0,0 +1,341 @@
...
+
+subroutine s2()
+!ERROR: No matching construct for EXIT statement
+  exit
+end subroutine s2
+
+subroutine s3()
+  level0: block
+!ERROR: No matching construct for CYCLE statement
+    cycle level0
```
tskeith: It's confusing to say there is not matching construct because there is a block construct with this name. I think it would be better to say that the CYCLE statement must name a DO construct.

lib/semantics/check-do.cc
psteinfeld: Thanks.  I'll change that.

lib/semantics/check-do.cc
psteinfeld: Thanks.  I'll change that.

lib/semantics/check-do.cc
```diff
@@ -28,12 +28,34 @@ namespace Fortran::semantics {
 
 using namespace parser::literals;
 
+static NamePtr GetNamePtr(const std::optional<parser::Name> &name) {
```
klausler: This should be generalized and put into lib/common/template.h among the other std::optional<> utilities if it's useful.

lib/semantics/check-do.cc
```diff
@@ -28,12 +28,34 @@ namespace Fortran::semantics {
...
+  if (name.has_value()) {
+    return &(name.value());
```
klausler: You can use `*` rather than `.value()` immediately after testing for the presence of the value if you like.

You don't have to parenthesize the operand of a prefix operator in C/C++ if the operand is a primary or an application of a postfix operator.  All postfix operators in C/C++ bind more tightly than any prefix operator does.

lib/semantics/check-do.cc
```diff
@@ -213,9 +235,9 @@ class DoConcurrentBodyEnforce {
     return false;
   }
 
-  void addName(const std::optional<parser::Name> &nm) {
-    if (nm.has_value()) {
-      names_.insert(nm.value().source);
+  void addName(NamePtr nm) {
```
klausler: `addName` is not an accessor, mutator, or obvious analog to a feature in a standard library, so it should be capitalized.

lib/semantics/check-do.cc
```diff
@@ -647,20 +645,131 @@ void DoChecker::Leave(const parser::DoConstruct &x) {
...
+    const char *enclosingStmtType, const ConstructNode &construct) const {
+  context_
+      .Say(*context_.location(), "%s must not leave a %s statement"_err_en_US,
+          stmtChecked, enclosingStmtType)
+      .Attach(GetNodePosition(construct), "The construct that was left"_en_US);
+}
+
+static parser::DoConstruct const *MaybeGetDoConstruct(
+    const ConstructNode &construct) {
+  if (const auto doNode{std::get_if<const parser::DoConstruct *>(&construct)}) {
```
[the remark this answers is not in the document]
psteinfeld: Thanks, Tim.  I'll make that change.
Interestingly (for me), the declaration
```
  const auto do1{std::get_if<const parser::DoConstruct *>(&construct)};
```
results in the following type for ```do1```:
```
const Fortran::parser::DoConstruct* const* const
```

The declaration
```
  const auto *do2{std::get_if<const parser::DoConstruct *>(&construct)};
```
results in the following type for ```do2```:
```
const Fortran::parser::DoConstruct* const*
```

The declaration
```
  const auto * const do3{std::get_if<const parser::DoConstruct *>(&construct)};
```
results in the following type for ```do3```:
```
const Fortran::parser::DoConstruct* const* const
```

lib/semantics/check-do.cc
```diff
@@ -213,9 +235,9 @@ class DoConcurrentBodyEnforce {
     return false;
   }
 
-  void addName(const std::optional<parser::Name> &nm) {
-    if (nm.has_value()) {
-      names_.insert(nm.value().source);
+  void addName(NamePtr nm) {
+    if (nm) {
```
klausler: Can `AddName()` be legitimately called with a null pointer?  If not, use `CHECK()` or `DEREF()` instead.

lib/semantics/check-do.cc
```diff
@@ -647,20 +645,131 @@ void DoChecker::Leave(const parser::DoConstruct &x) {
...
+  }
+  return nullptr;
+}
+
+// Check that CYCLE and EXIT statements do not cause flow of control to
+// leave DO CONCURRENT, CRITICAL, or CHANGE TEAM constructs.
+void DoChecker::CheckForBadLeave(
+    const char *stmtName, const ConstructNode &construct) const {
+  // C1135 and C1167
+  if (const auto doConstructPtr{MaybeGetDoConstruct(construct)}) {
```
klausler: If `MaybeGetDoConstruct()` return a pointer, use `const auto *doConstructPtr`.

lib/semantics/check-do.cc
```diff
@@ -647,20 +645,131 @@ void DoChecker::Leave(const parser::DoConstruct &x) {
...
+    if (doConstructPtr->IsDoConcurrent()) {
+      SayBadLeave(stmtName, "DO CONCURRENT", construct);
+    }
+    return;
```
klausler: Use `else` rather than `return`.  You aren't testing for exceptional conditions.

lib/semantics/check-do.cc
```diff
@@ -647,20 +645,131 @@ void DoChecker::Leave(const parser::DoConstruct &x) {
...
+  }
+  // C1135 and C1168
+  if (const auto criticalConstructPtr{
```
klausler: You aren't using the `auto` variable that you declare in the predicate.  When you compile this code, you should get a warning about unused variables.

lib/semantics/check-do.cc
```diff
@@ -647,20 +645,131 @@ void DoChecker::Leave(const parser::DoConstruct &x) {
...
+    return;
+  }
+  if (const auto changeTeamConstructPtr{
+          std::get_if<const parser::ChangeTeamConstruct *>(&construct)}) {
+    SayBadLeave(stmtName, "CHANGE TEAM", construct);
+    return;
+  }
+}
+
+static bool ConstructIsDoConcurrent(const ConstructNode &construct) {
```
klausler: This handy predicate could have been used in the immediately preceding function.

lib/semantics/check-do.cc
```diff
@@ -647,20 +645,131 @@ void DoChecker::Leave(const parser::DoConstruct &x) {
...
+
+static bool isExitStmt(const char *stmtType) {
+  return std::strncmp("EXIT", stmtType, 4) == 0;
+}
+
+static bool StmtMatchesConstruct(NamePtr stmtName, const char *stmtType,
+    NamePtr constructName, const ConstructNode &construct) {
+  bool inDoConstruct{MaybeGetDoConstruct(construct) != nullptr};
+  if (stmtName == nullptr) {
+    if (inDoConstruct) {
```
klausler: `if (x) { return true; } else { return false; }` can be implemented as `return x;`.  Boolean expressions don't have to be limited to use in statement predicates.

lib/semantics/check-do.cc
```diff
@@ -647,20 +645,131 @@ void DoChecker::Leave(const parser::DoConstruct &x) {
...
+    SayBadLeave("EXIT", "DO CONCURRENT", construct);
   }
 }
 
-// C1166
-void DoChecker::Enter(const parser::ExitStmt &) {
-  if (!context_.InsideDoConstruct()) {
-    context_.Say(
-        *context_.location(), "EXIT must be within a DO construct"_err_en_US);
+// Check nesting violations for a CYCLE or EXIT statement Loop up the nesting
```
klausler: Missing a period or colon.

lib/semantics/check-do.cc
klausler: Or a Boolean "isExit", which would make `isExitStmt()` vanish.

lib/semantics/check-do.cc
tskeith: Another advantage of this style (besides consistency) is that you don't need to include "Ptr" in the name because it's obvious it is a pointer.

lib/semantics/check-do.cc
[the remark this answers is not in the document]
tskeith: This is another case of not normally declaring local variables const. Declaring `*do3` to be const is much more important than declaring `do3` const. The latter does convey some information (i.e. that `do3` is not going to be assigned again) but the same would apply to most other local variables too.

lib/semantics/check-do.cc
```diff
@@ -647,20 +645,131 @@ void DoChecker::Leave(const parser::DoConstruct &x) {
...
+  if (stmtName == nullptr) {
+    if (inDoConstruct) {
+      return true;  // Unlabeled statements match all DO constructs
+    } else {
+      return false;  // Unlabeled statements match no non-DO constructs
+    }
+  } else if (!constructName) {
+    return false;  // name on CYCLE/EXIT, but not on the construct
+  } else if (constructName->source == stmtName->source) {
+    if (isExitStmt(stmtType)) {
```
vjayathirtha-nv: `return (isExitStmt(stmtType) || inDoConstruct);` ?

klausler:
Maybe without the superfluous parentheses.

lib/semantics/check-do.cc
```diff
@@ -647,20 +645,131 @@ void DoChecker::Leave(const parser::DoConstruct &x) {
...
-  if (!context_.InsideDoConstruct()) {
-    context_.Say(
-        *context_.location(), "EXIT must be within a DO construct"_err_en_US);
+// Check nesting violations for a CYCLE or EXIT statement Loop up the nesting
+// levels looking for a construct that matches the CYCLE or EXIT statment.  At
+// every construct, check for a violation.  If we find a match without finding
+// a violation, the check is complete.
+void DoChecker::CheckNesting(const char *stmtType, NamePtr stmtName) const {
+  const ConstructStack &stack{context_.constructStack()};
+  for (ConstructStack::const_reverse_iterator riter = stack.crbegin();
```
vjayathirtha-nv: can you make this `rIter`?

klausler: and use braces, and consider omitting the explicit type.

lib/semantics/check-do.h
```diff
@@ -34,6 +36,12 @@ class DoChecker : public virtual BaseChecker {
 
 private:
   SemanticsContext &context_;
+
+  void SayBadLeave(const char *stmtChecked, const char *enclosingStmt,
+      const ConstructNode &) const;
+  void CheckDoConcurrentExit(const char *s, const ConstructNode &) const;
```
klausler: Either use a name that conveys more information that `s`, or leave it out.

lib/semantics/semantics.h
```diff
@@ -52,7 +52,7 @@ class Symbol;
 using ConstructNode = std::variant<const parser::AssociateConstruct *,
     const parser::BlockConstruct *, const parser::CaseConstruct *,
     const parser::ChangeTeamConstruct *, const parser::CriticalConstruct *,
-    const parser::DoConstruct *, const parser::ForAllConstruct *,
+    const parser::DoConstruct *, const parser::ForallConstruct *,
```
klausler: What changed here?

lib/semantics/check-do.cc
psteinfeld: Sounds good to me.  I'll make that change.

lib/semantics/check-do.cc
psteinfeld: Yes.  I'll change it as suggested.

lib/semantics/check-do.h
psteinfeld: OK.  I'll make that change.

test/semantics/doconcurrent05.f90
psteinfeld: Yes.  That was my intention, but I screwed up.  I'll fix that.

test/semantics/doconcurrent05.f90
psteinfeld: In this test, the ```EXIT``` statement at line 63 causes two different ```DO CONCURRENT``` statements to be exited, one at line 57 and the other at line 58.  It doesn't come out in the test source, but each error message has an additional message attached to it that lists the line being exited.  Thus, the results that the use sees are two different message combinations that point to both errors.

test/semantics/doconcurrent06.f90
psteinfeld: For the same reason given for the test ```doconcurrent05.f90```.

lib/semantics/check-do.cc
psteinfeld: Thanks.  For me, it made the code a little more readable.  I'll move it.

test/semantics/doconcurrent05.f90
sscalpone: Why don't those additional messages appear in the test?

lib/semantics/check-do.cc
psteinfeld: Thanks, Peter.  I purposely chose to use the ```.value()``` because I thought that using ```*``` might look like an ordinary pointer dereference, especially since I was in the process of creating a pointer.  I'll remove the parentheses.

lib/semantics/check-do.cc
psteinfeld: Thanks.  I'll make that change.

lib/semantics/check-do.cc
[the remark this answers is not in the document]
psteinfeld: To complete the set, the declaration ```auto * const doNode``` results in the type
```
const Fortran::parser::DoConstruct* const* const
```
In this case, the source code seems to me to be the best reflection of the underlying type.

test/semantics/doconcurrent05.f90
psteinfeld: Currently, our test infrastructure does not support checking of these additional messages.

lib/semantics/check-do.cc
psteinfeld: ```AddName()``` gets called with a null pointer for statements that don't have names.

lib/semantics/check-do.cc
psteinfeld: Thanks.  I'll change this.

lib/semantics/check-do.cc
psteinfeld: Good catch.  Weirdly, I didn't get a warning.  I'm using g++ version 8.3.0.

lib/semantics/check-do.cc
psteinfeld: Do'h!  Good catch.  I'll fix it.

lib/semantics/check-do.cc
psteinfeld: Yep.  Tim pointed this out, too.

lib/semantics/check-do.cc
psteinfeld: Thanks.  I'll fix this.

lib/semantics/check-do.cc
```diff
@@ -647,20 +645,131 @@ void DoChecker::Leave(const parser::DoConstruct &x) {
...
+    return *doNode;
+  }
+  return nullptr;
+}
+
+// Check that CYCLE and EXIT statements do not cause flow of control to
+// leave DO CONCURRENT, CRITICAL, or CHANGE TEAM constructs.
+void DoChecker::CheckForBadLeave(
+    const char *stmtName, const ConstructNode &construct) const {
+  // C1135 and C1167
```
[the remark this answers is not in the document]
psteinfeld: Thanks, Jean.  I like your suggesting on using ```std::visit```.  I'll do it!

lib/semantics/check-do.h
psteinfeld: Thanks for catching this typo.  I'll remove the ```s```.

test/semantics/dosemantics11.f90
psteinfeld: Good point.  I was trying to combine the error reporting for CYCLE and EXIT statements, but it's clearly better to differentiate them.  I'll do that.

lib/semantics/check-do.cc
psteinfeld: All sounds good.  I'll change the code.

lib/semantics/check-do.cc
```diff
@@ -660,116 +658,115 @@ static parser::CharBlock GetNodePosition(const ConstructNode &construct) {
...
+      .Say("%s must not leave a %s statement"_err_en_US, stmtName,
+          enclosingStmtName)
       .Attach(GetNodePosition(construct), "The construct that was left"_en_US);
 }
 
-static parser::DoConstruct const *MaybeGetDoConstruct(
+static const parser::DoConstruct *MaybeGetDoConstruct(
     const ConstructNode &construct) {
-  if (const auto doNode{std::get_if<const parser::DoConstruct *>(&construct)}) {
+  if (auto *const doNode{
```
tskeith: `const auto *`

lib/semantics/check-do.cc
```diff
@@ -660,116 +658,115 @@ static parser::CharBlock GetNodePosition(const ConstructNode &construct) {
...
 
   // We haven't found a match in the enclosing constructs
-  context_.Say(*context_.location(),
-      "No matching construct for %s statement"_err_en_US, stmtType);
+  if (isExit) {
+    context_.Say(*context_.location(),
+        "No matching construct for EXIT statement"_err_en_US);
+  } else {
+    context_.Say(*context_.location(),
+        "No matching DO construct for CYCLE statement"_err_en_US);
```
tskeith: Can't you omit the first arg from these two calls to `Say()`.

lib/semantics/check-do.cc
```diff
@@ -660,116 +658,115 @@ static parser::CharBlock GetNodePosition(const ConstructNode &construct) {
...
 // C1135
 void DoChecker::Enter(const parser::CycleStmt &cycleStmt) {
-  CheckNesting("CYCLE", GetNamePtr(cycleStmt.v));
+  CheckNesting(/*isExit*/ false, GetPtrFromOptional(cycleStmt.v));
 }
 
 // C1167 and C1168
 void DoChecker::Enter(const parser::ExitStmt &exitStmt) {
-  CheckNesting("EXIT", GetNamePtr(exitStmt.v));
+  CheckNesting(/*isExit*/ true, GetPtrFromOptional(exitStmt.v));
```
tskeith: If you used an `enum class` instead of `bool` these calls would be self-documenting and you wouldn't need `GetStmtTypeName` up above because declaring it with `ENUM_CLASS` gives it to you for free.

lib/semantics/check-do.cc
psteinfeld: Thanks, Tim.  I'll change this.

lib/semantics/check-do.cc
psteinfeld: Yep.  I'll remove them.

lib/semantics/check-do.cc
psteinfeld: Sounds good.  I'll do it!

lib/common/template.h
```diff
@@ -111,6 +111,16 @@ std::optional<A> JoinOptional(std::optional<std::optional<A>> &&x) {
   return std::nullopt;
 }
 
+// Convert an std::optional to an ordinary pointer
+template<typename A>
+const A *GetPtrFromOptional(const std::optional<A> &x) {
+  if (x.has_value()) {
+    return &x.value();
```
klausler: `return &*x;`

lib/semantics/check-do.cc
```diff
@@ -27,13 +28,32 @@
 namespace Fortran::semantics {
 
 using namespace parser::literals;
+using namespace common;
```
klausler: Please don't use `using namespace` except for namespaces specifically limited for user-defined literals.

psteinfeld: Thanks, Peter.  Won't do.

lib/common/template.h
psteinfeld: Thanks, Peter.  I'll make this change.

lib/semantics/check-do.cc
```diff
@@ -647,20 +643,124 @@ void DoChecker::Leave(const parser::DoConstruct &x) {
...
+    if (StmtMatchesConstruct(stmtName, stmtType, constructName, construct)) {
+      CheckDoConcurrentExit(stmtType, construct);
+      return;  // We got a match, so we're finished checking
+    }
+    CheckForBadLeave(stmtType, construct);
+  }
+
+  // We haven't found a match in the enclosing constructs
+  if (isExit(stmtType)) {
+    context_.Say("No matching construct for EXIT statement"_err_en_US);
```
vjayathirtha-nv:
```suggestion
    context_.Say("No matching DO construct for %s statement"_err_en_US, EnumToString(stmtType));
```

This will avoid the if-else block. Assuming the `DO` should be there in both messages.

vjayathirtha-nv: If the `DO` was omitted on purposed disregard my comment.

lib/semantics/check-do.cc
```diff
@@ -647,20 +643,124 @@ void DoChecker::Leave(const parser::DoConstruct &x) {
...
+    const ConstructNode &construct) const {
+  context_
+      .Say("%s must not leave a %s statement"_err_en_US, EnumToString(stmtType),
+          enclosingStmtName)
+      .Attach(GetNodePosition(construct), "The construct that was left"_en_US);
+}
+
+static const parser::DoConstruct *MaybeGetDoConstruct(
+    const ConstructNode &construct) {
+  if (auto const *doNode{
```
tskeith: `const auto *`

lib/semantics/check-do.cc
```diff
@@ -647,20 +643,124 @@ void DoChecker::Leave(const parser::DoConstruct &x) {
...
+          std::get_if<const parser::DoConstruct *>(&construct)}) {
+    return *doNode;
+  }
+  return nullptr;
+}
+
+static bool ConstructIsDoConcurrent(const ConstructNode &construct) {
+  parser::DoConstruct const *doConstruct{MaybeGetDoConstruct(construct)};
```
tskeith: `const parser::DoConstruct *`

lib/semantics/check-do.cc
```diff
@@ -647,20 +643,124 @@ void DoChecker::Leave(const parser::DoConstruct &x) {
...
+static bool isExit(StmtType stmtType) { return stmtType == StmtType::EXIT; }
+
+static bool StmtMatchesConstruct(const parser::Name *stmtName,
+    StmtType stmtType, const parser::Name *constructName,
+    const ConstructNode &construct) {
+  bool inDoConstruct{MaybeGetDoConstruct(construct) != nullptr};
+  if (stmtName == nullptr) {
+    return inDoConstruct;  // Unlabeled statements match all DO constructs
+  } else if (constructName && constructName->source == stmtName->source) {
+    return isExit(stmtType) || inDoConstruct;
```
tskeith: It seems to me that `stmtType == StmtType::EXIT` is at least as clear as `isExit(stmtType)`.

If you keep `isExit` it should be `IsExit`.

lib/semantics/check-do.cc
```diff
@@ -647,20 +643,124 @@ void DoChecker::Leave(const parser::DoConstruct &x) {
...
+          enclosingStmtName)
+      .Attach(GetNodePosition(construct), "The construct that was left"_en_US);
+}
+
+static const parser::DoConstruct *MaybeGetDoConstruct(
+    const ConstructNode &construct) {
+  if (auto const *doNode{
+          std::get_if<const parser::DoConstruct *>(&construct)}) {
+    return *doNode;
+  }
```
klausler: `else`

lib/semantics/check-do.cc
psteinfeld: Yes, ```CYCLE``` statements must specifically match a ```DO``` construct, whereas ```EXIT``` statements can match a variety of constructs.

lib/semantics/check-do.cc
```diff
@@ -647,20 +643,124 @@ void DoChecker::Leave(const parser::DoConstruct &x) {
...
 }
 
+// Check nesting violations for a CYCLE or EXIT statement.  Loop up the nesting
+// levels looking for a construct that matches the CYCLE or EXIT statment.  At
+// every construct, check for a violation.  If we find a match without finding
+// a violation, the check is complete.
+void DoChecker::CheckNesting(
+    StmtType stmtType, const parser::Name *stmtName) const {
+  const ConstructStack &stack{context_.constructStack()};
+  for (auto rIter{stack.crbegin()}; rIter != stack.crend(); ++rIter) {
```
klausler: I understand that you're traversing `stack` from its most-recently-pushed entries to its oldest entries.  It would be less surprising if the loop used a decremented iterator and ran from `end()` back to `begin()` rather than using the rarely-used reversed iterator form `crbegin()`.  That is a facility that makes it easy to pass a reversed view of a container to an "algorithm" like `find_if()`, but one typically doesn't run into it in explicit looping situations.

```
for (auto iter{stack.cend()}; iter-- != stack.cbegin(); ) { ... }
```

lib/semantics/check-do.cc
psteinfeld: Done.

lib/semantics/check-do.cc
psteinfeld: Done.

lib/semantics/check-do.cc
psteinfeld: Thanks, Tim.  I'll remove and replace the function.

lib/semantics/check-do.cc
psteinfeld: Thanks, Peter.  I made this change.

lib/semantics/check-do.cc
psteinfeld: Cool!  I'll make this change.  Thanks, Peter.

PR #758, opened 2019-09-24
Improve module file reading and writing

tskeith:
Use `flock()` to lock `.mod` files during reading and writing to prevent
writing while a read is in progress and reading or writing when a write
is in progress. To achieve this, change module file writing to use C
stdio so that the file descriptor is readily available. `errno` seems
to be set more reliably as well.

Change module file reading to only read the file once; we used to
read it to verify the checksum and then again to parse it.
Instead, change `Parsing` so that we can get the file contents
after `Prescan()` and use that to verify the checksum. Also, it has
a mechanism for searching directories for files, so make use of that
instead of duplicating that functionality in `ModFileReader`.
This requires some changes to how errors are returned so they can
be reported in the right place. The "isModuleFile" flag is passed in
to `SourceFile` so that file locking can be done only for module files.


sscalpone:
I'm not too clear on the reason for the locking.  Isn't a mod file written to a tmp file & then, if the tmp mod file is different from the original, the tmp is renamed?  If that's the case, then any process that has the old file open will still be accessing the old file even after the old file is removed.

tskeith:
> I'm not too clear on the reason for the locking. Isn't a mod file written to a tmp file & then, if the tmp mod file is different from the original, the tmp is renamed? If that's the case, then any process that has the old file open will still be accessing the old file even after the old file is removed.

No, the mod file is written directly to its destination, which is why locking is necessary but a temp file is not.

What you're suggesting is:
1. If the file exists, read it and compare to what will be written
2. Write to a temp
3. Rename temp to correct name (`rename()` from stdio seems to be the only way to do this)

That does seem simpler -- I'll try it that way and see.

lib/semantics/mod-file.cc
```diff
@@ -591,52 +609,60 @@ std::ostream &PutLower(std::ostream &os, const std::string &str) {
 }
 
 // Write the module file at path, prepending header. Return false on error.
```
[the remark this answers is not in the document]
tskeith: Fixed

lib/semantics/mod-file.cc
```diff
@@ -591,52 +609,60 @@ std::ostream &PutLower(std::ostream &os, const std::string &str) {
...
   }
-  return !stream.get(c);
+  return std::fgetc(fp) == EOF;
 }
 
-// Compute a simple hash of the contents of a module file and
-// return it as a string of hex digits.
-// This uses the Fowler-Noll-Vo hash function.
-template<typename Iter> static std::string CheckSum(Iter begin, Iter end) {
+static std::string CheckSum(const std::string_view &contents) {
```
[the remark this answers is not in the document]
tskeith: Right, I think that was an accident. Fixed now.

lib/parser/source.cc
```diff
@@ -124,7 +125,13 @@ bool SourceFile::Open(std::string path, std::stringstream *error) {
   errno = 0;
   fileDescriptor_ = open(path.c_str(), O_RDONLY);
   if (fileDescriptor_ < 0) {
-    *error << "could not open " << errorPath << ": " << std::strerror(errno);
+    *error << "Could not open " << errorPath << ": " << std::strerror(errno);
+    return false;
+  }
+  if (isModuleFile_ && flock(fileDescriptor_, LOCK_SH | LOCK_NB) != 0) {
```
klausler: You could just lock all files, not just modules.

klausler: _Also, when locking fails, it seems harsh to abort compilation.  Why not try a `sleep(1)` and a few retries?_

lib/semantics/mod-file.cc
```diff
@@ -13,32 +13,45 @@
...
 #include <sys/types.h>
+#include <unistd.h>
 #include <vector>
 
 namespace Fortran::semantics {
 
 using namespace parser::literals;
 
-// The initial characters of a file that identify it as a .mod file.
+// The first line of a file that identify it as a .mod file.
```
klausler: Grammar

lib/semantics/mod-file.cc
```diff
@@ -13,32 +13,45 @@
...
 // The first three bytes are a Unicode byte order mark that ensures
 // that the module file is decoded as UTF-8 even if source files
 // are using another encoding.
-static constexpr auto magic{"\xef\xbb\xbf!mod$ v1 sum:"};
+static constexpr struct {
+  const char *bom{"\xef\xbb\xbf"};
```
klausler: These could be `static` members, and the lengths could then be calculated (or checked).

lib/semantics/mod-file.cc
```diff
@@ -590,53 +608,65 @@ std::ostream &PutLower(std::ostream &os, const std::string &str) {
...
-      return true;
+// Write the module file at path, prepending header. If an error occurs,
+// return an error message, otherwise nullptr.
+static const char *WriteFile(
+    const std::string &path, const std::string &contents) {
+  std::FILE *fp{std::fopen(path.c_str(), "a+")};
+  if (!fp) {
+    return std::strerror(errno);
+  }
+  if (flock(fileno(fp), LOCK_EX | LOCK_NB) != 0) {
```
klausler: Iterated retries might be better.

lib/semantics/mod-file.cc
```diff
@@ -590,53 +608,65 @@ std::ostream &PutLower(std::ostream &os, const std::string &str) {
...
+    fclose(fp);
+    return "Another compilation is currently using it";
+  }
+  std::size_t size = std::ftell(fp);
```
klausler: Are you using `=` to get a cast?  It's more clear to use `static_cast<>`.

lib/semantics/mod-file.cc
```diff
@@ -590,53 +608,65 @@ std::ostream &PutLower(std::ostream &os, const std::string &str) {
...
+  if (size == header.size() + contents.size()) {
+    if (FileContentsMatch(fp, header, contents)) {
+      fclose(fp);
+      return nullptr;
     }
-    stream.seekp(0);
+  }
+  std::rewind(fp);
+  ftruncate(fileno(fp), 0);
+  if (fwrite(header.c_str(), 1, header.size(), fp) == header.size() &&
```
klausler: `fwrite()` returns the number of *items* written, not the number of bytes; is this comparison correct?

tskeith: Yes, item size is 1, nitems is `header.size()`.

lib/semantics/mod-file.cc
tskeith: OK

lib/parser/source.cc
tskeith: Locking all files would be okay, except the error would have to be less specific. For .mod files we can say "another compilation is writing it". If it applies to any file, all we can say is "the file is locked" or something.

tskeith: I chose not to retry because this error would normally indicate a failure of the build process. I think it's more helpful to users to know about that than to try to plow through and get their file compiled.

lib/semantics/mod-file.cc
klausler: Consider reversing the arguments to `fwrite()` and then comparing its result to 1.

lib/semantics/mod-file.cc
tskeith: Fixed

lib/semantics/mod-file.cc
tskeith: Yes, much better that way.

lib/semantics/mod-file.cc
tskeith: OK, good idea.

lib/semantics/mod-file.cc
```diff
@@ -590,53 +608,65 @@ std::ostream &PutLower(std::ostream &os, const std::string &str) {
...
+      CheckSum(contents) + modHeader.terminator};
+  if (size == header.size() + contents.size()) {
+    if (FileContentsMatch(fp, header, contents)) {
+      fclose(fp);
+      return nullptr;
     }
-    stream.seekp(0);
+  }
+  std::rewind(fp);
+  ftruncate(fileno(fp), 0);
```
psteinfeld: I was playing around with these changes in an unrelated place, and I compiled mod-file.cc with g++ 8.3.0, and I got this message:
```
../lib/semantics/mod-file.cc:633:12: error: ignoring return value of â€˜int ftruncate(int, __off_t)â€™, declared with attribute warn_unused_result [-Werror=unused-result]
```
For some reason, this warning doesn't appear in your version of the file.  If I put in a check that the return value of ```ftruncate()==0```, the warning goes away.

tskeith: Checking the return value seems like a good idea, so I've made that change.

tskeith:
I've pushed a new version that uses a temp file instead of file locking. I'm interested in opinions on which is better.

lib/parser/parsing.h
```diff
@@ -73,6 +74,7 @@ class Parsing {
 private:
   Options options_;
   CookedSource cooked_;
+  const SourceFile *sourceFile_{nullptr};
```
klausler: Please add a comment that identifies the owner of this storage.

Maybe it doesn't need to be a permanent part of the state here.  What if `Prescan()` were to return this pointer instead?

lib/semantics/mod-file.cc
```diff
@@ -13,32 +13,43 @@
...
 // that the module file is decoded as UTF-8 even if source files
 // are using another encoding.
-static constexpr auto magic{"\xef\xbb\xbf!mod$ v1 sum:"};
+struct ModHeader {
+  static constexpr const char *bom{"\xef\xbb\xbf"};
+  static constexpr const char *magic{"!mod$ v1 sum:"};
+  static constexpr int magicLen{std::strlen(magic)};
+  static constexpr int sumLen{16};
+  static constexpr char terminator{'\n'};
+  static constexpr int len{magicLen + 1 + sumLen};
```
klausler: Is the byte order mark no longer included in the length?

lib/semantics/mod-file.cc
```diff
@@ -55,11 +66,11 @@ static std::ostream &PutAttrs(std::ostream &, Attrs,
 static std::ostream &PutAttr(std::ostream &, Attr);
 static std::ostream &PutType(std::ostream &, const DeclTypeSpec &);
 static std::ostream &PutLower(std::ostream &, const std::string &);
-static bool WriteFile(const std::string &, std::string &&);
+static int WriteFile(const std::string &, const std::string &);
```
klausler: What does the `int` result signify?  errno?

lib/semantics/mod-file.cc
```diff
@@ -590,84 +608,116 @@ std::ostream &PutLower(std::ostream &os, const std::string &str) {
...
+static Temp MkTemp(const std::string &path) {
+  auto length{path.length()};
+  auto dot{path.find_last_of("./")};
+  std::string suffix{dot < length && path[dot] == '.' ? path.substr(dot) : ""};
+  CHECK(length > suffix.length() &&
+      path.substr(length - suffix.length()) == suffix);
+  auto tempPath{path.substr(0, length - suffix.length()) + "XXXXXX" + suffix};
+  int fd{mkstemps(&tempPath[0], suffix.length())};
+  auto mask{umask(0777)};
+  umask(mask);
```
klausler: Why mess with umask if you're going to force the mode to 0600 anyway?

lib/semantics/mod-file.cc
```diff
@@ -590,84 +608,116 @@ std::ostream &PutLower(std::ostream &os, const std::string &str) {
...
+  auto header{std::string{ModHeader::bom} + ModHeader::magic +
+      CheckSum(contents) + ModHeader::terminator};
+  if (FileContentsMatch(path, header, contents)) {
+    return 0;
+  }
+  Temp temp{MkTemp(path)};
+  if (temp.fd < 0) {
+    return errno;
+  }
+  if (write(temp.fd, header.c_str(), header.size()) == -1 ||
```
klausler: It's better practice to check the return value against the byte count, so that short writes are caught as errors.

lib/semantics/mod-file.cc
```diff
@@ -590,84 +608,116 @@ std::ostream &PutLower(std::ostream &os, const std::string &str) {
...
-  if (!stream.get(c) || c != '\n') {
+  int fd{open(path.c_str(), O_RDONLY)};
+  if (fd < 0) {
     return false;
   }
-  for (std::size_t i{0}; i < contents.size(); ++i) {
-    if (!stream.get(c) || c != contents[i]) {
+  constexpr std::size_t bufSize{4096};
+  std::string buffer(bufSize, '\0');
+  if (read(fd, &buffer[0], hsize) == -1 ||
```
klausler: Beware of short reads.  It's better to compare the result of `read()` with the expected byte count.

lib/semantics/mod-file.cc
```diff
@@ -690,28 +740,33 @@ Scope *ModFileReader::Read(const SourceName &name, Scope *ancestor) {
...
-  }
   parser::Parsing parsing{context_.allSources()};
   parser::Options options;
   options.isModuleFile = true;
   options.features.Enable(parser::LanguageFeature::BackslashEscapes);
-  parsing.Prescan(*path, options);
+  options.searchDirectories = context_.searchDirectories();
+  auto path{ModFileName(name, ancestorName, context_.moduleFileSuffix())};
+  parsing.Prescan(path, options);
+  const auto *sourceFile{parsing.sourceFile()};
```
klausler: As mentioned above, if this is the only use of the new data member in `Parsing`, it would be a cleaner interface to return the pointer from `Prescan()` instead.

lib/semantics/mod-file.cc
```diff
@@ -590,84 +608,116 @@ std::ostream &PutLower(std::ostream &os, const std::string &str) {
...
+  }
+  for (auto remaining{csize};;) {
+    auto bytes{std::min(bufSize, remaining)};
+    auto got{read(fd, &buffer[0], bytes)};
+    if (got != static_cast<ssize_t>(bytes) ||
+        std::memcmp(&buffer[0], &contents[csize - remaining], bytes) != 0) {
+      close(fd);
       return false;
     }
+    if (got == 0 && remaining == 0) {
```
vjayathirtha-nv: Or just return right after `remaining` reaches 0

```
remaining -= bytes;
if (remaining == 0) {
      close(fd);
      return true;
}
```
It's not too different, just avoids one call to `read()`  and this if check gets simpler.

lib/semantics/mod-file.cc
tskeith: When writing we need to take into account the byte order mark but when reading using `Prescan` it is not included in the file contents that are returned. So there are two different header lengths to consider.

lib/semantics/mod-file.cc
tskeith: The file created by `mkstemps()` is 0600. These gymnastics are to set the permissions to 0666 modified by the current umask. The only way to query the umask is to set it to something, get the return value of that, and then set it back to what it was.

lib/semantics/mod-file.cc
tskeith: I think that will be wrong if `remaining` is 0 but there are still bytes in the file to be read.

vjayathirtha-nv: In that case, in the next iteration
`auto bytes = std::min(4096, 0)`
`auto got = read(fd, &buffer[0], 0)` 
Nothing with be read and read() will return 0.

vjayathirtha-nv: Probably needs an EOF() check to make sure `contents` is not just matching a prefix of the file.

lib/parser/parsing.h
tskeith: OK, I've changed `Prescan()`.

lib/semantics/mod-file.cc
tskeith: Yes. I've updated the comment on the function definition.

lib/semantics/mod-file.cc
tskeith: OK, done. `read()` as well.

lib/semantics/mod-file.cc
tskeith: Done

lib/semantics/mod-file.cc
tskeith: Right, but I don't know a better way to check for EOF than to do a read and get back 0. That's what will happen in the next iteration if we keep going when `remaining` is 0 but `got` is not.

PR #761, opened 2019-09-25
Fix clang build on MacOS

tskeith:
clang doesn't consider `std::strlen` of a constant to be a constexpr,
so just replace with the actual length.


lib/semantics/mod-file.cc
```diff
@@ -43,7 +43,7 @@ using namespace parser::literals;
 struct ModHeader {
   static constexpr const char *bom{"\xef\xbb\xbf"};
   static constexpr const char *magic{"!mod$ v1 sum:"};
-  static constexpr int magicLen{std::strlen(magic)};
+  static constexpr int magicLen{13};
```
klausler: Even better, declare `bom` and `magic` as arrays, use `sizeof`, and subtract one for the `NUL`.

PR #763, opened 2019-09-26
Add documentation/DriverAndFrontend.md

peterwaller-arm:
Add documentation describing the driver and frontend. I've also linked to it from the top of the overview, which seemed like a logical place to link to it from.

This is a work in progress and a request for comments.

I'm happy for this documentation to be merged as is, and we intend to update it on a regular basis.

I'm away on leave until 14th Oct. My colleagues may reply in the meantime.

See also: #762 ("Add flang -fc1").

### DIFF
--- a/documentation/DriverAndFrontend.md
+++ b/documentation/DriverAndFrontend.md
@@ -1,0 +1,215 @@
+The Flang driver and frontend option parsing
+============================================
+
+Status
+======
+
+As of September 2019, this work is draft.
+
+* Changes in libclangDriver to add --driver-mode=flang:
+  * RFC: http://lists.llvm.org/pipermail/cfe-dev/2019-June/062669.html.
+  * (In progress) Clang Revision: https://reviews.llvm.org/D63607
+  * (To be scheduled): Small changes to make pieces of libclangDriver
+    and libclangFrontend less clang-specific.
+
+* Introduction of new main() entrypoint bin/flang:
+  This is the part which chooses which binaries to invoke for the
+  frontend and linking.
+  * (In progress) at https://github.com/flang-compiler/f18/pull/759
+
+* Introduction of new fc1_main() entrypoint `bin/flang -fc1`
+  * (Done) Describing the vision and implementation in this document.
+  * (In progress) A straw-man implementation created by copying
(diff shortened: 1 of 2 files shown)


RichBarton-Arm (approved):
I think this is a nice overview of the driver project. I like the idea to create a minimum viable driver to replace the throwaway driver and re-use as much Clang code as we can without major surgery. I think the further refactoring to move generic code out of libClangDriver and investigate sharing DiagnosticHandlers and fs access classes can come along later and will be easier once the initial code is in place and once F18 makes it into LLVM.

I notice in your work in progress whittled statue in #762 that you have kept in all the diagnostics rendering code from Clang. On a conceptual level, it seems to me that having the new Flang frontend use existing Flang diagnostics code would be a more logical first step, so all of flang uses the same style of diagnostics. Obviously #762 is WIP, so is this a step you plan to take or would it be very hard to do this in the ex-clang frontend code and it would be easier to unify the two diagnostics handling code first?

CarolineConcatto:
Question:" On a conceptual level, it seems to me that having the new Flang frontend use existing Flang diagnostics code would be a more logical first step, so all of flang uses the same style of diagnostics." 
Reply: I believe we do not have ATM a class Flang Diagnostic implemented in Flang as it is in Clang. Most  errors and  messages that the compiler emits are with the method  Message &Say. As discussed in the future Flang would need a Diagnostic Client as the one we have in Clang.

Question:"Obviously #762 is WIP, so is this a step you plan to take or would it be very hard to do this in the ex-clang frontend code and it would be easier to unify the two diagnostics handling code first?"
Reply:Some progress of removing the dependency of Flang Driver Frontend from Clang Frontend  was made in the https://github.com/flang-compiler/f18/pull/759. We intend to do similar  for the https://github.com/flang-compiler/f18/pull/762 in the future.

CarolineConcatto:
@kiranchandramohan, @RichBarton-Arm , @peterwaller-arm and @DavidTruby 
I would also like to know what you think about creating another PR for this entry point. Lots of things have happened here, like rebase to lit, rebase to master, I did lots of push --force and so on. I really would like to start clean.

PR #767, opened 2019-09-30
Replace access through union with reinterpret_cast.

### LINKED ISSUE #747: Possible undefined behaviour
@klausler hi! :wave:.

I was going through the commits and noticed #666 / 34a917c761541d7325156baafab47966264f2a3c (clean up warnings) in this use of a union. I wonder if it is undefined behaviour according to the standard. I understand that you're only allowed to access the member of a union that was most recently assigned to.

In practice the compilers we currently support may not do anything which breaks this code, but I wanted to just query what the project's stance is on the use of undefined behaviour like this.

https://github.com/flang-compiler/f18/blob/2752d09c3a7f599c27de14fcef3846f9db39e97e/lib/evaluate/intrinsics-library-templates.h#L132-L146

The alternative seems to be to use `reinterpret_cast<>`, which I see the patch moved away from. The standard does appear to allow such conversion of function pointers, so long as "you put them back how you found them before you call them" from C++ standard [5.2.10 Reinterpret cast paragraph 6](http://eel.is/c++draft/expr.reinterpret.cast#6):

> A function pointer can be explicitly converted to a function pointer of a different type. The effect of calling a function through a pointer to a function type (8.3.5) that is not the same as the type used in the definition of the function is undefined. Except that converting a prvalue of type “pointer to T1” to the type “pointer to T2” (where T1 and T2 are function types) and back to its original type yields the original pointer value, the result of such a pointer conversion is unspecified.

/cc @DavidTruby

DavidTruby:
Fixes #747 
This avoids UB and doesn't cause warnings due to a special case in GCC for the type `void* (*) (void)`

### DIFF
--- a/lib/evaluate/intrinsics-library.h
+++ b/lib/evaluate/intrinsics-library.h
@@ -35,6 +35,7 @@ class FoldingContext;
 using TypeCode = unsigned char;
 
 template<typename TR, typename... TA> using FuncPointer = TR (*)(TA...);
+// This specific type signature prevents GCC complaining about function casts.
 using GenericFunctionPointer = void (*) (void);
 
 enum class PassBy { Ref, Val };
--- a/lib/evaluate/intrinsics-library.h
+++ b/lib/evaluate/intrinsics-library.h
@@ -35,7 +35,7 @@ class FoldingContext;
 using TypeCode = unsigned char;
 
 template<typename TR, typename... TA> using FuncPointer = TR (*)(TA...);
-using GenericFunctionPointer = FuncPointer<void *>;
+using GenericFunctionPointer = void (*) (void);
 
 enum class PassBy { Ref, Val };
 template<typename TA, PassBy Pass = PassBy::Ref> struct ArgumentInfo {
--- a/lib/evaluate/intrinsics-library-templates.h
+++ b/lib/evaluate/intrinsics-library-templates.h
@@ -129,23 +129,13 @@ template<typename TR, typename... ArgInfo> struct CallableHostWrapper {
 template<typename TR, typename... TA>
 inline GenericFunctionPointer ToGenericFunctionPointer(
     FuncPointer<TR, TA...> f) {
-  union {
-    GenericFunctionPointer gp;
-    FuncPointer<TR, TA...> fp;
-  } u;
-  u.fp = f;
-  return u.gp;
+  return reinterpret_cast<GenericFunctionPointer>(f);
 }
 
 template<typename TR, typename... TA>
 inline FuncPointer<TR, TA...> FromGenericFunctionPointer(
     GenericFunctionPointer g) {
-  union {
-    GenericFunctionPointer gp;
-    FuncPointer<TR, TA...> fp;
-  } u;
-  u.gp = g;
-  return u.fp;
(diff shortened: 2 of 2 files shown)


lib/evaluate/intrinsics-library.h
```diff
@@ -35,7 +35,7 @@ class FoldingContext;
 using TypeCode = unsigned char;
 
 template<typename TR, typename... TA> using FuncPointer = TR (*)(TA...);
-using GenericFunctionPointer = FuncPointer<void *>;
+using GenericFunctionPointer = void (*) (void);
```
klausler: Please add a comment to the effect that this exact type has magical properties for GCC.

DavidTruby:
> This seems to work with all of the various versions of compilers that I test with. I don't know whether replacing blatantly undefined behavior with a cast to a nonstandard special type is perfect solution for all time, but it's no worse and is somewhat more readable. Thanks!

I don't think the type is nonstandard, it's just a normal function type. I have no idea why it has magic behaviour in gcc though... I briefly looked at the warning and it seems like they specifically check for this type, _with_ `(void)` instead of `()` even though as far as I know the types are the same in C++. 
The only thing I can think is that it errors in the `()` case because those two types are different in C, but that seems like a bug on their end as they should be identical types in C++.
I agree it's not the cleanest solution in the world but I believe it's standards conformant and does make gcc be quiet. I think the "correct" answer in C++ is to use std::function but that has some overhead.
 I'll put a comment in about the magic type!

klausler:
How would `std::function<>` be used to implement a generic function pointer type for use in a name->function mapping?

DavidTruby:
I was under the mistaken belief that you could assign std::function<T> to std::function<void()> for any T and then get it back out with std::function<void>::target. 
It is possible to do this all with std::any and have it be safe and with no warnings, but the generated code is significantly worse than what we have here.

klausler:
> I was under the mistaken belief that you could assign std::function<T> to std::function<void()> for any T and then get it back out with std::function<void>::target.
> It is possible to do this all with std::any and have it be safe and with no warnings, but the generated code is significantly worse than what we have here.

`std::any` implies RTTI, yes?

DavidTruby:
> > I was under the mistaken belief that you could assign std::function to std::function<void()> for any T and then get it back out with std::function::target.
> > It is possible to do this all with std::any and have it be safe and with no warnings, but the generated code is significantly worse than what we have here.
> 
> `std::any` implies RTTI, yes?

It doesn't, they use a strange trick to avoid RTTI (basically it stores a pointer to a unique function that is generated to represent the type in the any, and when you ask for it back out it checks that the function representing the type you're asking for matches the one that's stored.)
I think std::variant works in the same way? Since that also has to know whether the type you ask for is the one stored.

klausler:
`variant` just stores an index.

I think that your PR constitutes an improvement over the code that is current present; but if you can make a generic name->`std::function<>` table work with entirely standard C++ code that conforms with our restrictions on exceptions and RTTI and works in all of our build environments, that might be even better.  I suggest that you merge this change and then replace it later if you come up with something.  If you do not have permission to merge this change, let me know.  Thank you.

PR #769, opened 2019-09-30
More semantic checking for DO constructs

psteinfeld:
This time I'm adding to the checks for constraint C1137, which states
that image control statements cannot appear in a DO CONCURRENT.  The
checks I added test to see if the DO CONCURRENT contains an ALLOCATE or
DEALLOCATE that references a coarray.

### DIFF
--- a/lib/semantics/check-do.cc
+++ b/lib/semantics/check-do.cc
@@ -24,6 +24,7 @@
 #include "../evaluate/tools.h"
 #include "../parser/message.h"
 #include "../parser/parse-tree-visitor.h"
+#include "../parser/tools.h"
 
 namespace Fortran::semantics {
 
@@ -126,6 +127,7 @@ public:
         "image control statement not allowed in DO CONCURRENT"_err_en_US);
   }
 
+  // more C1137 checks
   void Post(const parser::SyncAllStmt &) { NoImageControl(); }
   void Post(const parser::SyncImagesStmt &) { NoImageControl(); }
   void Post(const parser::SyncMemoryStmt &) { NoImageControl(); }
@@ -139,18 +141,14 @@ public:
   void Post(const parser::UnlockStmt &) { NoImageControl(); }
   void Post(const parser::StopStmt &) { NoImageControl(); }
 
-  void Post(const parser::AllocateStmt &) {
-    if (anyObjectIsCoarray()) {
-      context_.Say(currentStatementSourcePosition_,
-          "ALLOCATE coarray not allowed in DO CONCURRENT"_err_en_US);
-    }
+  // more C1137 checks
+  void Post(const parser::AllocateStmt &allocateStmt) {
+    CheckDoesntContainCoarray(allocateStmt);
   }
 
-  void Post(const parser::DeallocateStmt &) {
-    if (anyObjectIsCoarray()) {
-      context_.Say(currentStatementSourcePosition_,
-          "DEALLOCATE coarray not allowed in DO CONCURRENT"_err_en_US);
-    }
+  void Post(const parser::DeallocateStmt &deallocateStmt) {
+    CheckDoesntContainCoarray(deallocateStmt);  // C1137
+
     // C1140: deallocation of polymorphic objects
     if (anyObjectIsPolymorphic()) {
       context_.Say(currentStatementSourcePosition_,
@@ -219,6 +217,35 @@ public:
   }
 
 private:
+  // C1137 helper functions
+  void CheckAllocateObjectIsntCoarray(
+      const parser::AllocateObject &allocateObject, StmtType stmtType) {
+    const parser::Name &name{GetLastName(allocateObject)};
+    if (name.symbol && IsCoarray(*name.symbol)) {
+      context_.Say(name.source,
+          "%s coarray not allowed in DO CONCURRENT"_err_en_US,
+          EnumToString(stmtType));
+    }
+  }
+
+  void CheckDoesntContainCoarray(const parser::AllocateStmt &allocateStmt) {
+    const auto &allocationList{
+        std::get<std::list<parser::Allocation>>(allocateStmt.t)};
+    for (const auto &allocation : allocationList) {
+      const auto &allocateObject{
+          std::get<parser::AllocateObject>(allocation.t)};
+      CheckAllocateObjectIsntCoarray(allocateObject, StmtType::ALLOCATE);
+    }
+  }
+
+  void CheckDoesntContainCoarray(const parser::DeallocateStmt &deallocateStmt) {
+    const auto &allocateObjectList{
+        std::get<std::list<parser::AllocateObject>>(deallocateStmt.t)};
+    for (const auto &allocateObject : allocateObjectList) {
+      CheckAllocateObjectIsntCoarray(allocateObject, StmtType::DEALLOCATE);
+    }
+  }
+
   bool anyObjectIsCoarray() { return false; }  // FIXME placeholder
   bool anyObjectIsPolymorphic() { return false; }  // FIXME placeholder
   bool EndTDeallocatesCoarray() { return false; }  // FIXME placeholder
@@ -476,7 +503,7 @@ private:
     SymbolSet result;
     if (const auto *expr{GetExpr(expression)}) {
       for (const Symbol *symbol : evaluate::CollectSymbols(*expr)) {
-        if (const Symbol * root{GetAssociationRoot(*symbol)}) {
+        if (const Symbol * root{GetAssociationRoot(DEREF(symbol))}) {
           result.insert(root);
         }
       }
--- a/lib/semantics/check-do.h
+++ b/lib/semantics/check-do.h
@@ -26,9 +26,8 @@ struct ExitStmt;
 
 namespace Fortran::semantics {
 
-// To specify CYCLE and EXIT statements in semantic checking that's common to
-// both of them.
-ENUM_CLASS(StmtType, CYCLE, EXIT)
+// To specify different statement types used in semantic checking.
+ENUM_CLASS(StmtType, CYCLE, EXIT, ALLOCATE, DEALLOCATE)
 
 class DoChecker : public virtual BaseChecker {
 public:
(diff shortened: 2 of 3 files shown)


lib/semantics/check-do.cc
```diff
@@ -219,6 +214,46 @@ class DoConcurrentBodyEnforce {
...
+    if (IsCoarray(*symbol)) {
+      context_.Say(position,
+          "%s coarray not allowed in DO CONCURRENT"_err_en_US, stmtType);
+    }
+  }
+
+  void CheckAllocateContainsCoarray(const parser::AllocateStmt &allocateStmt) {
+    const std::list<parser::Allocation> &allocationList{
+        std::get<1>(allocateStmt.t)};
+    for (const parser::Allocation &allocation : allocationList) {
```
vjayathirtha-nv: you can use auto here and in the loop in `CheckDeallocateContainsCoarray()`

```suggestion
    for (auto &allocation : allocationList) {
```

lib/semantics/check-do.cc
```diff
@@ -139,18 +139,13 @@ class DoConcurrentBodyEnforce {
   void Post(const parser::UnlockStmt &) { NoImageControl(); }
   void Post(const parser::StopStmt &) { NoImageControl(); }
 
-  void Post(const parser::AllocateStmt &) {
-    if (anyObjectIsCoarray()) {
```
vjayathirtha-nv: `anyObjectIsCoarray()` was a place holder which was used in other places as well. Maybe all occurrences of this function should be replaced with whatever logic you inserted to check for Coarrays.

lib/semantics/check-do.cc
klausler: We like to be explicitly `const` though.

lib/semantics/check-do.cc
vjayathirtha-nv: It is used on line 175 of `check-do.cc` 
https://github.com/flang-compiler/f18/blob/fc53f3415ee67bf731287e44593fa51ef4fd82ec/lib/semantics/check-do.cc#L175

lib/semantics/check-do.cc
vjayathirtha-nv: Right, I've updated by suggestion.

psteinfeld: Varun's suggestion has a ```const``` in it.

I prefer not to use ```auto``` in this case because I think it's slightly more readable to explicitly state the type.  But in this case I could go either way since the type of ```allocation``` is pretty obvious.

lib/semantics/check-do.cc
psteinfeld: You're probably correct.  I'm working my way through all of the remaining semantic checks for DO constructs, and this change seemed pretty well contained, so I decided to submit a pull request now rather than adding more stuff.

lib/semantics/check-do.cc
```diff
@@ -219,6 +214,46 @@ class DoConcurrentBodyEnforce {
...
+            [&](const parser::Name &x) {
+              symbol = x.symbol;
+              position = x.source;
+            },
+            [&](const parser::StructureComponent &x) {
+              symbol = x.component.symbol;
+              position = x.component.source;
+            },
+        },
+        allocateObject.u);
```
jeanPerier: `parser::GetLastName` that returns a `parse::Name&` should do something similar here and could hide the details of `parser::AllocateObject` here as well as avoid the lambda definition.

psteinfeld: Thanks, Jean!  ```GetLastName()``` works like a champ.

lib/semantics/check-do.cc
```diff
@@ -219,6 +214,46 @@ class DoConcurrentBodyEnforce {
...
+    if (IsCoarray(*symbol)) {
+      context_.Say(position,
+          "%s coarray not allowed in DO CONCURRENT"_err_en_US, stmtType);
+    }
+  }
+
+  void CheckAllocateContainsCoarray(const parser::AllocateStmt &allocateStmt) {
+    const std::list<parser::Allocation> &allocationList{
+        std::get<1>(allocateStmt.t)};
```
tskeith: I think it's better to use `std::get` with a type rather than a number. It's shorter and avoids the magic number:
`const auto &allocationList{std::get<std::list<parser::Allocation>>};`

There are some in the next function too.

lib/semantics/check-do.cc
```diff
@@ -219,6 +214,46 @@ class DoConcurrentBodyEnforce {
...
+            },
+        },
+        allocateObject.u);
+    if (IsCoarray(*symbol)) {
+      context_.Say(position,
+          "%s coarray not allowed in DO CONCURRENT"_err_en_US, stmtType);
+    }
+  }
+
+  void CheckAllocateContainsCoarray(const parser::AllocateStmt &allocateStmt) {
```
tskeith: The name of this function suggests it's doing the opposite of what it actually does. Maybe `CheckDoesntContainCoarray()`?

lib/semantics/check-do.cc
```diff
@@ -219,6 +214,46 @@ class DoConcurrentBodyEnforce {
   }
 
 private:
+  void CheckAllocateObjectIsCoarray(
+      const parser::AllocateObject &allocateObject, const char *stmtType) {
```
tskeith: I think it would be better if `stmtType` were an enumeration, like with `CYCLE` and `EXIT`. It could even be the same type.

lib/semantics/check-do.cc
psteinfeld: Also, I changed the test to use nested types to exercise this.

lib/semantics/check-do.cc
psteinfeld: Thanks, Tim.  I made these changes.

lib/semantics/check-do.cc
psteinfeld: Thanks.  I've changed the names.

lib/semantics/check-do.cc
psteinfeld: Good idea.  I've made the associated changes.

lib/semantics/check-do.h
```diff
@@ -28,7 +28,7 @@ namespace Fortran::semantics {
 
 // To specify CYCLE and EXIT statements in semantic checking that's common to
 // both of them.
```
jeanPerier: Comment is outdated (should be broader than `CYCLE` and `EXIT` now)

lib/semantics/check-do.cc
```diff
@@ -139,18 +140,13 @@ class DoConcurrentBodyEnforce {
   void Post(const parser::UnlockStmt &) { NoImageControl(); }
   void Post(const parser::StopStmt &) { NoImageControl(); }
 
-  void Post(const parser::AllocateStmt &) {
-    if (anyObjectIsCoarray()) {
-      context_.Say(currentStatementSourcePosition_,
-          "ALLOCATE coarray not allowed in DO CONCURRENT"_err_en_US);
-    }
+  void Post(const parser::AllocateStmt &allocateStmt) {
+    CheckDoesntContainCoarray(allocateStmt);
```
jeanPerier: Maybe you could actually add a reference to C1137 here and in `Post(const parser::DeallocateStmt &)`. It is rather clear, but I guess it does not hurt adding it. Sorry for not commenting on this during the first review.

lib/semantics/check-do.cc
```diff
@@ -219,6 +215,34 @@ class DoConcurrentBodyEnforce {
...
+    const auto &allocateObjectList{
+        std::get<std::list<parser::AllocateObject>>(deallocateStmt.t)};
+    for (const auto &allocateObject : allocateObjectList) {
+      CheckAllocateObjectIsntCoarray(allocateObject, StmtType::DEALLOCATE);
+    }
+  }
+
   bool anyObjectIsCoarray() { return false; }  // FIXME placeholder
   bool anyObjectIsPolymorphic() { return false; }  // FIXME placeholder
   bool EndTDeallocatesCoarray() { return false; }  // FIXME placeholder
```
klausler: Are these now dead?

lib/semantics/check-do.h
psteinfeld: Thanks for catching this.  I'll update the comment.

lib/semantics/check-do.cc
psteinfeld: Good idea.  I'll add the references.

lib/semantics/check-do.cc
psteinfeld: Probably.  I was planning to complete all of the semantic checking for DO constructs and then clean things up.  Some or all of these functions will probably disappear at that time.

lib/semantics/check-do.cc
```diff
@@ -511,8 +500,10 @@ class DoContext {
     SymbolSet result;
     if (const auto *expr{GetExpr(expression)}) {
       for (const Symbol *symbol : evaluate::CollectSymbols(*expr)) {
-        if (const Symbol * root{GetAssociationRoot(*symbol)}) {
-          result.insert(root);
+        if (symbol) {
+          if (const Symbol * root{GetAssociationRoot(*symbol)}) {
```
tskeith: I believe that the set returned by `evaluate::CollectSymbols()` should never contain null pointers, so the check is not necessary. You could assert that's the case by doing `GetAssociateRoot(DEREF(symbol))`.

lib/semantics/check-do.cc
```diff
@@ -523,7 +514,7 @@ class DoContext {
   void CheckMaskIsPure(const parser::ScalarLogicalExpr &mask) const {
     SymbolSet references{GatherSymbolsFromExpression(mask.thing.thing.value())};
     for (const Symbol *ref : references) {
-      if (IsProcedure(*ref) && !IsPureProcedure(*ref)) {
+      if (ref && IsProcedure(*ref) && !IsPureProcedure(*ref)) {
```
tskeith: Same comment here. `ref` can't be null.

lib/semantics/check-do.cc
psteinfeld: Thanks, Tim.  Using ```DEREF()``` gives me a little peace of mind, since ```CollectSymbols()``` is implemented elsewhere.

lib/semantics/check-do.cc
psteinfeld: Yes, and in the case, it's easy to verify.

PR #772, opened 2019-10-02
Make the second argument to dist_schedule optional.

### LINKED ISSUE #770: [OpenMP] dist_schedule chunk_size should be optional
The following code fails to parse: `!$omp distribute dist_schedule(static)` when this should be allowed. `!$omp distribute dist_schedule(static, 2)` for example works fine. 
The chunk_size should be optional according to section 2.10.8 of the OpenMP grammar:
```2.10.8 dist-schedule-clause -> DIST_SCHEDULE (STATIC [ , chunk_size])```

DavidTruby:
Fixes #770


tskeith:
I think you need a change to `unparse.cc` too, otherwise this will come out as `DIST_SCHEDULE(STATIC,)`.

PR #774, opened 2019-10-02
Enable clang warnings used by LLVM

klausler:
Enable `-Wstring-conversion` and `-Wcovered-switch-default` warnings when building with clang.  Patch sources to dodge the warnings.

### DIFF
--- a/lib/common/idioms.h
+++ b/lib/common/idioms.h
@@ -76,7 +76,7 @@ template<typename... LAMBDAS> visitors(LAMBDAS... x)->visitors<LAMBDAS...>;
 
 // For switch statements whose cases have return statements for
 // all possibilities.  Clang emits warnings if the default: is
-// present, gcc emits warings if it is absent.
+// present, gcc emits warnings if it is absent.
 #if __clang__
 #define SWITCH_COVERS_ALL_CASES
 #else
--- a/lib/common/idioms.h
+++ b/lib/common/idioms.h
@@ -80,9 +80,8 @@ template<typename... LAMBDAS> visitors(LAMBDAS... x)->visitors<LAMBDAS...>;
 #if __clang__
 #define SWITCH_COVERS_ALL_CASES
 #else
-#define SWITCH_COVERS_ALL_CASES \
-  default: CRASH_NO_CASE; \
-           #endif
+#define SWITCH_COVERS_ALL_CASES default:;
+#endif
 
 // For cheap assertions that should be applied in production.
 // To disable, compile with '-DCHECK=(void)'
--- a/lib/common/idioms.h
+++ b/lib/common/idioms.h
@@ -74,6 +74,16 @@ template<typename... LAMBDAS> visitors(LAMBDAS... x)->visitors<LAMBDAS...>;
 // For switch statements without default: labels.
 #define CRASH_NO_CASE DIE("no case")
 
+// For switch statements whose cases have return statements for
+// all possibilities.  Clang emits warnings if the default: is
+// present, gcc emits warings if it is absent.
+#if __clang__
+#define SWITCH_COVERS_ALL_CASES
+#else
+#define SWITCH_COVERS_ALL_CASES \
+  default: CRASH_NO_CASE; \
+           #endif
+
 // For cheap assertions that should be applied in production.
 // To disable, compile with '-DCHECK=(void)'
 #ifndef CHECK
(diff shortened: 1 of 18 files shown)


lib/common/idioms.h
```diff
@@ -74,6 +74,16 @@ template<typename... LAMBDAS> visitors(LAMBDAS... x)->visitors<LAMBDAS...>;
 // For switch statements without default: labels.
 #define CRASH_NO_CASE DIE("no case")
 
+// For switch statements whose cases have return statements for
+// all possibilities.  Clang emits warnings if the default: is
+// present, gcc emits warings if it is absent.
```
tskeith: warings -> warnings

lib/common/idioms.h
```diff
@@ -74,6 +74,16 @@ template<typename... LAMBDAS> visitors(LAMBDAS... x)->visitors<LAMBDAS...>;
...
+#if __clang__
+#define SWITCH_COVERS_ALL_CASES
+#else
+#define SWITCH_COVERS_ALL_CASES \
+  default: CRASH_NO_CASE; \
+           #endif
```
tskeith: The backslash on the previous line shouldn't be there.

lib/parser/characters.cc
```diff
@@ -122,8 +122,9 @@ template<> EncodedCharacter EncodeCharacter<Encoding::UTF_8>(char32_t ucs) {
 EncodedCharacter EncodeCharacter(Encoding encoding, char32_t ucs) {
   switch (encoding) {
   case Encoding::LATIN_1: return EncodeCharacter<Encoding::LATIN_1>(ucs);
-  case Encoding::UTF_8: return EncodeCharacter<Encoding::UTF_8>(ucs);
-  default: CRASH_NO_CASE;
+  case Encoding::UTF_8:
+    return EncodeCharacter<Encoding::UTF_8>(ucs);
+    SWITCH_COVERS_ALL_CASES
```
tskeith: clang-format treates `SWITCH_COVERS_ALL_CASES` like it is part of the last case.
What do you think of putting it between the `switch` and the first `case` to make it clear it is separate?
```
  switch (encoding) {
    SWITCH_COVERS_ALL_CASES
  case Encoding::LATIN_1: return EncodeCharacter<Encoding::LATIN_1>(ucs);
  case Encoding::UTF_8: return EncodeCharacter<Encoding::UTF_8>(ucs);
  }
```

lib/parser/prescan.cc
```diff
@@ -58,7 +58,7 @@ static void NormalizeCompilerDirectiveCommentMarker(TokenSequence &dir) {
       return;
     }
   }
-  CHECK(!"compiler directive all blank");
+  CHECK(false && "compiler directive all blank");
```
tskeith: Could use `DIE` instead.

lib/semantics/resolve-names.cc
```diff
@@ -4824,8 +4824,7 @@ const DeclTypeSpec &ConstructVisitor::ToDeclTypeSpec(
 
       );
     }
-  case common::TypeCategory::Character:
-  default: CRASH_NO_CASE;
+  case common::TypeCategory::Character: CRASH_NO_CASE; SWITCH_COVERS_ALL_CASES
```
tskeith: The clang-format problem is even worse here.

lib/common/idioms.h
```diff
@@ -74,6 +74,15 @@ template<typename... LAMBDAS> visitors(LAMBDAS... x)->visitors<LAMBDAS...>;
 // For switch statements without default: labels.
 #define CRASH_NO_CASE DIE("no case")
 
+// For switch statements whose cases have return statements for
+// all possibilities.  Clang emits warnings if the default: is
+// present, gcc emits warnings if it is absent.
+#if __clang__
```
sscalpone: Please update the style guide.

lib/semantics/check-omp-structure.cc
```diff
@@ -585,9 +585,6 @@ void OmpStructureChecker::Enter(
     OmpClauseSet allowed{OmpClause::DEPEND};
     SetContextAllowed(allowed);
   } break;
-  default:
-    // TODO others
-    break;
```
DavidTruby: This switch statement should also have `SWITCH_COVERS_ALL_CASES` as per the updated style guide

RichBarton-Arm: @klausler looks like @DavidTruby's comment fell through the cracks. @DavidTruby - can you submit a separate PR for it?

PR #775, opened 2019-10-02
Fix #773: ambiguous REALA=3 should be an assignment, not a declaration

### LINKED ISSUE #773: using keywords as part of a variable name, fixed form
This is a low priority bug, probably in the scanner or parser for fixed form input:
$ cat funny2.f
----
	program foo
	real a = 3
	write(6,*) "reala is ", real a 
	write(6,*) "a is ", a 
	stop
	end
----
gfortran assigns "3" to the variable named "reala" as is should, but f18 assigns "3" to "a".
-----
~/test$ gfortran funny2.f
~/test$ ./a.out
 reala is    3.00000000    
 a is    4.21369473E+33
------
~/test$ f18 funny2.f 
pgf90 -Mbackslash -c -o funny2.o /tmp/f18-1818a.f90
pgf90 -Mbackslash funny2.o
~/test$ ./a.out
 reala is    8431095.00    
 a is    3.00000000

klausler:
When parsing a type-declaration-stmt without doubled colons, don't allow an initializer (which is prohibited by C806).  This resolves the ambiguous `REALA=3` in favor of an assignment statement.

### DIFF
--- a/lib/parser/grammar.h
+++ b/lib/parser/grammar.h
@@ -751,6 +751,7 @@ TYPE_PARSER(recovery(
 // R737 data-component-def-stmt ->
 //        declaration-type-spec [[, component-attr-spec-list] ::]
 //        component-decl-list
+// N.B. The standard requires double colons if there's an initializer.
 TYPE_PARSER(construct<DataComponentDefStmt>(declarationTypeSpec,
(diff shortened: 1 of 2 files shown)


sscalpone:
I prefer the title to describe the change instead of the side effect (fixing 775).

PR #776, opened 2019-10-04
More procedure reference semantics checking (mostly arguments)

klausler:
Implement checks of effective argument number, types, and ranks against the requirements of explicit and implicit interfaces.  Handles calls to external subprograms for which interfaces are known but not explicit gracefully with warnings.  Implements "same derived type" test for type compatibility (with some remaining TODO iterms).  Implements insertion of `PASS` arguments into argument lists, and the rearrangement of argument lists with keywords.  Implements a predicate for simple contiguity -- this will be used more in later work.

More argument checks are coming, but this branch was accumulating too many changes and is stable enough for review now.

Tested against f90_correct suite (and found some errors).

### DIFF
--- a/documentation/Preprocessing.md
+++ b/documentation/Preprocessing.md
@@ -14,6 +14,8 @@ Behavior common to (nearly) all compilers:
 * Fixed form right margin clipping does not apply to directive lines.
 * Macro names are not recognized as such when spaces are inserted
   into their invocations in fixed form.
+  This includes spaces at the ends of lines that have been clipped
+  at column 72 (or whatever).
 * Text is rescanned after expansion of macros and arguments.
 * Macros are not expanded within quoted character literals or
   quoted FORMAT edit descriptors.
@@ -74,6 +76,9 @@ Judgement calls, where precedents are unclear:
 * IBM claims to be ISO C compliant and therefore recognizes trigraph sequences.
 * Fortran comments in macro actual arguments should be respected, on
   the principle that a macro call should work like a function reference.
+* If a `#define` or `#undef` directive appears among continuation
+  lines, it may or may not affect text in the continued statement that
+  appeared before the directive.
 
 Behavior that few compilers properly support (or none), but should:
 -------------------------------------------------------------------
@@ -98,7 +103,7 @@ In short, a Fortran preprocessor should work as if:
 6. Other preprocessing directives are processed and macros expanded.
    Along the way, Fortran `INCLUDE` lines and preprocessor `#include` directives
    are expanded, and all these steps applied recursively to the introduced text.
-7. Any newly-created Fortran comments are removed.
+7. Any Fortran comments created by macro replacement are removed.
 
 Steps 5 and 6 are interleaved with respect to the preprocessing state.
 Conditional compilation preprocessing directives always reflect only the macro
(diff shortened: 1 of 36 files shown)


lib/semantics/mod-file.cc
```diff
@@ -41,10 +41,10 @@ using namespace parser::literals;
 // that the module file is decoded as UTF-8 even if source files
 // are using another encoding.
 struct ModHeader {
-  static constexpr const char *bom{"\xef\xbb\xbf"};
-  static constexpr const char *magic{"!mod$ v1 sum:"};
+  static constexpr const char bom[3 + 1]{"\xef\xbb\xbf"};
   static constexpr int magicLen{13};
-  static constexpr int sumLen{16};
+  static constexpr int sumLen{magicLen + 3};
```
tskeith: `sumLen` is not related to `magicLen`, it's just the number of characters in the checksum.

klausler: Sure; thanks.

documentation/Preprocessing.md
```diff
@@ -115,3 +120,100 @@ text.
...
+
+```
+f18
+| pgfortran
+| | ifort
+| | | gfortran
+| | | | xlf
+| | | | | nagfor
+| | | | | |
+. . . . . .   pp001.F  keyword macros
```
tskeith: Are the tests going to be added somewhere? Otherwise the test names aren't meaningful.

klausler: Right.  I'll slap copyright notices on them and push them out in another PR.

lib/evaluate/type.cc
```diff
@@ -109,42 +109,178 @@ bool DynamicType::IsAssumedLengthCharacter() const {
...
+    const semantics::Symbol &y,
+    SetOfDerivedTypePairs & /* inProgress - not yet used */) {
+  if (x.attrs() != y.attrs()) {
+    return false;
+  }
+  if (x.attrs().test(semantics::Attr::PRIVATE)) {
+    return false;
+  }
+  if (const auto *xObject{x.detailsIf<semantics::ObjectEntityDetails>()}) {
+    if (const auto *yObject{y.detailsIf<semantics::ObjectEntityDetails>()}) {
```
tskeith: `xObject` and `yObject` are not used so this will get a (fatal) warning.

lib/evaluate/characteristics.h
```diff
@@ -198,9 +235,16 @@ struct Procedure {
   bool operator==(const Procedure &) const;
 
   // Characterizes the procedure represented by a symbol, which may be an
-  // "unrestricted specific intrinsic function".
+  // "unrestricted specific intrinsic function".  The Boolean argument
+  // (when true) is used to treat an external function or subroutine
+  // in the same source file as if it did not have a known explicit interface
+  // when it is being called from a distinct program unit.
```
tskeith: What boolean argument?

klausler: The one that used to be there but no longer is.  I'll fix the comment.

test/semantics/call13.f90
```diff
@@ -0,0 +1,49 @@
...
+  class(*) :: classStar
+  type(*) :: typeStar
+
+  type :: pdt(len)
+    integer, len :: len
+  end type
+  type(pdt(1)) :: pdtx
+
+  !ERROR: The expression (int(implicit01(),kind=8)) cannot be used as a specification expression (reference to impure function 'implicit01')
+  real :: array(implicit01())  ! 15.4.2.2(2)
```
tskeith: I'm not sure it's worth including the expression in the error message: it might be long and (as in this case) it might not look like the original expression. As long as the source location is pointing at the expression, I think a message like this is better:
`Invalid specification expression: reference to impure function 'implicit01'`

lib/semantics/expression.cc
```diff
@@ -1437,9 +1438,21 @@ MaybeExpr ExpressionAnalyzer::Analyze(
...
-std::optional<ProcedureDesignator>
-ExpressionAnalyzer::AnalyzeProcedureComponentRef(
-    const parser::ProcComponentRef &pcr) {
+static const semantics::WithPassArg *GetPassInfo(
+    const semantics::Symbol &symbol) {
+  if (const auto *binding{symbol.detailsIf<semantics::ProcBindingDetails>()}) {
+    return static_cast<const semantics::WithPassArg *>(binding);
+  } else if (const auto *proc{
+                 symbol.detailsIf<semantics::ProcEntityDetails>()}) {
+    return static_cast<const semantics::WithPassArg *>(proc);
```
tskeith: This cast and the one above aren't necessary, are they?

lib/semantics/expression.h
```diff
@@ -311,24 +311,31 @@ class ExpressionAnalyzer {
...
+      const parser::ProcedureDesignator &, ActualArguments &&,
+      bool isSubroutine);
+
+  void CheckForBadRecursion(const semantics::Symbol &, parser::CharBlock);
   bool EnforceTypeConstraint(parser::CharBlock, const MaybeExpr &, TypeCategory,
       bool defaultKind = false);
-  MaybeExpr MakeFunctionRef(ProcedureDesignator &&, ActualArguments &&);
-  MaybeExpr MakeFunctionRef(CalleeAndArguments &&);
+  MaybeExpr MakeFunctionRef(ProcedureDesignator &&, ActualArguments &&,
+      parser::CharBlock callSite = {});
```
tskeith: It bugs me that `MakeFunctionRef` and `CheckForBadRecursion` have the `CharBlock` specifying the location at the end of the arguments whereas `CheckCall` and `EnforceTypeConstraint` have it at the beginning.

lib/evaluate/check-call.h
```diff
@@ -0,0 +1,35 @@
...
+
+// Constraint checking for procedure references
+
+#ifndef FORTRAN_EVALUATE_CHECK_CALL_H_
+#define FORTRAN_EVALUATE_CHECK_CALL_H_
+
+#include "call.h"
+#include "characteristics.h"
+#include "expression.h"
+#include <optional>
```
tskeith: Not needed?

lib/evaluate/check-call.cc
```diff
@@ -0,0 +1,199 @@
...
+
+static void RearrangeArguments(const characteristics::Procedure &proc,
+    ActualArguments &arg, parser::ContextualMessages &messages) {
+  CHECK(proc.HasExplicitInterface());
+  if (arg.size() < proc.dummyArguments.size()) {
+    arg.resize(proc.dummyArguments.size());
+  } else if (arg.size() > proc.dummyArguments.size()) {
+    messages.Say(
+        "%zd effective argument%s are passed in a reference to a procedure that expects only %zd"_err_en_US,
+        arg.size(), arg.size() == 1 ? "" : "s", proc.dummyArguments.size());
```
tskeith: A little suspect from the point of view of translatability.

lib/evaluate/check-call.cc
```diff
@@ -0,0 +1,199 @@
...
+          },
+          [&](const characteristics::AlternateReturn &) {
+            // TODO check alternate return
+          },
+      },
+      dummy.u);
+}
+
+static void RearrangeArguments(const characteristics::Procedure &proc,
+    ActualArguments &arg, parser::ContextualMessages &messages) {
```
tskeith: `args` would be a better name than `arg`.

lib/evaluate/check-call.h
```diff
@@ -0,0 +1,35 @@
...
+
+namespace Fortran::parser {
+class ContextualMessages;
+}
+
+namespace Fortran::evaluate {
+class FoldingContext;
+
+void CheckArguments(const characteristics::Procedure &, ActualArguments &,
+    FoldingContext &, bool treatingExternalAsImplicit = false);
```
tskeith: I think a comment explaining `treatingExternalAsImplicit` would be helpful.

lib/evaluate/check-call.cc
```diff
@@ -0,0 +1,199 @@
...
+void CheckArguments(const characteristics::Procedure &proc,
+    ActualArguments &arg, FoldingContext &context,
+    bool treatingExternalAsImplicit) {
+  parser::Messages buffer;
+  parser::ContextualMessages messages{context.messages().at(), &buffer};
+  FoldingContext localContext{context, messages};
+  if (proc.HasExplicitInterface()) {
+    RearrangeArguments(proc, arg, messages);
+    int index{0};
+    for (auto &x : arg) {
```
tskeith: As above, I don't think `arg` is a good name for a collection of arguments.
Given the presence of `dummy` in this function, renaming `x` to `actual` and `arg` to `actuals` or `actualArguments` would make it easier to understand.

lib/evaluate/check-expression.cc
```diff
@@ -237,4 +261,93 @@ template void CheckSpecificationExpr(const std::optional<Expr<SomeInteger>> &,
 template void CheckSpecificationExpr(
     const std::optional<Expr<SubscriptInteger>> &, parser::ContextualMessages &,
     const semantics::Scope &);
+
+// IsSimplyContiguous() -- 9.5.4
+class IsSimplyContiguousHelper
+  : public AnyTraverse<IsSimplyContiguousHelper, std::optional<bool>> {
+public:
+  using Result = std::optional<bool>;  // tri-state
```
tskeith: Why can't `Result` be just `bool`? When is does it have the value`std::nullopt`?

test/semantics/call13.f90
klausler: Will remove the expression.

lib/semantics/expression.h
klausler: Which one do you prefer?

lib/semantics/expression.cc
klausler: I wanted to be explicit about casting to a base class.  Will remove.

lib/evaluate/check-call.h
klausler: Replaced with forward references to classes.

lib/evaluate/check-call.cc
klausler: Ok.

lib/semantics/expression.h
tskeith: Most functions that have a `CharBlock` argument that represents a source location (e.g. the constructor for `Message`) has it first, so that would be my preference.

lib/evaluate/check-expression.cc
klausler: It's a tri-state value: false, true, or unknown.

lib/evaluate/check-call.cc
klausler: Renaming both to `actuals`.

lib/evaluate/check-call.cc
klausler: Yes, will reword.

lib/semantics/expression.h
klausler: Ok.  I'll implement the default value with an additional overload.

lib/evaluate/characteristics.h
```diff
@@ -89,13 +102,34 @@ class TypeAndShape {
 
 private:
   void AcquireShape(const semantics::ObjectEntityDetails &);
+  void AcquireLEN();
 
 protected:
   DynamicType type_;
+  std::optional<Expr<SomeInteger>> LEN_;
```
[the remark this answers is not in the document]
klausler: A `DynamicType` instance can't contain an arbitrary `CHARACTER` length expression, but `TypeAndShape` can.  This allows `TypeAndShape` to fully characterize `CHARACTER` expressions as well as symbols, so that actual arguments as in e.g. `CALL FOO(X//Y)` can be represented.

lib/evaluate/characteristics.h
```diff
@@ -89,13 +102,34 @@ class TypeAndShape {
...
+  Attrs attrs_;
 };
 
+template<typename T>
+std::optional<TypeAndShape> GetTypeAndShape(
+    const Expr<T> &expr, FoldingContext &context) {
+  if (auto type{expr.GetType()}) {
+    if (auto shape{GetShape(context, expr)}) {
+      TypeAndShape result{*type, std::move(*shape)};
+      if (type->category() == TypeCategory::Character) {
```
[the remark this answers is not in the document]
klausler: No, it is not redundant.  In this case, the `CHARACTER` length must be extracted from analysis of an expression.

PR #778, opened 2019-10-09
Generic name resolution in expression analysis

tskeith:
Implement the basics of resolving generic names in expressions.

`ExpressionAnalyzer::ResolveGeneric` maps the symbol for a generic
name to the specific procedure appropriate for the actual arguments.
Extract `CheckExplicitInterface` out of `CheckArguments` so that it
can be tried for each specific procedure of the generic as part of
the test to see which is compatible.

Note that it may be there is an elemental and non-elemental specific
procedure that is compatible with the actual arguments. In that case
the generic is resolved to the non-elemental one.

Test this by using generic functions in specification expressions
that must be written to module files. Verify how the generics were
resolved by looking at the generated `.mod` files.

There is more work to be done in this area: the passed-object dummy
argument is not considered and in some cases generated module files
are not correct.

### DIFF
--- a/lib/evaluate/check-call.cc
+++ b/lib/evaluate/check-call.cc
@@ -72,7 +72,7 @@ static void CheckImplicitInterfaceArg(
   }
 }
 
-static void CheckExplicitInterfaceArg(const ActualArgument &arg,
+static bool CheckExplicitInterfaceArg(const ActualArgument &arg,
     const characteristics::DummyArgument &dummy, FoldingContext &context) {
   std::visit(
       common::visitors{
@@ -95,9 +95,10 @@ static void CheckExplicitInterfaceArg(const ActualArgument &arg,
           },
       },
       dummy.u);
+  return true;  // TODO: return false when error detected
 }
 
-static void RearrangeArguments(const characteristics::Procedure &proc,
+static bool RearrangeArguments(const characteristics::Procedure &proc,
     ActualArguments &actuals, parser::ContextualMessages &messages) {
   CHECK(proc.HasExplicitInterface());
   if (actuals.size() < proc.dummyArguments.size()) {
@@ -106,6 +107,7 @@ static void RearrangeArguments(const characteristics::Procedure &proc,
     messages.Say(
         "Too many actual arguments (%zd) passed to procedure that expects only %zd"_err_en_US,
         actuals.size(), proc.dummyArguments.size());
+    return false;
   }
   std::map<std::string, ActualArgument> kwArgs;
   for (auto &x : actuals) {
@@ -117,6 +119,7 @@ static void RearrangeArguments(const characteristics::Procedure &proc,
           messages.Say(*x->keyword,
               "Argument keyword '%s=' appears on more than one effective argument in this procedure reference"_err_en_US,
               *x->keyword);
+          return false;
         }
         x.reset();
       }
@@ -133,6 +136,7 @@ static void RearrangeArguments(const characteristics::Procedure &proc,
             messages.Say(*x.keyword,
                 "Keyword argument '%s=' has already been specified positionally (#%d) in this procedure reference"_err_en_US,
                 *x.keyword, index + 1);
+            return false;
           } else {
             actuals[index] = std::move(x);
           }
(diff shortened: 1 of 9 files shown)


lib/evaluate/check-call.cc
```diff
@@ -146,42 +149,53 @@ static void RearrangeArguments(const characteristics::Procedure &proc,
...
+bool CheckExplicitInterface(const characteristics::Procedure &proc,
+    ActualArguments &actuals, FoldingContext &context) {
+  if (!RearrangeArguments(proc, actuals, context.messages())) {
+    return false;
+  }
+  int index{0};
+  for (auto &actual : actuals) {
+    const auto &dummy{proc.dummyArguments[index++]};
+    if (actual.has_value()) {
+      CheckExplicitInterfaceArg(*actual, dummy, context);
```
klausler: `CheckExplicitInterfaceArg` should probably now return a Boolean too, so that you can `return false;` when it encounters an error.

lib/semantics/expression.cc
```diff
@@ -1514,17 +1514,97 @@ auto ExpressionAnalyzer::AnalyzeProcedureComponentRef(
...
+      parser::Messages buffer;
+      parser::ContextualMessages messages{
+          context_.foldingContext().messages().at(), &buffer};
+      FoldingContext localContext{context_.foldingContext(), messages};
+      ActualArguments localActuals{actuals};
+      if (CheckExplicitInterface(*procedure, localActuals, localContext) &&
+          CheckCompatibleArguments(*procedure, localActuals)) {
+        auto *ptr{procedure->IsElemental() ? &result.elemental
+                                           : &result.nonElemental};
+        CHECK(*ptr == nullptr);
```
klausler: In an error recovery situation where a generic interface has non-distinguishable specific procedures, this `CHECK` could fail.

lib/semantics/expression.cc
```diff
@@ -1699,10 +1785,9 @@ std::optional<ActualArguments> ExpressionAnalyzer::AnalyzeArguments(
 static bool IsExternalCalledImplicitly(
     parser::CharBlock callSite, const ProcedureDesignator &proc) {
   if (const auto *symbol{proc.GetSymbol()}) {
-    return !callSite.empty() && symbol->has<semantics::SubprogramDetails>() &&
-        (symbol->owner().IsGlobal() ||
-            (symbol->owner().parent().IsGlobal() &&
-                !symbol->owner().sourceRange().Contains(callSite)));
+    return symbol->has<semantics::SubprogramDetails>() &&
+        symbol->owner().IsGlobal() &&
+        !symbol->scope()->sourceRange().Contains(callSite);
```
tskeith: I'm not positive this condition is correct. This change fixes some incorrect warnings I was getting when a generic resolved to a specific in another module. `symbol->owner()` was the module and it did not contain the call site.

lib/evaluate/check-call.cc
tskeith: OK

lib/semantics/expression.cc
tskeith: I could probably check for the error flag on the symbol to avoid this.
But the logic would be simpler if I don't bother with this consistency check.
Then I can return as soon as a compatible non-elemental specific is found.

lib/semantics/expression.cc
```diff
@@ -1514,17 +1514,97 @@ auto ExpressionAnalyzer::AnalyzeProcedureComponentRef(
...
+    if (std::optional<characteristics::Procedure> procedure{
+            characteristics::Procedure::Characterize(
+                ProcedureDesignator{*specific}, context_.intrinsics())}) {
+      parser::Messages buffer;
+      parser::ContextualMessages messages{
+          context_.foldingContext().messages().at(), &buffer};
+      FoldingContext localContext{context_.foldingContext(), messages};
+      ActualArguments localActuals{actuals};
+      if (CheckExplicitInterface(*procedure, localActuals, localContext) &&
+          CheckCompatibleArguments(*procedure, localActuals)) {
```
jeanPerier: It is not clear to me why `CheckExplicitInterface` does not imply `CheckCompatibleArguments`.  `CheckExplicitInterfaceArg` and `CheckCompatibleArgument` look very similar. Obviously, if you are doing it, there is a reason. I see that `CheckCompatibleArguments` seems to deal with elemental aspect when `CheckExplicitInterface` does not seem to do so.
Could you explain a bit more what is the difference between these two checks ?

tskeith: The goal here is just to narrow down the specific procedures to the right one, thought it still might not perfectly match the actual arguments. This is just an initial implementation -- it may take more work to get the best error messages from bad calls to generics procedures.

lib/semantics/expression.cc
```diff
@@ -1514,17 +1514,97 @@ auto ExpressionAnalyzer::AnalyzeProcedureComponentRef(
...
+static bool CheckCompatibleArguments(
+    const characteristics::Procedure &procedure,
+    const ActualArguments &actuals) {
+  bool isElemental{procedure.IsElemental()};
+  const auto &dummies{procedure.dummyArguments};
+  CHECK(dummies.size() == actuals.size());
+  for (std::size_t i{0}; i < dummies.size(); ++i) {
+    const characteristics::DummyArgument &dummy{dummies[i]};
+    const std::optional<ActualArgument> &actual{actuals[i]};
+    if (!actual || !CheckCompatibleArgument(isElemental, *actual, dummy)) {
```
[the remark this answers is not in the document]
tskeith: Yes, you're right, the check for actual was backwards. I'll fix this and add a test for this case.

PR #780, opened 2019-10-11
Changes to disallow image control statements in DO CONCURRENT

psteinfeld:
Most of these changes were already implemented.  The last remaining part was to check for calls to move_alloc with coarray arguments.  This set of changes implements that.  I also bundled other changes.  Specifically:

All of the code to detect image control statements was moved from check-do.cc to tools.cc so that it could be used by other semantic checking functions.

I added location information to the error messages for all DO semantics checks to highlight either the DO loop associated with the error or other relevant source locations.

I cleaned up the error messages associated with DO semantics so that they have more consistent grammar and punctuation.

I eliminated redundant checks for IEEE_GET_FLAG and IEEE_HALTING_MODE.

I removed the redundant test doconcurrent08.f90.


lib/semantics/tools.h
```diff
@@ -113,10 +113,49 @@ inline bool IsAssumedSizeArray(const Symbol &symbol) {
...
 bool IsAssumedLengthCharacterFunction(const Symbol &);
+// Is the symbol modifiable in this scope
 std::optional<parser::MessageFixedText> WhyNotModifiable(
     const Symbol &symbol, const Scope &scope);
-// Is the symbol modifiable in this scope
 bool IsExternalInPureContext(const Symbol &symbol, const Scope &scope);
+bool HasCoarray(const parser::Expr &expression);
+template<typename T>
+
+// Detection of image control statements.  If the parameter is not an image
```
klausler: This is way more code than necessary, and it conflates two separate problems: determining whether a given parse tree node type represents an image control statement, and mapping those node types to error messages (or strings to be used in a common error message).

lib/semantics/check-do.cc
```diff
@@ -454,6 +394,7 @@ class DoContext {
     auto &doStmt{
         std::get<parser::Statement<parser::NonLabelDoStmt>>(doConstruct.t)};
     currentStatementSourcePosition_ = doStmt.source;
+    context_.set_location(currentStatementSourcePosition_);
```
jeanPerier: `context` is not something owned by the `DoContext` visitor, if it is modified here, it should probably be restored afterwards to what it was so that the next visitor called by `SemanticsVisitor` does not have a bogus context.

In this case, it is most likely useless to do so, the location was just reset by `SemanticsVisitor` and there is most likely no visitor that will ever use the location before it is set again by `SemanticsVisitor`, but IMHO we should always restore upper level states that we modify and do not own just in case.

I would go for adding a method to `SemanticsContext` that uses `common::Restorer` and `common::ScopedSet` to locally modify `SemanticsContext.location_`.
See `parser::Message::SetLocation` for example:
https://github.com/flang-compiler/f18/blob/b5f48dfc461db37650a54aea3e29531b577fd871/lib/parser/message.h#L289
You may want second opinion on this proposal, maybe it is overkill and simply using `set_location` again at the end of your function is enough.

lib/semantics/tools.h
psteinfeld: Thanks, Peter and Jean.

With respect to reducing the amount of code -- I note that all of these statements are direct or indirect members of the variant ```parser::ExecutableConstruct```.  So my plan is to build the interface simply around that type.

I also plan to change the interface to not return a string and to simply return a boolean.  Specifically, here's the interface I plan to implement --
```bool IsImageControlStatement(parser::ExecutableConstruct);```

Let me know if there's a better way.

File: lib/semantics/tools.h [hunk 3]
```diff
@@ -113,10 +113,49 @@ inline bool IsAssumedSizeArray(const Symbol &symbol) {
 }
 bool IsAssumedLengthCharacter(const Symbol &);
 bool IsAssumedLengthCharacterFunction(const Symbol &);
+// Is the symbol modifiable in this scope
 std::optional<parser::MessageFixedText> WhyNotModifiable(
     const Symbol &symbol, const Scope &scope);
-// Is the symbol modifiable in this scope
 bool IsExternalInPureContext(const Symbol &symbol, const Scope &scope);
+bool HasCoarray(const parser::Expr &expression);
+template<typename T>
```
schweitzpgi: Not sure you really want comments between `template<typename T>` and the return type here.

lib/semantics/tools.h
```diff
@@ -113,10 +113,49 @@ inline bool IsAssumedSizeArray(const Symbol &symbol) {
...
+
+// Detection of image control statements.  If the parameter is not an image
+// control statement, you get nullopt.  Otherwise, you get a message that
+// gives the specifics of why the construct is not allowed.
+std::optional<parser::MessageFixedText> GetImageControlStmtMsg(const T &) {
+  return std::nullopt;
+}
+std::optional<parser::MessageFixedText> GetImageControlStmtMsg(
+    const parser::SyncAllStmt &);
+std::optional<parser::MessageFixedText> GetImageControlStmtMsg(
```
schweitzpgi: Roughly a dozen of these simply return the same error string with the construct name pasted on the front. You could put these all together in a template function that returns a generic "not allowed" message.

That is complicated by the calling of `GetImageControlStmtMsg` on anything, rather than selectively on something known to be an image control statement.  Separation of concerns here would be to have one method determine *if* the object is an offender and a distinct method that constructs a suitable error message/report.

lib/semantics/tools.h
psteinfeld: Thanks, Eric.  I'll clean this up.

lib/semantics/tools.h
psteinfeld: Thanks, Eric.

This is similar to Peter's comment above.  I'm currently working on making this change.

lib/semantics/check-do.cc
psteinfeld: Thanks for spotting this, Jean.  I didn't notice that calling ```set_location()``` would have these side effects.  The basic problem I'm trying to solve is to save the location of the ```DO CONCURRENT``` statement for use in error messages.  It seems to me that the most direct way to do that is just to put it into a data member of the ```DoChecker``` class.  Also, to avoid such side effects, the data member ```SemanticsContext &context_``` should be declared ```const```.

Let me know if you think there's a better way to go about this.

lib/semantics/check-do.cc
```diff
@@ -43,19 +43,23 @@ static const parser::Name *MaybeGetConstructName(
...
 class DoConcurrentBodyEnforce {
 public:
   DoConcurrentBodyEnforce(SemanticsContext &context) : context_{context} {}
   std::set<parser::Label> labels() { return labels_; }
   std::set<SourceName> names() { return names_; }
   template<typename T> bool Pre(const T &) { return true; }
-  template<typename T> void Post(const T &) {}
+  template<typename T> void Post(const T &node) {
+    if (const std::optional<parser::MessageFixedText> message{
+            GetImageControlStmtMsg(node)}) {  // C1137
```
DavidTruby: Can this constraint have a brief description as the ones below do?

lib/semantics/check-do.cc
```diff
@@ -117,89 +121,44 @@ class DoConcurrentBodyEnforce {
 
   // C1136
```
DavidTruby: Same here

lib/semantics/tools.cc
```diff
@@ -508,6 +509,139 @@ std::optional<parser::MessageFixedText> WhyNotModifiable(
...
+  return "SYNC TEAM is not allowed in this context"_err_en_US;
+}
+
+bool IsCoarray(const parser::AllocateObject &allocateObject) {
+  const parser::Name &name{GetLastName(allocateObject)};
+  if (name.symbol && IsCoarray(*name.symbol)) {
+    return true;
+  } else {
+    return false;
+  }
```
DavidTruby: This can be written as `return name.symbol && IsCoarray(*name.symbol);`

lib/semantics/tools.cc
```diff
@@ -508,6 +509,139 @@ std::optional<parser::MessageFixedText> WhyNotModifiable(
...
+}
+
+bool HasCoarray(const parser::Expr &expression) {
+  if (const auto *expr{GetExpr(expression)}) {
+    for (const Symbol *symbol : evaluate::CollectSymbols(*expr)) {
+      if (const Symbol * root{GetAssociationRoot(DEREF(symbol))}) {
+        if (IsCoarray(*root)) {
+          return true;
+        }
+      }
```
DavidTruby: Can this be written as
```
const Symbol* root{GetAssociationRoot(DEREF(symbol))};
return root && IsCoarray(*root);
```
to reduce nesting?

lib/semantics/check-do.cc
jeanPerier: You will probably not be able to declare it const because the `Say` methods are modifying the context.

Adding a member with the DO CONCURRENT location seems good to me. But if it is much more work than what you have now, I am not against modifying the context location as long as it is set back to what it was before.

lib/semantics/check-do.cc
psteinfeld: Certainly!  I'll flesh out all of these cryptic constraint notations.

lib/semantics/check-do.cc
psteinfeld: Yep.

lib/semantics/tools.cc
psteinfeld: Thanks.  I'll make that change.

lib/semantics/tools.cc
psteinfeld: I don't think that would work.  This test is inside a loop that covers all of the symbols from the expression.  Your suggested change would cause the loop to exit at the first symbol.

lib/semantics/check-do.cc
psteinfeld: Thanks, Jean!

lib/semantics/tools.cc
DavidTruby: Good point, I think my internal optimizer was a bit overzealous here!
We could use some C++17 niceness to reduce nesting here with the same behaviour:
```
if (const Symbol * root{GetAssociationRoot(DEREF(symbol))}; root && IsCoarray(*root)) {
  return true;
}
```
but it's not really necessary

psteinfeld: Yep.  For me, the nesting is more readable, even though it uses more lines.

lib/semantics/tools.cc
```diff
@@ -508,6 +509,169 @@ std::optional<parser::MessageFixedText> WhyNotModifiable(
...
+    }
+  }
+  return false;
+}
+
+bool IsImageControlStmt(
+    const parser::Statement<parser::ActionStmt> &actionStmt) {
+  return std::visit(
+      common::visitors{
+          [](const common::Indirection<parser::EventPostStmt> &) {
```
klausler: All of these `return true;` cases can be rolled into the `const auto &` default below and implemented with a `HasMember` test.

lib/semantics/check-do.cc
```diff
@@ -318,8 +269,9 @@ class DoConcurrentLabelEnforce {
 
   void checkLabelUse(const parser::Label &labelUsed) {
     if (labels_.find(labelUsed) == labels_.end()) {
-      context_.Say(currentStatementSourcePosition_,
-          "control flow escapes from DO CONCURRENT"_err_en_US);
+      SayWithDo(context_, currentStatementSourcePosition_,
+          std::move("Control flow escapes from DO CONCURRENT"_err_en_US),
```
jeanPerier: I do not think you need to explicitly move this. It is not wring, but since you are constructing the message inside `SayWithDo` call, it is a temporary and it should already be something that can be moved/copy elided by C++ compilers. Have you tried without the `std::move` ?
(This comment applies to other `SayWithDo` calls)

lib/semantics/tools.cc
```diff
@@ -508,6 +509,169 @@ std::optional<parser::MessageFixedText> WhyNotModifiable(
   }
 }
 
+bool IsCoarray(const parser::AllocateObject &allocateObject) {
```
tskeith: `IsCoarray` and some of the `IsImageControlStmt` overloads below should be declared static so that it is clear they are just helpers and not part of the public API.

lib/semantics/tools.cc
tskeith: Extending this idea, I think it would simplify the code to add an explicit visitor class to pass to the two calls to `std::visit`. This would encapsulate all of the code to implement `IsImageControlStatement` in one place.

Something like this:
```
struct ImageControlStmtHelper {
  using ImageControlStmts = std::variant<  ... >;
  template<typename T> bool operator()(const T &) {
    return common::HasMember<T, ImageControlStmts>;
  }
  template<typename T> bool operator()(const common::Indirection<T> &x) {
    return (*this)(x.value());
  }
  bool operator()(const parser::AllocateStmt &stmt) ...
  bool operator()(const parser::DeallocateStmt &stmt) ...
  bool operator()(const parser::CallStmt &stmt) ...
  bool operator()(const parser::Statement<parser::ActionStmt> &stmt) {
    return std::visit(*this, stmt.statement.u);
  }
};
bool IsImageControlStmt(const parser::ExecutableConstruct &construct) {
  return std::visit(ImageControlStmtHelper{}, construct.u);
}
```

lib/semantics/tools.cc
```diff
@@ -508,6 +509,169 @@ std::optional<parser::MessageFixedText> WhyNotModifiable(
...
+            actionStmt->statement.u)) {
+      return "ALLOCATE of a coarray is an image control statement"_en_US;
+    } else if (std::holds_alternative<
+                   common::Indirection<parser::DeallocateStmt>>(
+                   actionStmt->statement.u)) {
+      return "DEALLOCATE of a coarray is an image control statement"_en_US;
+    } else if (std::holds_alternative<common::Indirection<parser::CallStmt>>(
+                   actionStmt->statement.u)) {
+      return "MOVE_ALLOC of a coarray is an image control statement"_en_US;
+    }
```
tskeith: `std::visit` is clearer than multiple calls to `std::holds_alternative`.

lib/semantics/tools.cc
```diff
@@ -508,6 +509,169 @@ std::optional<parser::MessageFixedText> WhyNotModifiable(
...
+      }
+    }
+  }
+  return false;
+}
+
+const parser::CharBlock GetLocation(
+    const parser::ExecutableConstruct &executableConstruct) {
+  parser::CharBlock result{};
+  std::visit(
```
tskeith: I think this is better written as `return std::visit(`... so it's clear that each lambda is returning the ultimate result of `GetLocation`.

lib/semantics/tools.h
```diff
@@ -113,10 +113,21 @@ inline bool IsAssumedSizeArray(const Symbol &symbol) {
...
 std::optional<parser::MessageFixedText> WhyNotModifiable(
     const Symbol &symbol, const Scope &scope);
-// Is the symbol modifiable in this scope
 bool IsExternalInPureContext(const Symbol &symbol, const Scope &scope);
+bool HasCoarray(const parser::Expr &expression);
+
+// Analysis of image control statements
+bool IsImageControlStmt(const parser::ExecutableConstruct &);
+// Get the location of the image control statement in this ExecutableConstruct
+const parser::CharBlock GetLocation(const parser::ExecutableConstruct &);
```
tskeith: `GetLocation` is too generic a name for what this function does. `GetImageControlStmtLocation` would be better, and analogous to the next function.

psteinfeld: Thanks, Tim.  I'll make that change.

lib/semantics/tools.cc
psteinfeld: Thanks, Tim.  I'll make that change.

lib/semantics/tools.cc
psteinfeld: Thanks, Tim.  I'll change them.

lib/semantics/check-do.cc
psteinfeld: Thanks, Jean.  I removed the ```std::move()``, and things still seem to work.

lib/semantics/tools.cc
psteinfeld: Thanks, Tim and Peter.  I'll make this change.

lib/semantics/tools.cc
psteinfeld: Thanks, Tim.  I'll make this change.

lib/semantics/tools.cc
```diff
@@ -536,6 +537,150 @@ std::unique_ptr<parser::Message> WhyNotModifiable(parser::CharBlock at,
...
+            std::get<parser::ActualArg>(args.front().t)};
+        if (const auto *argExpr{
+                std::get_if<common::Indirection<parser::Expr>>(&actualArg.u)}) {
+          return HasCoarray(argExpr->value());
+        }
+      }
+    }
+  }
+  return false;
+}
```
tskeith: The four functions above can move into `ImageControlStmtHelper` as that is the only place they are called.
And the three overloads of `IsImageControlStmt` are only called in one place each, so they could be expanded in-line.

lib/semantics/tools.cc
```diff
@@ -536,6 +537,150 @@ std::unique_ptr<parser::Message> WhyNotModifiable(parser::CharBlock at,
...
+  bool operator()(const parser::Statement<parser::ActionStmt> &stmt) {
+    return std::visit(*this, stmt.statement.u);
+  }
+};
+
+bool IsImageControlStmt(const parser::ExecutableConstruct &construct) {
+  return std::visit(ImageControlStmtHelper{}, construct.u);
+}
+
+using ResultType = std::optional<parser::MessageFixedText>;
```
tskeith: This would be okay within the scope of `GetImageControlStmtCoarrayMsg` but I don't think it is appropriate at the file level. It's the result type of only one of the many functions declared here.

tskeith:
Please don't force-push while a review is going on. It makes it very difficult for reviewers to see only the new changes.

It's okay (desirable, even) to clean up the commits and force-push after the reviews are complete and just before merging.

lib/semantics/tools.cc
psteinfeld: Good point.  I'll move them into the helper class.

lib/semantics/tools.cc
psteinfeld: Sounds good.  I'll make that change.

PR #782, opened 2019-10-16
Further semantics checks for procedure references

klausler:
These checks enable `call03.f90` to pass.  Other changes in this PR deal with fall-out from enabling these checks on some of our correctness tests (some bugs were exposed, now fixed).  The checking code itself has migrated over to `lib/semantics`, which seems to be a more natural home.  Intrinsic functions that return `CHARACTER` lengths or indices have been reverted to returning (by default) `INTEGER` results of the default kind, rather than subscript-sized integers.


lib/evaluate/characteristics.h
```diff
@@ -95,8 +104,9 @@ class TypeAndShape {
   const Attrs &attrs() const { return attrs_; }
 
   int Rank() const { return GetRank(shape_); }
-  bool IsCompatibleWith(
-      parser::ContextualMessages &, const TypeAndShape &) const;
+  bool IsCompatibleWith(parser::ContextualMessages &, const TypeAndShape &,
+      const char * = "POINTER", const char * = "TARGET",
+      bool isElemental = false) const;
```
jeanPerier: `isElemental` might deserve a comment. Looking at the signature, I thought it was a way to ask the function to accept different shapes as long as one of the shape is a scalar, but looking at the way it is implemented and called, I now understand that it actually means that shape conformance test should be skipped, which means the caller is responsible to ensure that one of the shape is scalar when setting this flag.

documentation/Extensions.md
```diff
@@ -102,6 +103,9 @@ Extensions, deletions, and legacy features supported by default
 * Specific intrinsics AMAX0, AMAX1, AMIN0, AMIN1, DMAX1, DMIN1, MAX0, MAX1,
   MIN0, and MIN1 accept more argument types than specified. They are replaced by
   the related generics followed by conversions to the specified result types.
+* When a scalar CHARACTER actual argument of the same kind is known to
+  have a length shorter than the associated dummy argument, it is extended
+  on the right with blanks, similar to assignment.
```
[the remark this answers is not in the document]
klausler: It does imply a temporary, but I have not yet extended `ActualArgument` with the annotations that I plan to use to represent special processing of arguments before & after calls.  There will not be a copy back in these cases because the actual argument will no longer be a variable.

lib/evaluate/tools.cc
```diff
@@ -646,6 +647,47 @@ bool IsAssumedRank(const ActualArgument &arg) {
...
+bool IsNullPointer(const Expr<SomeDerived> &expr) {
+  const auto *call{std::get_if<FunctionRef<SomeDerived>>(&expr.u)};
+  return call && IsNullPointer(*call);
+}
+bool IsNullPointer(const Expr<SomeType> &expr) {
+  return std::visit(
+      common::visitors{
+          [](const NullPointer &) { return true; },
+          [](const ProcedureRef &call) { return IsNullPointer(call); },
+          [](const auto &) { return false; },
```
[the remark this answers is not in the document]
klausler: Right!  Thanks.

lib/evaluate/tools.cc
```diff
@@ -646,6 +647,47 @@ bool IsAssumedRank(const ActualArgument &arg) {
   }
 }
 
+// IsProcedurePointer()
+bool IsProcedurePointer(const Expr<SomeType> &expr) {
+  return std::visit(
+      common::visitors{
+          [](const NullPointer &) { return true; },
+          [](const ProcedureDesignator &) { return true; },
+          [](const ProcedureRef &) { return true; },
```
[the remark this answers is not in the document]
klausler: Because a `ProcedureRef` in a top-level `Expr<SomeType>` is used only to represent a call to a function returning a pointer to a procedure.

lib/semantics/expression.cc
```diff
@@ -1557,20 +1559,18 @@ static bool CheckCompatibleArguments(
...
     if (std::optional<characteristics::Procedure> procedure{
             characteristics::Procedure::Characterize(
                 ProcedureDesignator{*specific}, context_.intrinsics())}) {
-      parser::Messages buffer;
-      parser::ContextualMessages messages{
-          context_.foldingContext().messages().at(), &buffer};
-      FoldingContext localContext{context_.foldingContext(), messages};
       ActualArguments localActuals{actuals};
-      if (CheckExplicitInterface(*procedure, localActuals, localContext) &&
+      auto messages{CheckExplicitInterface(
```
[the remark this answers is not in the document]
klausler: I don't know why this didn't elicit an error, but both g++ and clang are happy with it.  I'll add the namespace prefix for future-proofing.

lib/semantics/check-call.cc
```diff
@@ -0,0 +1,501 @@
...
+          msg->Attach(concerns.finalProcedure->name(),
+              "Declaration of FINAL procedure"_en_US);
+        }
+      }
+    }
+    if (actualIsCoindexed && concerns.allocatable &&
+        dummy.intent != common::Intent::In && !dummyIsValue) {
+      // 15.5.2.4(6)
+      if (auto *msg{messages.Say(
+              "Coindexed actual argument with ALLOCATABLE ultimate component must be associated with a %s with VALUE or INTENT(IN) attributes"_err_en_US,
```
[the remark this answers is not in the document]
klausler: I have also run into this while testing, will fix.

lib/semantics/check-call.cc
```diff
@@ -0,0 +1,501 @@
...
+  bool actualIsAsynchronous{false};
+  bool actualIsVolatile{false};
+  const Symbol *actualFirstSymbol{evaluate::GetFirstSymbol(actual)};
+  if (actualFirstSymbol != nullptr) {
+    const Symbol &ultimate{actualFirstSymbol->GetUltimate()};
+    actualIsAsynchronous =
+        actualFirstSymbol->attrs().test(Attr::ASYNCHRONOUS) ||
+        ultimate.attrs().test(Attr::ASYNCHRONOUS);
+    actualIsVolatile = actualFirstSymbol->attrs().test(Attr::VOLATILE) ||
+        ultimate.attrs().test(Attr::VOLATILE);
```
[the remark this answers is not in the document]
klausler: Because a `USE` statement can't add `ASYNCHRONOUS` or `VOLATILE` attributes, and a `BLOCK` construct can add them but not remove them, I really should have to check only the original inbound symbol.  I also checked the ultimate one because I wasn't sure that `USE` associations would preserve all attributes.  I will remove the ultimate checks, they're just confusing.

lib/evaluate/characteristics.h
klausler: But if the reader is not familiar with `ELEMENTAL` procedures in Fortran, this may not be the best place to describe them.  What it is supposed to do here is skip the shape conformance test between an actual argument and a dummy, since an `ELEMENTAL` procedure's dummy arguments are required to be scalar, but the actual arguments are allowed to be arrays (and must be conformant with each other).

lib/semantics/check-call.cc
```diff
@@ -0,0 +1,501 @@
...
+    }
+    if (actualLastSymbol && actualLastSymbol->Rank() == 0 &&
+        !(dummy.type.type().IsAssumedType() && dummyIsAssumedSize)) {
+      messages.Say(
+          "Whole scalar actual argument may not be associated with a %s array"_err_en_US,
+          dummyName);
+    }
+    if (actualIsPolymorphic) {
+      messages.Say(
+          "Element of polymorphic array may not be associated with a %s array"_err_en_US,
```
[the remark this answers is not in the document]
klausler: Fixed.

lib/evaluate/characteristics.cc
```diff
@@ -98,28 +98,46 @@ std::optional<TypeAndShape> TypeAndShape::Characterize(
...
   const auto &len{that.LEN()};
   if (!type_.IsTypeCompatibleWith(that.type_)) {
     std::stringstream lenstr;
     if (len) {
       len->AsFortran(lenstr);
     }
-    messages.Say("Target type '%s' is not compatible with '%s'"_err_en_US,
-        that.type_.AsFortran(lenstr.str()), type_.AsFortran());
+    messages.Say(
+        "%1$s type '%2$s' is not compatible with %3$s type '%4$s'"_err_en_US,
```
tskeith: What does `$s` mean in the error format string?

klausler: It selects the actual argument to be formatted.  In this case, where `1$` through `4$` are used, the effect is the same as if they had not been present.  I inserted them here in response to a comment about easing internationalization.

lib/semantics/tools.h
```diff
@@ -84,6 +84,10 @@ bool CanBeTypeBoundProc(const Symbol *);
 const Symbol *FindUltimateComponent(
     const DerivedTypeSpec &type, std::function<bool(const Symbol &)> predicate);
 
+// Returns an immediate component of type that matches predicate, or nullptr.
+const Symbol *FindImmediateComponent(
+    const DerivedTypeSpec &, const std::function<bool(const Symbol &)> &);
```
tskeith: It might be helpful to define "immediate component" in the comment below that defined "ultimate component", etc.

klausler: Will do.

lib/semantics/check-call.cc
```diff
@@ -0,0 +1,475 @@
+// Copyright (c) 2019, NVIDIA CORPORATION.  All rights reserved.
```
tskeith: git seems to be treating this as a new file rather than a move, so I can't tell if there are changes to review here.

klausler: There are, it's not just a move.  I do not know why github failed to track the move on this file when it seemed to be able to do so on the other file that I moved (`check-call.h`).  I will mail you a diff.

PR #783, opened 2019-10-17
Add call semantics checks, pass call04.f90 test

klausler:
Here's some more call semantics checks.  Note that this branch is a branch of "pmk-calls", and that I have not yet merged that branch with master.

### DIFF
--- a/lib/semantics/check-declarations.cc
+++ b/lib/semantics/check-declarations.cc
@@ -114,18 +114,16 @@ void CheckHelper::Check(Symbol &symbol) {
     Check(object->shape());
     Check(object->coshape());
     if (object->isDummy() && symbol.attrs().test(Attr::INTENT_OUT)) {
-      if (const auto *type{object->type()}) {
-        if (const auto *derived{type->AsDerived()}) {
-          TypeInspector inspector;
-          inspector.Inspect(*derived);
-          if (const Symbol * bad{inspector.allocatableCoarray()}) {  // C846
-            if (auto *msg{messages_.Say(
-                    "An INTENT(OUT) dummy argument may not contain an ALLOCATABLE coarray"_err_en_US)}) {
-              msg->Attach(
-                  bad->name(), "Declaration of ALLOCATABLE coarray"_en_US);
-            }
-          }
-        }
+      if (FindUltimateComponent(symbol,
+              std::function<bool(const Symbol &)>{[](const Symbol &symbol) {
+                return IsCoarray(symbol) && IsAllocatable(symbol);
+              }})) {  // C846
+        messages_.Say(
+            "An INTENT(OUT) dummy argument may not be, or contain, an ALLOCATABLE coarray"_err_en_US);
+      }
+      if (IsOrContainsEventOrLockComponent(symbol)) {  // C847
+        messages_.Say(
+            "An INTENT(OUT) dummy argument may not be, or contain, EVENT_TYPE or LOCK_TYPE"_err_en_US);
       }
     }
   }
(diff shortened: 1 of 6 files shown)


lib/evaluate/characteristics.cc
```diff
@@ -310,6 +331,51 @@ std::optional<DummyArgument> DummyArgument::Characterize(
...
+std::optional<DummyArgument> DummyArgument::FromActual(
+    std::string &&name, const Expr<SomeType> &expr, FoldingContext &context) {
+  return std::visit(
+      common::visitors{
+          [&](const BOZLiteralConstant &) {
+            return std::make_optional<DummyArgument>(std::move(name),
+                DummyDataObject{
+                    TypeAndShape{DynamicType::TypelessIntrinsicArgument()}});
+          },
+          [&](const NullPointer &) { return std::optional<DummyArgument>{}; },
```
psteinfeld: I think that the "&" reference capture is not needed here.

lib/evaluate/shape.cc
```diff
@@ -349,8 +353,28 @@ Shape GetUpperBounds(FoldingContext &context, const NamedEntity &base) {
...
+              Shape shape;
+              int n{object.shape().Rank()};
+              NamedEntity base{symbol};
+              for (int dimension{0}; dimension < n; ++dimension) {
+                shape.emplace_back(GetExtent(context_, base, dimension));
+              }
+              return Result{shape};
+            }
+          },
+          [&](const semantics::EntityDetails &) {
```
psteinfeld: The reference capture specification is not needed here or below for the default case.

lib/evaluate/shape.cc
```diff
@@ -451,21 +468,80 @@ auto GetShapeHelper::operator()(const ProcedureRef &call) const -> Result {
...
+        if (auto shape{(*this)(call.arguments().at(0))}) {
+          if (shape->size() == 2) {
+            std::swap((*shape)[0], (*shape)[1]);
+            return shape;
+          }
+        }
       }
+    } else if (intrinsic->characteristics.value().attrs.test(characteristics::
+                       Procedure::Attr::NullPointer)) {  // NULL(MOLD=)
+      return (*this)(call.arguments());
```
psteinfeld: Should there be curly braces around this "return"?

lib/semantics/check-declarations.cc
```diff
@@ -88,31 +88,44 @@ void CheckHelper::Check(Symbol &symbol) {
...
+                return IsCoarray(symbol) && IsAllocatable(symbol);
+              }})) {  // C846
+        messages_.Say(
+            "An INTENT(OUT) dummy argument may not be, or contain, an ALLOCATABLE coarray"_err_en_US);
+      }
+      if (IsOrContainsEventOrLockComponent(symbol)) {  // C847
+        messages_.Say(
+            "An INTENT(OUT) dummy argument may not be, or contain, EVENT_TYPE or LOCK_TYPE"_err_en_US);
+      }
+    }
```
psteinfeld: Thanks for adding these checks!  They were on my list.

lib/evaluate/shape.cc
klausler: They are present.

lib/semantics/check-declarations.cc
```diff
@@ -88,31 +88,44 @@ void CheckHelper::Check(Symbol &symbol) {
...
     }
   }
   if (auto *object{symbol.detailsIf<ObjectEntityDetails>()}) {
     Check(object->shape());
     Check(object->coshape());
+    if (object->isDummy() && symbol.attrs().test(Attr::INTENT_OUT)) {
+      if (FindUltimateComponent(symbol,
+              std::function<bool(const Symbol &)>{[](const Symbol &symbol) {
+                return IsCoarray(symbol) && IsAllocatable(symbol);
+              }})) {  // C846
```
tskeith: Why not just: `if (FindUltimateComponent(symbol, [](const Symbol &symbol) {...}))`?

klausler: Sure.

PR #785, opened 2019-10-21
Implement uint128_t for impoverished build environments

klausler:
Actually, for better testing, this implementation of `common::uint128_t` is enabled for all environments because it's fast enough and should get as much testing as possible.  We can use the native `__uint128_t` type later, where available, by flipping a `#define` if we want.

### DIFF
--- a/lib/common/uint128.h
+++ b/lib/common/uint128.h
@@ -1,0 +1,253 @@
+// Copyright (c) 2019, NVIDIA CORPORATION.  All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Portable 128-bit unsigned integer arithmetic
+
+#ifndef FORTRAN_COMMON_UINT128_H_
+#define FORTRAN_COMMON_UINT128_H_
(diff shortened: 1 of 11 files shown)


lib/common/uint128.h
```diff
@@ -0,0 +1,273 @@
...
+  }
+  constexpr UnsignedInt128 operator^(UnsignedInt128 that) const {
+    return {high_ ^ that.high_, low_ ^ that.low_};
+  }
+
+  constexpr UnsignedInt128 operator<<(UnsignedInt128 that) const {
+    if (that >= 128) {
+      return {};
+    } else if (that == 0) {
+      return *this;
```
tskeith: The shift operators could be simplified by replacing the first two special cases with `if (that.high_ > 0) return {}`.

lib/common/uint128.h
```diff
@@ -0,0 +1,273 @@
...
+      std::uint64_t y0{that.low_ & mask32}, y1{that.low_ >> 32},
+          y2{that.high_ & mask32}, y3{that.high_ >> 32};
+      UnsignedInt128 x0y0{x0 * y0}, x0y1{x0 * y1}, x0y2{x0 * y2}, x0y3{x0 * y3};
+      UnsignedInt128 x1y0{x1 * y0}, x1y1{x1 * y1}, x1y2{x1 * y2};
+      UnsignedInt128 x2y0{x2 * y0}, x2y1{x2 * y1};
+      UnsignedInt128 x3y0{x3 * y0};
+      return x0y0 + ((x0y1 + x1y0) << 32) + ((x0y2 + x1y1 + x2y0) << 64) +
+          ((x0y3 + x1y2 + x2y1 + x3y0) << 96);
+    }
+  }
```
tskeith: Should some of these more complex functions be defined out-of-line?

klausler: They need to be `constexpr` so it doesn't really make a difference.

tskeith: Should some of these more complex functions be defined out-of-line?

PR #786, opened 2019-10-22
Resolve defined operators to specifics

tskeith:
Most of these changes involve moving code around so that it case be
used for `DefinedUnary` and `DefinedBinary`. The functional changes are
in the `Analyze` member functions for those cases where the arguments
are now analyzed, the generic is resolved, and a `FunctionRef` is
created.

Add `ArgumentAnalyzer` to handling building of the `ActualArguments`
of a call. This allows the code to be shared with the defined unary
and defined binary cases. Move `AnalyzeActualArgument` and
`AnalyzeActualArgument` into that class (renaming both to `Analyze`).

Create an overload of `GetCalleeAndArguments` for the `Name` case so it
can be used for defined ops where we don't have a `ProcedureDesignator`.

Remove the `Scope` argument from `ResolveGeneric` and `GetCalleeAndArguments`
and instead use the symbol's owner to the scope. The requires being
careful not to call `GetUltimate` too early so that we get the right
scope.

Move `IsGenericDefinedOp` to `tools.h` to make it available to the
new code.

### DIFF
--- a/lib/semantics/expression.cc
+++ b/lib/semantics/expression.cc
@@ -133,6 +133,28 @@ common::IfNoLvalue<MaybeExpr, WRAPPED> TypedWrapper(
   }
 }
 
+class ArgumentAnalyzer {
+public:
+  explicit ArgumentAnalyzer(ExpressionAnalyzer &context) : context_{context} {}
+  bool success() const { return success_; }
+  ActualArguments &&GetActuals() {
+    CHECK(success_);
+    return std::move(actuals_);
+  }
+  template<typename T> void Analyze(const T &x) {
+    actuals_.emplace_back(context_.Analyze(x));
+    success_ &= actuals_.back().has_value();
+  }
+  void Analyze(const parser::ActualArgSpec &, bool isSubroutine);
+
+private:
+  std::optional<ActualArgument> Analyze(const parser::Expr &);
+
+  ExpressionAnalyzer &context_;
+  ActualArguments actuals_;
+  bool success_{true};
+};
+
 // Wraps a data reference in a typed Designator<>, and a procedure
 // or procedure pointer reference in a ProcedureDesignator.
 MaybeExpr ExpressionAnalyzer::Designate(DataRef &&ref) {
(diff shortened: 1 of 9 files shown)


lib/semantics/expression.cc
```diff
@@ -133,6 +133,28 @@ common::IfNoLvalue<MaybeExpr, WRAPPED> TypedWrapper(
   }
 }
 
+class ArgumentAnalyzer {
+public:
+  ArgumentAnalyzer(ExpressionAnalyzer &context) : context_{context} {}
```
klausler: This constructor should be `explicit`.

lib/semantics/expression.cc
```diff
@@ -1559,17 +1581,17 @@ static bool CheckCompatibleArguments(
...
-  const auto &details{symbol.get<semantics::GenericDetails>()};
+  const auto &details{symbol.GetUltimate().get<semantics::GenericDetails>()};
   for (const Symbol *specific : details.specificProcs()) {
     if (std::optional<characteristics::Procedure> procedure{
             characteristics::Procedure::Characterize(
                 ProcedureDesignator{*specific}, context_.intrinsics())}) {
       ActualArguments localActuals{actuals};
       auto messages{semantics::CheckExplicitInterface(
-          *procedure, localActuals, GetFoldingContext(), scope)};
+          *procedure, localActuals, GetFoldingContext(), symbol.owner())};
```
klausler: Does this change cause `CheckExplicitInterface` to enforce modifiability checks on actual arguments from the perspective of the called procedure rather than the call site?  Will constraints on calls *from* `PURE` subprograms still work?

lib/semantics/expression.cc
```diff
@@ -1581,47 +1603,25 @@ const Symbol *ExpressionAnalyzer::ResolveGeneric(const Symbol &symbol,
   }
   if (elemental) {
     return elemental;
+  }
+  if (semantics::IsGenericDefinedOp(symbol)) {
+    Say("No specific procedure of generic operator '%s' matches the actual arguments"_err_en_US,
```
klausler: Maybe `OPERATOR('%s')`?

lib/semantics/expression.cc
tskeith: I believe these still work as before. Here is a test: it gets the same errors as before (indicated in comments).
```
module m
  real, protected :: x
  real :: y
  interface s
    subroutine s1(x)
      real, intent(out) :: x
    end
  end interface
end

subroutine test1
  use m
  !ERROR: No specific procedure of generic 's' matches the actual arguments
  call s(x)
end

pure subroutine test2
  use m
  !ERROR: No specific procedure of generic 's' matches the actual arguments
  call s(y)
end
```

tskeith: These could be improved. Ideally we would resolve `s` to `s1` and then report the real errors about why the actual arguments aren't okay. That may mean we can't use `CheckExplicitInterface` in resolving generics.

lib/semantics/expression.cc
tskeith: This is consistent with other errors about defined operators but it might be a good idea to change them all. But not part of this PR.

lib/semantics/expression.cc
klausler: I guess it works because the specific procedure is brought into the scope of the call via USE or host association.  Would it work if the specific procedure were internal?

lib/semantics/expression.cc
tskeith: OK, done

lib/semantics/expression.cc
```diff
@@ -2390,8 +2362,84 @@ MaybeExpr ExpressionAnalyzer::MakeFunctionRef(
...
+        context_.Say(expr.source,
+            "Non-intrinsic ELEMENTAL procedure cannot be passed as argument"_err_en_US);
+      }
+    }
+    if (auto coarrayRef{ExtractCoarrayRef(x)}) {
+      const Symbol &coarray{coarrayRef->GetLastSymbol()};
+      if (const semantics::DeclTypeSpec * type{coarray.GetType()}) {
+        if (const semantics::DerivedTypeSpec * derived{type->AsDerived()}) {
+          if (auto ptr{semantics::FindPointerUltimateComponent(*derived)}) {
+            if (auto *msg{context_.Say(expr.source,
```
psteinfeld: If we're emitting an error message, should we be returning ```std::nullopt```?

lib/semantics/expression.h
```diff
@@ -326,11 +326,12 @@ class ExpressionAnalyzer {
       const parser::Call &, bool isSubroutine);
   std::optional<characteristics::Procedure> CheckCall(
       parser::CharBlock, const ProcedureDesignator &, ActualArguments &);
-  const Symbol *ResolveGeneric(
-      const Symbol &, ActualArguments &, const semantics::Scope &);
+  const Symbol *ResolveGeneric(const Symbol &, ActualArguments &);
```
psteinfeld: ```ResolveGeneric()``` only seems to be used in expression.cc.  Should it be in expression.h?

tskeith: It needs to be a member function so that `Say` is available.

lib/semantics/expression.cc
tskeith: I'm not sure. This code isn't new, it's just been moved.

PR #788, opened 2019-10-22
Create `common::Reference<>`, use it to replace non-nullable pointers

klausler:
This PR is a proof-of-concept that some uses of non-nullable pointers can be replaced with `std::reference_wrapper<>` to gain some extra readability and remove some `DEREF()` checks (or places where we should have them but don't).

If this all seems good, I'll push ahead with this and use `reference_wrapper<>` in place of non-nullable pointers in more data structures.

### DIFF
--- a/lib/evaluate/tools.h
+++ b/lib/evaluate/tools.h
@@ -102,13 +102,12 @@ template<typename A> bool IsVariable(const A &x) {
 }
 
 // Predicate: true when an expression is assumed-rank
-bool IsAssumedRank(const semantics::Symbol &);
+bool IsAssumedRank(const Symbol &);
 bool IsAssumedRank(const ActualArgument &);
 template<typename A> bool IsAssumedRank(const A &) { return false; }
 template<typename A> bool IsAssumedRank(const Designator<A> &designator) {
-  if (const auto *symbol{
-          std::get_if<const semantics::Symbol *>(&designator.u)}) {
-    return IsAssumedRank(*symbol);
+  if (const auto *symbol{std::get_if<SymbolRef>(&designator.u)}) {
+    return IsAssumedRank(symbol->get());
   } else {
     return false;
   }
@@ -244,8 +243,8 @@ template<typename A> std::optional<NamedEntity> ExtractNamedEntity(const A &x) {
   if (auto dataRef{ExtractDataRef(x)}) {
     return std::visit(
         common::visitors{
-            [](const Symbol *symbol) -> std::optional<NamedEntity> {
-              return NamedEntity{*symbol};
+            [](SymbolRef &&symbol) -> std::optional<NamedEntity> {
+              return NamedEntity{symbol};
             },
             [](Component &&component) -> std::optional<NamedEntity> {
               return NamedEntity{std::move(component)};
@@ -296,8 +295,8 @@ template<typename A> std::optional<CoarrayRef> ExtractCoarrayRef(const A &x) {
 // extract and return that symbol, else null.
 template<typename A> const Symbol *UnwrapWholeSymbolDataRef(const A &x) {
   if (auto dataRef{ExtractDataRef(x)}) {
-    if (const Symbol **p{std::get_if<const Symbol *>(&dataRef->u)}) {
-      return *p;
+    if (const SymbolRef * p{std::get_if<SymbolRef>(&dataRef->u)}) {
+      return &p->get();
     }
   }
   return nullptr;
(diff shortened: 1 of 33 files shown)


tskeith:
This looks good to me. It solves a problem that has been annoying me. It appears to be mostly transparent to users of `SymbolSet`, apart from the need to call `get()` when using iterators.

I encountered one build problem, with GCC 8.3.0 on MacOS:
```
../../lib/evaluate/../semantics/symbol.h: In member function 'constexpr bool semantics::SymbolIsLess::operator()(const std::reference_wrapper<const semantics::Symbol>&, const std::reference_wrapper<const semantics::Symbol>&) const':
../../lib/evaluate/../semantics/symbol.h:700:25: error: call to non-'constexpr' function 'const SourceName& semantics::Symbol::name() const'
     return x.get().name().begin() < y.get().name().begin();
                         ^
```
Removing the constexpr fixes (or works around) the problem.

It doesn't happen with clang 7.0.1.

lib/semantics/symbol.h
```diff
@@ -691,5 +692,16 @@ inline bool ProcEntityDetails::HasExplicitInterface() const {
   }
   return false;
 }
+
+// SymbolSet
+struct SymbolIsLess {
+  constexpr bool operator()(const std::reference_wrapper<const Symbol> &x,
```
psteinfeld: Just like Tim, the ```constexpr``` on this line caused compilation problems for me running gcc 8.3.0 on my laptop.  Removing it made everything build and test correctly.

tskeith:
I think it would be worthwhile to introduce `using SymbolRef = std::reference_wrapper<const Symbol>` in `symbol.h`. (Or in a new file if we want it to be available without all of `symbol.h`.)
Then add `operator==` and maybe `operator<` for that, and we wouldn't need as many calls to `get()`.

klausler:
> I think it would be worthwhile to introduce `using SymbolRef = std::reference_wrapper<const Symbol>` in `symbol.h`. (Or in a new file if we want it to be available without all of `symbol.h`.)
> Then add `operator==` and maybe `operator<` for that, and we wouldn't need as many calls to `get()`.

I'll look into both of those suggestions.

tskeith:
I don't understand why the `CharBlock` changes are here, they're not related to using `std::reference_wrapper<>`.

Why do we want to pass `CharBlock`s by value rather than reference?

klausler:
> I don't understand why the `CharBlock` changes are here, they're not related to using `std::reference_wrapper<>`.
> 
> Why do we want to pass `CharBlock`s by value rather than reference?

I wanted to make `Symbol::name()` a `constexpr` member function and value semantics made that possible.  But it turned out not to be sufficient to make the comparator function `constexpr`.  So I could back those out if you prefer that these values have reference semantics, just let me know.

klausler:
Whoops, I've hit a major snag: `reference_wrapper<>` doesn't have a move constructor.  That makes it unusable in `variant<>` instances that need to support moves.  I may have to use a new `common::nonnull_ptr<>` custom class instead.

klausler:
If I have to define a custom form of `reference_wrapper` anyway to get move semantics, I might as well add `operator*` and `->` to it as well.

tskeith:
> I wanted to make `Symbol::name()` a `constexpr` member function and value semantics made that possible. But it turned out not to be sufficient to make the comparator function `constexpr`. So I could back those out if you prefer that these values have reference semantics, just let me know.

If it's worthwhile making `Symbol::name()` be constexpr then it's fine with me. Otherwise I don't see the benefit.

klausler:
> > I wanted to make `Symbol::name()` a `constexpr` member function and value semantics made that possible. But it turned out not to be sufficient to make the comparator function `constexpr`. So I could back those out if you prefer that these values have reference semantics, just let me know.
> 
> If it's worthwhile making `Symbol::name()` be constexpr then it's fine with me. Otherwise I don't see the benefit.

I prefer value semantics when they're sufficient, but it's an issue that's peripheral to this change, so I'll back that part out before merging.

lib/semantics/symbol.h
klausler: Can you please confirm whether the change that you wanted has been made, and clear the block for the merge if so?

lib/evaluate/tools.h
```diff
@@ -103,13 +102,12 @@ template<typename A> bool IsVariable(const A &x) {
...
-bool IsAssumedRank(const semantics::Symbol &);
+bool IsAssumedRank(const Symbol &);
 bool IsAssumedRank(const ActualArgument &);
 template<typename A> bool IsAssumedRank(const A &) { return false; }
 template<typename A> bool IsAssumedRank(const Designator<A> &designator) {
-  if (const auto *symbol{
-          std::get_if<const semantics::Symbol *>(&designator.u)}) {
-    return IsAssumedRank(*symbol);
+  if (const auto *symbol{std::get_if<SymbolRef>(&designator.u)}) {
+    return IsAssumedRank(symbol->get());
```
tskeith: It is probably moot now, but I don't understand how this could have worked before.

Now `symbol` is `const SymbolRef *` so `symbol->get()` is `const Symbol &`, as expected.

But before it looks like `symbol` would have been `const Symbol **` and so a `const Symbol *` would have been passed to `IsAssumedRank`, which expects `const Symbol &`.

klausler: `IsAssumedRank()` is a template function with a default catch-all case that returns `false`.  So the original code compiled but didn't work properly.

lib/semantics/symbol.h
psteinfeld: Oops.  Sorry for the delay.  All looks good.

klausler:
Reviewers, please take another look (if you like) before I merge.  I have used the new `Reference<>` type to replace the non-nullable pointers in `Scope`.  I have also reworked the component iterator facility because my changes had broken it and I couldn't figure out how it worked well enough to fix it.

Many non-nullable pointers remain in our data structures, and we can update them incrementally if we're happy with the way this initial set of changes works.

tskeith:
Still looks good to me.

Issue #781, opened 2019-10-16
DO CONCURRENT test for pure procedure components is wrong


psteinfeld:
Constraint C1139 prohibits impure procedures from being referenced in a DO CONCURRENT.  One way to do this is to use a procedure component.  But the existing test for purity of procedure components doesn't work.  The attached test contains DO CONCURRENT loops that reference both pure and impure functions.  F18 currently thinks that both of them are impure.  In contrast, Gnu Fortran only flags the impure function call.

psteinfeld:
[ProcComponentTest.f90.txt](https://github.com/flang-compiler/f18/files/3763943/ProcComponentTest.f90.txt)

psteinfeld:
[ProcComponentTest.f90.txt](https://github.com/flang-compiler/f18/files/3768826/ProcComponentTest.f90.txt)

psteinfeld:
The test case works now.

PR #789, opened 2019-10-23
Changes to check for calls to PURE procedure components

psteinfeld:
This addresses issue #781.  As far as I can tell, there's no way to
declare a PURE procedure component.  So I simplified the code in
check-do.cc to check for this condition and added a test for it.

### DIFF
--- a/lib/semantics/check-do.cc
+++ b/lib/semantics/check-do.cc
@@ -167,7 +167,7 @@ public:
         }
       }
     } else {
-      // C1139: check for an impure procedure component
+      // C1139: this a procedure component
       auto &component{std::get<parser::ProcComponentRef>(procedureDesignator.u)
                           .v.thing.component};
       if (component.symbol && !IsPureProcedure(*component.symbol)) {
--- a/lib/semantics/tools.cc
+++ b/lib/semantics/tools.cc
@@ -167,6 +167,12 @@ bool IsFunction(const Symbol &symbol) {
 }
 
 bool IsPureProcedure(const Symbol &symbol) {
+  if (const auto *procDetails{symbol.detailsIf<ProcEntityDetails>()}) {
+    if (const Symbol *procInterface{procDetails->interface().symbol()}) {
+      // procedure component with a PURE interface
+      return IsPureProcedure(*procInterface);
+    }
+  }
   return symbol.attrs().test(Attr::PURE) && IsProcedure(symbol);
 }
 
--- a/test/semantics/doconcurrent01.f90
+++ b/test/semantics/doconcurrent01.f90
@@ -204,3 +204,53 @@ call move_alloc(ca, cb)
     call move_alloc(pvar%type1_field%coarray_type0_field, qvar%type1_field%coarray_type0_field)
   end do
 end subroutine s6
+
+subroutine s7()
+  interface
+    pure integer function pf()
+    end function pf
+  end interface
+
+  type :: procTypeNotPure
+    procedure(notPureFunc), pointer, nopass :: notPureProcComponent
+  end type procTypeNotPure
+
+  type :: procTypePure
+    procedure(pf), pointer, nopass :: pureProcComponent
+  end type procTypePure
+
+  type(procTypeNotPure) :: procVarNotPure
+  type(procTypePure) :: procVarPure
+  integer :: ivar
+
+  procVarPure%pureProcComponent => pureFunc
+
+  do concurrent (i = 1:10)
+    print *, "hello"
+  end do
+
+  do concurrent (i = 1:10)
+    ivar = pureFunc()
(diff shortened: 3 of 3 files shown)


psteinfeld:
> 
> 
> The code looks good but I do not follow your conclusion for #781 that one cannot define pure procedure components. Why is `pureProcComponent` from your example in #781 not working ?

You're correct.  My example did not actually declare a PURE procedure.  As far as I can tell, the grammar does not allow for the PURE attribute to be associated with a procedure component.

jeanPerier:
> You're correct. My example did not actually declare a PURE procedure. As far as I can tell, the grammar does not allow for the PURE attribute to be associated with a procedure component.

Right, a procedure component cannot in itself have the pure attribute, but C1139 is about reference to procedure, and isn't a reference to a procedure component a reference to the procedure it is associated with (15.5.5.3 point 1. (*see below))? Then, if the characteristic of the proc-interface in the proc-component-def-stmt is pure, it should imply that all references to this proc-component will actually reference pure procedure (following procedure pointer assignment rule 10.2.2.4 point 3.).

All that to say, why not having `IsPureProcedure` looking at the proc-interface characteristics to determine whether or not references to it will reference pure procedures ?

(*) I am not entirely sure 15.5.5.3 point 1 applies to procedure components.

lib/semantics/check-do.cc
```diff
@@ -167,15 +167,10 @@ class DoConcurrentBodyEnforce {
...
-      // C1139: check for an impure procedure component
-      auto &component{std::get<parser::ProcComponentRef>(procedureDesignator.u)
-                          .v.thing.component};
-      if (component.symbol && !IsPureProcedure(*component.symbol)) {
-        SayWithDo(context_, currentStatementSourcePosition_,
-            "Call to an impure procedure component is not allowed"
-            " in DO CONCURRENT"_err_en_US,
-            doConcurrentSourcePosition_);
-      }
+      // C1139: this a procedure component, all of which are impure
```
klausler: I do not believe that the statement in the comment is correct.  Procedure components can have specific interfaces, those interfaces imply characteristics, and one of those characteristics is whether or not the procedure is `PURE`.

psteinfeld:
> 
> 
> > You're correct. My example did not actually declare a PURE procedure. As far as I can tell, the grammar does not allow for the PURE attribute to be associated with a procedure component.
> 
> Right, a procedure component cannot in itself have the pure attribute, but C1139 is about reference to procedure, and isn't a reference to a procedure component a reference to the procedure it is associated with (15.5.5.3 point 1. (*see below))? Then, if the characteristic of the proc-interface in the proc-component-def-stmt is pure, it should imply that all references to this proc-component will actually reference pure procedure (following procedure pointer assignment rule 10.2.2.4 point 3.).
> 
> All that to say, why not having `IsPureProcedure` looking at the proc-interface characteristics to determine whether or not references to it will reference pure procedures ?
> 
> (*) I am not entirely sure 15.5.5.3 point 1 applies to procedure components.

Thanks, Jean.

I hadn't noticed the ability to declare procedure components with specific interfaces, which themselves can contain the PURE attribute.  I'll fix up the test and the code to check for that.

lib/semantics/check-do.cc
psteinfeld: Thanks, Peter.  I hadn't previously noticed the ability to specify PURE procedure interfaces.  I'll fix the test and the code to check for this.

psteinfeld:
> 
> 
> > You're correct. My example did not actually declare a PURE procedure. As far as I can tell, the grammar does not allow for the PURE attribute to be associated with a procedure component.
> 
> Right, a procedure component cannot in itself have the pure attribute, but C1139 is about reference to procedure, and isn't a reference to a procedure component a reference to the procedure it is associated with (15.5.5.3 point 1. (*see below))? Then, if the characteristic of the proc-interface in the proc-component-def-stmt is pure, it should imply that all references to this proc-component will actually reference pure procedure (following procedure pointer assignment rule 10.2.2.4 point 3.).
> 
> All that to say, why not having `IsPureProcedure` looking at the proc-interface characteristics to determine whether or not references to it will reference pure procedures ?
> 
> (*) I am not entirely sure 15.5.5.3 point 1 applies to procedure components.

You're exactly correct.  I think that's the right thing to do.

lib/semantics/check-do.cc
```diff
@@ -167,7 +167,7 @@ class DoConcurrentBodyEnforce {
         }
       }
     } else {
-      // C1139: check for an impure procedure component
+      // C1139: this a procedure component, all of which are impure
```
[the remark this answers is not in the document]
psteinfeld: Yep.  I noticed that, too.  I've deleted it.

PR #795, opened 2019-10-25
[OpenMP] Test Infra update to support OpenMP symbol tests

ichoyjx:
During the symbol tests, OpenMP directives (start with `[!c*]$omp`) line needs to be saved for 1.f90 and 2.f90. Also moved `OPTIONS:` detection to common.sh as `$USER_OPTIONS`.

### DIFF
--- a/test/semantics/common.sh
+++ b/test/semantics/common.sh
@@ -14,8 +14,9 @@
 
 # Common functionality for test scripts
 # Process arguments, expecting source file as 1st; optional path to f18 as 2nd
-# Set: $F18 to the path to f18; $temp to an empty temp directory; and $src
-# to the full path of the single source argument
+# Set: $F18 to the path to f18; $temp to an empty temp directory; $src
+# to the full path of the single source argument; and $USER_OPTIONS to the
+# option list given in the $src file after string "OPTIONS:"
 
 PATH=/usr/bin:/bin
 
@@ -35,5 +36,7 @@ case $1 in
   (/*) src=$1 ;;
   (*) src=$(dirname $0)/$1 ;;
 esac
+USER_OPTIONS=`sed -n 's/^ *! *OPTIONS: *//p' $src`
+echo $USER_OPTIONS
 temp=`mktemp -d ./tmp.XXXXXX`
 [[ $KEEP ]] || trap "rm -rf $temp" EXIT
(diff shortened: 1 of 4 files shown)


test/semantics/test_symbols.sh
```diff
@@ -28,13 +28,15 @@ src2=$temp/2.f90
 src3=$temp/3.f90
 diffs=$temp/diffs
 
-# Strip out blank lines and all comments except "!DEF:" and "!REF:"
-sed -e 's/!\([DR]EF:\)/KEEP \1/' \
+# Strip out blank lines and all comments except
+# "!DEF:", "!REF:", "!$omp", "c$omp", and "*$omp"
+sed -e 's/!\([DR]EF:\)/KEEP \1/' -e 's/[!c*]\($omp\)/KEEP \1/' \
   -e 's/!.*//' -e 's/ *$//' -e '/^$/d' -e 's/KEEP \([DR]EF:\)/!\1/' \
+  -e 's/KEEP \($omp\)/!\1/' \
```
tskeith: This turns `c$omp` and `*$omp` into `!$omp`. Is that okay? What do those even mean?

ichoyjx: Yes. OpenMP directive accepts fixed forms (`!$omp`, `c$omp`, and `*$omp`; starts with column 1) and free forms (`!$omp`; mostly used). For testing purposes, all `1.f90`, `2.f90`, and `3.f90` will be using the same free form (`!$omp`), which is okay.

test/semantics/common.sh
```diff
@@ -35,5 +36,7 @@ case $1 in
   (/*) src=$1 ;;
   (*) src=$(dirname $0)/$1 ;;
 esac
+USER_OPTIONS=`sed -n 's/^ *! *OPTIONS: *//p' $src`
+echo $USER_OPTIONS
```
vjayathirtha-nv: Earlier if there were multiple `! OPTIONS:` lines the actual options would be dumped out on multiple lines.
Now they will be listed on the same line separated by a space.  Not saying that it's necessarily wrong, just noting the difference.

ichoyjx: Okay, thanks! I think it won't influence the results. Earlier it was only writing the options into a file and `cat` it during the actual compilation line, which also ended up with the same line.

test/semantics/test_symbols.sh
tskeith: I don't see the point of allowing `c$omp` and `*$omp` in the input when it's just going to be changed to `!$omp` anyway. Why not just require `!$omp`?

ichoyjx: Yeah, that makes sense. I will remove that.

PR #798, opened 2019-10-28
Exit with error status in case of missing listed test

ichoyjx:
For `test_any` tests, if a test is listed in the `CMakeLists.txt` but actually missing the source file, throw an error and fail the test.

### DIFF
--- a/test/semantics/test_any.sh
+++ b/test/semantics/test_any.sh
@@ -63,6 +63,7 @@ function internal_check() {
 
 gr=0
 for input in ${srcdir}/$*; do
+  [[ ! -f $input ]] && die "File not found: $input"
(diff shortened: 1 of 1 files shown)


sscalpone:
Doesn't actually throw an error.  It exits with an error status (if that's what `die` does.)

ichoyjx:
> Doesn't actually throw an error. It exits with error status (if that's what `die` does.)

Yes, bad title... It should exit with an error status. The error message emits during the verbose mode:

```
245: Test command: /scratch/jyang/pgi/f18/openmp-rel/test/semantics/test_any.sh "getsymbols06.f90" "/scratch/jyang/pgi/f18/openmp-rel/tools/f18/bin/f18"
245: Test timeout computed to be: 10000000
245: test_any.sh: File not found: /scratch/jyang/pgi/f18/openmp-rel/test/semantics/getsymbols06.f90
1/1 Test #245: getsymbols06.f90 .................***Failed    0.05 sec

0% tests passed, 1 tests failed out of 1
```

PR #801, opened 2019-10-30
More semantic checks for calls & declarations

klausler:
`call06.f90` and two new tests (`call14.f90` and `misc-declarations.f90`) now pass.

### DIFF
--- a/lib/semantics/tools.h
+++ b/lib/semantics/tools.h
@@ -54,13 +54,15 @@ bool IsGenericDefinedOp(const Symbol &);
 bool IsCommonBlockContaining(const Symbol &block, const Symbol &object);
 bool DoesScopeContain(const Scope *maybeAncestor, const Scope &maybeDescendent);
 bool DoesScopeContain(const Scope *, const Symbol &);
-bool IsUseAssociated(const Symbol *, const Scope &);
+bool IsUseAssociated(const Symbol &, const Scope &);
 bool IsHostAssociated(const Symbol &, const Scope &);
 bool IsDummy(const Symbol &);
 bool IsPointerDummy(const Symbol &);
 bool IsFunction(const Symbol &);
 bool IsPureProcedure(const Symbol &);
 bool IsPureProcedure(const Scope &);
+bool IsBindCProcedure(const Symbol &);
+bool IsBindCProcedure(const Scope &);
 bool IsProcedure(const Symbol &);
 bool IsProcName(const Symbol &symbol);  // proc-name
 bool IsVariableName(const Symbol &symbol);  // variable-name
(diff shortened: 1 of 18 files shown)


test/semantics/misc-declarations.f90
```diff
@@ -0,0 +1,49 @@
...
+! limitations under the License.
+
+! Miscellaneous constraint and requirement checking on declarations:
+! - 8.5.19 constraints on the VOLATILE attribute
+
+module m
+  type :: hasCoarray
+    real :: coarray[:]
+  end type
+  real :: coarray[:]
```
tskeith: I believe both coarrays need to be allocatable or have explicit cobounds.

test/semantics/misc-declarations.f90
```diff
@@ -0,0 +1,49 @@
...
+    external :: notData
+  end subroutine
+  subroutine C867
+    !ERROR: VOLATILE attribute may not apply to a coarray accessed by USE or host association
+    volatile :: coarray
+    !ERROR: VOLATILE attribute may not apply to a type with a coarray ultimate component accessed by USE or host association
+    volatile :: coarrayComponent
+  end subroutine
+  subroutine C868(coarray,coarrayComponent)
+    real, volatile :: coarray[:]
```
tskeith: Same thing here.

test/semantics/misc-declarations.f90
klausler: Right, that colon should be an asterisk.  And we should check C827 & C828.  Will do both.

lib/semantics/tools.h
```diff
@@ -111,6 +113,12 @@ inline bool IsOptional(const Symbol &symbol) {
 inline bool IsIntentIn(const Symbol &symbol) {
   return symbol.attrs().test(Attr::INTENT_IN);
 }
+inline bool IsIntentInOut(const Symbol &symbol) {
+  return symbol.attrs().test(Attr::INTENT_INOUT);
+}
+inline bool IsIntentOut(const Symbol &symbol) {
+  return symbol.attrs().test(Attr::INTENT_OUT);
+}
```
psteinfeld: Just curious.  What's the point of declaring functions to be ```inline```?

klausler: https://en.cppreference.com/w/cpp/language/inline

test/semantics/misc-declarations.f90
klausler: Adding checks for these constraints caught a whole load of additional incorrect usage in other tests, so thanks for noticing.

lib/semantics/tools.h
psteinfeld: The compiler is free to ignore "inline" and to inline functions that do not specify it.  Given that, I see no value to specifying a function as ```inline```.

If there is some value, what criteria should I use to decide to declare a function ```inline```?

klausler: A C++ compiler can't completely ignore `inline`; at the least, it works like `static` to prevent the name from becoming externally visible and causing link-time errors about multiple definitions.  Function definitions in header files outside class definitions need to be `static`, `inline`, or `constexpr`.

Issue #677, opened 2019-08-22
Missing constant evaluation of min, size, and array parameter


sscalpone:
```
k.f90:3:37: error: Must be a constant value
    integer, parameter, public :: n = min(3, 4)
                                      ^^^^^^^^^
k.f90:4:38: error: Must be a constant value
    integer, parameter, public :: sk = min(3, size(logical_kinds))
                                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^
k.f90:5:38: error: Must be a constant value
    integer, parameter, public :: lk = logical_kinds(min(3, size(logical_kinds)))
                                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
```
module k
  use, intrinsic :: iso_fortran_env
  integer, parameter, public :: n = min(3, 4)
  integer, parameter, public :: sk = min(3, size(logical_kinds))
  integer, parameter, public :: lk = logical_kinds(min(3, size(logical_kinds)))
end module k
```

jeanPerier:
Fixed by PR #803. 
Code from issue is now unparsed to:
```
MODULE k
 USE, INTRINSIC :: iso_fortran_env
 INTEGER, PARAMETER, PUBLIC :: n = 3_4
 INTEGER, PARAMETER, PUBLIC :: sk = 3_8
 INTEGER, PARAMETER, PUBLIC :: lk = 4_4
END MODULE k
```

PR #803, opened 2019-10-31
Implement MIN and MAX folding

jeanPerier:
* Use `Extremum<T>` `FoldOperation` to fold MIN and MAX:
* Fix `Extremum<T>` `FoldOperation`:
    * For character, the length is the one of the longest argument
    * For array of all types, Extremum<T> with `Ordering::Less` was behaving like `Ordering::Greater`. This is because the default `ApplyElementwise` for `Operation` was selected and it then called the `Extremum<T>` constructor without the ordering argument (which was an optional defaulted to Greater).
The fix defines a specific handler for `Extremum<T>` and make the ordering argument mandatory to prevent this kind of bug to pass f18 compilation in the future. I also checked not other `Operation` had constructors with optional arguments and `Extremum` was the only one. 
* Fix `intrinsic.cc` `Match` for MIN and MAX:
When provided with two arguments, `Match` was adding an empty 3rd optional actual argument. Later code working on MIN and MAX was not expecting this and failing. The fix prevents this empty argument from being created.

This commit fixes issue #677.

### DIFF
--- a/lib/evaluate/character.h
+++ b/lib/evaluate/character.h
@@ -70,6 +70,17 @@ public:
     return str;
   }
 
+  // Resize adds spaces on the right if the new size is bigger than the
+  // original, or by trimming the rightmost characters otherwise.
+  static Character Resize(const Character &str, std::size_t newLength) {
+    auto oldLength{str.length()};
+    if (newLength > oldLength) {
+      return str + Character(newLength - oldLength, Space());
+    } else {
+      return str.substr(0, newLength);
+    }
+  }
+
 private:
   // Following helpers assume that character encodings contain ASCII
   static constexpr CharT Space() { return 0x20; }
--- a/lib/evaluate/fold.cc
+++ b/lib/evaluate/fold.cc
@@ -527,6 +527,26 @@ Expr<Type<TypeCategory::Integer, KIND>> UBOUND(FoldingContext &context,
 }
 
 template<typename T>
+Expr<T> FoldMINorMAX(
+    FoldingContext &context, FunctionRef<T> &&funcRef, Ordering order) {
+  std::vector<Constant<T> *> constantArgs;
+  for (auto &arg : funcRef.arguments()) {
+    if (auto *cst{FoldConvertedArg<T>(context, arg)}) {
+      constantArgs.push_back(cst);
+    } else {
+      return Expr<T>(std::move(funcRef));
+    }
+  }
+  CHECK(constantArgs.size() > 0);
+  Expr<T> result{std::move(*constantArgs[0])};
+  for (std::size_t i{1}; i < constantArgs.size(); ++i) {
+    Extremum<T> extremum{order, result, Expr<T>{std::move(*constantArgs[i])}};
+    result = FoldOperation(context, std::move(extremum));
+  }
+  return result;
+}
+
+template<typename T>
 Expr<T> FoldOperation(FoldingContext &context, FunctionRef<T> &&funcRef) {
   ActualArguments &args{funcRef.arguments()};
   for (std::optional<ActualArgument> &arg : args) {
@@ -751,6 +771,8 @@ Expr<Type<TypeCategory::Integer, KIND>> FoldIntrinsicFunction(
         ScalarFunc<T, Int4>([&fptr](const Scalar<Int4> &places) -> Scalar<T> {
           return fptr(static_cast<int>(places.ToInt64()));
         }));
+  } else if (name == "max") {
+    return FoldMINorMAX(context, std::move(funcRef), Ordering::Greater);
   } else if (name == "maxexponent") {
     if (auto *sx{UnwrapExpr<Expr<SomeReal>>(args[0])}) {
       return std::visit(
@@ -772,6 +794,8 @@ Expr<Type<TypeCategory::Integer, KIND>> FoldIntrinsicFunction(
           },
           sx->u);
     }
+  } else if (name == "min") {
+    return FoldMINorMAX(context, std::move(funcRef), Ordering::Less);
   } else if (name == "precision") {
     if (const auto *cx{UnwrapExpr<Expr<SomeReal>>(args[0])}) {
       return Expr<T>{std::visit(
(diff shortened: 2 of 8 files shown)


lib/evaluate/character.h
```diff
@@ -70,6 +70,15 @@ template<int KIND> class CharacterUtils {
     return str;
   }
 
+  static Character Pad(const Character &str, std::size_t length) {
```
vjayathirtha-nv: It's not obvious from the `length` variable name if it refers to padding length or new `str` length.
`newLength` might make it clear.

lib/evaluate/intrinsics.cc
```diff
@@ -976,14 +976,16 @@ std::optional<SpecificCall> IntrinsicInterface::Match(
...
   // MAX and MIN (and others that map to them) allow their last argument to
   // be repeated indefinitely.  The actualForDummy vector is sized
   // and null-initialized to the non-repeated dummy argument count,
   // but additional actual argument pointers can be pushed on it
   // when this flag is set.
   bool repeatLastDummy{dummyArgPatterns > 0 &&
       dummy[dummyArgPatterns - 1].optionality == Optionality::repeats};
+  std::size_t nonRepeatedDummys{
+      repeatLastDummy ? dummyArgPatterns - 1 : dummyArgPatterns};
+  std::vector<ActualArgument *> actualForDummy(nonRepeatedDummys, nullptr);
```
vjayathirtha-nv: > // The actualForDummy vector is sized
> // and null-initialized to the non-repeated dummy argument count,
> // but additional actual argument pointers can be pushed on it
> // when this flag is set.

I think the comment about `actualForDummy` needs to be modified. Since additional arguments can't be pushed, and the vector is always constructed to the non-repeated dummy argument count.

vjayathirtha-nv: Also, the commit message could mention this change as well.

lib/evaluate/intrinsics.cc
```diff
@@ -976,14 +976,16 @@ std::optional<SpecificCall> IntrinsicInterface::Match(
...
   }
-  std::vector<ActualArgument *> actualForDummy(dummyArgPatterns, nullptr);
   // MAX and MIN (and others that map to them) allow their last argument to
   // be repeated indefinitely.  The actualForDummy vector is sized
   // and null-initialized to the non-repeated dummy argument count,
   // but additional actual argument pointers can be pushed on it
   // when this flag is set.
   bool repeatLastDummy{dummyArgPatterns > 0 &&
       dummy[dummyArgPatterns - 1].optionality == Optionality::repeats};
+  std::size_t nonRepeatedDummys{
```
klausler: English quirk: the plural of "dummy" is "dummies".

lib/evaluate/character.h
```diff
@@ -70,6 +70,15 @@ template<int KIND> class CharacterUtils {
...
+    auto oldLength{str.length()};
+    if (length > oldLength) {
+      return str + Character(length - oldLength, Space());
+    } else {
+      return str;
```
klausler: What if `length < oldLength`?

lib/evaluate/fold.cc
```diff
@@ -526,6 +526,26 @@ Expr<Type<TypeCategory::Integer, KIND>> UBOUND(FoldingContext &context,
...
+    FoldingContext &context, FunctionRef<T> &&funcRef, Ordering order) {
+  std::vector<Constant<T> *> constantArgs;
+  for (auto &arg : funcRef.arguments()) {
+    if (auto *cst{FoldConvertedArg<T>(context, arg)}) {
+      constantArgs.push_back(cst);
+    } else {
+      return Expr<T>(std::move(funcRef));
+    }
+  }
+  CHECK(constantArgs.size() >= 2);
```
klausler: What if a program contains `MIN(x)`?  Are you sure that we won't try to fold it later, or accept it as an extension?

test/evaluate/folding01.f90
```diff
@@ -100,4 +100,38 @@ module m
...
+  integer, parameter :: res_max_i = max(i1, i2, i3, i4, i5)
+  integer, parameter :: res_min_i = min(i1, i2, i3, i4, i5)
+  logical, parameter :: test_max_i2 = res_max_i.EQ.i2
+  logical, parameter :: test_min_i2 = res_min_i.EQ.i5
+
+  character(*), parameter :: c1 = "elephant", c2="elevator"
+  character(*), parameter :: c3 = "excalibur", c4="z", c5="epsilon"
+  character(*), parameter :: res_max_c = max(c1, c2, c3, c4, c5)
+  character(*), parameter :: res_min_c = min(c1, c2, c3, c4, c5)
+  ! lenght of result is length of longest arguments!
```
klausler: length*

lib/evaluate/fold.cc
jeanPerier: You are right, I do not need to enforce `>=2` when `>=1` is sufficient for what is done here. Changed.

lib/evaluate/character.h
jeanPerier: Right, `Pad` interface was bad. I have renamed the function to `Resize` and it now trims the rightmost characters in such case.

lib/evaluate/intrinsics.cc
jeanPerier: Thanks

lib/evaluate/intrinsics.cc
jeanPerier: Actually, I think the comment is now enforced (previously, it was not completely true that it was sized to the non-repeated dummy argument count, the repeated dummy argument was included).
Also I am not sure I follow your point that additional arguments can't be pushed.
I think they can be pushed and are. See line: https://github.com/flang-compiler/f18/blob/a4a91f2f9b396020208cf6e536d1d3687e1e7fd6/lib/evaluate/intrinsics.cc#L1030
The commit message mentioned something was fixed in `Match`, but was not explicit about what was the fix. I will name the array explicitly in the commit message (it is in the new commit and will get squashed before I merged).

lib/evaluate/character.h
jeanPerier: Thanks, that is indeed ambiguous. I went for `newLength`. Also, I changed the function name to `Resize` so that it can trim if the `newLength` is smaller.

lib/evaluate/intrinsics.cc
vjayathirtha-nv: I think I misread the code, what you said makes sense. Thanks.

PR #805, opened 2019-11-01
Relax the implementation of constraint C1302.

vdonaldson:
When a format item list can be unambiguously partitioned into individual
items even though one or more otherwise required comma separators are omitted,
generate a warning rather than an error.

### DIFF
--- a/lib/common/format.h
+++ b/lib/common/format.h
@@ -102,6 +102,30 @@ private:
     String,  // char-literal-constant or Hollerith constant
   };
 
+  bool mayHaveLeadingInt(TokenKind desc) {
+    switch (desc) {
+    case TokenKind::A:
+    case TokenKind::B:
+    case TokenKind::D:
+    case TokenKind::DT:
+    case TokenKind::E:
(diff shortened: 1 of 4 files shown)


lib/common/format.h
```diff
@@ -102,6 +102,30 @@ template<typename CHAR = char> class FormatValidator {
     String,  // char-literal-constant or Hollerith constant
   };
 
+  bool mayHaveLeadingInt(TokenKind desc) {
```
klausler: This is basically a set membership test; consider using `EnumSet`.

klausler: This member function should be marked `static` or (better, if possible) `static constexpr`.

vdonaldson: Yes, it is.  Doesn't EnumSet bring C++ library requirements with it?

klausler: Whether a dependence on the binary C++ runtime support library is created probably depends on what parts of `EnumSet` one uses.  I think you have a chance to just depend on things that are implemented in the standard headers in this case.

PR #806, opened 2019-11-01
Enable and pass test call07.f90

klausler:
More semantics checking for procedure references.  These checks are for `POINTER` dummy arguments when the corresponding actual argument is not a pointer.  The implementation of the checks reuses code for semantic analysis of pointer assignment statements, after generalizing it a bit.

### DIFF
--- a/lib/evaluate/characteristics.cc
+++ b/lib/evaluate/characteristics.cc
@@ -218,9 +218,9 @@ std::optional<DummyDataObject> DummyDataObject::Characterize(
     const semantics::Symbol &symbol) {
   if (const auto *obj{symbol.detailsIf<semantics::ObjectEntityDetails>()}) {
     if (auto type{TypeAndShape::Characterize(*obj)}) {
-      DummyDataObject result{*type};
+      std::optional<DummyDataObject> result{std::move(*type)};
       using semantics::Attr;
-      CopyAttrs<DummyDataObject, DummyDataObject::Attr>(symbol, result,
+      CopyAttrs<DummyDataObject, DummyDataObject::Attr>(symbol, *result,
           {
               {Attr::OPTIONAL, DummyDataObject::Attr::Optional},
               {Attr::ALLOCATABLE, DummyDataObject::Attr::Allocatable},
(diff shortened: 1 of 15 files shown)


lib/evaluate/characteristics.cc
```diff
@@ -84,7 +84,7 @@ std::optional<TypeAndShape> TypeAndShape::Characterize(
   if (auto type{DynamicType::From(object.type())}) {
     TypeAndShape result{std::move(*type)};
     result.AcquireShape(object);
-    return result;
+    return std::move(result);
```
tskeith: Isn't `std::move` unnecessary when returning a local?

klausler: In this case, the function result type is `optional<>`, so the `move` optimizes its construction.

klausler: However, some compilers are warning about a needless move, so I'll take it out.  Never mind.

lib/semantics/assignment.cc
```diff
@@ -26,210 +26,231 @@
...
-    }
-    if (error.has_value()) {
-      if (auto *msg{messages.Say(*error, lhs.name(), funcName)}) {
-        msg->Attach(lhs.name(), "Declaration of pointer"_en_US);
-        if (ultimate != nullptr) {
-          msg->Attach(ultimate->name(), "Declaration of function"_en_US);
+      if (error.has_value()) {
+        if (auto *msg{Say(*error, description_, funcName)}) {
+          if (ultimate != nullptr) {
+            msg->Attach(ultimate->name(), "Declaration of function"_en_US);
```
tskeith: I don't think this will have a source location if `ultimate` came from a `.mod` file.

klausler: I'll look into it.  If it's a problem here, it's a problem at other sites as well.

tskeith: `SayAlreadyDeclared` in `resolve-names.cc` has one solution to this problem: If the symbol is not use-associated it shows the declaration at the source location of the name. Otherwise it reports the module it was use-associated from.

PR #807, opened 2019-11-02
Resolve extended intrinsic operators

tskeith:
Enhance `ArgumentAnalyzer` to do most of the work for this.
For each kind of operator that might have a user-defined form we follow
this process:
- analyze the arguments
- if the types and shapes match the intrinsic operator do the usual processing
- otherwise attempt to interpret it as a user-defined operator with `TryDefinedOp`

When we fail to resolve an operator, produce different errors depending
on whether there is a user-defined operator available or not.
If there is, report that neither user-defined nor intrinsic operator
worked. If there is not, describe the rules for the intrinsic operator.
In either case, include the type(s) of the operand(s).

Most of the uses of `ArgumentAnalyzer` are in helper functions that
apply to classes of operators.
For consistency, rename `BinaryOperationHelper` to `NumericBinaryOperator`
and `LogicalHelper` to `LogicalBinaryHelper` and introduce `NumericUnaryHelper`
for unary `+` and `-`.  `.NOT.` and `//` are not implemented in helpers.

Replace `success_` with `fatalErrors_` in `ArgumentAnalyzer` for
consistency with `ExpressionAnalyzer`.

Add `NumericOperator` and `LogicalOperator` enums to `Fortran.h` to go
with `RelationalOperator`. Add `AddFortran` functions to each to convert
to a Fortran source string. `RelationalOperator` also has `AllFortranNames`
because there are multiple names for each operator. This replaces
`LogicalOperator` in `expression.h` and the string representation of
the operators in `formatting.cc`.

### DIFF
--- a/lib/semantics/expression.cc
+++ b/lib/semantics/expression.cc
@@ -47,7 +47,12 @@ using MaybeExpr =
 // and appears here in namespace Fortran::evaluate for convenience.
 namespace Fortran::evaluate {
 
+using common::NumericOperator;
 using common::TypeCategory;
+
+static inline std::string ToUpperCase(const std::string &str) {
+  return parser::ToUpperCaseLetters(str);
+}
 
 struct DynamicTypeWithLength : public DynamicType {
   explicit DynamicTypeWithLength(const DynamicType &t) : DynamicType{t} {}
@@ -136,23 +141,46 @@ common::IfNoLvalue<MaybeExpr, WRAPPED> TypedWrapper(
 class ArgumentAnalyzer {
 public:
   explicit ArgumentAnalyzer(ExpressionAnalyzer &context) : context_{context} {}
-  bool success() const { return success_; }
+  bool fatalErrors() const { return fatalErrors_; }
   ActualArguments &&GetActuals() {
-    CHECK(success_);
+    CHECK(!fatalErrors_);
     return std::move(actuals_);
   }
-  template<typename T> void Analyze(const T &x) {
-    actuals_.emplace_back(context_.Analyze(x));
-    success_ &= actuals_.back().has_value();
+  Expr<SomeType> GetAsExpr(std::size_t i) const {
+    return DEREF(actuals_.at(i).value().UnwrapExpr());
+  }
+  void Analyze(const common::Indirection<parser::Expr> &x) {
+    Analyze(x.value());
+  }
+  void Analyze(const parser::Expr &x) {
+    actuals_.emplace_back(AnalyzeExpr(x));
+    fatalErrors_ |= !actuals_.back().has_value();
   }
   void Analyze(const parser::ActualArgSpec &, bool isSubroutine);
 
+  bool IsIntrinsicRelational(RelationalOperator) const;
+  bool IsIntrinsicLogical() const;
+  bool IsIntrinsicNumeric() const;
+  bool IsIntrinsicConcat() const;
+
+  // Find and return a user-defined operator for opr or report an error.
+  // The provided message is used if there is no such operator.
+  MaybeExpr TryDefinedOp(const char *, parser::MessageFixedText &&);
+  MaybeExpr TryDefinedOp(RelationalOperator, parser::MessageFixedText &&);
+
 private:
-  std::optional<ActualArgument> Analyze(const parser::Expr &);
+  std::optional<ActualArgument> AnalyzeExpr(const parser::Expr &);
+  bool AreConformable() const;
+  const Symbol *FindDefinedOp(const char *) const;
+  std::optional<DynamicType> GetType(std::size_t) const;
+  void SayNoMatch(const char *);
+  std::string TypeAsFortran(std::size_t);
+  bool AnyUntypedOperand();
 
   ExpressionAnalyzer &context_;
   ActualArguments actuals_;
-  bool success_{true};
+  parser::CharBlock source_;
+  bool fatalErrors_{false};
 };
 
 // Wraps a data reference in a typed Designator<>, and a procedure
@@ -1713,7 +1741,7 @@ MaybeExpr ExpressionAnalyzer::AnalyzeCall(
   for (const auto &arg : std::get<std::list<parser::ActualArgSpec>>(call.t)) {
     analyzer.Analyze(arg, isSubroutine);
   }
-  if (analyzer.success()) {
+  if (!analyzer.fatalErrors()) {
     // TODO: map non-intrinsic generic procedure to specific procedure
     if (std::optional<CalleeAndArguments> callee{
             GetCalleeAndArguments(std::get<parser::ProcedureDesignator>(call.t),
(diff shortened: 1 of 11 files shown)


lib/semantics/expression.cc
```diff
@@ -136,23 +141,46 @@ common::IfNoLvalue<MaybeExpr, WRAPPED> TypedWrapper(
...
   ActualArguments &&GetActuals() {
-    CHECK(success_);
+    CHECK(!fatalErrors_);
     return std::move(actuals_);
   }
-  template<typename T> void Analyze(const T &x) {
-    actuals_.emplace_back(context_.Analyze(x));
-    success_ &= actuals_.back().has_value();
+  Expr<SomeType> GetAsExpr(std::size_t i) const {
+    return DEREF(actuals_[i].value().UnwrapExpr());
```
klausler: Might want to use `.at(i)` here for bounds checking.

lib/semantics/expression.cc
```diff
@@ -136,23 +141,46 @@ common::IfNoLvalue<MaybeExpr, WRAPPED> TypedWrapper(
...
+  bool fatalErrors() const { return fatalErrors_; }
   ActualArguments &&GetActuals() {
-    CHECK(success_);
+    CHECK(!fatalErrors_);
     return std::move(actuals_);
   }
-  template<typename T> void Analyze(const T &x) {
-    actuals_.emplace_back(context_.Analyze(x));
-    success_ &= actuals_.back().has_value();
+  Expr<SomeType> GetAsExpr(std::size_t i) const {
```
klausler: Could this return a `const &` and avoid the copy, or might the results outlive the source object?

lib/semantics/expression.cc
```diff
@@ -136,23 +141,46 @@ common::IfNoLvalue<MaybeExpr, WRAPPED> TypedWrapper(
...
   }
   void Analyze(const parser::ActualArgSpec &, bool isSubroutine);
 
+  bool IsIntrinsicRelational(RelationalOperator);
+  bool IsIntrinsicLogical();
+  bool IsIntrinsicNumeric();
+  bool IsIntrinsicConcat();
+
+  // Find and return a user-defined operator for opr or report an error.
+  // `error` is used if there is no such operator.
```
klausler: `error` doesn't appear as an argument name in the following prototypes.

lib/semantics/expression.cc
```diff
@@ -1798,51 +1826,44 @@ MaybeExpr ExpressionAnalyzer::Analyze(const parser::Expr::Parentheses &x) {
...
+  ArgumentAnalyzer analyzer{*this};
+  analyzer.Analyze(x.v);
+  if (analyzer.fatalErrors()) {
+    return std::nullopt;
+  } else if (analyzer.IsIntrinsicLogical()) {
+    return AsGenericExpr(
+        LogicalNegation(std::get<Expr<SomeLogical>>(analyzer.GetAsExpr(0).u)));
+  } else {
+    return analyzer.TryDefinedOp(
+        ".not.", "Operand of %s must be LOGICAL; have %s"_err_en_US);
```
klausler: Why is `.not.` in lower case while `LOGICAL` is upper case?  (rhetorical question, the usage just seems inconsistent)

lib/semantics/expression.cc
```diff
@@ -2400,12 +2423,82 @@ void ArgumentAnalyzer::Analyze(
     }
     actuals_.emplace_back(std::move(*actual));
   } else {
-    success_ = false;
+    fatalErrors_ = true;
+  }
+}
+
+bool ArgumentAnalyzer::IsIntrinsicRelational(RelationalOperator opr) {
```
klausler: Could this (and the following) be `const`?

lib/semantics/expression.cc
```diff
@@ -2439,6 +2532,68 @@ std::optional<ActualArgument> ArgumentAnalyzer::Analyze(
   }
 }
 
+bool ArgumentAnalyzer::AreConformable() const {
+  CHECK(!fatalErrors_ && actuals_.size() == 2);
+  return evaluate::AreConformable(*actuals_[0], *actuals_[1]);
+}
+
+const Symbol *ArgumentAnalyzer::FindDefinedOp(const char *opr) {
```
klausler: Could be `const`?

lib/semantics/expression.cc
```diff
@@ -2439,6 +2532,68 @@ std::optional<ActualArgument> ArgumentAnalyzer::Analyze(
...
+
+std::string ArgumentAnalyzer::TypeAsFortran(std::size_t i) {
+  if (std::optional<DynamicType> type{GetType(i)}) {
+    return type->category() == TypeCategory::Derived
+        ? "TYPE("s + type->AsFortran() + ')'
+        : type->category() == TypeCategory::Character
+            ? "CHARACTER(KIND="s + std::to_string(type->kind()) + ')'
+            : ToUpperCase(type->AsFortran());
+  } else {
+    return "untyped"s;
```
klausler: The suffix is optional here, if you don't want to use it, I think.

lib/semantics/expression.cc
```diff
@@ -2439,6 +2532,68 @@ std::optional<ActualArgument> ArgumentAnalyzer::Analyze(
...
+  }
+}
+
+bool ArgumentAnalyzer::AnyUntypedOperand() {
+  for (const auto &actual : actuals_) {
```
klausler: Might be more natural as a `std::find()`, might not.

lib/semantics/expression.cc
tskeith: Done

lib/semantics/expression.cc
tskeith: These get passed as r-value references to `NumericOperation`, etc. so we would need a copy anyway.

lib/semantics/expression.cc
tskeith: The generic symbol is created with the name "operator(.not.)" in lower case, from the cooked source. But when it's referenced in an error message it is converted to ".NOT.".

lib/semantics/expression.cc
tskeith: Yes, done

lib/semantics/expression.cc
tskeith: Done

lib/semantics/expression.cc
tskeith: OK

lib/semantics/expression.cc
tskeith: It is a little shorter, but more complex, so I left it as-is.

lib/semantics/expression.cc
tskeith: These get passed as r-value references to NumericOperation, etc. so we would need a copy anyway.

lib/semantics/expression.cc
tskeith: I've updated the comment.

PR #808, opened 2019-11-03
[OpenMP] Semantic check for Master construct

Sameeranjoshi:
Hi, I am unable to figure out the error for the master construct implementation.

**$ f18 -fopenmp hello.F90 -fdebug-semantics**

**fatal internal error**: CHECK(!ompContext_.empty()) failed at /home/pc/Workspace/tool-chain/F18/f18/lib/semantics/check-omp-structure.h(154)
Aborted (core dumped)

for following test: containing a clause after master construct(which is not expecting any clauses)

program hello
        implicit none
        !$OMP PARALLEL NUM_THREADS(4)
        !$OMP MASTER NUM_THREADS(4)
        print *, "Hello World"
        !$OMP END MASTER
        !$OMP END PARALLEL
        print *, "out of loop"
end program

Will a new case 
**end-master -> END MASTER** needed to add ?
Thanks


test/semantics/omp-clause-validity01.f90
```diff
@@ -377,6 +377,15 @@
     a = 3.14
   enddo
 
+! 2.13.1 master
+
+  !$omp parallel
+  !ERROR: NUM_THREADS clause is not allowed on the MASTER directive
```
Sameeranjoshi: No clause is allowed on master construct, currently just a single case is written considering num_threads as clause , do we need all tests with all the remaining clauses?

Sameeranjoshi:
Hi please check the test case.Thanks.

test/semantics/omp-clause-validity01.f90
DavidTruby: I think just checking one or a few is probably fine, writing one for every clause would be a large amount of effort for not a lot of gain

test/semantics/omp-clause-validity01.f90
```diff
@@ -377,6 +377,15 @@
     a = 3.14
   enddo
 
+! 2.13.1 master
+
```
DavidTruby: Could you also add a simple test here that doesn't have an error?

Sameeranjoshi:
Hi any pointers on **fatal internal error: CHECK(!ompContext_.empty())** core dumped issue?

ichoyjx:
> Hi any pointers on **fatal internal error: CHECK(!ompContext_.empty())** core dumped issue?

It's actually expected when people compile the code with OpenMP directives that we haven't implemented. (just like how you found the issue) Without the directive support, there will be no `ompContext` got pushed into the stack, that's why the `CHECK` fails.

Sameeranjoshi:
@ichoyjx is
https://github.com/flang-compiler/f18/pull/808/commits/b052a710629cc08cda27de579651dd5328e24b9f
above commit enough for supporting master construct?
I haven't allowed any of the clauses similar to workshare.

DavidTruby:
> @ichoyjx is
> [b052a71](https://github.com/flang-compiler/f18/commit/b052a710629cc08cda27de579651dd5328e24b9f)
> above commit enough for supporting master construct?
> I haven't allowed any of the clauses similar to workshare.

The above should be fine as the master construct doesn't support any clauses. By not listing them the semantic analyser will automatically pick up that none are allowed.

PR #809, opened 2019-11-04
Convert more non-nullable pointers to Reference<>s

klausler:


lib/semantics/symbol.h
```diff
@@ -304,14 +304,14 @@ class NamelistDetails {
...
-  void add_object(Symbol &object) { objects_.push_back(&object); }
+  std::list<SymbolRef> &objects() { return objects_; }
+  const std::list<SymbolRef> &objects() const { return objects_; }
+  void add_object(const Symbol &object) { objects_.emplace_back(object); }
   MaybeExpr bindName() const { return bindName_; }
   void set_bindName(MaybeExpr &&expr) { bindName_ = std::move(expr); }
 
 private:
-  std::list<Symbol *> objects_;
+  std::list<SymbolRef> objects_;
```
tskeith: I'm not sure there is a good reason this is a list rather than a vector, so we could use `SymbolVector` here.

lib/semantics/resolve-names.cc
```diff
@@ -1190,9 +1190,10 @@ void OmpVisitor::ResolveOmpObject(
               // 2.15.3 When a named common block appears in a list, it has the
               // same meaning as if every explicit member of the common block
               // appeared in the list
-              for (Symbol *object :
+              for (const Symbol &object :
                   symbol->get<CommonBlockDetails>().objects()) {
-                ResolveOmp(*object, ompFlag);
+                Symbol &mutableObject{*const_cast<Symbol *>(&object)};
```
tskeith: `CommonBlockDetails::objects_` should really be returning `std::list<common::Reference<Symbol>>`, then we wouldn't have to mess with const-correctness.

tskeith: If we leave it as a const_cast, I think `const_cast<Symbol &>(object)` is better.

lib/semantics/symbol.h
klausler: `SymbolVector` works (I tried it).

lib/semantics/resolve-names.cc
klausler: I've been bitten by casting a non-`const` reference type to a `const &`  causing data copies in ways that I don't fully understand, so I wrote it this way in order feel more confident that copies wouldn't happen.

I'll mess with the API into common blocks; maybe the think to do is to have a distinct interface for mutable symbol references as opposed to constant ones.

PR #814, opened 2019-11-05
Changes to check for constraint C1140 --  deallocation of polymorphic entities

psteinfeld:
Changes to check for constraint C1140 which prohibits deallocation of 
polymorphic entities in a DO CONCURRENT.  Deallocation can be caused by exiting
from a block where the entity is declared, from an assignment, and from direct
deallocation.

Section 9.7.3.2 specifies the situations that might cause deallocation of a polymorphic entity.  The ones that are applicable to a DO CONCURRENT are exiting from a block that declares such variables, intrinsic assignment, and an actual DEALLOCATE statement.  This section also specifies (paragraph 8) that deallocation of a derived type causes deallocation of all of its allocatable subobjects.

Section 10.2.1.3 specifies what happens during intrinsic assignment.  Paragraph 3 states If the variable is an allocated allocatable variable, it is deallocated if expr is an array of different shape, any corresponding length type parameter values of the variable and expr differ, or the variable is polymorphic and the dynamic type or any corresponding kind type parameter values of the variable and expr differ."  Thus, an allocatable polymorphic variable on the left hand side of an assignment statement gets deallocated.  Paragraph 13 states that "For a noncoarray allocatable component the
following sequence of operations is applied.
(1) If the component of the variable is allocated, it is deallocated."

Thus, a variable on the left-hand side of an assignment statement might have noncorray allocatable components.  Such components will be deallocated.

I added a function to tools.cc to check if a symbol is polymorphic.

I also added a function to check-do.cc to report errors along with a reference
to the declaration of the entity associated with error.  This function allows
for the possibility that the declaration of the entity is in a module.

### DIFF
--- a/lib/semantics/check-do.cc
+++ b/lib/semantics/check-do.cc
@@ -64,6 +64,125 @@ public:
   template<typename T> bool Pre(const T &) { return true; }
   template<typename T> void Post(const T &) {}
 
+  template<typename T> bool Pre(const parser::Statement<T> &statement) {
+    currentStatementSourcePosition_ = statement.source;
+    if (statement.label.has_value()) {
+      labels_.insert(*statement.label);
+    }
+    return true;
+  }
+
+  // C1140 -- Can't deallocate a polymorphic entity in a DO CONCURRENT.
+  // Deallocation can be caused by exiting a block that declares an allocatable
+  // entity, assignment to an allocatable variable, or an actual DEALLOCATE
+  // statement
+  //
+  // Note also that the deallocation of a derived type entity might cause the
+  // invocation of an IMPURE final subroutine.
+  //
+
+  // Predicate for deallocations caused by block exit and direct deallocation
+  static bool DeallocateAll(const Symbol &) { return true; }
+
+  // Predicate for deallocations caused by intrinsic assignment
+  static bool DeallocateNonCoarray(const Symbol &component) {
+    return !IsCoarray(component);
+  }
+
+  static bool WillDeallocatePolymorphic(const Symbol &entity,
+      const std::function<bool(const Symbol &)> &WillDeallocate) {
+    return WillDeallocate(entity) && IsPolymorphicAllocatable(entity);
+  }
+
+  // Is it possible that we will we deallocate a polymorphic entity or one
+  // of its components?
+  static bool MightDeallocatePolymorphic(const Symbol &entity,
+      const std::function<bool(const Symbol &)> &WillDeallocate) {
+    if (const Symbol * root{GetAssociationRoot(entity)}) {
+      // Check the entity itself, no coarray exception here
+      if (IsPolymorphicAllocatable(*root)) {
+        return true;
+      }
+      // Check the components
+      if (const auto *details{root->detailsIf<ObjectEntityDetails>()}) {
+        if (const DeclTypeSpec * entityType{details->type()}) {
+          if (const DerivedTypeSpec * derivedType{entityType->AsDerived()}) {
+            UltimateComponentIterator ultimates{*derivedType};
+            for (const auto &ultimate : ultimates) {
+              if (WillDeallocatePolymorphic(ultimate, WillDeallocate)) {
+                return true;
+              }
+            }
+          }
+        }
+      }
+    }
+    return false;
+  }
+
+  // Deallocation caused by block exit
+  // Allocatable entities and all of their allocatable subcomponents will be
+  // deallocated.  This test is different from the other two because it does
+  // not deallocate in cases where the entity itself is not allocatable but
+  // has allocatable polymorphic components
+  void Post(const parser::BlockConstruct &blockConstruct) {
+    const auto &endBlockStmt{
+        std::get<parser::Statement<parser::EndBlockStmt>>(blockConstruct.t)};
+    const Scope &blockScope{context_.FindScope(endBlockStmt.source)};
+    const Scope &doScope{context_.FindScope(doConcurrentSourcePosition_)};
+    if (DoesScopeContain(&doScope, blockScope)) {
+      for (auto &pair : blockScope) {
+        Symbol &entity{*pair.second};
+        if (IsAllocatable(entity) && !entity.attrs().test(Attr::SAVE) &&
+            MightDeallocatePolymorphic(entity, DeallocateAll)) {
+          context_.SayWithDecl(entity, endBlockStmt.source,
+              "Deallocation of a polymorphic entity caused by block"
+              " exit not allowed in DO CONCURRENT"_err_en_US);
+        }
+        // TODO: Check for deallocation of a variable with an IMPURE FINAL
+        // subroutine
+      }
+    }
+  }
+
+  // Deallocation caused by assignment
+  // Note that this case does not cause deallocation of coarray components
+  void Post(const parser::AssignmentStmt &stmt) {
+    const auto &variable{std::get<parser::Variable>(stmt.t)};
+    if (const Symbol * entity{GetLastName(variable).symbol}) {
+      if (MightDeallocatePolymorphic(*entity, DeallocateNonCoarray)) {
+        context_.SayWithDecl(*entity, variable.GetSource(),
+            "Deallocation of a polymorphic entity caused by "
+            "assignment not allowed in DO CONCURRENT"_err_en_US);
+        // TODO: Check for deallocation of a variable with an IMPURE FINAL
+        // subroutine
+      }
+    }
+  }
+
+  // Deallocation from a DEALLOCATE statement
+  // This case is different because DEALLOCATE statements deallocate both
+  // ALLOCATABLE and POINTER entities
+  void Post(const parser::DeallocateStmt &stmt) {
+    const auto &allocateObjectList{
+        std::get<std::list<parser::AllocateObject>>(stmt.t)};
+    for (const auto &allocateObject : allocateObjectList) {
+      const parser::Name &name{GetLastName(allocateObject)};
+      if (name.symbol) {
+        const Symbol &entity{*name.symbol};
+        const DeclTypeSpec *entityType{entity.GetType()};
+        if ((entityType && entityType->IsPolymorphic()) ||  // POINTER case
+            MightDeallocatePolymorphic(entity, DeallocateAll)) {
+          context_.SayWithDecl(entity, currentStatementSourcePosition_,
+              "Deallocation of a polymorphic entity not allowed in DO"
+              " CONCURRENT"_err_en_US);
+        }
+        // TODO: Check for deallocation of a variable with an IMPURE FINAL
+        // subroutine
+      }
+    }
+  }
+
   // C1137 -- No image control statements in a DO CONCURRENT
   void Post(const parser::ExecutableConstruct &construct) {
     if (IsImageControlStmt(construct)) {
@@ -77,14 +196,6 @@ public:
       }
       msg.Attach(doConcurrentSourcePosition_, GetEnclosingDoMsg());
     }
-  }
-
-  template<typename T> bool Pre(const parser::Statement<T> &statement) {
-    currentStatementSourcePosition_ = statement.source;
-    if (statement.label) {
-      labels_.insert(*statement.label);
-    }
-    return true;
   }
 
   // C1167 -- EXIT statements can't exit a DO CONCURRENT
@@ -199,7 +310,6 @@ private:
     return common::GetPtrFromOptional(std::get<0>(a.t));
   }
 
-  bool anyObjectIsPolymorphic() { return false; }  // FIXME placeholder
   bool fromScope(const Symbol &symbol, const std::string &moduleName) {
     if (symbol.GetUltimate().owner().IsModule() &&
         symbol.GetUltimate().owner().GetName().value().ToString() ==
@@ -303,14 +413,11 @@ public:
       if (IsVariableName(*symbol)) {
         const Scope &variableScope{symbol->owner()};
         if (DoesScopeContain(&variableScope, blockScope_)) {
-          context_
-              .Say(name.source,
-                  "Variable '%s' from an enclosing scope referenced in a DO "
-                  "CONCURRENT with DEFAULT(NONE) must appear in a "
-                  "locality-spec"_err_en_US,
-                  name.source)
-              .Attach(symbol->name(), "Declaration of variable '%s'"_en_US,
-                  symbol->name());
+          context_.SayWithDecl(*symbol, name.source,
+              "Variable '%s' from an enclosing scope referenced in DO "
+              "CONCURRENT with DEFAULT(NONE) must appear in a "
+              "locality-spec"_err_en_US,
+              symbol->name());
         }
       }
     }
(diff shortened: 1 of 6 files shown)


lib/semantics/check-do.cc
```diff
@@ -52,6 +52,20 @@ static void SayWithDo(SemanticsContext &context, parser::CharBlock stmtLocation,
   context.Say(stmtLocation, message).Attach(doLocation, GetEnclosingDoMsg());
 }
 
+static void SayWithDeclaration(SemanticsContext &context,
+    const parser::CharBlock &location, const parser::MessageFixedText &&message,
+    const parser::CharBlock &name, const Symbol &symbol) {
+  if (const auto *details{symbol.detailsIf<UseDetails>()}) {
+    context.Say(location, message, name)
```
tskeith: `message` should not be `const` and the uses should be `std::move(message)`.

lib/semantics/check-do.cc
```diff
@@ -52,6 +52,20 @@ static void SayWithDo(SemanticsContext &context, parser::CharBlock stmtLocation,
...
+        .Attach(details->location(),
+            "It is use-associated with '%s' in module '%s'"_err_en_US,
+            details->symbol().name(), details->module().name());
+  } else {
+    context.Say(location, message, name)
+        .Attach(symbol.name(), "Declaration of '%s'"_en_US, symbol.name());
+  }
+}
+
```
tskeith: This function seems generally useful, so I think it should be in `SemanticsContext` along with the `Say` it is calling. I suggest an interface like this:
```
template<typename... A>
void SayWithDecl(const Symbol &symbol, parser::CharBlock at,
    parser::MessageFixedText &&msg, A &&... args);
```

We use "Declaration" and "Decl" in names throughout the compiler, but "Decl" appears to be more common and has the advantage of being shorter.

lib/semantics/check-do.cc
```diff
@@ -64,6 +78,77 @@ class DoConcurrentBodyEnforce {
...
+        // TODO: Check for deallocation of a variable with an IMPURE FINAL
+        // subroutine
+      }
+    }
+  }
+
+  void Post(const parser::AssignmentStmt &stmt) {
+    const auto &variable{std::get<parser::Variable>(stmt.t)};
+    const auto &variableExpr{variable.typedExpr};
+    if (const SomeExpr * expr{GetExpr(variableExpr)}) {
```
tskeith: You can use `GetExpr(variable)` and eliminate `variableExpr`.

lib/semantics/check-do.cc
```diff
@@ -64,6 +78,77 @@ class DoConcurrentBodyEnforce {
...
+  }
+
+  void Post(const parser::DeallocateStmt &stmt) {
+    const auto &allocateObjectList{
+        std::get<std::list<parser::AllocateObject>>(stmt.t)};
+    for (const auto &allocateObject : allocateObjectList) {
+      const parser::Name &name{GetLastName(allocateObject)};
+      if (name.symbol) {
+        const Symbol &symbol{*name.symbol};
+        if (IsPolymorphic(symbol)) {
```
tskeith: Why does this function use this `IsPolymorphic()` while the one above uses `DynamicType::IsPolymorphic()`. Do they get different results in some cases?

lib/semantics/tools.cc
```diff
@@ -503,6 +503,13 @@ bool IsAssumedLengthCharacterFunction(const Symbol &symbol) {
   return symbol.has<SubprogramDetails>() && IsAssumedLengthCharacter(symbol);
 }
 
+bool IsPolymorphic(const Symbol &symbol) {
+  if (const DeclTypeSpec * type{symbol.GetType()}) {
+    return type->IsPolymorphic();
+  }
+  return false;
```
kiranchandramohan: Can this be rewritten as type && type->IsPolymorphic() ?

lib/semantics/check-do.cc
psteinfeld: Thanks!  I'll make that change.

lib/semantics/check-do.cc
```diff
@@ -64,6 +78,77 @@ class DoConcurrentBodyEnforce {
...
+    if (const SomeExpr * expr{GetExpr(variableExpr)}) {
+      const parser::Name &name{GetLastName(variable)};
+      if (const Symbol * symbol{name.symbol}) {
+        if (const std::optional<evaluate::DynamicType> &type{expr->GetType()}) {
+          if (type->IsPolymorphic()) {
+            SayWithDeclaration(context_, variable.GetSource(),
+                "Deallocation of a polymorphic entity caused by "
+                "assignment not allowed in DO CONCURRENT"_err_en_US,
+                parser::CharBlock{}, *symbol);
+          }
```
kiranchandramohan: Can you add a comment regarding the check here? Shouldn't there be checks regarding difference in types/size, existing allocation status (that might make it a runtime check though).

lib/semantics/check-do.cc
```diff
@@ -64,6 +78,77 @@ class DoConcurrentBodyEnforce {
...
+    const Scope &doScope{context_.FindScope(doConcurrentSourcePosition_)};
+    if (DoesScopeContain(&doScope, blockScope)) {
+      for (auto &pair : blockScope) {
+        Symbol &symbol{*pair.second};
+        if (IsPolymorphic(symbol) && IsAllocatable(symbol) &&
+            !symbol.attrs().test(Attr::SAVE)) {
+          SayWithDeclaration(context_, endBlockStmt.source,
+              "Deallocation of a polymorphic entity caused by block"
+              " exit not allowed in DO CONCURRENT"_err_en_US,
+              parser::CharBlock{}, symbol);
```
kiranchandramohan: Should this be a runtime check? How do we know that the allocatable variable was allocated?

lib/semantics/check-do.cc
psteinfeld: Thanks, Tim.  I'll make that change.

lib/semantics/check-do.cc
psteinfeld: Good question.  I'll change the code above to use the ```IsPolymorphic()``` function.

lib/semantics/tools.cc
psteinfeld: Yep.  But this section of code has been eliminated entirely in response to a comment from Tim.

lib/semantics/check-do.cc
psteinfeld: We don't know.  But note that constraint C1140 states that "A statement that might result in the deallocation of a polymorphic entity shall not appear within a DO CONCURRENT construct."  I interpret the use of the word "might" to say that we should check for this circumstance at compile time.

If this causes too many problems with existing Fortran code, we might want to downgrade this check to a warning.

klausler: There isn't much extant Fortran that uses `DO CONCURRENT`.

test/semantics/doconcurrent08.f90
```diff
@@ -0,0 +1,103 @@
+! Copyright (c) 2019, Arm Ltd.  All rights reserved.
```
tskeith: "Arm Ltd"?

psteinfeld: Yikes!  I copied the wrong copyright message.

lib/semantics/check-do.cc
psteinfeld: Thanks, Tim.  I'll do that.

test/semantics/doconcurrent08.f90
```diff
@@ -0,0 +1,103 @@
...
+
+  class(Base), allocatable :: polyVar
+  type(Base), allocatable :: nonPolyVar
+
+  allocate(ChildType:: polyVar)
+  allocate(nonPolyVar)
+
+  do concurrent (i = 1:10)
+    ! Deallocation of a polymorphic entity
+!ERROR: Deallocation of a polymorphic entity not allowed in a DO CONCURRENT
```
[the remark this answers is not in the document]
psteinfeld: Good spot, Kiran.  I'll make things more consistent.

lib/semantics/check-do.cc
psteinfeld: I added some comments to clarify where deallocation checks are needed.  Assignment statements may require other semantic checks, but this code is only checking for deallocation of polymorphic entities.

lib/semantics/semantics.h
```diff
@@ -135,6 +136,18 @@ class SemanticsContext {
   parser::Message &Say(parser::Message &&msg) {
     return messages_.Say(std::move(msg));
   }
+  template<typename... A>
```
klausler: I just basically wrote the same thing for use in `lib/evaluate`.

lib/semantics/check-do.cc
```diff
@@ -64,6 +64,76 @@ class DoConcurrentBodyEnforce {
...
+  // Deallocation caused by block exit
+  void Post(const parser::BlockConstruct &blockConstruct) {
+    const auto &endBlockStmt{
+        std::get<parser::Statement<parser::EndBlockStmt>>(blockConstruct.t)};
+    const Scope &blockScope{context_.FindScope(endBlockStmt.source)};
+    const Scope &doScope{context_.FindScope(doConcurrentSourcePosition_)};
+    if (DoesScopeContain(&doScope, blockScope)) {
+      for (auto &pair : blockScope) {
+        Symbol &symbol{*pair.second};
+        if (IsPolymorphic(symbol) && IsAllocatable(symbol) &&
```
kiranchandramohan: Will this check cover cases where there are components which are polymorphic and allocatable?

lib/semantics/semantics.h
psteinfeld: Hmmm.  Where's the best place for this function to live?

psteinfeld:
> 
> 
> Consider the following program. It has two do concurrent loops.
> -> I think the assignment in the first do concurrent loop would result in deallocation of the polymorphic variable 'p' in 'n'. This is not currently caught.
> -> The second one results in no deallocation and the current check is correct. Can this be added as a test?
> 
> ```
> program mn
>   type :: t1
>     integer :: x
>   end type
>   type, extends (t1) :: t2
>     class(t1), allocatable :: p
>   end type
> contains
>   subroutine s
>     type(t2) :: n
>     type(t2) :: m
>     allocate(n%p)
>     allocate(m%p)
>     do concurrent (i=1:100)
>       n = m
>     end do
>     do concurrent (i=1:100)
>       n%p%x = m%p%x
>     end do
>   end subroutine
> end program
> ```

Good catch, Kiran!  I'll add an equivalent test to deal with components of derived types that are polymorphic.

lib/semantics/check-do.cc
psteinfeld: Clearly polymorphic components of derived types aren't being handled correctly.  I'm not yet sure what needs to be changed.  Stay tuned ...

lib/semantics/semantics.h
klausler: The library dependence order is common -> parser -> evaluate -> semantics, so probably evaluate.  But leave it here for now, I'm unlikely to merge before you do, and I can clean it up later.

lib/semantics/check-do.cc
psteinfeld: I've pushed another commit that deals with types that have polymorphic components.  Please see if it fixes your test case.  It seems to work for me.

lib/semantics/semantics.h
```diff
@@ -135,6 +136,18 @@ class SemanticsContext {
...
+  void SayWithDecl(const Symbol &symbol, const parser::CharBlock &at,
+      parser::MessageFixedText &&msg, A &&... args) {
+    auto &message{Say(at, msg, symbol.name(), args...)};
```
tskeith: `msg` should be `std::move(msg)`.

lib/semantics/semantics.h
```diff
@@ -135,6 +136,18 @@ class SemanticsContext {
   parser::Message &Say(parser::Message &&msg) {
     return messages_.Say(std::move(msg));
   }
+  template<typename... A>
+  void SayWithDecl(const Symbol &symbol, const parser::CharBlock &at,
+      parser::MessageFixedText &&msg, A &&... args) {
```
tskeith: This implicitly includes the symbol name as an argument to the message before `args`. That isn't obvious so it should be documented. If the caller wants something else as the first argument in the message, they have to do something special with the placeholders in the message. (I don't know what, but I know the feature is there.)

I think it would be better to only pass `args` to `Say` and require the caller to include `symbol.name()` explicitly in the position they want (if at all).

lib/semantics/semantics.h
psteinfeld: Thanks, Tim.  I'll change this.

lib/semantics/check-do.cc
```diff
@@ -64,6 +64,76 @@ class DoConcurrentBodyEnforce {
...
+  // Deallocation caused by block exit
+  void Post(const parser::BlockConstruct &blockConstruct) {
+    const auto &endBlockStmt{
+        std::get<parser::Statement<parser::EndBlockStmt>>(blockConstruct.t)};
+    const Scope &blockScope{context_.FindScope(endBlockStmt.source)};
+    const Scope &doScope{context_.FindScope(doConcurrentSourcePosition_)};
+    if (DoesScopeContain(&doScope, blockScope)) {
+      for (auto &pair : blockScope) {
+        Symbol &symbol{*pair.second};
+        if (IsOrContainsPolymorphicComponent(symbol) && IsAllocatable(symbol) &&
```
kiranchandramohan: Shouldn't the allocatable check be for the polymorphic component?

lib/semantics/check-do.cc
```diff
@@ -64,6 +64,76 @@ class DoConcurrentBodyEnforce {
...
+        // subroutine
+      }
+    }
+  }
+
+  // Deallocation caused by assignement
+  void Post(const parser::AssignmentStmt &stmt) {
+    const auto &variable{std::get<parser::Variable>(stmt.t)};
+    if (const Symbol * symbol{GetLastName(variable).symbol}) {
+      if (IsOrContainsPolymorphicComponent(*symbol)) {
```
kiranchandramohan: Should have the allocatable check also.

lib/semantics/semantics.h
tskeith: It looks like at least this function could call `evaluate::AttachDeclaration` to attach the appropriate message.

lib/semantics/check-do.cc
psteinfeld: D'oh!  Good catch.  I'll add a test for the non-error case, too.

lib/semantics/check-do.cc
psteinfeld: Wow.  Thanks for pointing this out.  I hadn't realized that the ```ALLOCATABLE``` attribute could be associated with a component.

So deallocation of a polymorphic entity can happen when you have something that's allocatable that either is or contains a polymorphic entity.  I'll redo my checking code to test for this more complex condition and add tests to cover the relevant cases.

lib/semantics/semantics.h
psteinfeld: Done.

lib/semantics/semantics.h
psteinfeld: Thanks, Tim.  I'll use ```evaluate::AttachDeclaration```.

PR #815, opened 2019-11-06
Add support for logical abbreviations and .XOR.

tskeith:
Update the grammar to handle logical abbreviations (e.g. `.A.` for `.AND.`)
when the feature is enabled. Only support `.X.` when both XOR and
logical abbreviations are enabled.

Fix the driver to enable logical abbreviations with the
`-flogical-abbreviations` option. This was already documented in
`documentation/Extensions.md`.

Remove `parser::Expr::XOR` from the parse tree and immediately map
`.XOR.` to `.NEQV.` if that feature is enabled. This was already being
done during expression analysis anyway.

Add `LanguageFeatureControl::GetNames` to return all of the names of
a logical or relational operator, depending on which features are
enabled. Use these in both name resolution and expression analysis.
Add `Not` to `LogicalOperator` to help in those cases.

Fix handling of BOZ literals: A numeric operation with one real or
integer operand and the other a BOZ literal is intrinsic.
Also, unary plus with a BOZ literal operand is also intrinsic.

### DIFF
--- a/lib/common/Fortran.cc
+++ b/lib/common/Fortran.cc
@@ -34,22 +34,19 @@ const char *AsFortran(LogicalOperator opr) {
   case LogicalOperator::Or: return ".or.";
   case LogicalOperator::Eqv: return ".eqv.";
   case LogicalOperator::Neqv: return ".neqv.";
+  case LogicalOperator::Not: return ".not.";
   }
 }
 
 const char *AsFortran(RelationalOperator opr) {
-  return *AllFortranNames(opr).begin();
-}
-
-std::vector<const char *> AllFortranNames(RelationalOperator opr) {
   switch (opr) {
     SWITCH_COVERS_ALL_CASES
-  case RelationalOperator::LT: return {"<", ".lt."};
-  case RelationalOperator::LE: return {"<=", ".le."};
-  case RelationalOperator::EQ: return {"==", ".eq."};
-  case RelationalOperator::NE: return {"/=", ".ne.", "<>"};
-  case RelationalOperator::GE: return {">=", ".ge."};
-  case RelationalOperator::GT: return {">", ".gt."};
+  case RelationalOperator::LT: return "<";
+  case RelationalOperator::LE: return "<=";
+  case RelationalOperator::EQ: return "==";
+  case RelationalOperator::NE: return "/=";
+  case RelationalOperator::GE: return ">=";
+  case RelationalOperator::GT: return ">";
   }
 }
 
--- a/lib/common/features.cc
+++ b/lib/common/features.cc
@@ -1,0 +1,63 @@
+// Copyright (c) 2019, NVIDIA CORPORATION.  All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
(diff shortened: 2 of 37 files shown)


lib/common/Fortran.cc
```diff
@@ -34,22 +34,19 @@ const char *AsFortran(LogicalOperator opr) {
...
-std::vector<const char *> AllFortranNames(RelationalOperator opr) {
   switch (opr) {
     SWITCH_COVERS_ALL_CASES
-  case RelationalOperator::LT: return {"<", ".lt."};
-  case RelationalOperator::LE: return {"<=", ".le."};
-  case RelationalOperator::EQ: return {"==", ".eq."};
-  case RelationalOperator::NE: return {"/=", ".ne.", "<>"};
-  case RelationalOperator::GE: return {">=", ".ge."};
-  case RelationalOperator::GT: return {">", ".gt."};
+  case RelationalOperator::LT: return "<";
```
klausler: Are the alternative names intentionally removed?

klausler: Never mind, I get it.

lib/parser/grammar.h
```diff
@@ -178,12 +178,12 @@ constexpr auto namedIntrinsicOperator{
...
     ".NEQV." >> pure(DefinedOperator::IntrinsicOperator::NEQV) ||
     extension<LanguageFeature::XOROperator>(
-        ".XOR." >> pure(DefinedOperator::IntrinsicOperator::XOR)) ||
+        ".XOR." >> pure(DefinedOperator::IntrinsicOperator::NEQV)) ||
     extension<LanguageFeature::LogicalAbbreviations>(
         ".N." >> pure(DefinedOperator::IntrinsicOperator::NOT) ||
         ".A." >> pure(DefinedOperator::IntrinsicOperator::AND) ||
         ".O." >> pure(DefinedOperator::IntrinsicOperator::OR) ||
-        ".X." >> pure(DefinedOperator::IntrinsicOperator::XOR))};
+        ".X." >> pure(DefinedOperator::IntrinsicOperator::NEQV))};
```
klausler: This last line might need to be nested in two `extension<>` calls if you don't want to enable `.X.` unless both features have been enabled.

tskeith: Right -- I've fixed it.

lib/semantics/expression.cc
```diff
@@ -2436,10 +2441,28 @@ bool ArgumentAnalyzer::IsIntrinsicRelational(RelationalOperator opr) const {
...
 }
 
-bool ArgumentAnalyzer::IsIntrinsicNumeric() const {
-  return IsNumericTypeCategory(GetType(0)->category()) &&
-      (actuals_.size() == 1 ||
-          (AreConformable() && IsNumericTypeCategory(GetType(1)->category())));
+bool ArgumentAnalyzer::IsIntrinsicNumeric(NumericOperator opr) const {
+  std::optional<DynamicType> type0{GetType(0)};
+  if (actuals_.size() == 1) {
+    if (IsBOZLiteral(0)) {
```
psteinfeld: Not sure if it's significant, but the two calls to ```IsBOZLiteral(0)``` could be hoisted out of the outer ```if``` statement.

lib/common/features.cc
```diff
@@ -0,0 +1,63 @@
+// Copyright (c) 2018-2019, NVIDIA CORPORATION.  All rights reserved.
```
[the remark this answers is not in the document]
tskeith: Right, I'll fix it.

Issue #694, opened 2019-08-28
Unexpected error when compiling submodule


sscalpone:
```
% flang -c foo.f90
foo.f90:11:19: error: 'ma_create' was not declared a separate module procedure
    module function ma_create() result(this)
                    ^^^^^^^^^
```
```
module mm
  interface ma
    module function ma_create( ) result(this)
      integer :: this
    end function
  end interface
end module

submodule (mm) ma_submodule
  contains
  module function ma_create() result(this)
    integer :: this
  end function
end submodule
```

PR #817, opened 2019-11-08
Fix for #694 - Unexpected error when compiling submodule

kiranktp:
Check was missing to see if the module procedure is declared in a separate interface body before issuing the error.


test/semantics/separate_module_procs_1.f90
```diff
@@ -0,0 +1,61 @@
+! Copyright (c) 2018, NVIDIA CORPORATION.  All rights reserved.
```
kiranchandramohan: Change this to your company's (AMD?) copyright.

sscalpone:
We'll be moving away from copyrights in files soon.
https://llvm.org/foundation/relicensing/

Go ahead and use this if you'd like:

//===-- file/name - File description ----------------------------*- C++ -*-===//
//
// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
// See https://llvm.org/LICENSE.txt for license information.
// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
//
//===----------------------------------------------------------------------===//

lib/semantics/resolve-names.cc
```diff
@@ -2799,15 +2799,32 @@ bool SubprogramVisitor::BeginMpSubprogram(const parser::Name &name) {
...
 bool SubprogramVisitor::BeginSubprogram(
     const parser::Name &name, Symbol::Flag subpFlag, bool hasModulePrefix) {
+  bool isSeparateModuleProc = false;
+  // Check if Subprogram has Module Prefix and Subprogram is not inside Interface Block 
   if (hasModulePrefix && !inInterfaceBlock()) {
-    Say(name, "'%s' was not declared a separate module procedure"_err_en_US);
-    return false;
+    // Check if the Subprogram is declared as separate module procedures
+    auto *symbol{FindSymbol(name)};
+    if (symbol && symbol->has<SubprogramNameDetails>()) {
```
tskeith: Why do you need to do this check rather than just looking directly in the parent scope?

lib/semantics/resolve-names.cc
```diff
@@ -2799,15 +2799,32 @@ bool SubprogramVisitor::BeginMpSubprogram(const parser::Name &name) {
...
+      symbol = FindSymbol(currScope().parent(), name);
+    }
+    if (symbol)
+      isSeparateModuleProc = symbol->IsSeparateModuleProc();
+
+    // Issue error: Subprogram is not declared as separate module procedure
```
tskeith: This comment and some of the other just repeat what the code is doing and so is unnecessary. Just use inline comments to clarify the code (or make the code clearer so they aren't necessary).

test/semantics/separate_module_procs_2.f90
```diff
@@ -0,0 +1,41 @@
...
+! WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+! See the License for the specific language governing permissions and
+! limitations under the License.
+
+module m1
+  interface ma
+  end interface
+end module
+
+submodule (m1) ma_submodule
```
tskeith: `m1` should declare a separate module procedure or it can't be used as the ancestor module of a submodule. We don't detect this error yet, but should some day.

test/semantics/separate_module_procs_2.f90
```diff
@@ -0,0 +1,41 @@
...
+end module
+
+submodule (m2) mb_submodule
+  contains
+  !ERROR: 'mb_create_new_sub' was not declared a separate module procedure
+  module SUBROUTINE  mb_create_new_sub() 
+    integer :: this
+    print *, "Hello"
+  end SUBROUTINE mb_create_new_sub
+end submodule
```
tskeith: Here are some other cases that should probably be tested:
- the name is declared in an interface but without the module prefix
- the name is declared but not in an interface block
- the name is declared in a parent submodule rather than a module

test/semantics/separate_module_procs_1.f90
```diff
@@ -0,0 +1,61 @@
...
+!DEF: /m2/mb_submodule Module
+submodule (m2) mb_submodule
+  contains
+  !DEF: /m2/mb_submodule/mb_create_sub MODULE, PUBLIC (Subroutine) Subprogram
+  module subroutine  mb_create_sub
+    !DEF: /m2/mb_submodule/mb_create_sub/this ObjectEntity INTEGER(4)
+    integer this
+    print *, "Hello"
+  end subroutine mb_create_sub
+end submodule
```
tskeith: I'm not sure this test adds anything. I suggest adding your non-error cases to the other test and eliminating this one.

test/semantics/separate_module_procs_1.f90
kiranktp: Almost all files had NVIDIA copyright, so i did not try to change it.
I will go ahead and use the new license for the files that i added.

lib/semantics/resolve-names.cc
kiranktp: In couple of places before searching in parent scope, this check present. i could not get a scenario where this check is valid/invalid so i just retained it here as well. 
I will remove this check.

lib/semantics/resolve-names.cc
kiranktp: Sure. i would so..

test/semantics/separate_module_procs_1.f90
klausler: For what it's worth, we've been typically using hyphens (`-`) rather than underscores (`_`) in file names.

test/semantics/separate_module_procs_1.f90
kiranktp: I tried to add it under module file generation scenario, I was facing a lot of issues while adding sub module info. So i went ahead and added under symbol generation category. 
Can we have one more category under testing which just checks for clean compilation?

lib/semantics/resolve-names.cc
```diff
@@ -2799,15 +2799,32 @@ bool SubprogramVisitor::BeginMpSubprogram(const parser::Name &name) {
   return true;
 }
 
+// A subprogram declared with SUBROUTINE or function
+// If the subprogram is not in interface block, check if its declared as separate 
+// module procedures
 bool SubprogramVisitor::BeginSubprogram(
     const parser::Name &name, Symbol::Flag subpFlag, bool hasModulePrefix) {
+  bool isSeparateModuleProc = false;
```
klausler: We prefer braced initialization `{false}` when either form will work, as here.

lib/semantics/resolve-names.cc
```diff
@@ -2799,15 +2799,32 @@ bool SubprogramVisitor::BeginMpSubprogram(const parser::Name &name) {
...
+  // Check if Subprogram has Module Prefix and Subprogram is not inside Interface Block 
   if (hasModulePrefix && !inInterfaceBlock()) {
-    Say(name, "'%s' was not declared a separate module procedure"_err_en_US);
-    return false;
+    // Check if the Subprogram is declared as separate module procedures
+    auto *symbol{FindSymbol(name)};
+    if (symbol && symbol->has<SubprogramNameDetails>()) {
+      symbol = FindSymbol(currScope().parent(), name);
+    }
+    if (symbol)
```
klausler: We always use braces around the bodies of structured statements.

test/semantics/separate_module_procs_1.f90
tskeith: The error tests category handles clean compilation as well. If an unexpected error occurs the test fails. So you can add your non-error cases to the other test.

test/semantics/separate_module_procs_2.f90
kiranktp: I was not aware of this requirement. At least the documents that i read about sub modules, did not mention it. We can raise a new issue to handle this scenario.
I wrote this test case to show the error condition of not defining separate module procedure.
Tim, Could you please share any document which explains about this requirement?

test/semantics/separate_module_procs_2.f90
kiranktp: Thanks for the details Tim, I will try to cover these scenarios as well in this test case.

lib/semantics/resolve-names.cc
kiranktp: I will follow braced initialization.

test/semantics/separate_module_procs_2.f90
klausler: It follows directly from the standard.  If a module declares no separate module procedure, its submodules can have no effect on the program.  (See C.9.4(2): "A separate module procedure that is declared in a module to have public accessibility can be accessed by use association even if it is defined in a submodule. No other entity in a submodule can be accessed by use association.")

Put another way: if a submodule and its descendants define no separate module procedures apart from the SMPs that they themselves declare, they're all useless.  If a module declares no SMP, all of its submodules must be useless as a consequence.

tskeith: The main document you should be using is the Fortran 2018 standard. It says:
> C1412 (R1418) The ancestor-module-name shall be the name of a nonintrinsic module that declares a separate module procedure; ...

I'm not asking you to add that check, just to fix your test program so that the only errors in it are the ones you intend to be there.

tskeith:
Can you update the commit message with a description of the change and take out the "Change-Id".

lib/semantics/resolve-names.cc
```diff
@@ -2799,15 +2799,32 @@ bool SubprogramVisitor::BeginMpSubprogram(const parser::Name &name) {
   return true;
 }
 
+// A subprogram declared with SUBROUTINE or function
+// If the subprogram is not in interface block, check if its declared as separate
```
vjayathirtha-nv: `its` -> `it's`

lib/semantics/resolve-names.cc
```diff
@@ -2799,15 +2799,32 @@ bool SubprogramVisitor::BeginMpSubprogram(const parser::Name &name) {
...
+      symbol = FindSymbol(currScope().parent(), name);
+    }
+    if (symbol)
+      isSeparateModuleProc = symbol->IsSeparateModuleProc();
+
+    // Issue error: Subprogram is not declared as separate module procedure
+    if (!isSeparateModuleProc) {
+      Say(name, "'%s' was not declared a separate module procedure"_err_en_US);
+      return false;
+    }
```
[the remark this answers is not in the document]
klausler: We almost never use assignment in the middle of complicated expressions, though.  Don't sacrifice readability to save a line.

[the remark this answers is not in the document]
vjayathirtha-nv: I agree, I will modify my suggestion.

sscalpone:
@kiranktp Someday we will re-license these files to use the Apache 2.0 license with LLVM exceptions.  This license is the new standard LLVM license.  Do you give us permission to relicense your changes too?

kiranktp:
> @kiranktp Someday we will re-license these files to use the Apache 2.0 license with LLVM exceptions.  This license is the new standard LLVM license.  Do you give us permission to relicense your changes too?

Licencing should be same for all the files in the code base. So it is fine to relicense these changes as well.

PR #818, opened 2019-11-08
Check semantics for actual arguments associated with dummy procedures

klausler:
Extend, enable, and implement code to pass tests/semantics/call09.f90.


lib/evaluate/characteristics.h
```diff
@@ -129,6 +130,9 @@ struct DummyDataObject {
   explicit DummyDataObject(TypeAndShape &&t) : type{std::move(t)} {}
   explicit DummyDataObject(DynamicType t) : type{t} {}
   bool operator==(const DummyDataObject &) const;
+  bool operator!=(const DummyDataObject &that) const {
+    return !(*this == that);
+  }
```
tskeith: Is it possible to write a generic `!=` non-member function and avoid the need for all of these member functions?
```
template<typename T> bool operator!=(const T &x, const T &y) {
  return !(x == y);
}
```

lib/semantics/resolve-names.cc
```diff
@@ -5105,6 +5103,25 @@ const DeclTypeSpec &ConstructVisitor::ToDeclTypeSpec(
 
 // ResolveNamesVisitor implementation
 
+// Ensures that bare undeclared intrinsic procedure names passed as actual
+// arguments get recognized as being intrinsics.
+bool ResolveNamesVisitor::Pre(const parser::ActualArg &arg) {
+  if (const auto *expr{
+          std::get_if<common::Indirection<parser::Expr>>(&arg.u)}) {
+    if (const auto *designator{
+            std::get_if<common::Indirection<parser::Designator>>(
```
tskeith: There is a using-declaration for `Indirection` at the top of this file so these could be shortened to `std::get_if<Indirection<`... .

lib/evaluate/characteristics.h
klausler: I don't know.

lib/semantics/resolve-names.cc
klausler: I'll clean up all the needless `common::` prefixes in this file while I'm here.

lib/semantics/check-call.cc
```diff
@@ -437,6 +450,105 @@ static void CheckExplicitDataArg(const characteristics::DummyDataObject &dummy,
...
+    const auto *argProcSymbol{
+        argProcDesignator ? argProcDesignator->GetSymbol() : nullptr};
+    if (auto argChars{characteristics::DummyArgument::FromActual(
+            "actual argument", *expr, context)}) {
+      if (auto *argProc{
+              std::get_if<characteristics::DummyProcedure>(&argChars->u)}) {
+        characteristics::Procedure &argInterface{argProc->procedure.value()};
+        if (argInterface.attrs.test(
+                characteristics::Procedure::Attr::NullPointer)) {
+        }
```
tskeith: Is something supposed to happen in this `if`?

klausler: Not any more.

PR #821, opened 2019-11-10
[OpenMP] Semantic checks for Atomic construct

Sameeranjoshi:
Considering,
using OmpClauseSet = common::EnumSet<OmpClause, OmpClause_enumSize>;
Is a new EnumSet expected for OmpMemoryClause ? 
I added SEQ_CST in OmpClauseSet currently.

I am unclear about the design pattern for adding semantic checks for Executable directives. Any suggestion on that front?

Testcase1.F90 fails as expected if more than one seq_cst present whereas
Testcase2.F90 passes though I added allowedOnce{OmpClause::SEQ_CST}

**$ f18 -fopenmp -fdebug-semantics ~/temp/Testcase1.F90**
F90-S-0034-Syntax error at or near SEQ_CST (/tmp/f18-4c87.f90: 5)
  0 inform,   0 warnings,   1 severes, 0 fatal for atomic
$ cat Testcase1.F90 
      !$omp parallel num_threads(4)
      **!$omp atomic read seq_cst seq_cst**
                        i = 10
      !$omp end atomic

**$ f18 -fopenmp -fdebug-semantics ~/temp/Testcase2.F90**
$ cat Testcase2.F90 
      **!$omp atomic seq_cst read seq_cst**
                        i = 10
      !$omp end atomic


kiranchandramohan:
The error that you see (Syntax error at or near SEQ_CST)  is not the usual error that is printed when allowed once is specified for a clause. It should be something like the following,
ERROR: At most one SEQ_CST clause can appear on the ATOMIC directive

From a cursory look, the Memory Clause do not seem to checked. So you will probably have to add the following function and do some processing there.
void OmpStructureChecker::Enter(const parser::OmpMemoryClause &x) {

DavidTruby:
The syntax error must mean that having two SEQ_CST on the same atomic directive is being caught earlier in the parser rather than during semantic analysis. I guess that's not a problem, other than that the error message is less descriptive? It would seem weird to change the parser to allow it just so we can fail it in semantic analysis though.

With regards to adding new Enter functions, you will possibly need one for OpenMPAtomicConstruct for some checks, as we don't seem to have that either.

Sameeranjoshi:
How do you in case of AtomicConstruct verify the start and end of atomic directives, because in case of other constructs I see for example, struct OmpBeginBlockDirective and struct OmpEndBlockDirective and CheckMatching() was verifying the construct, where as in case of atomic?

Also using parser::Verbatim 
const auto &dir{std::get <parser::Verbatim >(x.t)};
for any of (atomic read | atomic write| ..... ) has "READ" or "WRITE" as verbatim, which gives the output message as

> ERROR: At most one SEQ_CST clause can appear on the READ directive

kiranchandramohan:
Is the matching check required for atomic construct? Will the parser fail to catch mismatching OpenMP directives involving '!$omp atomic' or 'end atomic' during parsing?

ichoyjx:
> Is the matching check required for atomic construct? Will the parser fail to catch mismatching OpenMP directives involving '!$omp atomic' or 'end atomic' during parsing?

Parser does the matching check for all the "!$omp end atomic". Either the end directive is not present, or it has to be "!$omp end atomic".

ichoyjx:
> The syntax error is because having two SEQ_CST on the same atomic directive is being caught earlier in the parser rather than during semantic analysis. I guess that's not a problem, other than that the error message is less descriptive? It would seem weird to change the parser to allow it just so we can fail it in semantic analysis though.
> 
> With regards to adding new Enter functions, you will possibly need one for OpenMPAtomicConstruct for some checks, as we don't seem to have that either.

The `F90-S-0034-Syntax error at or near SEQ_CST (/tmp/f18-4c87.f90: 5)` error is not from F18, it's from PGI compiler. F18 unparses the program into a temporary .f90 file and compiles it with available PGI compiler. For memory related clause, yes, no checks have been implemented yet.

Sameeranjoshi:
Hi, @ichoyjx https://github.com/flang-compiler/f18/pull/821/commits/b40e11a5bb82fc03606efcf1e596856272d227e8
are more checks required?

Sameeranjoshi:
@ichoyjx tried some checks for above comment.
Inside 
OmpStructureChecker::CheckAllowed() ------
Following condition founds to be failing

if ((GetContext().allowedOnceClauses.test(type) ||
          GetContext().allowedExclusiveClauses.test(type)) &&
      FindClause(type)) 
Which is preventing from printing the actual error message.

After debugging I found that, FindClause() doesn't find the respective entry in below multimap.
std::multimap<OmpClause, const parser::OmpClause *> clauseInfo

I see the clauseinfo is filled at end in CheckAllowed using
SetContextClauseInfo() where as the FindClause() checks before inserting the member in clauseinfo, which returns nullptr hence, the if condition fails.

I tried, creating OmpMemoryClause as an entity similar to OmpClause with adding 
OmpStructureChecker::Enter(const parser::OmpMemoryClause &){
    SetContextClause(x);
}
But this requires overloading of functions associated with OmpClause.
I am unable to figure out the next steps, any pointer to clear the road ahead?

kiranchandramohan:
Did you close because you have a different approach?

Sameeranjoshi:
No, the things were not moving forward since many days for the issue I was facing and I was a blocker for others interested folks to start working on atomic if they ever had any new implementation idea.
Thanks

PR #822, opened 2019-11-10
Remove explicit comparisons to `nullptr` and most usage of `has_value()`

klausler:
There seems to be a consensus that `if (p != nullptr)` and `if (p.has_value())` would be more readable if they were more simply written `if (p)`.

### DIFF
--- a/lib/parser/token-parsers.h
+++ b/lib/parser/token-parsers.h
@@ -140,16 +140,16 @@ public:
           continue;  // redundant; ignore
         }
       }
-      if (!at.has_value()) {
+      if (!at) {
         at = nextCh.Parse(state);
-        if (!at.has_value()) {
+        if (!at) {
           return std::nullopt;
         }
       }
       if (spaceSkipping) {
         if (**at == ' ') {
           at = nextCh.Parse(state);
-          if (!at.has_value()) {
+          if (!at) {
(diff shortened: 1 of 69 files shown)


lib/parser/token-parsers.h
```diff
@@ -289,7 +289,7 @@ struct BOZLiteral {
     space.Parse(state);
     const char *start{state.GetLocation()};
     std::optional<const char *> at{nextCh.Parse(state)};
-    if (!at.has_value()) {
+    if (!at) {
```
tskeith: I'm seeing a bunch of compilation errors with GCC 8.3.0. For example, on this line:
```
../../lib/parser/token-parsers.h:292:9: error: could not convert 'Fortran::parser::operator!<std::optional<const char*> >(at)' from 'Fortran::parser::NegatedParser<std::optional<const char*> >' to 'bool'
```
It doesn't complain about line 143, which looks the same.

klausler: Already fixed; will not merge until all builds and tests are clean.

tskeith: What was wrong? It looks okay to me.

klausler: `operator!()` was already overloaded in the context of parser combinators.

PR #823, opened 2019-11-13
Add CI file for cloud.drone.io

DavidTruby:
Add the configuration file needed to enable CI on cloud.drone.io for F18

Once this is merged someone with commit access can enable CI by logging in to cloude.drone.io with their github account, enabling CI for this repository and changing the config file in the settings from .drone.yml to .drone.star


.drone.star
```diff
@@ -0,0 +1,48 @@
...
+            ],
+
+        }
+
+def main(ctx):
+    return [
+        clang("amd64"),
+        clang("arm64"),
+        gcc("amd64"),
+        gcc("arm64"),
```
AlexisPerry: Looks like we're only testing on AMD and ARM architectures. How easy/hard would it be to add in, say, an x86 targeted build?

DavidTruby: It wouldn't be very hard to add an x86 build in principle, however I'm fairly sure you can't build f18 on x86 as the compile process for parsing.cc will run out before completing.
I might be wrong about this though and I can certainly try it!

DavidTruby: I just realised I may have been unclear/confused in this reply: I'm referring to x86-32 above. If you meant x86-64 in general, drone is just using amd64 to refer to the x86-64 architecture, it's not specifically talking about AMD cpus.

AlexisPerry: Thank you for the clarification.  The fact that amd64 doesn't mean AMD cpus is indeed confusing.

PR #825, opened 2019-11-15
Extend, enable, and pass test call10.f90 (restrictions on PURE subprograms)

klausler:
Some problems in other tests were revealed and repaired.

### DIFF
--- a/lib/semantics/assignment.cc
+++ b/lib/semantics/assignment.cc
@@ -300,15 +300,15 @@ struct WhereContext {
 
 class AssignmentContext {
 public:
-  explicit AssignmentContext(
-      SemanticsContext &c, parser::CharBlock at = parser::CharBlock{})
-    : context_{c}, messages_{at, &c.messages()} {}
+  explicit AssignmentContext(SemanticsContext &c) : context_{c} {}
   AssignmentContext(const AssignmentContext &c, WhereContext &w)
-    : context_{c.context_}, messages_{c.messages_}, where_{&w} {}
+    : context_{c.context_}, at_{c.at_}, where_{&w} {}
   AssignmentContext(const AssignmentContext &c, ForallContext &f)
-    : context_{c.context_}, messages_{c.messages_}, forall_{&f} {}
+    : context_{c.context_}, at_{c.at_}, forall_{&f} {}
 
   bool operator==(const AssignmentContext &x) const { return this == &x; }
+
+  void set_at(parser::CharBlock at) { at_ = at; }
 
   void Analyze(const parser::AssignmentStmt &);
   void Analyze(const parser::PointerAssignmentStmt &);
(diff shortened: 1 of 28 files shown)


lib/semantics/check-purity.cc
```diff
@@ -0,0 +1,79 @@
...
+// limitations under the License.
+
+#include "check-purity.h"
+#include "tools.h"
+#include "../parser/parse-tree.h"
+
+namespace Fortran::semantics {
+void PurityChecker::Enter(const parser::ExecutableConstruct &exec) {
+  if (InPureSubprogram() && IsImageControlStmt(exec)) {
+    context_.messages().Say(GetImageControlStmtLocation(exec),
```
tskeith: I think you can replace `context_.messages().Say` with `context_.Say`.

lib/semantics/check-io.cc
```diff
@@ -706,4 +723,11 @@ void IoChecker::CheckForProhibitedSpecifier(
   }
 }
 
+void IoChecker::CheckForPureSubprogram() const {  // C1597
+  if (context_.location() &&
+      FindPureProcedureContaining(&context_.FindScope(*context_.location()))) {
```
tskeith: Can `context_.location()` ever be null? It looks like this is only called for a *Stmt which is wrapped in a Statement, which would have provided the location.

lib/semantics/check-declarations.h
```diff
@@ -18,6 +18,7 @@
 #define FORTRAN_SEMANTICS_CHECK_DECLARATIONS_H_
 namespace Fortran::semantics {
 class SemanticsContext;
+class Symbol;
```
tskeith: ?

lib/semantics/check-io.cc
klausler: It's an `optional`.  Would you prefer a `CHECK()`?

lib/semantics/check-declarations.h
klausler: No longer needed; removed.

lib/semantics/check-purity.cc
klausler: Ok.

lib/semantics/check-io.cc
klausler: Done.

lib/semantics/check-declarations.cc
```diff
@@ -119,6 +147,23 @@ void CheckHelper::Check(const Symbol &symbol) {
       canHaveAssumedParameter |= symbol.has<AssocEntityDetails>();
     }
     Check(*type, canHaveAssumedParameter);
+    if (inPure && inFunction && IsFunctionResult(symbol)) {
+      if (derived && HasImpureFinal(*derived)) {  // C1584
+        messages_.Say(
+            "Result of PURE function may not have an impure FINAL subroutine"_err_en_US);
+      }
+      if (type && type->IsPolymorphic() && IsAllocatable(symbol)) {  // C1585
```
tskeith: `type` can't be null here, it was checked up above.

lib/semantics/check-purity.h
```diff
@@ -0,0 +1,45 @@
+// Copyright (c) 2018-2019, NVIDIA CORPORATION.  All rights reserved.
```
psteinfeld: Copyright date should be "2019".

lib/semantics/assignment.cc
```diff
@@ -337,27 +337,57 @@ class AssignmentContext {
...
-void AssignmentContext::Analyze(const parser::AssignmentStmt &) {
+void AssignmentContext::Analyze(const parser::AssignmentStmt &stmt) {
   if (forall_) {
     // TODO: Warn if some name in forall_->activeNames or its outer
     // contexts does not appear on LHS
   }
   // TODO: Fortran 2003 ALLOCATABLE assignment semantics (automatic
   // (re)allocation of LHS array when unallocated or nonconformable)
+
+  // C1596 checks for polymorphic deallocation in a PURE subprogram
```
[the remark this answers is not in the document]
klausler: That is not a correct reading of the standard.

lib/semantics/check-declarations.cc
```diff
@@ -107,6 +106,35 @@ void CheckHelper::Check(const Symbol &symbol) {
...
+    }
+    if (symbol.attrs().test(Attr::VOLATILE)) {
+      messages_.Say(
+          "A PURE subprogram may not have a variable with the VOLATILE attribute"_err_en_US);
+    }
+    if (IsProcedure(symbol) && !IsPureProcedure(symbol) && IsDummy(symbol)) {
+      messages_.Say(
+          "A dummy procedure of a PURE subprogram must be PURE"_err_en_US);
+    }
+    if (!IsDummy(symbol) && !IsFunctionResult(symbol)) {
```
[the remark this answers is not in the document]
klausler: I don't think that it does, and it certainly shouldn't.  Fortran has both polymorphic allocatable dummy arguments and polymorphic allocatable function results.

lib/semantics/assignment.cc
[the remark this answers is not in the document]
psteinfeld: Where did I get things wrong?  I'd like to know since I'm working on a similar test for DO CONCURRENT.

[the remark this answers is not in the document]
klausler: Assignments to allocatables do not unconditionally cause deallocation.  Only the case of array assignments to allocatable arrays with mismatching shapes will do so, as part of the reallocation process.

PR #826, opened 2019-11-15
Rename features.h

sscalpone:
Rename features.h because many linux systems have an include file called features.h which can complicate makefiles that use a search path.


klausler:
The guard macro in the header file needs to be updated if its name is changed.

The features described in the header are not really flang features; Fortran-features.h would be a better name.

Is this a real problem?  If so, should be rename all our header files?

sscalpone:
> The guard macro in the header file needs to be updated if its name is changed.
> 
> The features described in the header are not really flang features; Fortran-features.h would be a better name.
> 
> Is this a real problem? If so, should be rename all our header files?

There are other conflicts, in lib/evaluate, e.g. complex.h.  I'm willing to wait for a complaint.

tskeith:
I don't understand. When is this a problem?

sscalpone:
> I don't understand. When is this a problem?

When building f18 in the llvm tree, compiling parse-tree.cc, there's eventually an include of the system include file "functional" from idioms.h, which eventually includes /usr/include/.../os_defines.h which includes features.h.  Instead of /usr/include/features.h, the compiler chooses the f18 features.h instead.  This is with gcc 8.3 on a regular linux system.

PR #827, opened 2019-11-15
Fix bad deferences in ProgramTree

tskeith:
We weren't handling MainProgram with no ProgramStmt correctly in
ProgramTree. When building it we were dereferencing an empty optional.
And in ResolveSpecificationParts we were dereferencing a null pointer.


psteinfeld:
The changes look good.  I tried writing a test program that would fail without these changes, but I couldn't figure out how to cause the null pointer dereference.

PR #829, opened 2019-11-15
Disable arm64 gcc builds in CI until they can be fixed

DavidTruby:
Currently the arm64 gcc builds in CI are getting stuck and I can't work out why.
I think we should disable these to get CI working again until they're fixed, which I'll have a look at on my own branch.

### DIFF
--- a/.drone.star
+++ b/.drone.star
@@ -43,6 +43,6 @@ def main(ctx):
         clang("amd64"),
         clang("arm64"),
         gcc("amd64"),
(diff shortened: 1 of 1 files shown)


DavidTruby:
Just to clarify, I'm proposing disabling these as it looks like a false alarm (which is bad). I can use gcc 9 on arm64 locally so it seems to be some function of docker/drone that I am going to look into, but we shouldn't have failing CI with false alarms in the mean time.

PR #830, opened 2019-11-15
Add design doc draft for lowering array expression to FIR

jeanPerier:
This is an early draft of the document describing how array expressions will be lowered to FIR.

Some knowledge of FIR may be required to understand part of what is written inside, but other parts should be generic enough to be understood without understanding too much about it.
See [Fir Language Reference](https://github.com/schweitzpgi/f18/blob/f18/documentation/FIRLangRef.md) to learn more about FIR.
It most likely still contains many typoz, I apologize, I will sanitize it on Monday, but I wanted to post this before the week-end.

### DIFF
--- a/documentation/ArrayExprLoweringToFIR.md
+++ b/documentation/ArrayExprLoweringToFIR.md
@@ -1,0 +1,303 @@
+# Array Expression Lowering
+This document describes how Fortran array expressions will be lowered from their `Fortran::evaluate::Expr` tree representation to a sequence of FIR operations.
+Some higher-level design regarding array expression evaluation is presented in [Array Expression Evaluation Design Document](documentation/ArrayComposition.md).
+The objectives of array expression lowering to FIR are to produce a sequence of FIR operations that:
+1.  Implements Fortran Semantics and f18 semantic extensions
+2.  Exposes enough structure so that array optimizations can later be performed on FIR
+3.  Either implements, or makes it possible to implement as FIR transformations the expressions rewriting exposed in [Array Expression Evaluation Design Document](documentation/ArrayComposition.md).
+## Discussions about the objectives
+### Implementing Fortran and f18 extension semantics 
+Rules to evaluate expressions are listed in Fortran 2018 section 10. For what matters to lowering to FIR, one of the most important part is section 10.1.4 regarding expression evaluation and 10.1.7 regarding evaluation of operands.
+One of the take-away is that Fortran standard gives a lot of freedom to the compiler regarding the order of evaluations of operands and whether they are evaluated or not. The compiler is also allowed to perform mathematically equivalent rewriting of expressions (see 10.1.5.2.4 point 3.).
+A few rules to restrain this freedom during expression lowering to FIR and keep the roles of different compilation phases clear are listed below.
+#### Preserving numerical properties
+##### Rule The lowering of expression to FIR will not perform any expression rewriting while lowering that could lead to different numerical results than the one specified by the semantic context.
+##### Rationale
+Numerical stability is important for many Fortran coders and some others valuate performance more. Options will be required to control numerical stability requirements during compilation. Starting to do a lowering that may change numerical properties will lead to augmenting the number of possible paths in expression lowering to respect these numerical options. Instead, these transformations should be perform directly on `evaluate::Expr` or later on the FIR operations.
+##### Example
+`x + y + z` is not numerically equivalent to `z + x + y`, although it is mathematically.  Lowering to FIR shall not modify the order of the additions in such case.
+##### Plans to enforce the rule while lowering to FIR
+
+This mainly translate in preserving the `evaluate::Expr` tree structure in the FIR operation sequence structure. The operands order should be respected as well as the dependencies. In case the lowering of an expression does not introduce the need to store some sub expressions, the dependency tree of the SSA value holding the result value should match the `evaluate::Expr` tree.
+Concretely, because there is not a one to one match between FIR operations and all Fortran operations and intrinsic procedures. Some rewriting must be done while lowering expressions.
+For instance REAL MAX intrinsics might be lowered to `fir.fcmp` comparison and a MLIR `std.select` operation. In such case, the lowering should ensure that choices and combinations of FIR operations match with the floating-point behavior that is guaranteed for `MAX` given the context (that may include where it is in a program and some compile time options).
+Numerical requirements will need to be accessible in such cases so that lowering can select the right operations and/or propagate the requirements as flags on the operations.
+
+#### Scalar operands evaluation in elemental expressions
+##### Rule
+Scalar operands in elemental expressions must be evaluated only once.
+##### Rational
+10.1.10 point 2. says a scalar in an elemental expression must be treated as an array with _all elements equal_ to the scalar.
+This means that if the compiler cannot prove that re-evaluating such operands will yield the same result, it must be evaluated only once while evaluating the elemental expression. In other cases, a lowering that re-evaluate the scalar sub-expression for every element would be standard compliant. 
+However, in general, skipping useless evaluation should be a good thing regarding performance. So, rather than to do an analysis regarding whether or scalar operand must not be re-evaluated, it is easier and performance friendly to request that they should be evaluated only once.
+##### Example
+In `array + foo(x)` where `foo(x)` is scalar, foo will only be called once.
+##### Plans to enforce the rule while lowering to FIR
+The evaluation of scalar operands will be inserted before the loop evaluating the elemental expression. In the example above, the generated FIR should look like:
+```
+%1 = fir.call foo %x
+fir.loop %i 1 to n {
+  %2 = fir.extract_value %array %i 
+  %3 = fir.fadd  %2 %1
+  %4 = fir.coordinate_of %result %i
+  fir.store %3 to %4
+} 
+```
+#### Short-circuiting
+##### Rule
+Expression lowering to FIR must insert short-circuiting of operand evaluation that are guaranteed to not be performed in contexts defined in by f18 semantic extensions. Other short-circuiting opportunities should be left to later optimization phases.
+##### Rational
+If it needs to be guaranteed something will not be evaluated if unneeded, then that is semantic related, and it is unsafe to assume later FIR transformation and processing will respect this if short-circuiting was not explicitly inserted while lowering the expression.
+However, in general, the decision of whether to short-circuit an evaluation or not should be left to the optimizers based on the cost of the useless operation compared to the cost of the potential branch. Therefore, expression lowering should not add short-circuits that are not required because these cannot always be reversed in FIR (unless FIR transformations take the liberty to violate semantics which they probably should not).
+Note that it is unclear how well FIR transformation will be able to introduce short-circuits that are rather obvious looking at the `evaluate::Expr` tree, so this rule might get revisited.
+So far, the required short-circuits are:
+- `b.OR.SE`, `b.AND.SE` -> do not evaluate `SE` when `b` is false and `SE` has side effects.
+- `MERGE(SE1, SE2, b)` -> when `b` is a scalar, either `SE1` or `SE2` evaluation has side effects.
+Side effects should be understood in the broad sense (array accesses can cause segfaults, floating point operations can raise signals).
+##### Example
+In `b.AND.foo()` the call to `foo` shall not be performed if `b` is false.
+##### Plans to enforce the rule while lowering to FIR
+It is possible to use `fir.where` to implement a simple if/else logic without having to build a control flow graph using `br_cond` and basic blocks. Although both of these implementation are equivalent, `fir.where` gives more structural guarantees than basic block logic and is expected to interact in better way with optimizations. It is also more readable.
+However, with a `fir.where`, the result of such expression will need to be materialized (with basic blocks logic, one could have taken advantages of phi nodes scalar values, but FIR/LLVM transformation should anyway get rid of scalar temporary in such cases).
+The fir for the example above will look like:
+```
+%1 = load %b
+fir.where %1 {
+// b true case  
+  %1 = fir.call foo
+   fir.store %1 tmp 
+} otherwise {
+  %2 = constant 0
+  fir.store %2 tmp
+}
+```
+### Exposing enough structure so that array optimizations can be performed on FIR
+Because there are no FIR array optimizations yet, it is not entirely defined what this requirement implies yet. However, one of the objectives of MLIR is to be able to represent array layouts and access patterns as affine expressions. These affine expressions can then be composed allowing generalized loop and data transformations.
+Of course, this only make sense for array expressions that have such affine properties, mainly elemental and transformational operations. There is not much one can do to transform opaque function calls involving arrays into some affine operation.
+#### FIR operations to express array operations
+In FIR, affine expressions can be implemented using the `fir.loop`, `fir.where`, `fir.extract_value` and `fir.coordinate_of` operations in combinations with the `fir.array<>` type and the `fir.box<>` type that can hold affine maps encoding linear access patterns (see [FIR  Language reference](documentation/FIRLangRef.md)).
+#### Views over array in FIR
+In FIR, one can avoid copying array sections by creating `fir.box` with that holds the memory reference to the base variable as well as the subscript describing the re-mapping.
+For instance, `A(10:20:2, :)` where `A` is `REAL, DIMENSION(20, 10)` can be written in FIR as:
+```
+//%a is the memory reference to A storage, it is of type !fir.ref<fir.array<20:5:fir.real<4>>>
+%0 = fir.gendims 2*m+1, 4*m, 2
+%1 = fir.gendims 1, n, 1
+%asection = fir.embox %a, %0, %1
+// %asection has type !fir.box<fir.ref<fir.array<5:10:fir.real<4 >>>>
+```
+Then the element (i, j) of `A(10:20:2, :)`  can be accessed with:
+```
+// fir.extract_value takes care of applying the index mapping of the array section over A.
+%asectionij = fir.extract_value %asection, %i, %j
+// % asectionij has type !fir.real<4>
+```  
+No copy into a temporary was needed to implement the example above. `fir.box` allows one to represent array and manipulate subsections in FIR without materializing them. It can be seen as if we had created a Fortran pointer over `A` (except such object would have had the type `%fir.box<fir.ptr< fir.array<5:10:fir.real<4 >>>`, notice the `fir.ptr` instead of `fir.ref`).
+
+One notable difference between using `fir.box` and materializing a section inside a temporary is that its value its not “fixed”, the evaluation of the section is delayed until it is needed. This introduce the risk that `A` variable is modified in-between the array emboxing and its usage. In the context of expression evaluation, the only thing that could modify another variable appearing inside the expression evaluation are function call (potentially hidden in user defined operators).
+Fortran Standard prohibit this by stating in section 10.1.4 point 2
+> [..] Except in those cases: the evaluation of a function reference shall neither affect nor be affected by the evaluation of any other entity within the statement
+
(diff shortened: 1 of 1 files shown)


documentation/ArrayExprLoweringToFIR.md
```diff
@@ -0,0 +1,303 @@
...
+Concretely, because there is not a one to one match between FIR operations and all Fortran operations and intrinsic procedures. Some rewriting must be done while lowering expressions.
+For instance REAL MAX intrinsics might be lowered to `fir.fcmp` comparison and a MLIR `std.select` operation. In such case, the lowering should ensure that choices and combinations of FIR operations match with the floating-point behavior that is guaranteed for `MAX` given the context (that may include where it is in a program and some compile time options).
+Numerical requirements will need to be accessible in such cases so that lowering can select the right operations and/or propagate the requirements as flags on the operations.
+
+#### Scalar operands evaluation in elemental expressions
+##### Rule
+Scalar operands in elemental expressions must be evaluated only once.
```
vjayathirtha-nv: Can this be extended to `pure` functions?

documentation/ArrayExprLoweringToFIR.md
```diff
@@ -0,0 +1,303 @@
...
+##### Rational
+If it needs to be guaranteed something will not be evaluated if unneeded, then that is semantic related, and it is unsafe to assume later FIR transformation and processing will respect this if short-circuiting was not explicitly inserted while lowering the expression.
+However, in general, the decision of whether to short-circuit an evaluation or not should be left to the optimizers based on the cost of the useless operation compared to the cost of the potential branch. Therefore, expression lowering should not add short-circuits that are not required because these cannot always be reversed in FIR (unless FIR transformations take the liberty to violate semantics which they probably should not).
+Note that it is unclear how well FIR transformation will be able to introduce short-circuits that are rather obvious looking at the `evaluate::Expr` tree, so this rule might get revisited.
+So far, the required short-circuits are:
+- `b.OR.SE`, `b.AND.SE` -> do not evaluate `SE` when `b` is false and `SE` has side effects.
```
vjayathirtha-nv: When `b` is true in the `.OR.` expr and `b` is false in the `.AND.` expr

documentation/ArrayExprLoweringToFIR.md
```diff
@@ -0,0 +1,303 @@
+# Array Expression Lowering
+This document describes how Fortran array expressions will be lowered from their `Fortran::evaluate::Expr` tree representation to a sequence of FIR operations.
+Some higher-level design regarding array expression evaluation is presented in [Array Expression Evaluation Design Document](documentation/ArrayComposition.md).
+The objectives of array expression lowering to FIR are to produce a sequence of FIR operations that:
+1.  Implements Fortran Semantics and f18 semantic extensions
+2.  Exposes enough structure so that array optimizations can later be performed on FIR
+3.  Either implements, or makes it possible to implement as FIR transformations the expressions rewriting exposed in [Array Expression Evaluation Design Document](documentation/ArrayComposition.md).
+## Discussions about the objectives
```
vdonaldson: Are there two potentially separable parts to this lowering process?

I'm wondering if there could (and should) be a first part that is the official lowering from the front end to FIR that leaves arrays, array sections, and scalars as arrays, array sections, and scalars, possibly emboxing whichever of those require that in order to be represented in FIR.  If necessary, any object that requires emboxing could be repeatedly emboxed as often as needed to satisfy correctness requirements, possibly on the expectation that one or more downstream CSE passes could eliminate some of those as applicable.

Once valid "high-level" FIR is generated, array-aware optimization can be performed, with lowering to fir loops done somewhere downstream, possibly as late as during generation of LLVM IR.  If there are two (or more) choices for doing this second lowering step, there is a possibility of doing both, as determined on a case by case basis.

documentation/ArrayExprLoweringToFIR.md
```diff
@@ -0,0 +1,303 @@
...
+Side effects should be understood in the broad sense (array accesses can cause segfaults, floating point operations can raise signals).
+##### Example
+In `b.AND.foo()` the call to `foo` shall not be performed if `b` is false.
+##### Plans to enforce the rule while lowering to FIR
+It is possible to use `fir.where` to implement a simple if/else logic without having to build a control flow graph using `br_cond` and basic blocks. Although both of these implementation are equivalent, `fir.where` gives more structural guarantees than basic block logic and is expected to interact in better way with optimizations. It is also more readable.
+However, with a `fir.where`, the result of such expression will need to be materialized (with basic blocks logic, one could have taken advantages of phi nodes scalar values, but FIR/LLVM transformation should anyway get rid of scalar temporary in such cases).
+The fir for the example above will look like:
+```
+%1 = load %b
+fir.where %1 {
```
vdonaldson: Minor - Double check SSA index numbering?

documentation/ArrayExprLoweringToFIR.md
```diff
@@ -0,0 +1,303 @@
...
+Of course, this only make sense for array expressions that have such affine properties, mainly elemental and transformational operations. There is not much one can do to transform opaque function calls involving arrays into some affine operation.
+#### FIR operations to express array operations
+In FIR, affine expressions can be implemented using the `fir.loop`, `fir.where`, `fir.extract_value` and `fir.coordinate_of` operations in combinations with the `fir.array<>` type and the `fir.box<>` type that can hold affine maps encoding linear access patterns (see [FIR  Language reference](documentation/FIRLangRef.md)).
+#### Views over array in FIR
+In FIR, one can avoid copying array sections by creating `fir.box` with that holds the memory reference to the base variable as well as the subscript describing the re-mapping.
+For instance, `A(10:20:2, :)` where `A` is `REAL, DIMENSION(20, 10)` can be written in FIR as:
+```
+//%a is the memory reference to A storage, it is of type !fir.ref<fir.array<20:5:fir.real<4>>>
```
vdonaldson: Minor - Does FIR/MLIR use _x_ to separate triplet component values, rather than Fortran's _:_ separator?
I believe the extent of a 10:20:2 slice triplet is 6, rather than 5.
Some of the details for this example, such as references to _m_ and _n_, may be from a later example rather than this one.

documentation/ArrayExprLoweringToFIR.md
```diff
@@ -0,0 +1,303 @@
...
+Its main drawback is that a lot of temporaries may be created, and it is not entirely sure later optimizations will be able to get rid of them and to merge the loops. Even if it is, it may consume “optimization time and resource” to undo something that could have been done right the first time with only a little more work.
+
+This could be implemented as a single pass over the expression tree where the lowering for each expression node first asks for the lowering of its operands and expect to get a memory reference that materializes it (or an SSA value if the operand is scalar). In case the node being processed is an elemental operation, a `fir.loop` is generated to implement it. If a node is an array section or a call to a transformational, a box is created over the data reference or the result of the transformational argument expression evaluation.
+
+Short-circuits are generated also generated outside of any fir-loop which already is the case of other operands so nothing on top of generating the `fir.where` needs to be done.
+
+## Further Features that Expression Lowering
```
psteinfeld: I don't understand this title.  Did you mean to say "Further Features of Expression Lowering"?

documentation/ArrayExprLoweringToFIR.md
psteinfeld: I think it would be good so summarize the objectives here before going into detail.  I think that there should be four objectives, two of which are described in this document:
  1) To preserve Fortran semantics
  2) To generate code that is able to take advantage of existing and future LLVM/MLIR optimizations.
  3) To generate code that can be parallelized.  I'm too ignorant of LLVM/MLIR to know if this is the same as the previous objective.
  4) To enable other tools to help programmers debug/optimize/measure/understand the generated code.  At the very least, we'll want to be able to map generated instructions and data back to statements and data declarations and references.

documentation/ArrayExprLoweringToFIR.md
```diff
@@ -0,0 +1,303 @@
+# Array Expression Lowering
+This document describes how Fortran array expressions will be lowered from their `Fortran::evaluate::Expr` tree representation to a sequence of FIR operations.
+Some higher-level design regarding array expression evaluation is presented in [Array Expression Evaluation Design Document](documentation/ArrayComposition.md).
```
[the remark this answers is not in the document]
jeanPerier: Indeed, Thanks

documentation/ArrayExprLoweringToFIR.md
jeanPerier: For array sections, that is more or less what is happening, sub-expressions that are array sections are emboxed (*). However, in general FIR does not provide elemental operations, so we must lower to `fir.loop` during the lowering (for instance,  `fir.fadd` for floating-point add expects only scalar so we cannot lower delay lowering of `A+B` to a `fir.loop` in FIR).

(*) There is a discussion below that current `fir.box` being based on memory reference, it is generally unsafe to leave array section as box if it is the expression top level because it is hard to guarantee the memory reference will not been modified when we propagate the box outside of the statement.

documentation/ArrayExprLoweringToFIR.md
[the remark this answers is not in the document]
jeanPerier: Thanks, the previous was a bug.

jeanPerier: Do you mean, if we have `x = foo() + A + foo()` where we know `foo` is `pure`, we should only call `foo` once for the whole expression and not twice ?
If so, you are correct we can remove redundant calls to pure procedure if we can prove they have the same arguments.
However, to me it is an optimization and we should not attempt to do this while lowering. FIR still has knowledge of procedure `pure` attribute. I think Eric already has an experimental pass to remove these duplicate calls.
So lowering the proposed lowering would still lower `foo() + A + foo()` to:
```
%1 = fir.call foo
%2 = fir.call foo
fir.loop %1 = 1 to n {
  %3 = fir.extract_value %a %i
  %4 = fir.addf %1 %3
  %5 = fir.addf %4 %2
  %6 = fir.coordinate_of %res %i
  fir.store %5 to %6
}
```
And the optimizations would be responsible to remove the redundant call to `foo`.

The idea is to keep the lowering as simple as possible if we know the optimization will later be applied on the generated FIR. Here, a later pass would anyway need to be done to remove redundant pure calls in-between expressions.

documentation/ArrayExprLoweringToFIR.md
jeanPerier: Thanks

documentation/ArrayExprLoweringToFIR.md
```diff
@@ -0,0 +1,303 @@
...
+So far, the required short-circuits are:
+- `b.OR.SE`, `b.AND.SE` -> do not evaluate `SE` when `b` is false and `SE` has side effects.
+- `MERGE(SE1, SE2, b)` -> when `b` is a scalar, either `SE1` or `SE2` evaluation has side effects.
+Side effects should be understood in the broad sense (array accesses can cause segfaults, floating point operations can raise signals).
+##### Example
+In `b.AND.foo()` the call to `foo` shall not be performed if `b` is false.
+##### Plans to enforce the rule while lowering to FIR
+It is possible to use `fir.where` to implement a simple if/else logic without having to build a control flow graph using `br_cond` and basic blocks. Although both of these implementation are equivalent, `fir.where` gives more structural guarantees than basic block logic and is expected to interact in better way with optimizations. It is also more readable.
+However, with a `fir.where`, the result of such expression will need to be materialized (with basic blocks logic, one could have taken advantages of phi nodes scalar values, but FIR/LLVM transformation should anyway get rid of scalar temporary in such cases).
```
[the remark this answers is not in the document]
jeanPerier: PHI nodes are related to phi functions in IR with static single assignment form property. Phi functions are used to decide what value an SSA node has in a basic block depending on which other basic block it was entered from. This is to implement something like `if (b){i=1} else {i=2}; j=i;)` without using memory operation for the variable `i`.
Now, it is good that you raise this because LLVM uses phi nodes (https://github.com/tensorflow/mlir/blob/master/g3doc/LangRef.md#blocks) but MLIR uses block arguments (https://github.com/tensorflow/mlir/blob/master/g3doc/LangRef.md#blocks) which are IMHO easier to understand.

So I have rewritten this to:
"one could have taken advantages of [MLIR basic block arguments](https://github.com/tensorflow/mlir/blob/master/g3doc/LangRef.md#blocks) to avoid materialization"

FYI, the illustration of `if (b){i=1} else {i=2}; j=i;)` with basic block arguments where we do need to materialize `i` in memory could be:
```
    cond_br %b ^bb1(1) ^bb2(2)
^bb1(%i : i32)  // %i is a block argument, and we would need a PHI node for it in LLVM IR.
    fir.store %i to %j
```

documentation/ArrayExprLoweringToFIR.md
jeanPerier: You are correct regarding all you points, thanks

documentation/ArrayExprLoweringToFIR.md
```diff
@@ -0,0 +1,303 @@
...
+Fortran Standard prohibit this by stating in section 10.1.4 point 2
+> [..] Except in those cases: the evaluation of a function reference shall neither affect nor be affected by the evaluation of any other entity within the statement
+
+The cases mentioned do not apply inside an expression. However, it is in general unsafe to propagate such box outside of an expression evaluation. This means that if `A(10:20:2, :)` if a top level expression, it is probably safer to materialize into a temporary storage and leave later optimizations determine if it is safe to keep skip the materialization.
+This means for the Fortan code:
+```
+B = A(1:10, :)
+PRINT *m, B
+```
+The bridge to FIR will not replace refences to `B` by the `fir.box` encoding ` A(1:10, :)` because it cannot ensure `A` was not modified in-between without an analysis of A potential definition in-between that does not belong to the bridge towards FIR.
```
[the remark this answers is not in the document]
jeanPerier: OK, I have reworded to:
"The bridge to FIR will not replace the reference to `B` inside the `PRINT` statement by a `fir.box` encoding ` A(1:10, :)`. It would be correct here, but dangerous in general. The reason is that the value represented by this `fir.box` is not constant. This value is the one of `A(1:10,:)` at the moment the `fir.box` is accessed with `fir.extract_value`. Emboxing an array section has the side effect of delaying its evaluation. 
In general, the bridge cannot easily ensure `A` was not modified in-between `B=A(1:10, :)` and `PRINT *m, B`. It would require an analysis of `A` variable definitions that do not belong to the bridge towards FIR."

documentation/ArrayExprLoweringToFIR.md
```diff
@@ -0,0 +1,303 @@
...
+
+An alternative, that requires FIR language modifications, would be to allow `fir.box` to embox array SSA values (`fir.array<T>`) instead of only memory references (like `fir.ref<fir.array<T>>` ).  The value-based aspect would remove such need of performing re-definition analysis to propagate boxes. 
+#### Lowering transformational intrinsic procedures to `fir.box`
+`fir.box` can also be used to implement some transformational intrinsic because it can describe more than array subsection in case it is used with an mlir affine map instead of susbscripts.
```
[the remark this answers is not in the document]
jeanPerier: Thanks, I reworded a bit starting from your version.

documentation/ArrayExprLoweringToFIR.md
```diff
@@ -0,0 +1,303 @@
...
+
+More can be read about MLIR affine maps in the [Affine Dialect Language Reference](https://github.com/tensorflow/mlir/blob/master/g3doc/Dialects/Affine.md)
+#### Expression storage in FIR
+
+A `fir.loop` is more or less a regular counted loop and it is not returning an ssa value that would be the result of the array expression associated to the loop. Hence, it can only have effects on the program through memory accesses. `fir.store` operations in the `fir.loop` must store the array result elements into some memory reference.
+This implies that every `fir.loop` computing an array expression must be given a memory reference to some storage can store the array result. The required size of such storage may know at compile time or may be dynamic.
+Hence, expression lowering will have to handle storage for expressions and may have to introduce dynamic memory allocation and free operations (`fir.allocmem` and `fir.freemem`).
+To simplify the design, we will assume the entry point of array expression lowering is provided with a memory-reference that can store the result. Whether this reference is a temporary and whether it is dynamic should not affect the lowering.
```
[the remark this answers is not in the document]
jeanPerier: I have added sentence: "For instance, when lowering assignment `X = Y + B`, it is the part of the bridge that lowers `parser::AssignStmt` that should be responsible for the storage required to store `Y + B`. This is to de-correlate as much as possible expression lowering from the statements in which expressions may appear."


To answer you question: in some ways, yes, there is something before expression lowering that would do this. However it is not a pass done from the expression lowering facility that would that, but the lowering of the statement where the expression appears (it may require doing some kind of pass over the expression to get some information like the shape with `evaluate::GetSahpe`).

If  `parser::AssignStmt` lowering (that is not done by expression lowering)  allocates a dynamic temporary for the right-hand side evaluation, then it should also be the one the one inserting the clean-up after the copy to the left hand side variable is done. But this should be transparent to expression lowering.

I have some other ideas to do this de-correlation. It would involve the expression facility returning or accepting lambdas that would apply the array expressions given the array indexes passed as argument to the lambda. But that is still in aplha version in my head.

documentation/ArrayExprLoweringToFIR.md
```diff
@@ -0,0 +1,303 @@
...
+However, expression lowering may have to introduce storages to materialize sub-expressions. In such case, expression lowering will be responsible to create the storage and to release them if necessary.
```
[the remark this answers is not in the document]
jeanPerier: Thanks for "storage locations".
If I get your question correctly, yes, there is a difference. Most SSA values created while lowering elemental expressions will be scalar values. These scalar SSA values will be stored into the created storage locations for the arrays.

documentation/ArrayExprLoweringToFIR.md
```diff
@@ -0,0 +1,303 @@
...
+This implies that every `fir.loop` computing an array expression must be given a memory reference to some storage can store the array result. The required size of such storage may know at compile time or may be dynamic.
+Hence, expression lowering will have to handle storage for expressions and may have to introduce dynamic memory allocation and free operations (`fir.allocmem` and `fir.freemem`).
+To simplify the design, we will assume the entry point of array expression lowering is provided with a memory-reference that can store the result. Whether this reference is a temporary and whether it is dynamic should not affect the lowering.
+However, expression lowering may have to introduce storages to materialize sub-expressions. In such case, expression lowering will be responsible to create the storage and to release them if necessary. 
+To simplify, expression lowering will not try to do any kind of clever temporary re-use or pooling. It is expected later transformations should do so based on the dependencies and lifetime of the created temporaries.
```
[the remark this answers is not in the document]
jeanPerier: I am not sure LLVM can do this kind of transformations with _array_ storage locations. But before LLVM, FIR and MLIR transformations should.  Eric has done a "Memory to Register" FIR pass that transforms scalar local variable with a storage location into SSA values (I am not sure it currently works on `fir.ref<fir.array<>>`, but it probably should in the future).

MLIR/FIR is meant to be able to do such high transformations, but you are pointing to an area that is still a bit blurry to me and will probably takes some trial and errors. I am unable to provide a clear list of requirements yet (that is why the applying the functional approach while doing expression lowering, which would minimize such temporary storage locations, seems safer to me.)

documentation/ArrayExprLoweringToFIR.md
jeanPerier: Oups, something like this, yes. It is now: "Possible Future Improvements of Array Expression Lowering"

documentation/ArrayExprLoweringToFIR.md
jeanPerier: I feel like it is slightly repeating what is just above for some points. I agree I said nothing about parallelism and tools though, I will think a bit more about it.

documentation/ArrayExprLoweringToFIR.md
schweitzpgi: Re: "an experimental pass"

That is correct. There is already a CSE pass that merges duplicate pure function calls.

jeanPerier:
It would not bring much to merge this document in the main branch at the moment.

PR #831, opened 2019-11-16
Threadprivate

Sameeranjoshi:


ichoyjx:
Hi, what's the rationale for this change? For the structural wise, I don't think there is something needs to be checked for `threadprivate`. If you see some `omp-symbol05.f90` test, there are some tests there.

Sameeranjoshi:
Thanks for pointing the test. I was trying for semantic checks for threadprivate, but couldn't figure out.
Any suggestion what needs to be looked for the construct? Thanks for review.

PR #832, opened 2019-11-18
Change images used for CI to solve build issues

DavidTruby:


DavidTruby:
It seems that the fedora docker images being used were causing the random build failures. I can't work out why this would be the case but I've run 6 builds with these new images with no failures so hopefully this fixes the issue.

DavidTruby:
This doesn't seem fixed. More investigation needed I fear

PR #833, opened 2019-11-19
For call11.f90: more checks on PURE subprograms and TBP bindings

klausler:
Extend, enable, and pass the test cases in call11.f90.  Add and pass new bindings01.f90 test for other constraints and restrictions on type-bound procedures and their overrides.

### DIFF
--- a/lib/evaluate/tools.cc
+++ b/lib/evaluate/tools.cc
@@ -765,4 +765,33 @@ parser::Message *AttachDeclaration(
   }
   return message;
 }
+
+class FindImpureCallHelper
+  : public AnyTraverse<FindImpureCallHelper, std::optional<std::string>> {
+  using Result = std::optional<std::string>;
+  using Base = AnyTraverse<FindImpureCallHelper, Result>;
+
+public:
+  explicit FindImpureCallHelper(const IntrinsicProcTable &intrinsics)
+    : Base{*this}, intrinsics_{intrinsics} {}
+  using Base::operator();
+  Result operator()(const ProcedureRef &call) const {
+    if (auto chars{characteristics::Procedure::Characterize(
+            call.proc(), intrinsics_)}) {
+      if (chars->attrs.test(characteristics::Procedure::Attr::Pure)) {
+        return std::nullopt;
+      }
+    }
+    return call.proc().GetName();
+  }
+
+private:
+  const IntrinsicProcTable &intrinsics_;
+};
+
+std::optional<std::string> FindImpureCall(
+    const IntrinsicProcTable &intrinsics, const Expr<SomeType> &expr) {
+  return FindImpureCallHelper{intrinsics}(expr);
+}
+
 }
--- a/lib/evaluate/tools.h
+++ b/lib/evaluate/tools.h
@@ -817,5 +817,10 @@ parser::Message *SayWithDeclaration(
     MESSAGES &messages, const Symbol *symbol, A &&... x) {
   return AttachDeclaration(messages.Say(std::forward<A>(x)...), symbol);
 }
+
+// Check for references to impure procedures; returns the name
+// of one to complain about, if any exist.
+std::optional<std::string> FindImpureCall(
+    const IntrinsicProcTable &, const Expr<SomeType> &);
 }
 #endif  // FORTRAN_EVALUATE_TOOLS_H_
--- a/lib/semantics/assignment.cc
+++ b/lib/semantics/assignment.cc
@@ -259,7 +259,7 @@ using MaskExpr = evaluate::Expr<evaluate::LogicalResult>;
 // and some number of active WHERE statements/constructs.  WHERE can nest
 // in FORALL but not vice versa.  Pointer assignments are allowed in
 // FORALL but not in WHERE.  These constraints are manifest in the grammar
-// and don't need to be rechecked here, since they cannot appear in the
+// and don't need to be rechecked here, since errors cannot appear in the
 // parse tree.
 struct Control {
   Symbol *name;
@@ -289,8 +289,8 @@ struct ForallContext {
 };
 
 struct WhereContext {
-  explicit WhereContext(MaskExpr &&x) : thisMaskExpr{std::move(x)} {}
-
+  WhereContext(MaskExpr &&x, const WhereContext *o, const ForallContext *f)
+    : outer{o}, forall{f}, thisMaskExpr{std::move(x)} {}
   const WhereContext *outer{nullptr};
   const ForallContext *forall{nullptr};  // innermost enclosing FORALL
   std::optional<parser::CharBlock> constructName;
@@ -308,7 +308,10 @@ public:
 
   bool operator==(const AssignmentContext &x) const { return this == &x; }
 
-  void set_at(parser::CharBlock at) { at_ = at; }
+  void set_at(parser::CharBlock at) {
+    at_ = at;
+    context_.set_location(at_);
+  }
 
   void Analyze(const parser::AssignmentStmt &);
   void Analyze(const parser::PointerAssignmentStmt &);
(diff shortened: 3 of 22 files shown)


lib/evaluate/tools.h
```diff
@@ -817,5 +817,9 @@ parser::Message *SayWithDeclaration(
     MESSAGES &messages, const Symbol *symbol, A &&... x) {
   return AttachDeclaration(messages.Say(std::forward<A>(x)...), symbol);
 }
+
+// Check for references to impure procedures
```
tskeith: Can the comment include a mention of what is returned?

lib/semantics/resolve-names.cc
```diff
@@ -3812,6 +3813,9 @@ void DeclarationVisitor::Post(
     }
     if (auto *s{MakeTypeSymbol(bindingName, ProcBindingDetails{*procedure})}) {
       SetPassNameOn(*s);
+      if (GetAttrs().test(Attr::DEFERRED)) {
+        context().SetError(*s);
```
tskeith: This could be written `context().SetError(*s, GetAttrs().test(Attr::DEFERRED));`

lib/semantics/check-declarations.cc
```diff
@@ -106,6 +107,41 @@ void CheckHelper::Check(const Symbol &symbol) {
...
+    CheckBinding(symbol);
+    return;
+  }
+  if (const auto *details{symbol.detailsIf<DerivedTypeDetails>()}) {
+    CHECK(symbol.scope());
+    CHECK(symbol.scope()->symbol() == &symbol);
+    CHECK(symbol.scope()->IsDerivedType());
+    if (symbol.attrs().test(Attr::ABSTRACT) &&
+        (symbol.attrs().test(Attr::BIND_C) ||
+            (details && details->sequence()))) {
```
jeanPerier: Why not using `IsExtensibleType` here  also ?

lib/semantics/check-declarations.cc
```diff
@@ -106,6 +107,41 @@ void CheckHelper::Check(const Symbol &symbol) {
...
+      if (!IsExtensibleType(parentDerived)) {
+        messages_.Say("The parent type is not extensible"_err_en_US);
+      }
+      if (!symbol.attrs().test(Attr::ABSTRACT) && parentDerived &&
+          parentDerived->typeSymbol().attrs().test(Attr::ABSTRACT)) {
+        ScopeComponentIterator components{*parentDerived};
+        for (const Symbol &component : components) {
+          if (component.attrs().test(Attr::DEFERRED)) {
+            if (symbol.scope()->find(component.name()) ==
+                symbol.scope()->end()) {
```
jeanPerier: This code is only looking in the current derived type `symbol` for a binding, but I think there may be issue here if the deferred was actually defined in an intermediate derived type. e.g:
```
module m0

  type, abstract :: t0
  contains
    procedure(swap), deferred, nopass :: q
  end type

  type, abstract, extends(t0) :: t1
  contains
    procedure, nopass :: q => swap
  end type

  ! Error raised here
  type, extends(t1) :: t2
  end type

contains
  subroutine swap(i, j)
    integer :: i, j, tmp
    tmp = i
    i = j
    j = tmp
  end subroutine

end module
```
f18 raises an error:
`Non-ABSTRACT extension of ABSTRACT derived type 't1' lacks a binding for DEFERRED procedure 'q'`
I think it should not since a binding for `q` was defined in `t1`.

lib/semantics/resolve-names.cc
klausler: Sure, but that looks to me like it could *clear* a previously-set error flag if the predicate were satisfied.  Which it probably doesn't, but I think it's more clear as written.

lib/semantics/tools.cc
```diff
@@ -362,6 +362,56 @@ const Symbol *FindFunctionResult(const Symbol &symbol) {
...
+            if (found->owner().kind() == Scope::Kind::DerivedType) {
+              return found;
+            }
+          }
+        }
+      }
+    }
+  }
+  return nullptr;
+}
```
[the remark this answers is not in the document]
klausler: They can be, yes.  I mistook `FindSymbol`'s behavior of checking parent scopes for that of another function that checks parent types.

lib/semantics/resolve-names.cc
```diff
@@ -3607,8 +3607,9 @@ void DeclarationVisitor::Post(const parser::DerivedTypeStmt &x) {
   symbol.ReplaceName(name.source);
   derivedTypeInfo_.type = &symbol;
   PushScope(Scope::Kind::DerivedType, &symbol);
+  CHECK(currScope().symbol() == &symbol);  // pmk
```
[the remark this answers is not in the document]
klausler: That's left-over debugging code; thanks.

lib/semantics/check-declarations.cc
klausler: Good point, thanks.

lib/semantics/check-declarations.cc
klausler: Because I don't have a `DerivedTypeSpec`.

lib/evaluate/tools.cc
```diff
@@ -765,4 +765,32 @@ parser::Message *AttachDeclaration(
...
+class FindImpureCallHelper
+  : public AnyTraverse<FindImpureCallHelper, std::optional<std::string>> {
+  using Result = std::optional<std::string>;
+  using Base = AnyTraverse<FindImpureCallHelper, Result>;
+
+public:
+  explicit FindImpureCallHelper(const IntrinsicProcTable &intrinsics)
+    : Base{*this}, intrinsics_{intrinsics} {}
+  using Base::operator();
+  Result operator()(const ProcedureDesignator &pd) const {
```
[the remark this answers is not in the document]
klausler: You're right, will fix.

lib/semantics/assignment.cc
```diff
@@ -363,18 +374,11 @@ void AssignmentContext::Analyze(const parser::AssignmentStmt &stmt) {
...
-                "Deallocation of polymorphic object '%s' is not permitted in a PURE subprogram"_err_en_US,
-                last->name());
-          }
-        }
+      if (type->IsPolymorphic() && FindPureProcedureContaining(rhs.source)) {
+        Say(at_,
+            "Deallocation of polymorphic object is not permitted in a PURE subprogram"_err_en_US);
       }
       if (type->category() == TypeCategory::Derived &&
           !type->IsUnlimitedPolymorphic() /* TODO */ &&
```
psteinfeld: In the code that checks for deallocation of components, I don't see anything that checks to see if the components are coarrays.  Section 10.2.1.3, paragraph 13 says that only noncoarray components will be deallocated.

klausler: Copy that; thanks.

lib/semantics/symbol.h
```diff
@@ -582,8 +582,23 @@ class Symbol {
...
+      return true;
+    } else if (name() != that.name()) {
+      return false;
+    } else if (const auto *object{detailsIf<ObjectEntityDetails>()}) {
+      if (const auto *thatObject{that.detailsIf<ObjectEntityDetails>()}) {
+        return object->isDummy() && thatObject->isDummy();
+      }
+    }
+    return false;
+  }
```
tskeith: This definition of `operator==` is not consistent with `operator<`. So you could have two symbols, `x` and `y,` such that `x < y` and `x == y`. It probably wouldn't work to put them in the same `std::set` or `std::map`.

If think it would be better to add a new function for this version of equivalence.

klausler: Ok, fine.  I'm going to have to replace all the names of the `operator==` instances in expressions that end up here, too.

klausler: Done in branch pmk-call12, work in progress.

PR #834, opened 2019-11-20
Add a check for DO step expressions being zero

psteinfeld:
This is prohibited by Section 11.1.7.4.1, paragraph 1.  Since this is
not a constraint, I made it a warning.  Since it's a warning, the test I
added does not contain an error message.

Note also that we allow for REAL step expressions.  But the check I
added only works for INTEGER step expressions.

### DIFF
--- a/lib/semantics/check-do.cc
+++ b/lib/semantics/check-do.cc
@@ -464,7 +464,6 @@ private:
     if (isReal && !warn) {
       // No messages for the default case
     } else if (isReal && warn) {
-      // TODO: Mark the following message as a warning when we have warnings
       context_.Say(sourceLocation, "DO controls should be INTEGER"_en_US);
     } else {
       SayBadDoControl(sourceLocation);
@@ -510,7 +509,11 @@ private:
     CheckDoExpression(bounds.lower);
     CheckDoExpression(bounds.upper);
     if (bounds.step) {
-      CheckDoExpression(bounds.step.value());
+      CheckDoExpression(*bounds.step);
+      if (IsZero(*bounds.step)) {
+        context_.Say(bounds.step->thing.value().source,
+            "DO step expression should not be zero"_en_US);
+      }
     }
   }
 
@@ -670,6 +673,10 @@ private:
         if (const auto &expr{
                 std::get<std::optional<parser::ScalarIntExpr>>(c.t)}) {
           HasNoReferences(indexNames, *expr);
+          if (IsZero(*expr)) {
+            context_.Say(expr->thing.thing.value().source,
+                "DO CONCURRENT step expression should not be zero"_err_en_US);
+          }
         }
       }
     }
--- a/lib/semantics/tools.h
+++ b/lib/semantics/tools.h
@@ -272,6 +272,11 @@ template<typename T> std::optional<std::int64_t> GetIntValue(const T &x) {
   }
 }
 
+template<typename T> bool IsZero(const T &expr) {
+  auto value{GetIntValue(expr)};
+  return value && *value == 0;
+}
+
 // Derived type component iterator that provides a C++ LegacyForwardIterator
 // iterator over the Ordered, Direct, Ultimate or Potential components of a
 // DerivedTypeSpec. These iterators can be used with STL algorithms
--- a/test/semantics/dosemantics02.f90
+++ b/test/semantics/dosemantics02.f90
@@ -14,6 +14,9 @@
 
 ! C1121 -- any procedure referenced in a concurrent header must be pure
 
+! Also, check that the step expressions are not zero.  This is prohibited by
+! Section 11.1.7.4.1, paragraph 1.
+
 SUBROUTINE do_concurrent_c1121(i,n)
   IMPLICIT NONE
   INTEGER :: i, n, flag
@@ -28,3 +31,25 @@ SUBROUTINE do_concurrent_c1121(i,n)
       i = 35
     END FUNCTION random
 END SUBROUTINE do_concurrent_c1121
+
+SUBROUTINE s1()
+  INTEGER, PARAMETER :: constInt = 0
+
+  ! Warn on this one for backwards compatibility
+  DO 10 I = 1, 10, 0
+  10 CONTINUE
+
+  ! Warn on this one for backwards compatibility
+  DO 20 I = 1, 10, 5 - 5
+  20 CONTINUE
+
+  ! Error, no compatibility requirement for DO CONCURRENT
+!ERROR: DO CONCURRENT step expression should not be zero
+  DO CONCURRENT (I = 1 : 10 : 0)
+  END DO
+
+  ! Error, this time with an integer constant
+!ERROR: DO CONCURRENT step expression should not be zero
+  DO CONCURRENT (I = 1 : 10 : constInt)
+  END DO
+end subroutine s1
--- a/test/semantics/dosemantics05.f90
+++ b/test/semantics/dosemantics05.f90
@@ -46,7 +46,7 @@ subroutine s1()
   ! References in this DO CONCURRENT are OK since there's no DEFAULT(NONE)
   ! locality-spec
   associate (avar => ivar)
-    do concurrent (i = 1:2:0) shared(jvar)
+    do concurrent (i = 1:2) shared(jvar)
       ivar = 3
       ivar = ivar + i
       block
@@ -61,6 +61,7 @@ subroutine s1()
   end associate
   
   associate (avar => ivar)
+!ERROR: DO CONCURRENT step expression should not be zero
     do concurrent (i = 1:2:0) default(none) shared(jvar) local(kvar)
 !ERROR: Variable 'ivar' from an enclosing scope referenced in DO CONCURRENT with DEFAULT(NONE) must appear in a locality-spec
       ivar =  &


lib/semantics/check-do.cc
```diff
@@ -395,6 +395,20 @@ class DoContext {
     }
   }
 
+  // Step expressions can't be zero (Section 11.1.7.4.1, paragraph 1).
+  // Make this a warning since it's not a constraint violation
+  void CheckZeroStep(const parser::ScalarExpr &stepExpression) {
+    MaybeExpr stepExpr{evaluate::Fold(
+        context_.foldingContext(), AnalyzeExpr(context_, stepExpression))};
+    if (stepExpr) {
```
klausler: `if (MaybeExpr stepExpr{...}) { ... `

lib/semantics/check-do.cc
```diff
@@ -395,6 +395,20 @@ class DoContext {
...
+      std::optional<std::int64_t> stepValue{evaluate::ToInt64(*stepExpr)};
+      if (stepValue && stepValue.value() == 0) {
```
klausler: `if (auto stepValue{...}; *stepValue == 0) { ... `

lib/semantics/check-do.cc
```diff
@@ -404,6 +418,7 @@ class DoContext {
     CheckDoExpression(bounds.upper);
     if (bounds.step) {
       CheckDoExpression(bounds.step.value());
+      CheckZeroStep(bounds.step.value());
```
klausler: After checking for the presence of a value immediately beforehand, you don't have to use `.value()`.

lib/semantics/check-do.cc
psteinfeld: Thanks.  I'll change this.

lib/semantics/check-do.cc
psteinfeld: Thanks.  I'll change this.

lib/semantics/check-do.cc
psteinfeld: Thanks.  I'll change this.

psteinfeld:
> 
> 
> Should you add a similar check to `DO CONCURRENT` and `FORALL` _concurrent-header_ processing? How about subscript triplets in array sections? Implied DO loops in I/O statements and array constructors?

Good point.  I'll look through the standard for similar opportunities and add them where appropriate.

lib/semantics/check-do.cc
psteinfeld: Is the call to ```evaluate::ToInt64(*stepExpr)``` guaranteed to return something that I can dereference?

klausler: No; that's why it's optional.

lib/semantics/check-do.cc
```diff
@@ -395,6 +395,20 @@ class DoContext {
...
+  // Step expressions can't be zero (Section 11.1.7.4.1, paragraph 1).
+  // Make this a warning since it's not a constraint violation
+  void CheckZeroStep(const parser::ScalarExpr &stepExpression) {
+    MaybeExpr stepExpr{evaluate::Fold(
+        context_.foldingContext(), AnalyzeExpr(context_, stepExpression))};
+    if (stepExpr) {
+      std::optional<std::int64_t> stepValue{evaluate::ToInt64(*stepExpr)};
+      if (stepValue && stepValue.value() == 0) {
+        context_.Say(stepExpression.thing.value().source,
+            "DO step expression should not be zero"_en_US);
```
vdonaldson: Constraint violations and program requirement violations are both equally bad (Clause 4.2).  The difference is that the processor (compiler) is *required* to diagnose the former, but not the latter.  This is usually because it is difficult or impossible to diagnose all program requirement violations, not because they are necessarily less problematic.  In cases where as an aid to the programmer it *is* possible to diagnose a program requirement violation, the violation should usually be an error, just like a constraint.  In the absence of an -Mstandard option, both constraint and program requirement violations could possibly appear as warnings rather than errors if the violation doesn't necessarily create a problem with how the program should be interpreted.  In this case, if a step value of zero is likely to result in an infinite loop, that seems like it should be a hard error.

psteinfeld: Thanks for the clarification, Val.  I'll update the comment.

test/semantics/dosemantics02.f90
```diff
@@ -28,3 +31,13 @@ IMPURE FUNCTION random() RESULT(i)
       i = 35
     END FUNCTION random
 END SUBROUTINE do_concurrent_c1121
+
+SUBROUTINE s1()
+  ! Warn since no constrains are violated
```
kiranchandramohan: Nit: constraints

test/semantics/dosemantics02.f90
```diff
@@ -28,3 +31,13 @@ IMPURE FUNCTION random() RESULT(i)
       i = 35
     END FUNCTION random
 END SUBROUTINE do_concurrent_c1121
+
+SUBROUTINE s1()
+  ! Warn since no constrains are violated 
+  DO 20 I = 1, 10, 0
+  20 CONTINUE
+
```
kiranchandramohan: Maybe one more test with a constant parameter variable.

psteinfeld: Great suggestion!  I'll do that.

lib/semantics/check-do.cc
sscalpone: You might have
```
integer, parameter :: debug_step = 0
if(debug_step > 0) then
  do i = 1, N, debug_step  !  Want compilation not to fail
```

psteinfeld: I've searched the standard, and I found 12 places where prohibit zero values for various constructs -- 
Section 9.5.3.3.2, paragraph 2
Section 11.1.7.4.1, paragraph 1, point (1)
Section 11.1.7.4.2, paragraph 1, point (1)
Section 13.7.2.1, paragraph 6
Section 13.7.2.2, paragraph 1
Section 13.7.2.3.2, paragraph 1
Section 13.7.2.3.6, paragraph 2
Section 13.7.2.4, paragraph 1
Section 13.7.5.1, paragraph 1
Section 16.9.118, paragraph 3
Section 16.9.135, paragraph 3
Section 16.9.136, paragraph 3
I agree that the step expression for a normal DO construct is a context where a warning is preferable to an error.  What about these other contexts, though?  For example, what about the step expression for a DO CONCURRENT?

vdonaldson: A program violation, if diagnosed, should generally be an error.

psteinfeld: Thanks for the feedback, Val.  I have yet to look at these in detail, but my default will be to call them errors, starting with DO CONCURRENT.

test/semantics/dosemantics02.f90
psteinfeld: Thanks, Kiran.  I'll fix this.

lib/semantics/tools.h
```diff
@@ -248,6 +248,19 @@ template<typename T> std::optional<std::int64_t> GetIntValue(const T &x) {
   }
 }
 
+template<typename T> bool IsZero(const T &expr) {
+  const std::optional<std::int64_t> value{GetIntValue(expr)};
```
klausler: Why `const`?

And why not `auto`?

lib/semantics/tools.h
```diff
@@ -248,6 +248,19 @@ template<typename T> std::optional<std::int64_t> GetIntValue(const T &x) {
...
+  return value && *value == 0;
+}
+
+template<typename T>
+static void SayIfZero(T &expr, SemanticsContext &context,
```
klausler: Why is this template function `static` and the preceding one not?

Why not `const T &expr`?

lib/semantics/tools.h
psteinfeld: No good reason.  I'll change it.

lib/semantics/tools.h
psteinfeld: No good reason.  I'll change them.

psteinfeld:
> 
> 
> Can you catch this in implied do loops?
> print *, (/ (i, i=1,4,0) /)

The code in this pull request doesn't catch this case.  It only deals with normal and CONCURRENT DO loops.  I looked through the standard for other situations where zero values are prohibited and created issue #838 to cover them.

lib/semantics/tools.h
```diff
@@ -248,6 +248,19 @@ template<typename T> std::optional<std::int64_t> GetIntValue(const T &x) {
   }
 }
 
+template<typename T> static bool IsZero(const T &expr) {
+  auto value{GetIntValue(expr)};
+  return value && *value == 0;
+}
+
+template<typename T>
```
klausler: You could replace the two sites that call this template function with a simple expansion of its body and turn 11 lines into six, and probably improve readability.

psteinfeld: Thanks, Peter.  I've tightened up the code to emit a message if the expression is zero.

klausler: I wasn't suggesting that you should remove `IfZero()`.  It's a useful abstraction.  I was suggesting that you remove `SayIfZero()`.

psteinfeld: Oops.  My misunderstanding.  I've pushed another change that I believe accomplishes what your original comment suggested.

PR #835, opened 2019-11-21
Enable, extend, and pass call12.f90 test; revert some recent changes

klausler:
Recent changes to `semantics::Symbol::operator=()` were reverted.

Implement checks in C1594 that prevent some operations in `PURE` subprograms.


lib/semantics/assignment.cc
```diff
@@ -563,6 +558,117 @@ void AssignmentContext::CheckForImpureCall(
...
+            "A PURE subprogram may not copy the value of '%s' because it is %s and has the POINTER component '%s'"_err_en_US,
+            base->name(), why, *pointer);
+      }
+    }
+  }
+}
+
+void AssignmentContext::CheckForPureContext(
+    const evaluate::Expr<evaluate::SomeType> &lhs,
+    const evaluate::Expr<evaluate::SomeType> &rhs, parser::CharBlock source,
```
psteinfeld: "lhs" and "rhs" could be defined as "&SomeExpr".

lib/semantics/assignment.cc
```diff
@@ -563,6 +558,117 @@ void AssignmentContext::CheckForImpureCall(
...
+          "A PURE subprogram may not define a coindexed object"_err_en_US);
+    } else if (const Symbol * base{GetFirstSymbol(lhs)}) {
+      CheckDefinabilityInPureScope(messages, *base, scope);
+    }
+    if (isPointerAssignment) {
+      if (const Symbol * base{GetFirstSymbol(rhs)}) {
+        if (const char *why{
+                WhyBaseObjectIsSuspicious(*base, scope)}) {  // C1594(3)
+          evaluate::SayWithDeclaration(messages, base,
+              "A PURE subprogram may not use '%s' as the target of pointer assignment because it is %s"_err_en_US,
```
psteinfeld: This line is longer than 80 characters.

psteinfeld: As are other lines that contain message string literals.

lib/semantics/tools.cc
```diff
@@ -70,13 +70,14 @@ const Scope *FindProgramUnitContaining(const Symbol &symbol) {
...
 }
 
-const Scope *FindPureProcedureContaining(const Scope *scope) {
-  scope = FindProgramUnitContaining(*scope);
-  while (scope) {
+const Scope *FindPureProcedureContaining(const Scope &start) {
+  // N.B. We only need to examine the innermost containing program unit
+  // because an internal subprogram of a PURE subprogram must also
+  // be PURE (C1592).
+  if (const Scope *scope{FindProgramUnitContaining(start)}) {
```
psteinfeld: It looks like this file needs to have clang-format run on it.

test/semantics/call12.f90
```diff
@@ -23,58 +27,62 @@ module m
   type :: hasPtr
     real, pointer :: p
   end type
+  type :: hasCoarray
+    real :: co[*]
+  end type
  contains
-  pure subroutine msubr
-    !ERROR: A PURE subprogram must not define a host-associated object
+  pure function test(ptr, in, hpd)
```
psteinfeld: One of the circumstances that can cause the illegal deallocation of a a polymorphic entity is peculiar to PURE procedures.  It's described in section 9.7.3.2, paragraph 5:

If a function whose result is allocatable or has an allocatable subobject is referenced in the specification part of a scoping unit, and the function reference is executed, an allocatable result and any allocated allocatable subobject of the result is deallocated before execution of the executable constructs of the scoping unit.

I'm not sure if your code covers this case, but it would be good to have a test for it.

lib/semantics/assignment.cc
```diff
@@ -563,6 +558,117 @@ void AssignmentContext::CheckForImpureCall(
   }
 }
 
+// C1594 checks
+static bool IsPointerDummyOfPureFunction(const Symbol &x) {
+  return IsPointerDummy(x) && FindPureProcedureContaining(x.owner()) &&
+      x.owner().symbol() && IsFunction(*x.owner().symbol());
```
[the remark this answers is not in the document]
klausler: `x.owner()` returns a `Scope &`, so no.

lib/semantics/assignment.cc
```diff
@@ -563,6 +558,117 @@ void AssignmentContext::CheckForImpureCall(
...
+        return pointer.BuildResultDesignatorName();
+      }
+    }
+  }
+  return std::nullopt;
+}
+
+// Checks C1594(5,6)
+void CheckCopyabilityInPureScope(parser::ContextualMessages &messages,
+    const evaluate::Expr<evaluate::SomeType> &expr, const Scope &scope) {
```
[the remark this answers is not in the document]
klausler: Yes.  Do you want me to do so?

lib/semantics/assignment.cc
```diff
@@ -563,6 +558,117 @@ void AssignmentContext::CheckForImpureCall(
...
+  if (const Symbol * base{GetFirstSymbol(expr)}) {
+    if (const char *why{WhyBaseObjectIsSuspicious(*base, scope)}) {
+      if (auto pointer{GetPointerComponentDesignatorName(expr)}) {
+        evaluate::SayWithDeclaration(messages, base,
+            "A PURE subprogram may not copy the value of '%s' because it is %s and has the POINTER component '%s'"_err_en_US,
```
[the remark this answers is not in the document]
klausler: I am deferring the reformatting of messages into multiple lines for the moment.

lib/semantics/assignment.cc
klausler: Yes.  Do you want me to do so?

lib/semantics/assignment.cc
klausler: I know.  I am deferring the reformatting of long message lines.

test/semantics/call12.f90
klausler: Doesn't C1584 cover that case?  A pure function can't return a value whose finalization might reference an impure subroutine, and only references to pure functions may appear in specification expressions (10.1.11(4)).

lib/semantics/assignment.cc
klausler: `SomeExpr &`; done.

lib/semantics/assignment.cc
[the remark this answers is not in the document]
psteinfeld: Yes, please.  I think it's more readable.

test/semantics/call12.f90
psteinfeld: Actually, the case I was thinking about is covered by C1585.  Furthermore, this case is already covered by test call10.f90.

So, never mind ...

PR #836, opened 2019-11-21
Fold MERGE, use it in ISO_FORTRAN_ENV

klausler:
Reduce use of intrinsic type KIND codes in ISO_FORTRAN_ENV by implementing constant folding of the `MERGE` intrinsic function, and then using it to massage the results of `SELECTED_REAL_KIND()` & al. to construct the definitions in the module.

### DIFF
--- a/module/iso_fortran_env.f90
+++ b/module/iso_fortran_env.f90
@@ -19,20 +19,113 @@ include '../runtime/magic-numbers.h' ! for IOSTAT= error/end code values
 
 module iso_fortran_env
 
-  integer, parameter :: atomic_int_kind = 8
-  integer, parameter :: atomic_logical_kind = 8
+  implicit none
 
-  integer, parameter :: character_kinds(*) = [1, 2, 4]
-  integer, parameter :: int8 = 1, int16 = 2, int32 = 4, int64 = 8, int128 = 16
-  integer, parameter :: integer_kinds(*) = [int8, int16, int32, int64, int128]
+  integer, parameter :: atomic_int_kind = selected_int_kind(18)
+  integer, parameter :: atomic_logical_kind = atomic_int_kind
+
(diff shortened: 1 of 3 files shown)


module/iso_fortran_env.f90
```diff
@@ -19,20 +19,110 @@
...
+    int16 = merge(selectedInt16, merge(-2, -1, selectedInt16 >= 0), &
+                  digits(int(0,kind=safeInt16)) == 15), &
+    int32 = merge(selectedInt32, merge(-2, -1, selectedInt32 >= 0), &
+                  digits(int(0,kind=safeInt32)) == 31), &
+    int64 = merge(selectedInt64, merge(-2, -1, selectedInt64 >= 0), &
+                  digits(int(0,kind=safeInt64)) == 63), &
+    int128 = merge(selectedInt128, merge(-2, -1, selectedInt128 >= 0), &
+                   digits(int(0,kind=safeInt128)) == 127)
+  integer, parameter :: integer_kinds(*) = [ &
+    [(int8, integer :: j=1, count([int8==selectedInt8]))], &
```
tskeith: Why not `count([int8 >= 0])`? That's how it's done for `character_kinds` and `logical_kinds` but not `integer_kinds` and `real_kinds`.

klausler: Ok.

klausler: Thanks for pointing this out -- both `integer_kinds` and `real_kinds` can include kinds other than the standard ones, since they're meant to include all of the kinds supported on the "processor".  Fixing this with a complicated search loop written as an array constructor exposed a bug in array constructor expression analysis, and it's good that it was caught now.

klausler:
Please take another look, this PR now also contains corrections to array constructor expression analysis problems that were exposed while changing the module.

tskeith:
Looks good to me.

PR #841, opened 2019-11-23
Analyze defined assignments

tskeith:
Commits 1-4 and 6 are refactorings and should not result in functional changes.
The rest perform checks on defined assignments and when successful insert an analyzed form of the assignment into the parse tree.

Note: For lowering to FIR the analyzed form (retrieved by `GetAssignment` in `semantics/tools.h`) should be used.

### DIFF
--- a/lib/semantics/check-declarations.cc
+++ b/lib/semantics/check-declarations.cc
@@ -450,18 +450,11 @@ void CheckHelper::SayNotDistinguishable(const SourceName &name,
   evaluate::AttachDeclaration(msg, proc2);
 }
 
-static bool ConflictsWithIntrinsicAssignment(
-    const DummyDataObject &arg0, const DummyDataObject &arg1) {
-  auto cat0{arg0.type.type().category()};
-  auto cat1{arg1.type.type().category()};
-  int rank0{arg0.type.Rank()};
-  int rank1{arg1.type.Rank()};
-  if (cat0 == TypeCategory::Derived || (rank1 > 0 && rank0 != rank1)) {
-    return false;
-  } else {
-    return cat0 == cat1 ||
-        (IsNumericTypeCategory(cat0) && IsNumericTypeCategory(cat1));
-  }
+static bool ConflictsWithIntrinsicAssignment(const Procedure &proc) {
+  auto lhs{std::get<DummyDataObject>(proc.dummyArguments[0].u).type};
+  auto rhs{std::get<DummyDataObject>(proc.dummyArguments[1].u).type};
+  return Tristate::No ==
+      IsDefinedAssignment(lhs.type(), lhs.Rank(), rhs.type(), rhs.Rank());
 }
 
 // Check if this procedure can be used for defined assignment (see 15.4.3.4.3).
@@ -476,9 +469,7 @@ bool CheckHelper::CheckDefinedAssignment(
   } else if (!CheckDefinedAssignmentArg(specific, proc.dummyArguments[0], 0) |
       !CheckDefinedAssignmentArg(specific, proc.dummyArguments[1], 1)) {
     return false;  // error was reported
-  } else if (ConflictsWithIntrinsicAssignment(
-                 std::get<DummyDataObject>(proc.dummyArguments[0].u),
-                 std::get<DummyDataObject>(proc.dummyArguments[1].u))) {
+  } else if (ConflictsWithIntrinsicAssignment(proc)) {
     msg = "Defined assignment subroutine '%s' conflicts with"
           " intrinsic assignment"_err_en_US;
   } else {
--- a/lib/semantics/check-declarations.cc
+++ b/lib/semantics/check-declarations.cc
@@ -26,6 +26,10 @@
 
 namespace Fortran::semantics {
 
+using evaluate::characteristics::DummyArgument;
+using evaluate::characteristics::DummyDataObject;
+using evaluate::characteristics::Procedure;
+
 class CheckHelper {
 public:
   explicit CheckHelper(SemanticsContext &c) : context_{c} {}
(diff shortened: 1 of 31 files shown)


lib/semantics/tools.h
```diff
@@ -57,6 +58,14 @@ const DeclTypeSpec *FindParentTypeSpec(const Symbol &);
 // Return the Symbol of the variable of a construct association, if it exists
 const Symbol *GetAssociationRoot(const Symbol &);
 
+enum class Tristate { No, Yes, Maybe };
+inline Tristate ToTristate(bool x) { return x ? Tristate::Yes : Tristate::No; }
```
tskeith: Should this be move to `common/idioms.h` or somewhere like that? There are a couple of other places we use `std::optional<bool>` where this might be better.

klausler: How is `Tristate` better than using `optional<bool>`, which gets the job done with standard language features?

tskeith: With `std::optional<bool>` the three states aren't accessed symmetrically: you have to check for the first (`std::nullopt`) before you can check for the other two. In this case I have three equally valid states -- one shouldn't be distinguished.

I started with `std::optional<bool>` here but I thought it made the code less clear.

lib/semantics/check-declarations.cc
```diff
@@ -352,6 +281,251 @@ void CheckHelper::CheckValue(
...
+    CheckSpecificsAreDistinguishable(symbol, details, *procs);
+  }
+}
+
+// Check that the specifics of this generic are distinguishable from each other
+void CheckHelper::CheckSpecificsAreDistinguishable(const Symbol &generic,
+    const GenericDetails &details, const std::vector<Procedure> &procs) {
+  const SymbolVector &specifics{details.specificProcs()};
+  auto count{specifics.size()};
+  if (specifics.size() < 2) {
```
psteinfeld: Should you just use "count" rather than "specifics.size()"?

lib/semantics/check-declarations.cc
```diff
@@ -352,6 +281,251 @@ void CheckHelper::CheckValue(
...
+  const SymbolVector &specifics{details.specificProcs()};
+  const auto &bindingNames{details.bindingNames()};
+  std::optional<std::vector<Procedure>> procs{Characterize(specifics)};
+  if (!procs) {
+    return;
+  }
+  bool ok{true};
+  if (details.kind().IsAssignment()) {
+    for (std::size_t i{0}; i < specifics.size(); ++i) {
+      auto save{messages_.SetLocation(bindingNames[i])};
```
psteinfeld: I don't understand this use of "save".  It's being initialized, but I don't see it used anywhere.  But I don't get a compilation warning.

lib/semantics/symbol.cc
```diff
@@ -586,4 +586,28 @@ void TypeParamDetails::set_type(const DeclTypeSpec &type) {
...
+bool GenericKind::IsOperator() const {
+  return IsDefinedOperator() || IsIntrinsicOperator();
+}
+
+std::string GenericKind::ToString() const {
+  return std::visit(
+      common::visitors{
+          [&](const OtherKind &x) { return EnumToString(x); },
+          [&](const DefinedIo &x) { return EnumToString(x); },
+          [&](const auto &x) { return common::EnumToString(x); },
```
psteinfeld: I think you can do without the reference captures on these lambdas.

lib/semantics/check-declarations.cc
```diff
@@ -352,6 +281,251 @@ void CheckHelper::CheckValue(
...
+      if (InFunction()) {  // C1583
+        messages_.Say(
+            "non-POINTER dummy argument of PURE function must be INTENT(IN) or VALUE"_err_en_US);
+      } else if (IsIntentOut(symbol)) {
+        if (const DeclTypeSpec * type{details.type()}) {
+          if (type && type->IsPolymorphic()) {  // C1588
+            messages_.Say(
+                "An INTENT(OUT) dummy argument of a PURE subroutine may not be polymorphic"_err_en_US);
+          } else if (const DerivedTypeSpec * derived{type->AsDerived()}) {
+            if (FindUltimateComponent(*derived, [](const Symbol &x) {
```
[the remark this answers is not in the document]
tskeith: I didn't write this code, only moved it around, so I don't want to change it in this PR.

lib/semantics/check-declarations.cc
tskeith: Right, fixed.

lib/semantics/check-declarations.cc
tskeith: `save` is an instance of `common::Restorer`. When it goes out of scope it undoes what `SetLocation` did. That way you don't need separate save and restore operations here and run the risk of forgetting the restore.

I've changed the name of the variable to `restorer` to make its purpose clearer.

lib/semantics/symbol.cc
tskeith: Right

lib/semantics/expression.cc
```diff
@@ -2540,6 +2549,43 @@ MaybeExpr ArgumentAnalyzer::TryDefinedOp(
   return TryDefinedOp(oprs[0], std::move(error));
 }
 
+std::optional<ProcedureRef> ArgumentAnalyzer::TryDefinedAssignment() {
+  using semantics::Tristate;
+  const auto &lhs{GetAsExpr(0)};
+  const auto &rhs{GetAsExpr(1)};
```
jeanPerier: Isn't `GetAsExpr` returning values ? I think it might be better to change `GetAsExpr` so it would return `Expr<SomeExpr>&` and so that the expression can also be moved instead of being copied after the  analysis of intrinsic assignments at:

https://github.com/flang-compiler/f18/blob/35a31138f08ff2d41e3fb016f3e33bfde4cd835e/lib/semantics/expression.cc#L1801

tskeith: OK, done.

klausler: The name should probably be changed to reflect the altered semantics so that there's less risk of failing to consider the lifetime of the referent.

tskeith: OK, I made it `MoveExpr`.

klausler: Now it returns an rvalue reference to the expression, which remains in place in the actual argument structure; the expression is not actually moved, despite the new `MoveExpr` name.

If you'd like to leave it the expression in place, I think you should grab a `const &` reference to it via some API whose name indicates that a reference has been created.  If you want to get a copy of the expression, the original API did that.

tskeith: I don't see where the the result of `MoveExpr` is being copied now. What would I have to do to fix it?

If I use just `GetAsExpr`, returning `const Expr<SomeType> &`, then in each of the places I need an r-value reference (`BinaryLogicalOperation`, `Relate`, etc.) I'll need to use `common::Clone` or something to make a copy. That doesn't seem like a good approach.

klausler: Then it might be more simple to revert `GetAsExpr` to be the point that makes the copy.

PR #844, opened 2019-11-26
Fix cmake path: minor change

arjunsuresh1987:
Cmake minor fix for building f18


lib/semantics/expression.h
```diff
@@ -100,7 +100,7 @@ struct SetExprHelper {
     if constexpr (ConstraintTrait<T>) {
       Set(x.thing);
     } else {
-      static_assert("bad type");
+      static_assert(static_cast<bool>("bad type"));
     }
```
DavidTruby: As an observation could this whole section not be written as follows?
```
static_assert(ConstraintTrait<T>, "bad type");
Set(x.thing);
```
The original static assert looks like a mistake as it will never get triggered (a non-0 value is always true!)
I suspect the original should say static_assert(false, "bad type"), in which case the cast wouldn't be necessary.

I realise you didn't write this original code though

CMakeLists.txt
```diff
@@ -25,7 +25,9 @@ if( GCC )
 endif()
 if(BUILD_WITH_CLANG)
   set(CMAKE_CXX_COMPILER "${BUILD_WITH_CLANG}/bin/clang++")
```
sscalpone: Why not change this line too?

CMakeLists.txt
```diff
@@ -25,7 +25,9 @@ if( GCC )
...
-  set(CMAKE_CC_COMPILER "${BUILD_WITH_CLANG}/bin/clang")
+  file(TO_CMAKE_PATH "${BUILD_WITH_CLANG}" CLANG_PATH)
```
sscalpone: Why is this needed?  Why not use TO_NATIVE_PATH?

sscalpone:
@arjunsuresh1987  Thanks for the PR!  What compiler and operating system are you working with?

lib/semantics/expression.h
arjunsuresh1987: Thank you David for the reply. As you mentioned that static_assert does not make sense. But the assert when corrected causes compilation to fail (assertion fails). There must be a logical error in the implementation here.

CMakeLists.txt
arjunsuresh1987: TO_NATIVE_PATH is not working for me if I use 

-DBUILD_WITH_CLANG=~/llvm

If I expand ~/llvm to the full absolute path it works. 

I'm on Ubuntu 19.04 and not sure if this will work on Windows though.

CMakeLists.txt
arjunsuresh1987: Sorry. This line needs to be removed. I'll do it if the change is fine.

arjunsuresh1987:
You are welcome @sscalpone . I'm using Clang 10.0 on Ubuntu 19.04

lib/semantics/expression.h
klausler: I probably meant to type `static_assert(!"bad_type");`.

arjunsuresh1987: static_assert(!"bad_type"); //should always fail right as it equals static_assert(false)? static_assert(false, "bad_type") should be better as it outputs the message too. But the main problem is this assetion itself is failing.

klausler: Of course.  Are you also fixing that?

CMakeLists.txt
sscalpone: Yes, the CMakeList change is fine if you remove the duplicate assignment.

lib/semantics/expression.h
arjunsuresh1987: I suppose changing the static_assert to dynamic assert is the right way here. If that change is fine, I can do and amend to the pull request.

CMakeLists.txt
arjunsuresh1987: Thank you. I'll do an amend to the PR.

arjunsuresh1987:
I have removed the changes to the static_assert as it is causing error. Will file it as a separate issue. Have added only the cmake changes now.

lib/semantics/expression.h
DavidTruby: Changing to a runtime `assert`is guaranteed to cause a runtime failure too, since we're definitely calling the function with an invalid T if the static_assert is failing.

klausler: If the assert is (or should be) failing at compilation time, then there's a problem that needs to be addressed!

arjunsuresh1987: Yes definitely. I see that types Names and ConstantValue are not being handled and are causing the assert failure. I'll add it as a new issue/PR.

PR #846, opened 2019-11-27
Check functions that implement defined operators

tskeith:
Section 15.4.3.4.2 specifies restrictions on functions that may be used
to implement an "extended-intrinsic-op". These checkw are implemented in
`CheckHelper::CheckDefinedOperator`.

Move `IsIntrinsicRelational` et al. to `semantics/tools.h` so that
the same logic is used to check both dummy and actual arguments.

Fix up tests that had errors that are now detected.

### DIFF
--- a/lib/semantics/check-declarations.cc
+++ b/lib/semantics/check-declarations.cc
@@ -65,7 +65,7 @@ private:
   std::optional<std::vector<Procedure>> Characterize(const SymbolVector &);
   bool CheckDefinedOperator(const SourceName &, const GenericKind &,
       const Symbol &, const Procedure &);
-  bool CheckNumberOfArgs(std::optional<parser::MessageFixedText> &,
+  std::optional<parser::MessageFixedText> CheckNumberOfArgs(
       const GenericKind &, std::size_t);
   bool CheckDefinedOperatorArg(
       const SourceName &, const Symbol &, const Procedure &, std::size_t);
@@ -516,8 +516,8 @@ bool CheckHelper::CheckDefinedOperator(const SourceName &opName,
   } else if (proc.functionResult->IsAssumedLengthCharacter()) {
     msg = "%s function '%s' may not have assumed-length CHARACTER(*)"
           " result"_err_en_US;
-  } else if (!CheckNumberOfArgs(msg, kind, proc.dummyArguments.size())) {
-    // error is in msg
+  } else if (auto m{CheckNumberOfArgs(kind, proc.dummyArguments.size())}) {
+    msg = std::move(m);
   } else if (!CheckDefinedOperatorArg(opName, specific, proc, 0) |
       !CheckDefinedOperatorArg(opName, specific, proc, 1)) {
     return false;  // error was reported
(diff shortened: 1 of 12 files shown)


lib/semantics/check-declarations.cc
```diff
@@ -457,6 +469,137 @@ static bool ConflictsWithIntrinsicAssignment(const Procedure &proc) {
...
+    msg = "%s function '%s' conflicts with intrinsic operator"_err_en_US;
+  } else {
+    return true;  // OK
+  }
+  SayWithDeclaration(specific, std::move(msg.value()),
+      parser::ToUpperCaseLetters(opName.ToString()), specific.name());
+  return false;
+}
+
+// If the number of arguments is wrong for this intrinsic operator, return
```
klausler: Returning just an optional message (whose presence signifies an error) would be a cleaner interface.

lib/semantics/check-declarations.cc
```diff
@@ -457,6 +469,137 @@ static bool ConflictsWithIntrinsicAssignment(const Procedure &proc) {
...
+    const Symbol &symbol, const Procedure &proc, std::size_t pos) {
+  if (pos >= proc.dummyArguments.size()) {
+    return true;
+  }
+  auto &arg{proc.dummyArguments.at(pos)};
+  std::optional<parser::MessageFixedText> msg;
+  if (arg.IsOptional()) {
+    msg = "In %s function '%s', dummy argument '%s' may not be"
+          " OPTIONAL"_err_en_US;
+  } else if (const auto *dataObject{std::get_if<DummyDataObject>(&arg.u)};
```
klausler: If you transpose the then & else parts here, it would avoid the unusual `nullptr` test in the predicate.

lib/semantics/check-declarations.cc
```diff
@@ -457,6 +469,137 @@ static bool ConflictsWithIntrinsicAssignment(const Procedure &proc) {
...
+  } else {
+    msg = "%s function '%s' must have one or two dummy arguments"_err_en_US;
+  }
+  return false;
+}
+
+bool CheckHelper::CheckDefinedOperatorArg(const SourceName &opName,
+    const Symbol &symbol, const Procedure &proc, std::size_t pos) {
+  if (pos >= proc.dummyArguments.size()) {
+    return true;
```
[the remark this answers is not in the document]
tskeith: For unary operators we only want to check the first argument, for binary operators both arguments. So the check for "is it unary" could happen where this is called to avoid calling it when pos == 1, or it can happen here. I chose here to simplify the caller.

lib/semantics/check-declarations.cc
tskeith: It is unusual, but it makes the structure of the `if`s simpler. This way it's a series of: check for error and set error message. Changing it as you suggest requires nesting checks inside of this one.

If you feel strongly about avoiding this form of `if` I'm happy to change it. Otherwise I'll leave it as-is.

lib/semantics/check-declarations.cc
tskeith: OK, I've made that change.

PR #847, opened 2019-12-02
Forward references to derived types

klausler:
Implement forward references to derived types, including parameterized derived types.

Allowing forward references to unparameterized derived types is straightforward (a flag was added to `DerivedTypeDetails` to indicate them, and a state flag in `DeclTypeSpecVisitor` allows them), but enabling forward references to PDTs and getting it right involved some restructuring of existing code so that validation of type parameter lists could be deferred to the time of type instantiation.  So some larger functions were broken up into smaller ones, and many were converted into member functions and relocated into different C++ source files.

Fixes issue #573.

### DIFF
--- a/lib/semantics/resolve-names.cc
+++ b/lib/semantics/resolve-names.cc
@@ -89,7 +89,7 @@ private:
   // map_ contains the mapping between letters and types that were defined
   // by the IMPLICIT statements of the related scope. It does not contain
   // the default Fortran mappings nor the mapping defined in parents.
-  std::map<char, const DeclTypeSpec *> map_;
+  std::map<char, common::Reference<const DeclTypeSpec>> map_;
 
   friend std::ostream &operator<<(std::ostream &, const ImplicitRules &);
   friend void ShowImplicitRule(std::ostream &, const ImplicitRules &, char);
@@ -173,11 +173,18 @@ public:
     if (!maybeExpr) {
       return std::nullopt;
     }
+    auto exprType{maybeExpr->GetType()};
     auto converted{evaluate::ConvertToType(symbol, std::move(*maybeExpr))};
     if (!converted) {
-      Say(source,
-          "Initialization expression could not be converted to declared type of '%s'"_err_en_US,
-          symbol.name());
+      if (exprType) {
+        Say(source,
+            "Initialization expression could not be converted to declared type of '%s' from %s"_err_en_US,
+            symbol.name(), exprType->AsFortran());
+      } else {
+        Say(source,
+            "Initialization expression could not be converted to declared type of '%s'"_err_en_US,
+            symbol.name());
+      }
       return std::nullopt;
     }
     return FoldExpr(std::move(*converted));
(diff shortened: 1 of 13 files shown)


lib/semantics/type.cc
```diff
@@ -38,22 +39,253 @@ void DerivedTypeSpec::ReplaceScope(const Scope &scope) {
...
+    }
+  }
+}
+
+void DerivedTypeSpec::EvaluateParameters(
+    evaluate::FoldingContext &foldingContext) {
+  if (evaluated_) {
+    return;
+  }
+  evaluated_ = true;
```
tskeith: Is it necessary that `CookParameters` has been called so that `FindParameter` works correctly? If so, it might be a good idea to assert that here.

lib/semantics/type.cc
```diff
@@ -38,22 +39,253 @@ void DerivedTypeSpec::ReplaceScope(const Scope &scope) {
...
+        if (auto converted{evaluate::ConvertToType(symbol, SomeExpr{*expr})}) {
+          SomeExpr folded{
+              evaluate::Fold(foldingContext, std::move(*converted))};
+          if (auto *intExpr{std::get_if<SomeIntExpr>(&folded.u)}) {
+            paramValue->SetExplicit(std::move(*intExpr));
+            continue;
+          }
+        }
+        std::stringstream fortran;
+        fortran << *expr;
```
tskeith: I think this should use `expr->AsFortran(fortran);`. Even though they may be equivalent now, `AsFortran` implies a guarantee (i.e. that its output is Fortran) that `operator<<` doesn't.

lib/semantics/type.cc
```diff
@@ -38,22 +39,253 @@ void DerivedTypeSpec::ReplaceScope(const Scope &scope) {
...
+        if (auto *msg{messages.Say("Value of type parameter '%s' (%s) is not "
+                                   "convertible to its type"_err_en_US,
+                name, fortran.str())}) {
+          msg->Attach(name, "declared here"_en_US);
```
tskeith: Is it worthwhile using `evaluate::AttachDeclaration` here?

lib/semantics/symbol.h
```diff
@@ -228,10 +228,12 @@ class DerivedTypeDetails {
   const std::list<SourceName> &paramNames() const { return paramNames_; }
   const SymbolVector &paramDecls() const { return paramDecls_; }
   bool sequence() const { return sequence_; }
+  bool forward() const { return forward_; }
```
tskeith: I think this deserves a comment -- it's not obvious what it means.

lib/semantics/type.cc
klausler: Or I could call it.

lib/semantics/symbol.h
klausler: I'll improve the name is `isForwardReferenced()`.

lib/semantics/resolve-names.cc
```diff
@@ -4443,8 +4431,17 @@ std::optional<DerivedTypeSpec> DeclarationVisitor::ResolveDerivedType(
...
-    return std::nullopt;
+    if (allowForwardReferenceToDerivedType()) {
+      Symbol &forward{MakeSymbol(InclusiveScope(), name.source, Attrs{})};
+      DerivedTypeDetails details;
+      details.set_forward();
+      forward.set_details(std::move(details));
+      Resolve(name, forward);
+      symbol = &forward;
+    } else {
+      Say(name, "Derived type '%s' not found"_err_en_US);
```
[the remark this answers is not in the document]
klausler: If it could be forward-referenced in this context, we wouldn't emit this message.

PR #848, opened 2019-12-02
Dodge GCC 7.2 bug

klausler:
GCC 7.2 has a bug that I've had to work around elsewhere in the f18 codebase -- in `visitors` lists with `auto` catch-alls, the `auto` case is sometimes incorrectly selected rather than an explicit match.

### DIFF
--- a/lib/semantics/symbol.cc
+++ b/lib/semantics/symbol.cc
@@ -708,10 +708,22 @@ bool GenericKind::IsOperator() const {
 
 std::string GenericKind::ToString() const {
   return std::visit(
-      common::visitors{
-          [](const OtherKind &x) { return EnumToString(x); },
-          [](const DefinedIo &x) { return EnumToString(x); },
-          [](const auto &x) { return common::EnumToString(x); },
+      common::visitors {
+        [](const OtherKind &x) { return EnumToString(x); },
+            [](const DefinedIo &x) { return EnumToString(x); },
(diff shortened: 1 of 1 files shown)


lib/semantics/symbol.cc
```diff
@@ -708,10 +708,22 @@ bool GenericKind::IsOperator() const {
...
+            },
+            [](const common::LogicalOperator &x) {
+              return common::EnumToString(x);
+            },
+            [](const common::RelationalOperator &x) {
+              return common::EnumToString(x);
+            },
+#else
+            [](const auto &x) { return common::EnumToString(x); },
+#endif
```
tskeith: It would probably work to use a single lambda with `std::is_same_v` tests. In that case it would be good to comment why `common::visitors` is being avoided.

For example,
```
  return std::visit(
      [](const auto &x) {
        using T = std::decay_t<decltype(x)>;
        if constexpr (std::is_same_v<T, OtherKind> || std::is_same_v<T, DefinedIo>) {
          return EnumToString(x);
        } else {
          return common::EnumToString(x);
        }
      },
      u);
```

klausler: Yes, I tried it that way as well, and it works, but eventually we'll deprecate GCC 7.2, and after the conditional patch is removed the code will look the way we really want it to look.

Issue #843, opened 2019-11-25
`Include` statement is ignored


fghanim:
Whenever a Fortran source has an `include` for a header file, the `include` statement is ignored. 
- If the header file exists, it is not included. When the `-funparse` command line option is used, the generated output does not contain the `include`.
- If the header file does not exist, no errors are generated.

tskeith:
The second point is a bug -- an error should be reported if the file is not found.

For the first one, the unparsed output is not expected to contain the `include` line. It contains the unparsed form of the Fortran source included from the file.

fghanim:
Maybe I should explain the first point more. The expected behavior is what you say, if you include a header file, the contents of that file is included instead.
only that's not what happens. nothing is included from that file. Not even the things that are used.

So based on that, I assumed that given that `include` in fortran is a statement, not a preprocessor directive. Maybe flang parsed it like other fortran statements and so it didn't replace it, with the intention that the other compiler will handle it - which means unparsing should produce it back, and that didn't happen either. and so I mentioned it.

klausler:
```
f18 λ cat a.f90
include 'b.f90'
end
f18 λ cat b.f90
print *, 'works'
f18 λ f18 a.f90
f18 λ ./a.out
 works
```

sscalpone:
Using  Peter's example from above with the -funparse option:
```
% f18 -funparse a.f90
 PRINT *, "works"
END PROGRAM
```

fghanim:
I stand corrected. It turns out I was including a file with the wrong path. So when I didn't get an error, I mistakenly thought It was being ignored, when in reality it was just the same bug in the second point from my original post.

PR #850, opened 2019-12-03
Fix bug #843, bad recovery from failed opens of source files

klausler:


tskeith:
There is no issue 853. Is this fixing #843?

klausler:
> There is no issue 853. Is this fixing #843?

Yes, sorry.

PR #853, opened 2019-12-05
Prevent multiple CI runs on a single architecture

DavidTruby:
I believe that the CI failures might be due to Drone running multiple builds on the same machine if you have multiple commits waiting for CI. I've added here a concurrency limit, meaning that new commits will wait for previous commits CI to run before starting.

This would explain why the problem didn't appear in my fork and only manifested in the upstream repo: effectively because the upstream repo has a lot more traffic on it than my own fork.

### DIFF
--- a/.drone.star
+++ b/.drone.star
@@ -1,7 +1,8 @@
 def clang(arch):
     return {
         "kind": "pipeline",
-        "name": "%s-clang" % arch,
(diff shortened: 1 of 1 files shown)


DavidTruby:
The CI build failed for this commit, so this can't have been the problem unfortunately.

PR #854, opened 2019-12-05
Add script to flatten git history for llvm monorepo submission

peterwaller-arm:
## [Click here for the latest description of intent posted to flang-dev](http://lists.llvm.org/pipermail/flang-dev/2019-December/000122.html)

Please note this pull request is not intended to be merged, it is here for review only.

Please see `flatten.cpp`, which contains significant self-documentation. `flatten.sh` was an earlier shell script with a similar intent, but is now obsolete, and just left here for reference.

In summary, this program takes as input the upstream llvm-project monoreo and the f18 project. As output, it generates from the f18 history a new series of commits following on from the HEAD of the llvm-project repository. Additionally, the commits are rewritten so that it appears as though all of the project history happened inside the `flang/` directory.

The key reason to do this is to eliminate merge commits. Secondarily, we improve the commit messages so that references to pull requests and issues point back into this repository correctly and will continue to work into the future, rather than referencing arbitrary unrelated issues llvm-project.

To achieve the flattening we follow the first parent history of f18. When a merge commit is encountered, the second-parent is rebased onto the first-parent. This is of course "dangerous", because the commits on the second-parent lineage will not have the same checkout as the original commits. To mitigate this, we make certain that by the end of the rebase, the checkouts are the same as the original master.

There are a few cases where they are not, and in those cases the second-parent history is squashed into a single commit.

However, it is guaranteed that the checkouts at the f18 merge commits will be equivalent, within the flang directory.

### DIFF
--- a/flatten.cpp
+++ b/flatten.cpp
@@ -1,0 +1,916 @@
+// Copyright (c) 2019, Arm Ltd.  All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Reminder: no warranty with this program. I recommend using a fresh checkout.
+
+// Compile with:
+//   clang++ -Wall -Werror -O2 flatten.cpp -lgit2
+// Run with f18 in PWD or argv[1]:
+//   time ./a.out ~/.local/src/github.com/flang-compiler/f18/
+
+// To get a rewritten history, do this:
+//
+//   sudo apt install -y libgit2-dev # or equivalent
+//   git clone https://github.com/flang-compiler/f18
+//   git remote add llvm-project https://github.com/llvm/llvm-project
+//   git fetch llvm-project
+//   clang++ -DREPLACE_REFERENCES -Wall -Werror -O2 flatten.cpp -lgit2
+//   ./a.out
+
+// Inputs:
+// * a ref called origin/master, representing f18 history
+// * (optionally) a ref called llvm-project/master, representing llvm upstream
+// * (optionally) branches called rebase-{12 digit merge sha}, representing the
+//   manual rebase of tricky cases.
+//
+// Outputs:
+// * A branch called rewritten-history-v2, with a linearized f18 history.
+// * A branch called rewritten-history-v2-llvm-project-merge, representing the
+//   renaming of the project under /flang/ and taking llvm-project/master as the
+//   new parent for the (original) root commit.
+//
+// This program is meant to be idempotent and should not write to the working
+// directory, it simply takes refs as input and produces them as output.
+
+// Key concepts:
+//
+// * The checkout that git gives you for a commit is called a "tree", which is
+//   determined by a recursive checksum of the directory structure. If two
+//   commits have the same tree ("treesame"), then they are by definition
+//   equivalent when you check them out.
+//
+// * Lineage of the "master" branch is taken by following the first parents of
+//   each commit. To see this in git log, run `git log --first-parent`. This
+//   effectively ignores the second-parent history (i.e. commits that happened
+//   on branches).
+//
+// * By construction it is arranged that the trees of the first-parent history
+//   are preserved. This means "the code on the master branch is the same before
+//   and after rewrite".
+//
+// * Preserving the non-first-parent commits is trickier, and requires a rebase.
+//
+// * If nothing changed on the master branch during a feature branch, a rebase
+//   will not change the trees of the feature branch, so trees of those commits
+//   will still be the same. It's like rewriting the merge as a fast-forward.
+//
+// * However, if something happened on the master branch during the feature
+//   branch, then a rebase *must* create new trees. This implies code which
+//   might not build. As an example, imagine a case where a class is renamed on
+//   master, and the old name is used in the feature branch (until it's fixed at
+//   some point by the time it is merged).
+//
+// * By the end of the rebase, we assert that the trees are the same as those
+//   merged into master. So code in the middle of the rebased feature branch may
+//   not build, but at least the overall result of the feature branch will be as
+//   good as master was. Thankfully this is relatively rare.
+//
+// * If a branch exists called rebase-{sha of merge commit}, that branch is
+//   substituted in place of the merge commit. This allows manually rebasing
+//   tricky merges.
+//
+// * For the non-treesame, we can take a second-order diff (diff-of-diff)
+//   comparing those commits before and after rewrite, and ensure that only line
+//   numbers and context changed. This is almost totally the case.
+
+// Using the following script, it is possible to see whether non-TREESAME
+// patches still have the same diff, modulo blank lines, by taking a
+// second-order diff.
+//
+// git log --grep=TREESAME --invert-grep  --format="%h %(trailers:key=Original-commit)" rewritten-history-v2 |
+//   sed -n 's|Original-commit: flang-compiler/f18@||p' |
+//   while read NEW ORIG
+//   do
+//     echo ORIG NEW: $ORIG $NEW
+//     git show $ORIG > a
+//     git show $NEW > b
+//     sed -r -i \
+//       -e 's/@@ .* @@/@@ Numbers @@/g' \
+//       -e '/^(commit|index) .*/d' \
+//       -e '/Original-commit.*/d' \
+//       -e '/^\s$/d' \
+//        a b
+//     git diff --color --no-index a b
+//   done |& less -SR
+
+#include <array>
+#include <cstdio>
+#include <cstring>
+#include <map>
+#include <sstream>
+
+#ifndef NO_REPLACE_REFERENCES
+#include <regex>
+#endif
+
+#include <sys/stat.h>
+#include <sys/types.h>
+
+#include <git2.h>
+#include <git2/sys/commit.h>
+
+void check(int error, const char *message, const char *extra) {
+    const git_error *err;
+    const char *msg = "", *spacer = "";
+
+    if (!error)
+        return;
+
+    if ((err = giterr_last()) != NULL && err->message != NULL) {
+        msg = err->message;
+        spacer = " - ";
+    }
+
+    if (extra)
+        fprintf(stderr, "%s '%s' [%d]%s%s\n", message, extra, error, spacer, msg);
+    else
+        fprintf(stderr, "%s [%d]%s%s\n", message, error, spacer, msg);
+
+    exit(1);
+}
+
+int n_conflicts = 0, n_discards = 0;
+
+// Copy src string to dst string, rewriting issue references.
+char *rewrite_issue_references(char *dst, const char *src) {
+#ifndef NO_REPLACE_REFERENCES
+    const char *src_end = src + strlen(src);
+    // return src_end;
+    char *new_end = std::regex_replace(
+        dst, src, src_end,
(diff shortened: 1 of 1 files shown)


flatten.sh
```diff
@@ -0,0 +1,118 @@
...
+
+# Known deficiencies:
+#
+# * The commit messages are ugly. These can be tidied up fairly easily once a
+#   format is agreed.
+#
+# * Individual commits within PRs are lost. This is somewhat fundamental, and
+#   could be worked around with some effort; it would require rebasing a number
+#   of patches, which is risky and time consuming, so instead the working tree
+#   is taken as-is from the merge commit.
```
tskeith: I think this is a serious short-coming. Squashing together all of the commits in a pull request can lose a lot of information. See the new commit 5e2cd5eb066062b7bd6d7874817d5213187bccb2 for an example. It combines several commits, some of are refactorings and some are function changes.

peterwaller-arm: I agree that isn't great.

The only alternative I'm aware of is to do effectively an interactive rebase, which would also flatten history, discarding only the (otherwise empty) merge commits. As far as I can tell, that would require some effort to deal with various bits of the history where commits can't be replayed without the merge commit. I don't see why this couldn't be done in principle, but it would require additional engineering effort.

RichBarton-Arm: @peterwaller-arm mentioned to me this evening that he might have a mitigation to this in that the new squashed commits could link back to the original F18 git shas that went into them. So the F18 git history in LLVM monorepo is straightline (and cheaply done) but has some commits that combine things together but with the original commits, that can be clicked through to in github, to help separate them.

peterwaller-arm:
I've posted an update here: http://lists.llvm.org/pipermail/flang-dev/2019-December/000105.html

Key changes:
* New way of preserving commits, such that we preserve many more commits.
* Added some trailer headers and `Reviewed-on` links.

The script got quite a bit more complicated, but I think we can have a base level of confidence that the history is not significantly messed up because the trees of the mainline commits are the same.

There are a couple of known deficiencies:

* There are 66 commits from 6 merges which have been sqaushed into 6 commits. As far as I currently know, fixing this would require a manual rebase of those 66 commits.
* Original-Commit is not consistently set, it's missing for commits which have been rebased. This is a smallish bug that could be fixed.
* There are some "merge commits" (with 1 parent) in the rewritten history which complicate things. The history is still linear, it just has these extra commits. I haven't yet had time to think about what to do with them but I believe there is likely to be a simple solution.

flatten.sh
peterwaller-arm: The next update does a better job. Now there are only 6 merge commits which get squashed. As described elsewhere, fixing those would require manual work. Please take another look.

flatten.sh
```diff
@@ -0,0 +1,277 @@
...
+        # some interesting topologies in the original history.
+        ORIG_SHA=$SHA
+        # echo SHA=$SHA
+        # echo ORIG_SHA=$ORIG_SHA
+        ((N_PRESERVED++)) || true
+        {
+            git show --no-patch --format=%B "$SHA"
+            # echo Original-Commit: flang-compiler/flang@${SHA}
+            make_reviewed_on ${merge_commit}
+        } > message.txt
```
tskeith: I have one small suggestion here: It's good to have the link back to the original pull request, but I think it should be easier to tell which PR a commit is from, for example, in the output of `git log --oneline`. We've lost the information that the commits were grouped together in a pull request.

My suggestion is to prefix the title with something like "PR123: " when the commit comes from pull request 123.

peterwaller-arm:
I have pushed an update including the C rewrite.

The full description is in the [mailing list post](http://lists.llvm.org/pipermail/flang-dev/2019-December/000122.html).

Other key links:


* Tree, merged with LLVM: 
https://github.com/peterwaller-arm/f18/tree/rewritten-history-v2-llvm-project-merge

* Rewritten history: 
https://github.com/peterwaller-arm/f18/commits/rewritten-history-v2-llvm-project-merge

* Rewritten history without llvm merge: 
https://github.com/peterwaller-arm/f18/commits/rewritten-history-v2

flatten.sh
peterwaller-arm: In the latest flatten.cpp now pushed, I wrote it as `[flang-compiler/f18#NNN]` so that the cross reference will not reference the LLVM repository, but this repository.

peterwaller-arm:
@tskeith: Aaron Smith requested on the llvm-dev mailing list that the commit subjects don't include `[flang-compiler/f18#123]` but just say `[flang]`, with the rationale that nothing is lost -- and his comments come after we now retain the merge commits. Do you think we can change the prefix to `[flang]` now? Individual commits won't show the PR number in the title, but it should be clear from the position of the empty merge commits, and the commit message body does have the "Reviewed-on" trailer.

tskeith:
> @tskeith: Aaron Smith requested on the llvm-dev mailing list that the commit subjects don't include `[flang-compiler/f18#123]` but just say `[flang]`, with the rationale that nothing is lost -- and his comments come after we now retain the merge commits. Do you think we can change the prefix to `[flang]` now? Individual commits won't show the PR number in the title, but it should be clear from the position of the empty merge commits, and the commit message body does have the "Reviewed-on" trailer.

So the history would look like one or more commits that went into PR 123 followed by an empty commit titled "Merge pull request flang-compiler#123 from ...", right? That seems fine to me.

But looking at your rewritten history, there seem to be a few that don't look like that. For example, there is a commit for PR #867 but no merge commit. Why is that?

peterwaller-arm:
> But looking at your rewritten history, there seem to be a few that don't look like that. For example, there is a commit for PR #867 but no merge commit. Why is that?

```
$ git log --oneline --graph --decorate b3bb698086a..3d05d39e363
*   3d05d39e363 Merge pull request #868 from schweitzpgi/remove-flatten
|\  
| * d7554c76050 remove some old files
* | 61bb9db531a Flang relicense: LICENSE file and one change (#867)
* | ae4868e572c Updated call10.f90 with more test for #794-Check for PURE procedures containing image control statements (#857)
* |   53545d7436c Merge pull request #865 from flang-compiler/jpr-fix-cplx-power
|\ \  
| * | 9a84bcb9ac3 Fix INTGER**COMPLEX regression introduced by PR 858 Fix bad copy past that had introduced an unitialized value use.
|  /  
* | 4771f8cb789 Fix file path in cmake (#844)
|/  
* d1b6e3cd2c4 Merge pull request #864 from flang-compiler/pmk-fix
* e58c4e53c2f Revert recent addition of ActualArgument::PassedObject
```

The above hitstory shows the commit which "merged" 61bb9db531a is actually a cherry-pick/squash commit. So where github shows "merged" in the user interface, I believe those PRs get a commit message on master which looks like `<Pull request title> (#<PR number>)`, and not actually a merge commit. There are a fair number of those in that form.

That is actually a case where I haven't generated "Reviewed-on" headers, because they aren't distinguishable from commits where someone writes `(#<issue number>)` from the history alone, I would also need to obtain a list of numbers which are pull requests to achieve that.

peterwaller-arm:
> I would also need to obtain a list of numbers which are pull requests to achieve that.

For now I don't propose to fix up these edge cases, unless it seems particularly important. I have the impression they are relatively rare, but they do represent a gap of some unknown size in the `Reviewed-on` trailer headers.

peterwaller-arm:
I got a few minutes at a computer. There are 32 such squash commits. I was able to verify using the github API that all commits with subjects ending in "(#[0-9]+)" are indeed pull requests.

One-liner to grab PR numbers: `curl https://api.github.com/repos/flang-compiler/f18/pulls?state=all'&'page='[1-23]' | grep -v '_curl_' | jq -c -s '.[][].number'`

I'm assuming no such commits will land between now and when we do a rewrite which name an issue number in that position rather than a PR, but the worst that happens is that we generate incorrect Reviewed-on lines for them.

A few straightforward changes above:

* These commits now have a Reviewed-on reference.
* Subjects no longer begin `[flang-compiler/f18#nnn]`
* I switched the sense of the tree-same-pre-rewrite header. Now only commits which don't result in the same checkout are marked. I thought this was obviously the better way around in hindsight, since it means there is a hint there on the commits which need it that they aren't equal. It also makes them easier to search for by grepping the commit log in the pager. We have't discussed this header very much. I think it is nice to leave a hint that these commits aren't quite what they used to be before the rebase. Hopefully if it ever comes up, it is recognized as such and searched for, which leads back to this PR. We could just drop it if it nobody else thinks this is useful.

peterwaller-arm:
And here's the latest history:

https://github.com/peterwaller-arm/f18/commits/rewritten-history-v3

peterwaller-arm:
Two proposed changes:

* I note that the MLIR commit was merged in as a second root, so llvm/llvm-project@aed0d21a62d does not have any parents. We could do the same. See [http://lists.llvm.org/pipermail/llvm-dev/2020-January/137960.html](this post for further detail).

* There was a suggestion to prefix flang commits with `[flang]`, which seems reasonable to me. I removed `[flang-compiler/f18#nnn]` but forgot to re-add that.

peterwaller-arm:
I've added a few commits above to tweak minor things that came up. Now we prepend all the flang commits with `[Flang]`, as is considered polite in LLVM.

The program now generates the merge commit itself. So the program now directly produces the commit we intend to push into LLVM.

In contrast to previously where we based the flang history on top of the latest llvm history, the latest program now generates a zero-parent first commit. The effect of this is that if I pick a random flang commit [it only has the flang folder in it](https://github.com/peterwaller-arm/f18/tree/9bafcc671d2c08ce4ce7f5769e82b7ecd9722c55):

Screenshot from 2020-01-09 14-04-36

We don't have to do it this way, but it is how MLIR did it, and it makes some sense to me because up to that point the trees were independent. I raised this on the mailing list [but no-one commented either way](http://lists.llvm.org/pipermail/llvm-dev/2020-January/137960.html). I think it's best to follow MLIR's lead.

#### Here is the new flang history lineage (second-parent of the merge): https://github.com/peterwaller-arm/f18/commits/rewritten-history-v4-llvm-project-merge%5E2

Screenshot from 2020-01-09 14-05-31

#### Here is the llvm lineage after the merge: https://github.com/peterwaller-arm/f18/commits/rewritten-history-v4-llvm-project-merge -- note that git's display of commits appears to sort them in date order, rather than topologically, so the flang commits appear interleaved with the llvm commits:

Screenshot from 2020-01-09 14-06-08

[...]

Screenshot from 2020-01-09 14-06-37

flatten.cpp
```diff
@@ -0,0 +1,916 @@
...
+            "\n"
+            "The history of f18 was rewritten to:\n"
+            "\n"
+            "* Put the code under /flang/.\n"
+            "* Linearize the history.\n"
+            "* Rewrite commit messages so that issue and PR numbers point to the old repository.\n"
+            "\n"
+            "Updates: flang-compiler/f18#876 (submission into llvm-project)\n"
+            "Mailing-list: http://lists.llvm.org/pipermail/llvm-dev/2020-January/137989.html ([llvm-dev] Flang landing in the monorepo - next Monday!)\n"
+            "Mailing-list: http://lists.llvm.org/pipermail/llvm-dev/2019-December/137661.html ([llvm-dev] Flang landing in the monorepo)\n";
```
peterwaller-arm: @tskeith This is the commit message of the merge. I chose to include some extra references here so that people unfamiliar with what's going on have something to go by when they see the merge in their histories. Please review this, which is new.

tskeith: > @tskeith This is [the commit message](https://github.com/peterwaller-arm/f18/commit/527fdd0649249bd076681cf0c2230eb096832f7f) of the merge. I chose to include some extra references here so that people unfamiliar with what's going on have something to go by when they see the merge in their histories. Please review this, which is new.

That looks okay to me.

RichBarton-Arm:
@sscalpone Do you think it makes sense to merge this PR into master so we have the script available in source code when the upstreaming to LLVM time comes?

peterwaller-arm:
In the above force push:

* Rebased against current master.
* Removed obsolete flatten.sh.
* Squashed the history to a single commit.

I re-ran the script against the current llvm-project and f18 HEADs and everything continued continues to pass the sniff test.

RichBarton-Arm:
@sscalpone - this is ready to merge.

PR #855, opened 2019-12-05
Fix problems with passed-object arguments...

klausler:
... by deferring the identification of their indices in the dummy argument list,
simplifying their representation, completing the representation
of their actual arguments, and (while I'm here) resolving
calls to type-bound procedures whose bindings are known at
compilation time.

Button up class ActualArgument by making remaining data
members private and adding accessors & mutators.

### DIFF
--- a/lib/evaluate/call.cc
+++ b/lib/evaluate/call.cc
@@ -59,12 +59,11 @@ int ActualArgument::Rank() const {
 }
 
 bool ActualArgument::operator==(const ActualArgument &that) const {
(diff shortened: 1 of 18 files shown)


tskeith:
This would be easier to review if the refactoring were separate from the functional changes.

klausler:
> This would be easier to review if the refactoring were separate from the functional changes.

Probably so.  In this case, it wasn't clear that the movement of code and adjustments to data structures were doing to be necessary until I had used them to successfully fix the problem.

PR #856, opened 2019-12-06
Allow assignment between INTEGER and LOGICAL as extension

klausler:
GNU Fortran, PGI, and Intel Fortran accept intrinsic assignments of LOGICAL to INTEGER and vice versa.  Extend f18 to also allow these assignments as an extension, with the usual optional warnings and ability to disable with `-Mstandard`.  If the program has defined `ASSIGNMENT(=)` procedures that handle these cases, the extension is overridden.

### DIFF
--- a/documentation/Extensions.md
+++ b/documentation/Extensions.md
@@ -109,6 +109,8 @@ Extensions, deletions, and legacy features supported by default
 * When a dummy argument is `POINTER` or `ALLOCATABLE` and is `INTENT(IN)`, we
   relax enforcement of some requirements on actual arguments that must otherwise
   hold true for definable arguments.
+* Assignment of `LOGICAL` to `INTEGER` and vice versa (but not other types).
+  The values are normalized.
 
 Extensions supported when enabled by options
 --------------------------------------------
@@ -140,7 +142,7 @@ Extensions and legacy features deliberately not supported
 * Defining an explicit interface for a subprogram within itself (PGI only)
 * USE association of a procedure interface within that same procedure's definition
 * NULL() as a structure constructor expression for an ALLOCATABLE component (PGI).
-* Conversion of LOGICAL to INTEGER.
+* Conversion of LOGICAL to INTEGER in expressions.
 * IF (integer expression) THEN ... END IF  (PGI/Intel)
 * Comparsion of LOGICAL with ==/.EQ. rather than .EQV. (also .NEQV.) (PGI/Intel)
 * Procedure pointers in COMMON blocks (PGI/Intel)
--- a/lib/common/Fortran-features.h
+++ b/lib/common/Fortran-features.h
@@ -33,7 +33,7 @@ ENUM_CLASS(LanguageFeature, BackslashEscapes, OldDebugLines,
     Hollerith, ArithmeticIF, Assign, AssignedGOTO, Pause, OpenMP,
     CruftAfterAmpersand, ClassicCComments, AdditionalFormats, BigIntLiterals,
     RealDoControls, EquivalenceNumericWithCharacter, AdditionalIntrinsics,
-    AnonymousParents, OldLabelDoEndStatements)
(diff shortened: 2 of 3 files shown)


klausler:
I should clarify that this change only affects intrinsic assignments; it does not allow mixed INTEGER & LOGICAL operands in expressions, LOGICAL actual arguments for INTEGER dummy arguments, or other plausible scenarios.

documentation/Extensions.md
```diff
@@ -140,7 +141,7 @@ Extensions and legacy features deliberately not supported
 * Defining an explicit interface for a subprogram within itself (PGI only)
 * USE association of a procedure interface within that same procedure's definition
 * NULL() as a structure constructor expression for an ALLOCATABLE component (PGI).
-* Conversion of LOGICAL to INTEGER.
+* Conversion of LOGICAL to INTEGER in expressions.
```
tskeith: The other compilers that support this (PGI/Intel) also allow assignment between LOGICAL and other numeric types. If we're not supporting it we should mention that here.

jeanPerier:
Looks good to me, I just have a question regarding how this extension should be implemented.

When integer and logical are the same size `TRANSFER(source=logical, mold=integer)` makes the most sens, but the extension allows assigning logical to integer of different sizes and vice-versa (e.g. `integer(4) = logical(2)`).  pgfortan and ifort seem to also support this. There are different alternatives one can think of to do such conversions:
- `INT(TRANSFER(source=logical(2), mold=integer(2)), 4)` (looks like this is what pgfortran is doing)
- `TRANSFER(source=logical(2), mold=integer(4))`, gives different results compare to pgfortran  (so probably a bad idea).
- `TRANSFER(source=LOGICAL(logical(2), logical(4)), mold=integer(4))` I do not see a difference with the first one (though I am not a 100% sure), so maybe that is actually what pgfortran is doing.

Will semantic analysis of assignments insert these conversions on top the expressions or is it expected lowering should decide how to make these conversions in assignments ?

klausler:
> Looks good to me, I just have a question regarding how this extension should be implemented.
> 
> When integer and logical are the same size `TRANSFER(source=logical, mold=integer)` makes the most sens, but the extension allows assigning logical to integer of different sizes and vice-versa (e.g. `integer(4) = logical(2)`). pgfortan and ifort seem to also support this. There are different alternatives one can think of to do such conversions:
> 
> * `INT(TRANSFER(source=logical(2), mold=integer(2)), 4)` (looks like this is what pgfortran is doing)
> * `TRANSFER(source=logical(2), mold=integer(4))`, gives different results compare to pgfortran  (so probably a bad idea).
> * `TRANSFER(source=LOGICAL(logical(2), logical(4)), mold=integer(4))` I do not see a difference with the first one (though I am not a 100% sure), so maybe that is actually what pgfortran is doing.
> 
> Will semantic analysis of assignments insert these conversions on top the expressions or is it expected lowering should decide how to make these conversions in assignments ?

I would like us to try a more principled interpretation of the feature with normalization, by treating `logical=integer` as `logical=integer/=0` and `integer=logical` as `integer=merge(TRUTH,0,logical)`.  I don't think that programs using these assignments really want to move raw bits.  I could be wrong.

tskeith:
> I would like us to try a more principled interpretation of the feature with normalization, by treating `logical=integer` as `logical=integer/=0` and `integer=logical` as `integer=merge(TRUTH,0,logical)`. I don't think that programs using these assignments really want to move raw bits. I could be wrong.

I agree. I think it is better if you can't get non-canonical LOGICAL values unless you use TRANSFER.

It looks like this is how ifort interprets it, whereas pgfortran transfers the bits. (The others don't allow it.) So there is no consensus as to what it should mean.

klausler:
gfortran and ifort copy bits when assigning LOGICAL to INTEGER, but normalizes INTEGER to canonical LOGICAL truth values.

pgf90 copies bits in both directions.

Issue #794, opened 2019-10-25
Check for PURE procedures containing image control statements


psteinfeld:
Constraint C1159 states that a PURE subprogram shall not contain an image control statement.

This check can be modeled after a similar check for image control statements appearing in DO CONCURRENT statements.  The code to imitated is in check-do.cc in the "Post()" function for "parser::ExecutableConstruct".

kiranktp:
lib/semantics/check-purity.cc does the check now: 


$ cat sync_image_pure.f90
PURE FUNCTION DOUBLE(X)
  REAL, INTENT(IN) :: X
integer :: img, nimgs, i[*], tmp
                                 ! implicit sync all
  img = this_image()
nimgs = num_images()
   i = img                       ! i is ready to use

 if ( img .eq. 1 ) then
   sync images( nimgs )          ! explicit sync 1 with last img
    tmp = i[ nimgs ]
   sync images( nimgs )          ! explicit sync 2 with last img
      i = tmp
 end if


 if ( img .eq. nimgs ) then
   sync images( 1 )              ! explicit sync 1 with img 1
    tmp = i[ 1 ]
   sync images( 1 )              ! explicit sync 2 with img 1
      i = tmp
 end if
 write (*,*) img, i
                                 ! all other images wait here
END FUNCTION DOUBLE
$F18_HOME/bin/f18 -fdebug-semantics sync_image_pure.f90
sync_image_pure.f90:5:9: error: Procedure referenced in PURE subprogram 'double' must be PURE too
    img = this_image()
          ^^^^^^^^^^^^
sync_image_pure.f90:6:9: error: Procedure referenced in PURE subprogram 'double' must be PURE too
  nimgs = num_images()
          ^^^^^^^^^^^^
sync_image_pure.f90:10:4: error: An image control statement may not appear in a PURE subprogram
     sync images( nimgs )          ! explicit sync 1 with last img
     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
sync_image_pure.f90:12:4: error: An image control statement may not appear in a PURE subprogram
     sync images( nimgs )          ! explicit sync 2 with last img
     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
sync_image_pure.f90:18:4: error: An image control statement may not appear in a PURE subprogram
[...]

PR #857, opened 2019-12-06
Added a test case for #794-Check for PURE procedures containing image…

kiranktp:
Added a test case to validate presence of image control statements in PURE procedures.

### DIFF
--- a/test/semantics/call10.f90
+++ b/test/semantics/call10.f90
@@ -192,7 +192,36 @@ module m
   pure subroutine s13
     !ERROR: An image control statement may not appear in a PURE subprogram
     sync all ! C1599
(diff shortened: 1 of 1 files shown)


tskeith:
The tests look good, but I think they should go into `call10.f90`. That's where SYNC ALL is already being tested with a TODO for the other cases.

kiranktp:
I have removed the newly added test file by moving the checks to call10.f90 file.

sscalpone:
@kiranktp Please squash and make a single commit.

kiranktp:
> @kiranktp Please squash and make a single commit.

I have squashed multiple commit and made it a single commit.

PR #858, opened 2019-12-06
Fix COMPLEX-REAL `/` and `**` operand promotion

jeanPerier:
For `**` and `/`, the real operand was always converted to the complex operand type.
The highest precision should be used instead. This fix converts the real to a complex of the same kind before applying the promotion rules for complex operands.

### DIFF
--- a/lib/evaluate/tools.cc
+++ b/lib/evaluate/tools.cc
@@ -184,6 +184,20 @@ Expr<SomeReal> GetComplexPart(const Expr<SomeComplex> &z, bool isImaginary) {
       z.u);
 }
 
+// Convert REAL to COMPLEX of the same kind. Preserving the real operand kind
+// and then applying complex operand promotion rules allows the result to have
+// the highest precision of REAL and COMPLEX operands as required by Fortran
+// 2018 10.9.1.3.
+Expr<SomeComplex> PromoteRealToComplex(Expr<SomeReal> &&someX) {
+  return std::visit(
+      [](auto &&x) {
+        using RT = ResultType<decltype(x)>;
+        return AsCategoryExpr(ComplexConstructor<RT::kind>{
+            std::move(x), AsExpr(Constant<RT>{Scalar<RT>{}})});
+      },
+      std::move(someX.u));
+}
+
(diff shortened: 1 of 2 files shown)


klausler:
Where in the Fortran 2018 standard is the kind of the result of an intrinsic numeric operation defined?  The standard says that operands that differ in type or kind from the type or kind of the result are converted to the type & kind of the result, but it doesn't define the kind of the result anywhere that I can find.

lib/evaluate/tools.cc
```diff
@@ -230,8 +241,15 @@ std::optional<Expr<SomeType>> MixedComplexLeft(
         std::move(zx.u)));
   } else if (defaultRealKind != 666) {  // dodge unused parameter warning
     // (a,b) ** x -> (a,b) ** (x,0)
-    Expr<SomeComplex> zy{ConvertTo(zx, std::move(iry))};
-    return Package(PromoteAndCombine<OPR>(std::move(zx), std::move(zy)));
+    if constexpr (RCAT == TypeCategory::Integer) {
+      Expr<SomeComplex> zy{ConvertTo(zx, std::move(iry))};
+      return Package(PromoteAndCombine<OPR>(std::move(zx), std::move(zy)));
+    } else {
+      static_assert(RCAT == TypeCategory::Real,
```
klausler: The asserts seem unnecessary; if the operand is not real, the call to `PromoteRealToComplex` will fail anyway.

jeanPerier: Right, removed.

jeanPerier:
> Where in the Fortran 2018 standard is the kind of the result of an intrinsic numeric operation defined? The standard says that operands that differ in type or kind from the type or kind of the result are converted to the type & kind of the result, but it doesn't define the kind of the result anywhere that I can find.
> 
> EDIT: I found it, in 10.1.9.3 paragraph 4. The kind of the result of a mixed "real * complex" operation must be the kind with the greatest decimal precision, so "double precision * complex" produces a "double complex" for example.

It also took me more than 10mins to go from table 10.2 to section 10.9.1.3 using ctrl-f on "kind"... From your comment, it seems it is not only me and this info is hard to find so I added a ref to the section in a comment.

Issue #658, opened 2019-08-14
error: file is empty


sscalpone:
There's a real application that has an empty file in its makefile.

PGI issues a warning.  Intel makes a remark.  Both create object files with no content.
```
% f18 b.f90
../f18/build/tools/f18/bin/f18: could not scan b.f90
error: file is empty
```

klausler:
Fixed.

PR #859, opened 2019-12-06
Accept source files with no program units (#658)

klausler:
Fortran requires a "program" to contain at least one program-unit.  Accept an effectively empty source file as an extension.


sscalpone:
Is this extension listed in Extensions.md?

klausler:
> Is this extension listed in Extensions.md?

It is so now.

PR #860, opened 2019-12-07
Semantic checks to see if a DO variable is modified

psteinfeld:
This is driven by Section 11.1.7.4.3, paragraph 2, which states:
Except for the incrementation of the DO variable that occurs in step (3), the 
DO variable shall neither be redefined nor become undefined while the DO 
construct is active.

Sections 19.6.5 and 19.6.6 describe the events that can cause variables to
become defined and undefined.  Section 19.6.7 describes the contexts that imply
the redefinition or undefinition of a variable.  For all of these situations
that are reasonably easy to discover at compile time, we should emit messages.

I added infrastructure to SemanticsContext to track active DO variables and
the source locations where they appear in DO statements.  I also added code
to tools.[h,cc] to check to see if a DO variable is already defined, and, if
so, to emit an error message along with a reference to the relevant DO
construct.  I also added calls to several places where variables are defined
to determine if the definitions are happening in the context of an active DO
construct.

### DIFF
--- a/lib/semantics/check-allocate.cc
+++ b/lib/semantics/check-allocate.cc
@@ -531,6 +531,7 @@ bool AllocationCheckerHelper::RunChecks(SemanticsContext &context) {
             "Allocatable object declared here with rank %d"_en_US, rank_);
     return false;
   }
+  context.CheckDoVarRedefine(name_);
   return RunCoarrayRelatedChecks(context);
 }
 
--- a/lib/semantics/check-deallocate.cc
+++ b/lib/semantics/check-deallocate.cc
@@ -29,6 +29,8 @@ void DeallocateChecker::Leave(const parser::DeallocateStmt &deallocateStmt) {
               } else if (!IsAllocatableOrPointer(*symbol)) {  // C932
                 context_.Say(name.source,
                     "name in DEALLOCATE statement must have the ALLOCATABLE or POINTER attribute"_err_en_US);
+              } else {
+                context_.CheckDoVarRedefine(name);
               }
             },
             [&](const parser::StructureComponent &structureComponent) {
--- a/lib/semantics/check-do.cc
+++ b/lib/semantics/check-do.cc
@@ -24,6 +24,27 @@ namespace Fortran::semantics {
 
 using namespace parser::literals;
 
+using Bounds = parser::LoopControl::Bounds;
+
+static const std::list<parser::ConcurrentControl> &GetControls(
+    const parser::LoopControl &loopControl) {
+  const auto &concurrent{
+      std::get<parser::LoopControl::Concurrent>(loopControl.u)};
+  const auto &header{std::get<parser::ConcurrentHeader>(concurrent.t)};
+  return std::get<std::list<parser::ConcurrentControl>>(header.t);
+}
+
+static const Bounds &GetBounds(const parser::DoConstruct &doConstruct) {
+  auto &loopControl{doConstruct.GetLoopControl().value()};
+  return std::get<Bounds>(loopControl.u);
+}
+
+static const parser::Name &GetDoVariable(
+    const parser::DoConstruct &doConstruct) {
+  const Bounds &bounds{GetBounds(doConstruct)};
+  return bounds.name.thing;
+}
+
 // Return the (possibly null)  name of the construct
 template<typename A>
 static const parser::Name *MaybeGetConstructName(const A &a) {
@@ -428,6 +449,38 @@ class DoContext {
 public:
   DoContext(SemanticsContext &context) : context_{context} {}
 
+  // Mark this DO construct as a point of definition for the DO variables
+  // or index-names it contains.  If they're already defined, emit an error
+  // message.  We need to remember both the variable and the source location of
+  // the variable in the DO construct so that we can remove it when we leave
+  // the DO construct and use its location in error messages.
+  void DefineDoVariables(const parser::DoConstruct &doConstruct) {
+    if (doConstruct.IsDoNormal()) {
+      context_.ActivateDoVariable(GetDoVariable(doConstruct));
+    } else if (doConstruct.IsDoConcurrent()) {
+      if (const auto &loopControl{doConstruct.GetLoopControl()}) {
+        const auto &controls{GetControls(*loopControl)};
+        for (const parser::ConcurrentControl &control : controls) {
+          context_.ActivateDoVariable(std::get<parser::Name>(control.t));
+        }
+      }
+    }
+  }
+
+  // Called at the end of a DO construct to deactivate the DO construct
+  void ResetDoVariables(const parser::DoConstruct &doConstruct) {
+    if (doConstruct.IsDoNormal()) {
+      context_.DeactivateDoVariable(GetDoVariable(doConstruct));
+    } else if (doConstruct.IsDoConcurrent()) {
+      if (const auto &loopControl{doConstruct.GetLoopControl()}) {
+        const auto &controls{GetControls(*loopControl)};
+        for (const parser::ConcurrentControl &control : controls) {
+          context_.DeactivateDoVariable(std::get<parser::Name>(control.t));
+        }
+      }
+    }
+  }
+
   void Check(const parser::DoConstruct &doConstruct) {
     if (doConstruct.IsDoConcurrent()) {
       CheckDoConcurrent(doConstruct);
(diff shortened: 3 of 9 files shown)


lib/semantics/semantics.cc
```diff
@@ -208,6 +208,28 @@ void SemanticsContext::PopConstruct() {
   constructStack_.pop_back();
 }
 
+void SemanticsContext::ActivateDoVariable(
+    const Symbol *doVariable, const parser::CharBlock location) {
```
tskeith: These `doVariable` arguments should be `const Symbol &` rather than `const Symbol *`. I don't think they are ever null.

lib/semantics/semantics.h
```diff
@@ -172,6 +177,7 @@ class SemanticsContext {
 
   bool CheckError(bool);
   ConstructStack constructStack_;
+  std::map<const Symbol *, const parser::CharBlock> activeDoVariables_;
```
tskeith: If the do variable symbols are changed to refs, the key for this map should then be `SymbolRef`.

lib/semantics/tools.h
```diff
@@ -175,6 +175,15 @@ const DeclTypeSpec &FindOrInstantiateDerivedType(Scope &, DerivedTypeSpec &&,
...
 void ProcessParameterExpressions(DerivedTypeSpec &, evaluate::FoldingContext &);
 
+// Check to see if a variable being redefined is a DO variable.  If so, emit
+// a message
+void SayDoVarRedefine(
+    const Symbol &, SemanticsContext &, const parser::CharBlock &);
+void CheckDoVarRedefine(
+    const Symbol &, SemanticsContext &, const parser::CharBlock &);
+void CheckDoVarRedefine(const parser::Variable &, SemanticsContext &);
+void CheckDoVarRedefine(const parser::Name &, SemanticsContext &);
```
tskeith: It might make more sense to make these be member functions of `SemanticContext` as they depend on the active do variable there. It seems like that would make the new API simpler: just Activate/DeactivateDoVariable and CheckDoVarRedefine.

lib/semantics/check-do.h
```diff
@@ -19,9 +19,16 @@
...
 namespace Fortran::parser {
-struct DoConstruct;
+struct AssignmentStmt;
+struct ConnectSpec;
 struct CycleStmt;
+struct DoConstruct;
 struct ExitStmt;
+struct InquireSpec;
+struct IoControlSpec;
+struct IoControlSpec;
```
jeanPerier: `IoControlSpec` seems duplicated.

lib/semantics/semantics.cc
psteinfeld: Thanks, Tim.  I'll make that change.

lib/semantics/semantics.h
psteinfeld: Sounds good.

lib/semantics/tools.h
psteinfeld: Good suggestion.  I'll do that.

lib/semantics/check-do.h
psteinfeld: D'oh!  Good catch.  I'll fix that.

lib/semantics/check-deallocate.cc
```diff
@@ -35,6 +35,8 @@ void DeallocateChecker::Leave(const parser::DeallocateStmt &deallocateStmt) {
               } else if (!IsAllocatableOrPointer(*symbol)) {  // C932
                 context_.Say(name.source,
                     "name in DEALLOCATE statement must have the ALLOCATABLE or POINTER attribute"_err_en_US);
+              } else {
+                CheckDoVarRedefine(name, context_);
```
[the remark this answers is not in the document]
psteinfeld: Thanks, Jean.

With respect to other similar instances of variables/names not being definable, I'll poke around and see what I can find in the standard and make a proposal for review.  I may want to deal with generalizing the definability analysis to a future pull request.

With respect to not including allocation, my original analysis was based on sections 19.6.5 and 19.6.6, which specify the events that case variables to become defined or undefined.  I neglected to include section 19.6.7, which specifies that contexts which imply definition or undefinition of variables.  In addition to allocation, I also neglected to deal with actual arguments passed as INTENT(OUT) or INTENT(INOUT).  I'll include tests and implementation for the cases described in section 19.6.7.

lib/semantics/check-do.cc
```diff
@@ -434,6 +455,40 @@ class DoContext {
...
+  // or index-names it contains.  If they're already defined, emit an error
+  // message.  We need to remember both the variable and the source location of
+  // the variable in the DO construct so that we can remove it when we leave
+  // the DO construct and use its location in error messages.
+  void DefineDoVariables(const parser::DoConstruct &doConstruct) {
+    if (doConstruct.IsDoNormal()) {
+      context_.ActivateDoVariable(GetDoVariable(doConstruct));
+    } else if (doConstruct.IsDoConcurrent()) {
+      auto &loopControl{doConstruct.GetLoopControl()};
+      if (loopControl) {
```
tskeith: You could write these two lines as:
`if (const auto &loopControl{doConstruct.GetLoopControl()}) {`

You should use `const auto &` rather than just `auto &`.

lib/semantics/check-do.cc
```diff
@@ -434,6 +455,40 @@ class DoContext {
...
+    }
+  }
+
+  // Called at the end of a DO construct to deactivate the DO construct
+  void ResetDoVariables(const parser::DoConstruct &doConstruct) {
+    if (doConstruct.IsDoNormal()) {
+      context_.DeactivateDoVariable(GetDoVariable(doConstruct));
+    } else if (doConstruct.IsDoConcurrent()) {
+      auto &loopControl{doConstruct.GetLoopControl()};
+      if (loopControl) {
```
tskeith: Same as previous comment.

lib/semantics/semantics.cc
```diff
@@ -208,6 +209,70 @@ void SemanticsContext::PopConstruct() {
   constructStack_.pop_back();
 }
 
+void SemanticsContext::SayDoVarRedefine(
+    const Symbol &variable, const parser::CharBlock &location) {
```
tskeith: Other "Say" functions that take a `CharBlock` for the location of the message have that as the first argument, so it would be consistent.

lib/semantics/semantics.cc
```diff
@@ -208,6 +209,70 @@ void SemanticsContext::PopConstruct() {
...
+  const parser::CharBlock doLoc{GetDoVariableLocation(variable)};
+  Say(location, "Cannot redefine DO variable %s"_err_en_US, variable.name())
```
tskeith: Names from the program that are mentioned in error messages are enclosed in single quotes. It makes messages easier to read.

lib/semantics/semantics.cc
```diff
@@ -208,6 +209,70 @@ void SemanticsContext::PopConstruct() {
...
+  }
+}
+
+bool SemanticsContext::IsActiveDoVariable(const Symbol &variable) {
+  return activeDoVariables_.find(variable) != activeDoVariables_.end();
+}
+
+parser::CharBlock SemanticsContext::GetDoVariableLocation(
+    const Symbol &variable) {
+  if (IsActiveDoVariable(variable)) {
```
tskeith: When would this be called on something that isn't an active do variable? If this actually did return an empty CharBlock to `SayDoVarRedefine` we wouldn't get a good error message location.

lib/semantics/check-do.cc
psteinfeld: Thanks, Tim.  I'll change that.

lib/semantics/check-do.cc
psteinfeld: Yep.

lib/semantics/semantics.cc
psteinfeld: Thanks.  I'll add the quotes.

lib/semantics/semantics.cc
psteinfeld: GetDoVariableLocation() gets called in two situations.  One is to get the location for error messages.  The other is to deactivate a DO variable at the end of a DO loop.  In the case of nested DO loops that use the same variable, we only want to deactivate the loop variable for the outer loop when we exit it, not when we exit the inner loop.  So DeactivateDoVariable() calls GetDoVariableLocation() to make sure that its deactivating the correct variable.

I don't think GetDoVariableLocation() will be called for an error message when there's not an empty location for the DO variable.  But I can certainly add a check that will verify that.

lib/semantics/check-allocate.cc
```diff
@@ -620,6 +620,7 @@ bool AllocationCheckerHelper::RunCoarrayRelatedChecks(
         "Allocatable object must not be coindexed in ALLOCATE"_err_en_US);
     return false;
   }
+  context.CheckDoVarRedefine(name_);
```
jeanPerier: A minor point: this check is not related to coarrays yet is is done in `RunCoarrayRelatedChecks`. To avoid misleading future maintainers, I would move this check just before the following line instead: https://github.com/flang-compiler/f18/blob/867dd045de033bd092ae476a852e57f3b4bd3213/lib/semantics/check-allocate.cc#L542

psteinfeld: Thanks, Jean.  That's a better place.  I'll move it.

PR #862, opened 2019-12-09
Outline parsers for types (reduce f18 build memory requirements)

klausler:
Instantiate the per-type `std::optional<T> Parser<T>::Parse()` functions in a new *.cc file rather than as `inline` functions.


lib/parser/CMakeLists.txt
```diff
@@ -27,6 +27,7 @@ add_library(FortranParser
   source.cc
   token-sequence.cc
   tools.cc
+  type-parsers.cc
```
tskeith: I think you forgot to add this file.

lib/parser/type-parsers.cc
```diff
@@ -0,0 +1,37 @@
...
+      ->std::optional<resultType> { \
+    static constexpr auto parser{(pexpr)}; \
+    return parser.Parse(state); \
+  }
+
+#define TYPE_CONTEXT_PARSER(contextText, pexpr) \
+  TYPE_PARSER(CONTEXT_PARSER((contextText), (pexpr)))
+
+#include "grammar.h"
+#include "openmp-grammar.h"
```
DavidTruby: Can we split this up further so that the main grammar and openmp grammar are instantiated in different files? Or maybe even more fine-grained division than that?

klausler:
Please take another look.  All of the per-type parsers have been partitioned into a collection of C++ source files.

sscalpone:
I measured a dramatic reduction in peak and overall memory usage.  Only two files are now using over 2G, fold.cc and resolve-names.cc, compared to 15 in the previous iteration.  The overall memory reduction is about 30% less than before.

Compilation times are down too, although the sum of all compilation times is roughly the same. Fewer files require over 30 seconds.  fold.cc takes 122 seconds to compile on my system, about 2x the next nearest files resolve-names.cc and expression.cc.  These three are also the biggest memory users with this change).

klausler:
These changes only affect the parser.

klausler:
These changes drop the wall-clock time to build f18 from scratch on a 14-core server with 512GiB of RAM from 7m30s down to 6m using clang-8.

kiranchandramohan:
Thanks @klausler for working on this. The CI that @DavidTruby set up seems to pass with your changes. It is a good sign that the build time memory usage has improved.

In fact I can see that build time memory usage for most files have dropped below 2.5GB. The following files still show large memory usage though,
fold.cc: 6.5 GB
expression.cc: 3.1 GB
resolve-names.cc: 2.9 GB

As others have reported, I have also not observed an improvement in build time. Wall clock build time is around 25/26 minutes.

Note: For this experiment, I did a few builds on my macbook (3.1 GHz Core i7, OS mojave) running on 4 threads in parallel. I have my chrome browser open with many tabs.

tskeith:
> However, I would add that this _dramatically_ reduces incremental build times for most changes: for example if I change check-omp-semantics.cc on master my incremental build takes 2.5 minutes on my machine, whereas on this PR it takes 30 seconds.

I'm not seeing any improvement in that scenario (20 seconds in both cases for Release build). Are you doing Debug builds? This wouldn't affect compile time so it must be the link time is reduced. (I'm assuming you meant check-omp-structure.cc.)

DavidTruby:
I got those results with both debug and release builds, and can still replicate them when comparing this branch with master, using clang 9. I'm not sure why I would be seeing this specifically on AArch64 vs x86 but perhaps that has something to do with it?

tskeith:
I don't see a difference with clang 7.0 either for release builds.
I'm touching check-omp-structure.cc and building f18:
gcc 8.3: 17.0-17.5 compile time, 2.0-2.5 link time
clang 7.0: 25.0-25.5 compile time, 2.0-2.5 link time

This is on a Macbook Pro, 2.6Ghz Core i7, 32G memory

kiranchandramohan:
@klausler @sscalpone @DavidTruby : The CI failed for this commit and couple of other PRs. All of them seem to be related to fold.cc. So this file seems to have become the new bottleneck. Can similar techniques be used to split fold.cc to reduce peak memory consumption?

klausler:
> @klausler @sscalpone @DavidTruby : The CI failed for this commit and couple of other PRs. All of them seem to be related to fold.cc. So this file seems to have become the new bottleneck. Can similar techniques be used to split fold.cc to reduce peak memory consumption?

I haven't looked into build times on other files, so I don't know.

PR #863, opened 2019-12-09
Resolve calls to type-bound generic names

tskeith:
Extend `ResolveGeneric` to handle calls to procedure components by
passing in the data-ref that is used as the passed-object argument.

`AddPassArg` takes care of adding a placeholder for the passed object.
This is shared by the generic and non-generic cases of calls to
procedure components.


klausler:
Be advised, I'm in the progress of reverting my recent addition of `ActualArgument::PassedObject` to the representation.  It turned out to complicate functions that need access to the passed-object data reference but only had the actual arguments, not the procedure designator.  I'll update your changes if they go in before mine do.

PR #869, opened 2019-12-13
remove stale optimizer files

schweitzpgi:
Removing the old header files


vjayathirtha-nv:
Did you intend to also delete `include/fir/Transforms/MemToReg.h` ?
That's the only file that'll remain under `include/fir` after this change.

PR #871, opened 2019-12-16
Fix remaining failures in expression semantics

klausler:
Apply implicit typing to names in COMMON that appear in specification expressions

Extend semantic analysis of expressions to catch missing cases

Fix statement function semantics, add degree trig intrinsics

Add GetUltimate to rewrite of bare namelist

Expression analysis must now succeed (barring previous fatal error) and will emit an internal error message when it fails that includes a dump of the expression's parse tree.

### DIFF
--- a/lib/semantics/expression.cc
+++ b/lib/semantics/expression.cc
@@ -24,19 +24,14 @@
 #include "../evaluate/fold.h"
 #include "../evaluate/tools.h"
 #include "../parser/characters.h"
+#include "../parser/dump-parse-tree.h"
 #include "../parser/parse-tree-visitor.h"
 #include "../parser/parse-tree.h"
 #include <algorithm>
 #include <functional>
 #include <optional>
 #include <set>
-
-// #define CRASH_ON_FAILURE 1
-// #define DUMP_ON_FAILURE 1
-#if DUMP_ON_FAILURE
-#include "../parser/dump-parse-tree.h"
-#include <iostream>
-#endif
+#include <sstream>
 
 // Typedef for optional generic expressions (ubiquitous in this file)
 using MaybeExpr =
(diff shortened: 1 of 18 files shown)


lib/semantics/expression.cc
```diff
@@ -2279,10 +2273,12 @@ MaybeExpr ExpressionAnalyzer::ExprOrVariable(const PARSED &x) {
 }
 
 MaybeExpr ExpressionAnalyzer::Analyze(const parser::Expr &expr) {
+  auto save{GetContextualMessages().SetLocation(expr.source)};
```
tskeith: You renamed a similar `save` variable to `restorer` in `AnalyzeCall` so I think that would be a better name here too.

lib/semantics/symbol.h
```diff
@@ -78,12 +78,15 @@ class SubprogramDetails {
   const std::vector<Symbol *> &dummyArgs() const { return dummyArgs_; }
   void add_dummyArg(Symbol &symbol) { dummyArgs_.push_back(&symbol); }
   void add_alternateReturn() { dummyArgs_.push_back(nullptr); }
+  const MaybeExpr &stmtFunction() const { return stmtFunction_; }
+  void set_stmtFunction(SomeExpr &&expr) { stmtFunction_ = std::move(expr); }
```
tskeith: Is this used for anything yet?

klausler: No.  The CG will need it, though.  And I suppose that folding could use it for statement functions in which each dummy argument appears at most once in the body (or whose actual argument can be safely re-evaluated).

lib/semantics/expression.cc
klausler: Will do here & everywhere else in lib/.

lib/semantics/symbol.h
klausler: Also, the presence of the statement function's expression is the only distinguishing characteristic by which semantics could determine that a subprogram is a statement function, and exclude it from use in various circumstances (e.g., passing the statement function as an actual argument, using the statement function in a generic interface, using the statement function as the target of a pointer assignment, &c.) which I'm sure that we still need to do.

PR #872, opened 2019-12-16
Support for type-bound operators and assignment

tskeith:
Add `ArgumentAnalyzer::FindBoundOp` to look for an operator or
assignment definition in the type of each operand. Then `TryBoundOp`
checks if it is actually applicable.

Change ResolveGeneric to handle type-bound operators: the `adjustActuals`
function passed in handles the difference between these and normal
type-bound procedures. For operators, either operand may be the passed-
object argument. For procedures we know which one it is.

Extract `GetDerivedTypeSpec`, `GetBindingResolution`, and
`OkLogicalIntegerAssignment` into separate functions to simplify the
logic of the calling functions.

### DIFF
--- a/lib/semantics/expression.cc
+++ b/lib/semantics/expression.cc
@@ -913,16 +913,6 @@ static std::optional<Component> CreateComponent(
   return std::nullopt;
 }
 
-static const semantics::DerivedTypeSpec *GetDerivedTypeSpec(
-    const std::optional<DynamicType> &type) {
-  if (type && type->category() == TypeCategory::Derived) {
-    if (!type->IsUnlimitedPolymorphic()) {
-      return &type->GetDerivedTypeSpec();
-    }
-  }
-  return nullptr;
-}
-
 // Derived type component references and type parameter inquiries
 MaybeExpr ExpressionAnalyzer::Analyze(const parser::StructureComponent &sc) {
   MaybeExpr base{Analyze(sc.base)};
--- a/lib/semantics/expression.cc
+++ b/lib/semantics/expression.cc
@@ -178,9 +178,11 @@ public:
 private:
   MaybeExpr TryDefinedOp(
       std::vector<const char *>, parser::MessageFixedText &&);
+  MaybeExpr TryBoundOp(const Symbol &, int passIndex);
   std::optional<ActualArgument> AnalyzeExpr(const parser::Expr &);
   bool AreConformable() const;
-  Symbol *FindDefinedOp(const char *) const;
+  const Symbol *FindBoundOp(parser::CharBlock, int passIndex);
+  bool OkLogicalIntegerAssignment(TypeCategory lhs, TypeCategory rhs);
   std::optional<DynamicType> GetType(std::size_t) const;
   int GetRank(std::size_t) const;
   bool IsBOZLiteral(std::size_t i) const {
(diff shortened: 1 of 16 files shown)


lib/semantics/expression.cc
```diff
@@ -183,9 +183,11 @@ class ArgumentAnalyzer {
 private:
   MaybeExpr TryDefinedOp(
       std::vector<const char *>, parser::MessageFixedText &&);
+  MaybeExpr TryBoundOp(const Symbol &, int passIndex);
   std::optional<ActualArgument> AnalyzeExpr(const parser::Expr &);
   bool AreConformable() const;
-  Symbol *FindDefinedOp(const char *) const;
+  const Symbol *FindBoundOp(parser::CharBlock, int passIndex);
+  bool OkLogicalIntegerAssignment(TypeCategory, TypeCategory rhsCat);
```
klausler: Should both of the formal arguments have names?  Or neither?

lib/semantics/expression.cc
```diff
@@ -915,6 +918,16 @@ static std::optional<Component> CreateComponent(
   return std::nullopt;
 }
 
+static const semantics::DerivedTypeSpec *GetDerivedTypeSpec(
```
klausler: I think this function should be in lib/evaluate/type.{h,cc}, and some calls to `GetDerivedTypeSpec()` in the codebase that are protected by similar predicates could be replaced with calls to this utility.  Also, I think you should squash the predicates together, they don't have to be distinct `if` statements, but that's just my preference.

lib/semantics/expression.cc
```diff
@@ -1577,12 +1585,28 @@ static void AddPassArg(ActualArguments &actuals, Expr<SomeDerived> &&expr,
...
+    }
   }
   actuals.emplace(iter, std::move(passed));
 }
 
+// Return the compile-time resolution of a procedure binding, if possible.
+static const Symbol *GetBindingResolution(
+    const std::optional<DynamicType> &baseType, const Symbol &component) {
+  const auto *binding{component.detailsIf<semantics::ProcBindingDetails>()};
+  if (!binding) {
```
klausler: Consider structuring this differently; the test on `binding` could be part of an `if` statement that declares it.

lib/semantics/expression.cc
```diff
@@ -1680,9 +1700,9 @@ static bool CheckCompatibleArguments(
 
 // Resolve a call to a generic procedure with given actual arguments.
 // If it's a procedure component, base is the data-ref to the left of the '%'.
+// adjustActuals is call on procedure bindings to handle pass arg.
```
klausler: nit: grammar

lib/semantics/expression.cc
```diff
@@ -1691,7 +1711,9 @@ const Symbol *ExpressionAnalyzer::ResolveGeneric(const Symbol &symbol,
                 ProcedureDesignator{specific}, context_.intrinsics())}) {
       ActualArguments localActuals{actuals};
       if (specific.has<semantics::ProcBindingDetails>()) {
-        AddPassArg(localActuals, common::Clone(base.value()), specific);
+        if (!(adjustActuals.value())(specific, localActuals)) {
```
klausler: Those inner parentheses confused me -- are they required?

lib/semantics/expression.cc
tskeith: Good idea, this simplifies the code in a bunch of places.

PR #875, opened 2019-12-18
Select Case - C1145 case-expr shall be of type character, integer, or…

kiranktp:
C1145 : case-expr shall be of type character, integer, or logical

### DIFF
--- a/lib/semantics/check-select-stmt.cc
+++ b/lib/semantics/check-select-stmt.cc
@@ -1,0 +1,28 @@
+//===--- check-select-stmt.cc - Checker for select-case, select-rank, select-type ---------===
+//
+// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===--------------------------------------------------------------------------------------===
+
+#include "check-select-stmt.h"
+#include "tools.h"
+#include "../parser/message.h"
+#include "../parser/parse-tree.h"
+
+namespace Fortran::semantics {
+
+void SelectStmtChecker::Leave(const parser::SelectCaseStmt &selectcaseStmt) {
+  const auto &parsedExpr{std::get<parser::Scalar<parser::Expr>>(selectcaseStmt.t).thing};
+  if (const auto *expr{GetExpr(parsedExpr)}) {
+    if (auto type{expr->GetType()}) {
+      if (type->category() == TypeCategory::Real) {
+        context_.Say(parsedExpr.source, // C1145
+            "Select case expression shall be of type character, integer, or logical"_err_en_US);
+      }
+    }
+  }
+}
+
+}  // namespace Fortran::semantics
--- a/lib/semantics/CMakeLists.txt
+++ b/lib/semantics/CMakeLists.txt
@@ -31,6 +31,7 @@ add_library(FortranSemantics
   check-purity.cc
   check-return.cc
   check-stop.cc
+  check-select-stmt.cc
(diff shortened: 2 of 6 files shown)


lib/semantics/check-select-stmt.cc
```diff
@@ -0,0 +1,28 @@
...
+#include "../parser/message.h"
+#include "../parser/parse-tree.h"
+
+namespace Fortran::semantics {
+
+void SelectStmtChecker::Leave(const parser::SelectCaseStmt &selectcaseStmt) {
+  const auto &parsedExpr{std::get<parser::Scalar<parser::Expr>>(selectcaseStmt.t).thing};
+  if (const auto *expr{GetExpr(parsedExpr)}) {
+    if (auto type{expr->GetType()}) {
+      if (type->category() == TypeCategory::Real) {
```
tskeith: What if the expression is complex or a derived type? Also, what if it is type-less (i.e. a BOZ literal)?

lib/semantics/check-select-stmt.cc
```diff
@@ -0,0 +1,28 @@
...
+        context_.Say(parsedExpr.source, // C1145
+            "Select case expression shall be of type character, integer, or logical"_err_en_US);
```
tskeith: Other messages use "must be" rather than "shall be" in errors like this.

tskeith:
Please format your code with clang-format 7.

lib/semantics/check-select-stmt.cc
kiranktp: Thanks. Check must be for character, integer and logical type instead of other remaining types. I will incorporate this change.

lib/semantics/check-select-stmt.cc
kiranktp: I picked up the error message as is from standard's document. I will make it "must be".

sscalpone:
Hi @kiranktp.  Please add tests not only for more types, but for other conditions too, such as ranges, overlapping ranges, missing case values, and duplicate case values.

kiranktp:
> Hi @kiranktp.  Please add tests not only for more types, but for other conditions too, such as ranges, overlapping ranges, missing case values, and duplicate case values.

Hi, Those will be C1146, C1147, C1148, C1149 Constraints.  I will add them subsequently. I wanted this new checker to be in place first. Post that, others (sameeran/Anshu) from AMD will push in the changes for select Rank and select Type construct.

sscalpone:
@kiranktp  Would like us to merge this change now to ease that sharing with Sameeran and Anshu?

Sameeranjoshi:
Hi , using clang-format-7 from ubuntu packages, like below
clang-format-7  -style=llvm -dump-config > ~/.clang-format
clang-format-7  -style=llvm file.cc

1] I see many differences in the formatted code from the rest of the coding style followed in other files, is there anything more to change in .clang-format file?

Thanks

klausler:
> Hi , using clang-format-7 from ubuntu packages, like below
> clang-format-7 -style=llvm -dump-config > ~/.clang-format
> clang-format-7 -style=llvm file.cc
> 
> 1] I see many differences in the formatted code from the rest of the coding style followed in other files, is there anything more to change in .clang-format file?
> 
> Thanks

There is a `.clang-format` file in the topmost directory of the repository.

kiranktp:
> @kiranktp  Would like us to merge this change now to ease that sharing with Sameeran and Anshu?

I have inc

kiranktp:
> @kiranktp  Would like us to merge this change now to ease that sharing with Sameeran and Anshu?

Hi Steve, I have Incorporated review comments and added support for C1146, C1147, C1148, Request Tim and Klausler to review.

PR #879, opened 2019-12-19
Save CallStmt::typedCall

tskeith:
When `ExpressionAnalyzer::AnalyzeCall` processed a subroutine it was
always returning std::nullopt. Change it to return a `ProcedureRef`
wrapped in an `Expr` so that it can be saved in `CallStmt::typedCall`.

### DIFF
--- a/lib/semantics/expression.cc
+++ b/lib/semantics/expression.cc
@@ -1822,9 +1822,8 @@ MaybeExpr ExpressionAnalyzer::Analyze(
 }
 
 void ExpressionAnalyzer::Analyze(const parser::CallStmt &callStmt) {
-  auto expr{AnalyzeCall(callStmt.v, true)};
-  if (auto *procRef{UnwrapExpr<ProcedureRef>(expr)}) {
-    callStmt.typedCall.reset(new ProcedureRef{*procRef});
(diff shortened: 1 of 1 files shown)


lib/semantics/expression.cc
```diff
@@ -1822,9 +1822,8 @@ MaybeExpr ExpressionAnalyzer::Analyze(
 }
 
 void ExpressionAnalyzer::Analyze(const parser::CallStmt &callStmt) {
-  auto expr{AnalyzeCall(callStmt.v, true)};
-  if (auto *procRef{UnwrapExpr<ProcedureRef>(expr)}) {
-    callStmt.typedCall.reset(new ProcedureRef{*procRef});
+  if (auto expr{AnalyzeCall(callStmt.v, true)}) {
```
klausler: Are you sure that this part of the change is needed?  The test on `expr` should not be required; the `Unwrap` routines should handle vacant optional values.  And the result of the `Unwrap` should not be dereferenced unconditionally -- please use a `DEREF` at least.

tskeith: Right, it's better to check the result of UnwrapExpr than of AnalyzeCall. I've fixed it.

psteinfeld:
Thanks, Tim and Peter.  My test works now.

PR #881, opened 2019-12-19
Add initial CODE_OWNERS file

RichBarton-Arm:
First cut at adding a CODE_OWNERS file for F18 in preparation for it to land in the LLVM monorepo. It follows the same format as the LLVM equivalent file.

Reminder of code owner duties - https://llvm.org/docs/DeveloperPolicy.html#code-owners

A few comments:
- Are there any large areas that we think need a code owner not listed here?
- Is there a sensible code owner for "anything not covered by anyone else" ? @klausler ?
- I tried to look at who contributed the most commits to each subdirectory when picking a code owner. I know the commits only tell part of the story, so am open to alternative suggestions.

### DIFF
--- a/CODE_OWNERS.TXT
+++ b/CODE_OWNERS.TXT
@@ -9,10 +9,6 @@ beautification by scripts. The fields are: name (N), email (E), web-address
 (S) and (I) IRC handle. Each entry should contain at least the (N), (E) and
 (D) fields.
 
-N: Kiran Chandramohan
-E: kiran.chandramohan@arm.com
-D: OpenMP lowering
-
 N: Tim Keith
 E: tkeith@nvidia.com
 D: Semantic Analysis (lib/semantics)
@@ -20,7 +16,3 @@ D: Semantic Analysis (lib/semantics)
 N: Peter Klausler
 E: pklausler@nvidia.com
 D: Parser (lib/parser)
-
-N: Eric Schweitz
-E: eschweitz@nvidia.com
-D: FIR (lib/fir), Fortran lowering (lib/burnside)
--- a/CODE_OWNERS.TXT
+++ b/CODE_OWNERS.TXT
@@ -1,0 +1,26 @@
+This file is a list of the people responsible for ensuring that patches for a
+particular part of Flang are reviewed, either by themself or by someone else.
+They are also the gatekeepers for their part of Flang, with the final word on
+what goes in or not.
+
+The list is sorted by surname and formatted to allow easy grepping and
+beautification by scripts. The fields are: name (N), email (E), web-address
+(W), PGP key ID and fingerprint (P), description (D), snail-mail address
+(S) and (I) IRC handle. Each entry should contain at least the (N), (E) and
+(D) fields.
+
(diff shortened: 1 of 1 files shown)


kiranchandramohan:
Since there is no code currently for OpenMP Lowering should we have that entry?

RichBarton-Arm:
It's a fair point @kiranchandramohan - should we also have FIR and lowering as there is no code yet?
I think I am inclined to agree, so have pushed a v2.

klausler:
I don't see why anybody would want to have their name on this list.

RichBarton-Arm:
Hi @klausler and Happy New Year

Firstly, It is not my intention to sign anyone up for something that they don't want to do. If you don't want to be on the list then you won't be on the list.

I'd like to explain why it is important to have this list. Commit access in the LLVM project is open and reviews are distributed rather than being strictly hierarchical with a gatekeeper controlling commit access and reviewing each commit on the way in. Everyone is encouraged to give a review and all review comments are required to be at least acknowledged if not acted upon. 

The approach has the primary advantage that it allows the project to move quickly rather than having reviews backed up behind a busy gatekeeper. It also makes the cost of being a gatekeeper lower freeing up time for these experts to do other things. It also encourages non-expert review which, if done bravely, we hope will lead to clearer code with lower barrier to entry.

The disadvantage of this approach is that changes not reviewed by the absolute expert can introduce problems that may have been spotted by the expert in review before commit. The code owner role is designed to be one of the balances against the totally open season for changes (alongside adding high-quality regression tests alongside your commits and post-commit review with reverting changes when required.)

Unless we have a concrete proposal against, Flang will have the same access rights as the rest of the LLVM project once it lands so it is important we start thinking about how to keep the review quality high.

There are many ways in which the Code Owner role is fulfilled in LLVM presently to proactively reviewing each patch in pre-commit to relying mostly on post-commit and weighing in when needed on pre-commit and RFCs. That said the essence of the role seems to me to be: a) understand in depth how that part of the codebase works and b) do lots of reviews for patches on said part of the codebase. This seems to be a good match for your current activities in the Flang github already so nominating you as a code owner seems to be a simple formalisation of what you already do so well.

Anyhow, I don't want to put you down if you don't want to be down. Have I convinced you to change your mind and stay on the list?

klausler:
A post-commit code review policy trades off quality to produce a perception of faster development.  That seems contrary to the best interests of our users.  No, thanks.

RichBarton-Arm:
Hi @klausler 
I have posted a patch that removes you from the list.

You share a reasonable counter argument against post-commit review and for pre-commit. The llvm project also prefers pre-commit review, although post-commit is possible in certain circumstances and by certain people, for example the code owner. The exact balance of how much pre-commit vs post-commit review is used is up to the communities that work on each project and the code owners heavily influence that decision.

sscalpone:
Hi @RichBarton-Arm  The NVIDIA team asked that I be a code-reviewer for flang/*.  Perhaps we can get @schweitzpgi to sign up for lib/fir et al.

RichBarton-Arm:
We eventually did this under #1066 instead

PR #883, opened 2019-12-20
Misc fixes from SPEC

klausler:
This PR rolls up some small patches that help build SPEC CPU 2008 with semantics enabled and fixed-point unparsed output.  There's more to come, but some of the codes are now clean.

### DIFF
--- a/lib/evaluate/type.cc
+++ b/lib/evaluate/type.cc
@@ -54,28 +54,38 @@ static bool IsDescriptor(const ProcEntityDetails &details) {
   return details.HasExplicitInterface();
 }
 
-bool IsDescriptor(const Symbol &symbol0) {
-  const Symbol &symbol{evaluate::ResolveAssociations(symbol0)};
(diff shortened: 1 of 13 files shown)


lib/evaluate/type.cc
```diff
@@ -54,28 +54,37 @@ static bool IsDescriptor(const ProcEntityDetails &details) {
...
+            }
+            return false;
+          },
+          [](const SubprogramDetails &d) {
+            return d.isFunction() && IsDescriptor(d.result());
+          },
+          [](const UseDetails &d) { return IsDescriptor(d.symbol()); },
+          [](const auto &) { return false; },
+      },
+      symbol.details());
```
tskeith: Do we need a case for `HostAssocDetails`?

klausler: > Do we need a case for `HostAssocDetails`?

Seems like a good idea.

PR #884, opened 2019-12-20
Convert INTEGER scalar actual argument expressions to the kinds of dummies

klausler:
When passing the result of `SIZE()` to a dummy argument that is default `INTEGER`, automatically convert the value from the larger kind of `INTEGER` that we use in order to support large arrays.

(This PR implements a more general actual argument expression conversion, and it emits a warning when truncation may occur.  It does not apply to actual arguments that are variables, nor does it apply to arrays.)

### DIFF
--- a/lib/semantics/check-call.cc
+++ b/lib/semantics/check-call.cc
@@ -101,6 +101,33 @@ static void PadShortCharacterActual(evaluate::Expr<evaluate::SomeType> &actual,
   }
 }
 
+// Automatic conversion of different-kind INTEGER scalar actual
+// argument expressions (not variables) to INTEGER scalar dummies.
+// We return nonstandard INTEGER(8) results from intrinsic functions
+// like SIZE() by default in order to facilitate the use of large
+// arrays.  Emit a warning when downconverting.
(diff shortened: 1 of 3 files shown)


lib/semantics/check-call.cc
```diff
@@ -101,6 +101,33 @@ static void PadShortCharacterActual(evaluate::Expr<evaluate::SomeType> &actual,
...
+      GetRank(dummyType.shape()) == 0 && GetRank(actualType.shape()) == 0 &&
+      !evaluate::IsVariable(actual)) {
+    auto converted{
+        evaluate::ConvertToType(dummyType.type(), std::move(actual))};
+    CHECK(converted);
+    actual = std::move(*converted);
+    if (dummyType.type().kind() < actualType.type().kind()) {
+      messages.Say(
+          "Actual argument scalar expression of type Integer(%d) was converted to smaller dummy argument type Integer(%d)"_en_US,
+          actualType.type().kind(), dummyType.type().kind());
```
tskeith: For consistency with other messages this should use "INTEGER(%d)".

klausler: Done.

PR #885, opened 2019-12-20
Two small fixes

klausler:
As Jean pointed out to me (merci), `REAL(z)` is defined for complex argument `z` to return by default the same kind of real as the argument, not default real.

Also, fix up expression formatting some more.  Sorry about the churn here, it's harder to get right than I thought it would be.  Note that parentheses introduced into the formatted output solely to honor operator precedence rules are distinguished from explicit parentheses by the addition of space characters around the enclosed subexpression.


klausler:
Actually, the extra spaces mentioned in the PR header are not preserved when unparsed output is reparsed and unparsed again by f18, so I'll have to remove them to prevent spurious mismatches.

PR #889, opened 2019-12-23
Select Case - Added support for C1145, C1146, C1147, C1148, C1149

kiranktp:
Select Case - Added support for C1144, C1145, C1146, C1147, C1148, C1149

C1144  (R1140) If the select-case-stmt of a case-construct specifies a case-construct-name, the corresponding end-select-stmt shall specify the same case-construct-name. If the select-case-stmt of a case-construct does not specify a case-construct-name, the corresponding end-select-stmt shall not specify a case-construct-name. If a case-stmt specifies a case-construct-name, the corresponding select-case-stmt shall specify the same case-construct-name
C1145 case-expr shall be of type character, integer, or logical.
C1146  (R1140) No more than one of the selectors of one of the CASE statements shall be DEFAULT.
C1147  (R1140) For a given case-construct, each case-value shall be of the same type as case-expr. For character type, the kind type parameters shall be the same; character length differences are allowed.
C1148  (R1140) A case-value-range using a colon shall not be used if case-expr is of type logical.
C1149  (R1140) For a given case-construct, there shall be no possible value of the case-expr that matches more than one case-value-range


lib/semantics/check-select-stmt.cc
```diff
@@ -0,0 +1,103 @@
...
+      }
+    } else {
+      const auto &casevaluerangelist{
+          std::get<std::list<parser::CaseValueRange>>(caseselector.u)};
+      for (const auto &casevalues : casevaluerangelist) {
+        if (std::holds_alternative<parser::Scalar<parser::ConstantExpr>>(
+                casevalues.u)) {
+          const auto &constcase{
+              std::get<parser::Scalar<parser::ConstantExpr>>(casevalues.u)
+                  .thing};
```
tskeith: You can check if casevalues.u holds a ConstantExpr and get it all in one step like this:
```
if (const auto *constcase{std::get_if<parser::Scalar<parser::ConstantExpr>>(&casevalues.u)}) {
```

lib/semantics/check-select-stmt.cc
```diff
@@ -0,0 +1,103 @@
...
+    const parser::CaseConstruct &selectcaseConstruct) {
+  const auto &selectcaseStmt{
+      std::get<parser::Statement<parser::SelectCaseStmt>>(
+          selectcaseConstruct.t)};
+  const auto &parsedExpr{
+      std::get<parser::Scalar<parser::Expr>>(selectcaseStmt.statement.t).thing};
+  TypeCategory selectcasestmttype;
+  if (const auto *expr{GetExpr(parsedExpr)}) {
+    if (auto type{expr->GetType()}) {
+      selectcasestmttype = type->category();
```
tskeith: selectcasestmttype doesn't need to be declared until this point, where you can declare and initialize it.

lib/semantics/check-select-stmt.cc
```diff
@@ -0,0 +1,103 @@
...
+      if ((selectcasestmttype != TypeCategory::Integer) &&
+          (selectcasestmttype != TypeCategory::Character) &&
+          (selectcasestmttype != TypeCategory::Logical)) {
+        // C1145 case-expr shall be of type character, integer, or logical.
+        context_.Say(parsedExpr.source,
+            "Select case expression must be of type character, integer, or logical"_err_en_US);
+      }
+    }
+  }
```
tskeith: What happens if selectcasestmttype has not been set in the if statement?

lib/semantics/check-select-stmt.cc
```diff
@@ -0,0 +1,103 @@
...
+          }
+        }
+      }
+    }
+  }
+}
+
+void SelectStmtChecker::CheckSelectCaseType(const TypeCategory &expectedtype,
+    const Fortran::parser::ConstantExpr &constcase,
+    const Fortran::parser::CharBlock &src) {
```
tskeith: `Fortran::` is unnecessary here.

lib/semantics/semantics.cc
```diff
@@ -118,7 +119,8 @@ using StatementSemanticsPass1 = ExprChecker;
 using StatementSemanticsPass2 = SemanticsVisitor<AllocateChecker,
     ArithmeticIfStmtChecker, AssignmentChecker, CoarrayChecker,
     DeallocateChecker, DoChecker, IfStmtChecker, IoChecker, NullifyChecker,
-    OmpStructureChecker, PurityChecker, ReturnStmtChecker, StopChecker>;
+    OmpStructureChecker, PurityChecker, ReturnStmtChecker, StopChecker,
+    SelectStmtChecker>;
```
tskeith: This list of checkers was in alphabetical order.

lib/semantics/check-select-stmt.cc
```diff
@@ -0,0 +1,103 @@
...
+namespace Fortran::semantics {
+
+void SelectStmtChecker::Leave(
+    const parser::CaseConstruct &selectcaseConstruct) {
+  const auto &selectcaseStmt{
+      std::get<parser::Statement<parser::SelectCaseStmt>>(
+          selectcaseConstruct.t)};
+  const auto &parsedExpr{
+      std::get<parser::Scalar<parser::Expr>>(selectcaseStmt.statement.t).thing};
+  TypeCategory selectcasestmttype;
```
tskeith: The style used in the rest of the project is to capitalize words after the first in local variable names. So `selectCaseStmtType` rather than `selectcasestmttype`, etc.

lib/semantics/check-select-stmt.cc
kiranktp: I did try once to use get_if, as I remember I was getting an error saying this version of get_if is unavailable. I do not remember if had done some mistake while using that. I will try to check and replace.

lib/semantics/check-select-stmt.cc
kiranktp: I had the same question. But I didn't think of a scenario where the select statement is not having any type. It will be either intrinsic type or derived type. In both cases the variable will be set to some type. If you are aware of any scenario, please do let me know. I will check and do the necessary changes.

lib/semantics/semantics.cc
kiranktp: Oops, I didn't pay attention to the alphabetical order. Thanks for pointing it out.

lib/semantics/check-select-stmt.cc
kiranktp: Will do this change..

lib/semantics/check-select-stmt.cc
tskeith: If there is an error in the expression, GetExpr or GetType might not return a value.
Also, if the expression is a BOZLiteral it doesn't have a type. I'm not sure what should happen in that case -- maybe it should be an error.

lib/semantics/check-select-stmt.cc
klausler: Also for the function argument.

lib/semantics/check-select-stmt.cc
klausler: There are other expressions that don't have types, like the name of a subroutine.  So you need to ensure that `selectCaseStmtType` is not left uninitialized.  You could make `selectCaseStmtType` a `std::optional<>` to make it clear when it is set and when it is not.

lib/semantics/check-select-stmt.cc
```diff
@@ -0,0 +1,103 @@
...
+        // C1145 case-expr shall be of type character, integer, or logical.
+        context_.Say(parsedExpr.source,
+            "Select case expression must be of type character, integer, or logical"_err_en_US);
+      }
+    }
+  }
+
+  const auto &caselist{
+      std::get<std::list<parser::CaseConstruct::Case>>(selectcaseConstruct.t)};
+  bool defaultcasefound = false;
```
klausler: We conventionally use braced initialization for safety.

lib/semantics/check-select-stmt.cc
```diff
@@ -0,0 +1,103 @@
...
+    const auto &caseselector{
+        std::get<parser::CaseSelector>(casestmt.statement.t)};
+    if (std::holds_alternative<parser::Default>(caseselector.u)) {
+      if (!defaultcasefound) {
+        defaultcasefound = true;
+      } else {
+        // C1146 (R1140) No more than one of the selectors of one of the CASE
+        //       statements shall be DEFAULT.
+        context_.Say(casestmt.source,
+            "Not more than one of the selectors of case statements must be default"_err_en_US);
```
klausler: When phrased in this way, the word "may" is a better choice than "must".

lib/semantics/check-select-stmt.cc
```diff
@@ -0,0 +1,103 @@
...
+      for (const auto &casevalues : casevaluerangelist) {
+        if (std::holds_alternative<parser::Scalar<parser::ConstantExpr>>(
+                casevalues.u)) {
+          const auto &constcase{
+              std::get<parser::Scalar<parser::ConstantExpr>>(casevalues.u)
+                  .thing};
+          CheckSelectCaseType(selectcasestmttype, constcase, casestmt.source);
+        } else {
+          if (selectcasestmttype == TypeCategory::Logical) {
+            // C1148  (R1140) A case-value-range using a colon shall not be used
```
klausler: We typically don't quote the entire text of a constraint as a comment -- it should be clear from the text of the error message what you're doing.  Just put the `C1148` in a comment at the end of the `if` line or somewhere else nearby so that a person may search the sources for that constraint identifier and find the code quickly.

lib/semantics/check-select-stmt.cc
```diff
@@ -0,0 +1,103 @@
+//= check-select-stmt.cc - Checker for select-case, select-rank, select-type ==
```
klausler: The code below seems to only be checking `SELECT CASE`, not `SELECT RANK` or `SELECT TYPE`.  Aren't those two already done in name resolution?  If this code is going to check `SELECT RANK` or `SELECT TYPE`, please add a `TODO` comment to that effect; otherwise, don't mention them here.

lib/semantics/check-select-stmt.h
```diff
@@ -0,0 +1,27 @@
+//= check-select-stmt.h - Checker for select-case, select-rank, select-type ==
```
klausler: My comment on the first line of the `.cc` file also applies here.

test/semantics/select-case.f90
```diff
@@ -0,0 +1,101 @@
...
+         print*, "Well done" 
+      case default
+         print*, "Invalid grade" 
+   end select
+
+   select case (grade3)
+      case (.true.) 
+       name = 'true'
+      case (.false.)
+        name = 'false'
```
klausler: I think that Fortran weirdly prohibits a case range for `LOGICAL`.

kiranktp:
I didn't face any merge conflict on my branch when I did a rebase.
Here it is mentioning as a merge conflict, I do not have permission to resolve the conflict here. 
How to fix this?

lib/semantics/check-select-stmt.cc
kiranktp: Thanks for the suggestion Klausler, Now I have made selectCaseStmtType a std::optional<>.

lib/semantics/check-select-stmt.cc
kiranktp: Now i have kept only the constraint number in the comment, as the warning/error gives the rest of the message.

lib/semantics/check-select-stmt.cc
kiranktp: I, Sameeran and Anchu are working on SELECT CASE, SELECT RANK AND SELECT TYPE.
In my code i have added TODO for SELECT RANK AND SELECT TYPE.
We will update this as and when we add support for these constructs.

kiranktp:
> I didn't face any merge conflict on my branch when I did a rebase.
> Here it is mentioning as a merge conflict, I do not have permission to resolve the conflict here.
> How to fix this?

Apologies for the blunder I did last evening in commit. I have corrected that and pushed the changes for review. Please do reivew and let me know if any thing more needs to be changed.

lib/semantics/check-select-stmt.h
```diff
@@ -0,0 +1,27 @@
...
+
+namespace Fortran::semantics {
+class SelectStmtChecker : public virtual BaseChecker {
+public:
+  SelectStmtChecker(SemanticsContext &context) : context_{context} {}
+  void Leave(const parser::CaseConstruct &);
+
+private:
+  SemanticsContext &context_;
+  bool isValidSelectCaseType(const TypeCategory &, const parser::ConstantExpr &,
```
tskeith: The function name should start with a capital letter.

lib/semantics/check-select-stmt.cc
```diff
@@ -0,0 +1,239 @@
...
+void SelectStmtChecker::Leave(
+    const parser::CaseConstruct &selectCaseConstruct) {
+  const auto &selectCaseStmt{
+      std::get<parser::Statement<parser::SelectCaseStmt>>(
+          selectCaseConstruct.t)};
+  const auto &parsedExpr{
+      std::get<parser::Scalar<parser::Expr>>(selectCaseStmt.statement.t).thing};
+  std::optional<TypeCategory> selectCaseStmtType;
+  bool validCaseStmtType{false};
+  if (const auto *expr{GetExpr(parsedExpr)}) {
```
tskeith: If there is an error in the expression (i.e. GetExpr  returns null), you shouldn't also report an error about the type of the expression. It looks like that's what would happen.

lib/semantics/check-select-stmt.cc
```diff
@@ -0,0 +1,239 @@
...
+  const auto &caseList{
+      std::get<std::list<parser::CaseConstruct::Case>>(selectCaseConstruct.t)};
+  bool defaultCaseFound{false};
+  using var_t = std::variant<std::int64_t, bool, std::string>;
+  using IntType = std::vector<
+      std::pair<std::optional<std::int64_t>, std::optional<std::int64_t>>>;
+  using StringType = std::vector<
+      std::pair<std::optional<std::string>, std::optional<std::string>>>;
+  std::set<var_t> caseVals;
+  std::pair<std::set<var_t>::iterator, bool> ret;
```
tskeith: You should declare an initialize `ret` where it is used. And give it a name that means something.

lib/semantics/check-select-stmt.cc
```diff
@@ -0,0 +1,239 @@
...
+    context_.Say(parsedExpr.source,
+        "SELECT CASE expression must be of type CHARACTER, INTERGER, OR LOGICAL"_err_en_US);
+  } else {
+    validCaseStmtType = true;
+  }
+
+  const auto &caseList{
+      std::get<std::list<parser::CaseConstruct::Case>>(selectCaseConstruct.t)};
+  bool defaultCaseFound{false};
+  using var_t = std::variant<std::int64_t, bool, std::string>;
```
tskeith: Don't use `_t` for type names. Can you give this a better name? What does it represent?

lib/semantics/check-select-stmt.cc
tskeith: Maybe just return in that case?

lib/semantics/tools.h
```diff
@@ -265,6 +265,22 @@ template<typename T> std::optional<std::int64_t> GetIntValue(const T &x) {
...
 
+template<typename T> std::optional<bool> GetBoolValue(const T &x) {
+  if (const auto *expr{GetExpr(x)}) {
+    return evaluate::ToBool(*expr);
+  } else {
+    return std::nullopt;
+  }
+}
+
+template<typename T> std::optional<std::string> GetString(const T &x) {
```
tskeith: For consistency with the two preceding function this should be GetStringValue.

test/semantics/select-case.f90
```diff
@@ -0,0 +1,167 @@
+!===--- select-case.f90 - Test select case constraints -------------------===
+!
+! Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+! See https://llvm.org/LICENSE.txt for license information.
+! SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+!
+!===------------------------------------------------------------------------===
```
tskeith: Apparently we don't need this header on test files.

lib/semantics/check-select-stmt.cc
```diff
@@ -0,0 +1,239 @@
...
+    const auto &caseStmt{
+        std::get<parser::Statement<parser::CaseStmt>>(cases.t)};
+    const auto &caseSelector{
+        std::get<parser::CaseSelector>(caseStmt.statement.t)};
+    if (std::holds_alternative<parser::Default>(caseSelector.u)) {
+      if (!defaultCaseFound) {
+        defaultCaseFound = true;
+      } else {  // C1146 (R1140)
+        context_.Say(caseStmt.source,
+            "Not more than one of the selectors of case statements may be default"_err_en_US);
```
tskeith: I  suggest replacing "case statements" with "SELECT CASE" or "SELECT CASE statement" like in other errors.

test/semantics/select-case.f90
```diff
@@ -0,0 +1,167 @@
...
+      case (40)
+         print*, "Excellent!"
+      case (90)
+         print*, "Excellent!"
+      case (91:99)
+         print*, "Very good!"
+      !ERROR: SELECT CASE statement value must not match more than one case-value-range
+      case (81:90)
+         print*, "Very good!"
+      !ERROR: SELECT CASE statement value must not match more than one case-value-range
```
tskeith: This error would be more helpful if it could specify at least one value that overlaps another case and had an attached message that referred to that other case.

lib/semantics/check-select-stmt.cc
```diff
@@ -0,0 +1,239 @@
...
+      }
+    }
+  }
+}
+
+bool SelectStmtChecker::isValidSelectCaseType(const TypeCategory &expectedType,
+    const parser::ConstantExpr &constCase, const parser::CharBlock &src) {
+  if (const auto *caseval{GetExpr(constCase)}) {
+    if (auto type{caseval->GetType()}) {
+      if (type->category() != expectedType) {  // C1147 (R1140)
```
tskeith: C1147 also says that character kinds must match. Are you checking that?

lib/semantics/check-select-stmt.cc
kiranktp: > Maybe just return in that case?

Yes, I would just modify to return if GetExpr returns null instead of issuing an error for the type of the expression.

lib/semantics/check-select-stmt.cc
```diff
@@ -0,0 +1,239 @@
...
+  } else {
+    validCaseStmtType = true;
+  }
+
+  const auto &caseList{
+      std::get<std::list<parser::CaseConstruct::Case>>(selectCaseConstruct.t)};
+  bool defaultCaseFound{false};
+  using var_t = std::variant<std::int64_t, bool, std::string>;
+  using IntType = std::vector<
+      std::pair<std::optional<std::int64_t>, std::optional<std::int64_t>>>;
```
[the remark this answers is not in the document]
kiranktp: I will move these types and operations to a new class.

lib/semantics/tools.h
kiranktp: Thanks for the comments! I will incoprate these.

test/semantics/select-case.f90
kiranktp: I had thought of this earlier. gfortran does this way.
Is there any way, where two different souce location can be pointed out for one diagnostic?

test/semantics/select-case.f90
kiranktp: I will remove the header.

test/semantics/select-case.f90
tskeith: You can attach an informational message with a different source location to the error message. See `SayAlreadyDeclared` in `resolve-names.cc` for an example.

lib/semantics/check-select-stmt.cc
kiranktp: I will give it meaningful name to these.

lib/semantics/check-select-stmt.cc
kiranktp: I have handled character kinds as well.

lib/semantics/check-select-stmt.cpp
```diff
@@ -0,0 +1,341 @@
...
+      ReportOverlapCaseValError(
+          std::get<2>(*selectCaseStringRangeValsIter), source);
+    }
+  }
+
+private:
+  using selectCaseStmtTypes =
+      std::vector<std::pair<std::variant<std::int64_t, bool, std::string>,
+          parser::CharBlock>>;
+  selectCaseStmtTypes selectCaseVals;
```
tskeith: Data members of classes should have names that end with "_".

lib/semantics/check-select-stmt.cpp
```diff
@@ -0,0 +1,341 @@
...
+  selectCaseStmtTypes::iterator selectCaseValsIter;
```
tskeith: What is this for?

lib/semantics/check-select-stmt.cpp
```diff
@@ -0,0 +1,341 @@
...
+public:
+  SelectCaseHelper(parser::Messages &messages) : messages_{messages} {}
+
+  void ReportOverlapCaseValError(
+      parser::CharBlock &source1, parser::CharBlock &source2) {
+    // C1149
+    messages_
+        .Say(source2,
+            "SELECT CASE statement value at this location overlaps with below location"_err_en_US)
+        .Attach(source1, ""_err_en_US);
```
tskeith: I don't think an empty message is a good idea. Why not specify the value that occurs in both CASE statements?

lib/semantics/check-select-stmt.cpp
```diff
@@ -0,0 +1,341 @@
...
+    }
+  }
+  if (!selectCaseStmtType) {
+    return;
+  }
+  if ((selectCaseStmtType.value() != TypeCategory::Integer) &&
+      (selectCaseStmtType.value() != TypeCategory::Character) &&
+      (selectCaseStmtType.value() != TypeCategory::Logical)) {  // C1145
+    context_.Say(parsedExpr.source,
+        "SELECT CASE expression must be of type CHARACTER, INTERGER, OR LOGICAL"_err_en_US);
```
tskeith: It would be better just to return here. I don't think there is any point in checking the cases when the expression is the wrong type.
Also, "INTERGER" -> "INTEGER".

lib/semantics/check-select-stmt.cpp
```diff
@@ -0,0 +1,341 @@
...
+          selectCaseConstruct.t)};
+  const auto &parsedExpr{
+      std::get<parser::Scalar<parser::Expr>>(selectCaseStmt.statement.t).thing};
+  std::optional<TypeCategory> selectCaseStmtType;
+  int selectCaseStmtKind;
+  bool validCaseStmtType{false};
+  if (const auto *expr{GetExpr(parsedExpr)}) {
+    if (auto type{expr->GetType()}) {
+      selectCaseStmtType = type->category();
+      selectCaseStmtKind = type->kind();
```
tskeith: Why not just save the type rather than selectCaseStmtType and SelectCaseStmtKind? That lets the reader know that the category and kind are tied together. `type->category()` is clearer and shorter than `selectCaseStmtType.value()`.

lib/semantics/check-select-stmt.cpp
```diff
@@ -0,0 +1,341 @@
...
+
+private:
+  using selectCaseStmtTypes =
+      std::vector<std::pair<std::variant<std::int64_t, bool, std::string>,
+          parser::CharBlock>>;
+  selectCaseStmtTypes selectCaseVals;
+  selectCaseStmtTypes::iterator selectCaseValsIter;
+  using selectCaseIntRangeType =
+      std::vector<std::tuple<std::optional<std::int64_t>,
+          std::optional<std::int64_t>, parser::CharBlock>>;
```
tskeith: If you defined a struct for the members of the vector the data members would have meaningful names and you wouldn't have to use std::get<0>, etc.

lib/semantics/check-select-stmt.cpp
tskeith: Instead of this vector, how about using three maps, keyed by int64_t, bool, and std::string?

kiranktp: Thanks. I have modified the data members now.

lib/semantics/check-select-stmt.cpp
kiranktp: Since case expression value can be either a case value  or case value range.
To store case value of one of the three allowed types (along with the location info), Either we can have separate a vector of pairs (value, location) for each type, or vector of pairs of variant of 3 types and the source location. So I choose the later option.
Any better approach to save these?

lib/semantics/check-select-stmt.cpp
kiranktp: This is how the error looks:
test4.f90:18:7: error: SELECT CASE statement value at this location overlaps with below location
        case (81:90)
        ^^^^^^^^^^^^
test4.f90:13:7:
        case (90)
        ^^^^^^^^^
The location and the values are already part of this error message. So I left it blank.

lib/semantics/check-select-stmt.cpp
kiranktp: We continue to check for duplicate default case only. We do not consider any other cases.
Here is the condition which checks this: 
242     } else if (validCaseStmtType) {
I didn't notice spell mistake. I will correct them.

lib/semantics/check-select-stmt.cpp
kiranktp: Thank you. This is a better approach. I will modify to keep the type instead of category and kind.

lib/semantics/check-select-stmt.cpp
kiranktp: I tried to avoid creating multiple containers for different types. so used variant.

lib/semantics/check-select-stmt.cpp
kiranktp: Since only first two members of tuples are being used frequently, I thought it is a overhead to create a struct and handle instead of using tuple. Should I modify them and use struct instead?

lib/semantics/check-select-stmt.cpp
klausler: Do not emit a blank message.  Add some descriptive text.

lib/semantics/check-select-stmt.cpp
klausler: All of the cases in a given `SELECT CASE` construct will use the same type.

lib/semantics/check-select-stmt.cpp
klausler: Use a struct or class.

lib/semantics/check-select-stmt.cpp
kiranktp: Thank you. I will add similar message.

lib/semantics/check-select-stmt.cpp
kiranktp: Ok. I will replace tuple with a struct.

lib/semantics/check-select-stmt.cpp
kiranktp: As @tskeith suggested, i will use three maps, keyed with these three types.

kiranktp:
Hi Tskeith, Klausler, 
I have incorporated all review comments, Could you please review the new changes?

include/flang/evaluate/fold.h
```diff
@@ -96,5 +96,56 @@ std::optional<std::int64_t> ToInt64(const std::optional<A> &x) {
...
+
+template<typename A> std::optional<bool> ToBool(const std::optional<A> &x) {
+  if (x) {
+    return ToBool(*x);
+  } else {
+    return std::nullopt;
+  }
+}
+
+// When an expression is a constant character, ToString() extracts its value.
```
klausler: How does this template function make sense for `KIND` values other than 1?  Information can be lost if other kinds of strings are converted to the default kind.

lib/evaluate/fold-integer.cpp
```diff
@@ -630,6 +630,32 @@ std::optional<std::int64_t> ToInt64(const Expr<SomeType> &expr) {
   }
 }
 
+std::optional<std::string> ToString(const Expr<SomeCharacter> &expr) {
```
klausler: None of these functions have anything to do with integers; `fold-integer.cpp` seems like a confusing home for them.

lib/semantics/check-select-stmt.cpp
```diff
@@ -0,0 +1,316 @@
...
+
+namespace Fortran::semantics {
+
+class SelectCaseHelper {
+
+public:
+  SelectCaseHelper(parser::Messages &messages) : messages_{messages} {}
+
+  void ReportOverlapCaseValError(
+      parser::CharBlock &source1, parser::CharBlock &source2) {
```
klausler: These arguments don't need to be references, but if they are, they should be `const`.  But simple value arguments are best for `CharBlock`.

lib/semantics/check-select-stmt.cpp
```diff
@@ -0,0 +1,316 @@
...
+    // C1149
+    messages_
+        .Say(source2,
+            "SELECT CASE statement value at this location overlaps with the below location"_err_en_US)
+        .Attach(source1,
+            "SELECT CASE statement value overlaps with the above location"_err_en_US);
+  }
+
+  void Insert(std::int64_t intVal, parser::CharBlock source) {
+    auto selectCaseIntRangeValsIter =
```
klausler: We conventionally use modern `{`braced`}` initialization for variables for safety, rather than the older `=` style.

lib/semantics/check-select-stmt.cpp
```diff
@@ -0,0 +1,316 @@
...
+        std::find_if(selectCaseIntRangeVals_.begin(),
+            selectCaseIntRangeVals_.end(), [&intVal](const auto &mem) {
+              return ((!mem.fromVal_ || (intVal >= mem.fromVal_.value())) &&
+                  (!mem.toVal_ || (intVal <= mem.toVal_.value())));
+            });
+    if (selectCaseIntRangeValsIter != selectCaseIntRangeVals_.end()) {
+      ReportOverlapCaseValError(selectCaseIntRangeValsIter->source_, source);
+    }
+    auto selectCaseIntValsInsertStatus =
```
klausler: Here, too.

lib/semantics/check-select-stmt.cpp
```diff
@@ -0,0 +1,316 @@
...
+        selectCaseIntVals_.insert(std::make_pair(intVal, source));
+    if (selectCaseIntValsInsertStatus.second == false) {
```
klausler: The logical negation operator `!` would be better idiomatic C/C++ usage here rather than an equality test with `false`.

lib/semantics/check-select-stmt.cpp
```diff
@@ -0,0 +1,316 @@
...
+      ReportOverlapCaseValError(
+          selectCaseIntValsInsertStatus.first->second, source);
+    }
+  }
+  void Insert(std::string strVal, parser::CharBlock source) {
```
klausler: All of my comments from the previous function apply to this one, and probably the one after.  Since they look like they're nearly identical, I think that you could abstract them into a template and then you'd have to fix the problems only once.

lib/semantics/check-select-stmt.cpp
```diff
@@ -0,0 +1,316 @@
...
+  }
+  void Insert(bool logicalVal, parser::CharBlock source) {
+    auto selectCaseLogicalValsInsertStatus =
+        selectCaseLogicalVals_.insert(std::make_pair(logicalVal, source));
+    if (selectCaseLogicalValsInsertStatus.second == false) {
+      ReportOverlapCaseValError(
+          selectCaseLogicalValsInsertStatus.first->second, source);
+    }
+  }
+  void Insert(std::optional<std::int64_t> lowerVal,
```
klausler: Abstract these into a template rather than coping and pasting the code, please.

lib/semantics/check-select-stmt.cpp
```diff
@@ -0,0 +1,316 @@
...
+        parser::CharBlock source)
+      : fromVal_(fromVal), toVal_(toVal), source_(source) {}
+    std::optional<T> fromVal_;
+    std::optional<T> toVal_;
+    parser::CharBlock source_;
+  };
+
+  std::map<bool, parser::CharBlock> selectCaseLogicalVals_;
+  std::map<std::int64_t, parser::CharBlock> selectCaseIntVals_;
+  std::map<std::string, parser::CharBlock> selectCaseStringVals_;
```
klausler: There are two other supported kinds of `CHARACTER` that can be used in `SELECT CASE`.

test/semantics/select-case.f90
```diff
@@ -0,0 +1,172 @@
...
+   end select
+   print*, "Your grade is ", name
+
+   select case (name)
+      case default
+         print*, "Invalid grade"
+      case ('now')
+      case ('test')
+         print*, "Well done"
+   end select
```
klausler: Tests are needed for other kinds of `CHARACTER`.

include/flang/evaluate/fold.h
```diff
@@ -96,5 +96,56 @@ std::optional<std::int64_t> ToInt64(const std::optional<A> &x) {
...
+template<int KIND>
+std::optional<std::string> ToString(
+    const Expr<Type<TypeCategory::Character, KIND>> &expr) {
+  if (auto scalar{
+          GetScalarConstantValue<Type<TypeCategory::Character, KIND>>(expr)}) {
+    auto str{*scalar};
+    // Type of str can be std::string or std::u16string or std::u32string
+    std::string newStr(str.begin(), str.end());
+    return newStr;
```
tskeith: I think it's clearer if you just write: `return std::string{scalar->begin(), scalar->end()};`

lib/semantics/check-select-stmt.cpp
```diff
@@ -0,0 +1,316 @@
...
+                       (intRangeValue.fromVal_.value() >= mem.fromVal_))) &&
+                  (!mem.toVal_ ||
+                      (intRangeValue.fromVal_ &&
+                          (intRangeValue.fromVal_.value() <= mem.toVal_)))) ||
+              ((!mem.fromVal_ ||
+                   (intRangeValue.toVal_ &&
+                       (intRangeValue.toVal_.value() >= mem.fromVal_))) &&
+                  (!mem.toVal_ ||
+                      (intRangeValue.toVal_ &&
+                          (intRangeValue.toVal_.value() <= mem.toVal_)))));
```
tskeith: Perhaps some of these complicated predicates could be simplified by creating functions that check if a value is in the range of a SelectCaseRangeType.

lib/semantics/check-select-stmt.cpp
```diff
@@ -0,0 +1,316 @@
...
+        std::find_if(selectCaseIntRangeVals_.begin(),
+            selectCaseIntRangeVals_.end(), [&intVal](const auto &mem) {
+              return ((!mem.fromVal_ || (intVal >= mem.fromVal_.value())) &&
+                  (!mem.toVal_ || (intVal <= mem.toVal_.value())));
+            });
+    if (selectCaseIntRangeValsIter != selectCaseIntRangeVals_.end()) {
+      ReportOverlapCaseValError(selectCaseIntRangeValsIter->source_, source);
+    }
+    auto selectCaseIntValsInsertStatus =
+        selectCaseIntVals_.insert(std::make_pair(intVal, source));
```
tskeith: `selectCaseIntVals_.emplace(intVal, source)` is better.

lib/semantics/check-select-stmt.cpp
```diff
@@ -0,0 +1,316 @@
...
+  }
+
+private:
+  template<typename T> struct SelectCaseRangeType {
+    SelectCaseRangeType(std::optional<T> fromVal, std::optional<T> toVal,
+        parser::CharBlock source)
+      : fromVal_(fromVal), toVal_(toVal), source_(source) {}
+    std::optional<T> fromVal_;
+    std::optional<T> toVal_;
+    parser::CharBlock source_;
```
tskeith: We only use the trailing underscore to mark private class data members. In this case `fromVal` etc. is fine.

lib/semantics/check-select-stmt.cpp
tskeith: I still think it would be better just to return. A user that is trying to do select case on some other type needs to rethink what they are doing.

lib/semantics/check-select-stmt.cpp
```diff
@@ -0,0 +1,316 @@
...
+          const auto &upper{rangeCase.upper};
+          if (upper &&
+              !IsValidSelectCaseType(selectCaseStmtType.value(),
+                  upper.value().thing, caseStmt.source)) {
+            continue;
+          }
+
+          if (selectCaseStmtType.value().category() == TypeCategory::Integer) {
+            const auto lowerVal = lower
+                ? std::optional(GetIntValue(lower->thing).value())
```
tskeith: Why not just `GetIntValue(lower->thing)`?

test/semantics/select-case.f90
```diff
@@ -0,0 +1,172 @@
...
+      case default
+         print*, "Invalid grade"
+   end select
+
+   select case (grade3)
+      case default
+         print*, "Invalid grade"
+      case (.true.)
+         print*, "Well done"
+      !ERROR: Not more than one of the selectors of SELECT CASE statement may be default
```
tskeith: I think "DEFAULT" would be better than "default".

lib/semantics/check-select-stmt.cpp
```diff
@@ -0,0 +1,316 @@
...
+          context_.Say(
+              src, "SELECT CASE value must be of type CHARACTER"_err_en_US);
+        } else {  // TypeCategory::Logical
+          context_.Say(
+              src, "SELECT CASE value must be of type LOGICAL"_err_en_US);
+        }
+      } else if (expectedType.category() == TypeCategory::Character) {
+        if (expectedType.kind() != type->kind()) {
+          context_.Say(src,
+              "SELECT CASE value kind must be same as SELECT CASE expression kind"_err_en_US);
```
tskeith: You could say what kind it is supposed to be. Like above where you say what type it is supposed to be.

test/semantics/select-case.f90
```diff
@@ -0,0 +1,172 @@
...
+      case (:30)
+         print*, "Excellent!"
+      case (40)
+         print*, "Excellent!"
+      case (90)
+         print*, "Excellent!"
+      case (91:99)
+         print*, "Very good!"
+      !ERROR: SELECT CASE statement value at this location overlaps with the below location
+      case (81:90)
```
tskeith: I'll repeat my suggestion to tell the user at least one of the overlapping values. Also, I think "below" and "above" are confusing because the "below" location is referring to the location of the message, but the actual location of the source is above.

I suggest a pair of messages like this:
CASE value '90' matches a previous CASE statement
Previous CASE statement matching '90'

test/semantics/select-case.f90
```diff
@@ -0,0 +1,172 @@
...
+      case default
+         print*, "Invalid grade"
+   end select
+
+   select case (grade3)
+      case default
+         print*, "Invalid grade"
+      case (.true.)
+      !ERROR: SELECT CASE value must be of type LOGICAL
+      case (3)
```
tskeith: This is referring to a case-stmt so it would be to say just "CASE value must be of type LOGICAL".

include/flang/evaluate/fold.h
kiranktp: KIND value is associated with the type. 
Eg: constCase

KIND => type{constCase->thing->GetType()}->kind()
Value => GetStringValue(constCase->thing).value()

KIND value is not lost upon conversion of Value from "u16string/u16string/string" to "string".
KIND value is still available with the expression type.

include/flang/evaluate/fold.h
kiranktp: As you suggested, I will replace this.

lib/evaluate/fold-integer.cpp
kiranktp: Thanks Klausler, I have moved them to fold-character.cpp and fold-logical.cpp files..

lib/semantics/check-select-stmt.cpp
kiranktp: I have removed the references to keep it uniform with other usages.

lib/semantics/check-select-stmt.cpp
kiranktp: I have used the braces for initialization.

lib/semantics/check-select-stmt.cpp
kiranktp: I have used the braces for initialization.

lib/semantics/check-select-stmt.cpp
kiranktp: I have done this changes.

lib/semantics/check-select-stmt.cpp
kiranktp: As suggested by you and Tim, I will abstract out the common code and keep it separate as a function.

lib/semantics/check-select-stmt.cpp
kiranktp: I have done this change.

test/semantics/select-case.f90
kiranktp: I have modified this.

test/semantics/select-case.f90
kiranktp: Thanks Tim, I will do the necessary changes.

lib/semantics/check-select-stmt.cpp
kiranktp: I have replaced insert(std::make_pair.. with emplace(..

lib/semantics/check-select-stmt.cpp
kiranktp: gfortran allows, Kind 1 and 4 for character type. I couldn't find any reference in the fortran 2018 standard for the character Kinds which are valid. What are the character kinds f18 supports?

lib/semantics/check-select-stmt.cpp
kiranktp: In select case, the upper range and lower range values are optional in certain cases. 
Eg: case (2:)
      case (:1)
The range pair has to be saved in the same fashion. so std::optional is being used.

lib/semantics/check-select-stmt.cpp
kiranktp: I will modify to display the expected kind for character.

test/semantics/select-case.f90
kiranktp: I will update the error message to show the overlapping values.

include/flang/evaluate/fold.h
klausler: I'm not worried about the kind value.  I'm worried about the value of the string.  You can't convert wide-character strings to 8-bit character strings without losing information.

lib/semantics/check-select-stmt.cpp
klausler: 1, 2, & 4 at the moment.

lib/semantics/check-select-stmt.cpp
klausler: You miss the point.  `GetIntValue()` returns an `optional`.  Why not use that result directly?

include/flang/evaluate/fold.h
kiranktp: I am not very good in C++. I tried different approaches to avoid converting the string from one kind to another for my requirement. I got only one solution working.
Let me know if you anyone has a better approach than the implemented one.

lib/semantics/check-select-stmt.cpp
kiranktp: I have taken care of all three character kinds without conversion.

lib/semantics/check-select-stmt.cpp
kiranktp: I have abstracted all the common code and created template functions for the same.

lib/semantics/check-select-stmt.cpp
kiranktp: Thanks for the detailed review Klauser. I was missing the point, GetIntValue() returns an optional.

test/semantics/select-case.f90
kiranktp: I have added check for all 3 character kinds, 1, 2 and 4 in the test case.

lib/semantics/check-select-stmt.cpp
```diff
@@ -0,0 +1,385 @@
...
+      const auto upperVal =
+          upper ? GetU16StringValue(upper->thing) : std::nullopt;
+      Insert(lowerVal, upperVal, source);
+    } else if (kind_ == 4) {
+      const auto lowerVal =
+          lower ? GetU32StringValue(lower->thing) : std::nullopt;
+      const auto upperVal =
+          upper ? GetU32StringValue(upper->thing) : std::nullopt;
+      Insert(lowerVal, upperVal, source);
+    }
```
kiranchandramohan: assert/check here?

lib/semantics/check-select-stmt.cpp
```diff
@@ -0,0 +1,385 @@
...
+      }
+    } else if (kind_ == 2) {
+      if (auto val{GetU16StringValue(x)}) {
+        Insert(val.value(), source);
+      }
+    } else if (kind_ == 4) {
+      if (auto val{GetU32StringValue(x)}) {
+        Insert(val.value(), source);
+      }
+    }
```
kiranchandramohan: assert/check here?

include/flang/semantics/tools.h
```diff
@@ -261,6 +261,55 @@ template<typename T> std::optional<std::int64_t> GetIntValue(const T &x) {
...
+    if (const auto *charExpr{
+            evaluate::UnwrapExpr<evaluate::Expr<evaluate::SomeCharacter>>(
+                *expr)}) {
+      return ToString(
+          std::get<evaluate::Expr<evaluate::Type<TypeCategory::Character, 4>>>(
+              (*charExpr).u));
+    }
+  }
+  return std::nullopt;
+}
```
kiranchandramohan: Can these 3 be a single function?

lib/semantics/check-select-stmt.cpp
```diff
@@ -0,0 +1,385 @@
...
+  }
+  if ((selectCaseStmtType.value().category() != TypeCategory::Integer) &&
+      (selectCaseStmtType.value().category() != TypeCategory::Character) &&
+      (selectCaseStmtType.value().category() !=
+          TypeCategory::Logical)) {  // C1145
+    context_.Say(parsedExpr.source,
+        "SELECT CASE expression must be of type CHARACTER, INTEGER, OR LOGICAL"_err_en_US);
+  } else {
+    validCaseStmtTypeCategory = selectCaseStmtType.value().category();
+  }
```
kiranchandramohan: This check does not seem to be handling if the select case expression is a derived type.
```
program mn
  type scores
    integer :: val
  end type
  type (scores) :: score
  select case (score)
  case(scores(100))
     print *, "century"
  end select
end
```

test/semantics/select-case.f90
```diff
@@ -0,0 +1,176 @@
...
+         print*, "Excellent!"
+      case (90)
+         print*, "Excellent!"
+      case (91:99)
+         print*, "Very good!"
+      !ERROR: CASE value 81:90 matches a previous CASE statement
+      case (81:90)
+         print*, "Very good!"
+      !ERROR: CASE value :80 matches a previous CASE statement
+      !ERROR: CASE value :80 matches a previous CASE statement
```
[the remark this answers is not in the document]
kiranktp: Here is how the error looks as of now.

select-case.f90:123:7: error: CASE value 81:90 matches a previous CASE statement
        case (81:90)
        ^^^^^^^^^^^^
select-case.f90:118:7: Previous CASE statement matching 81:90
        case (90)
        ^^^^^^^^^
Error is consecutive and has the same case value in both the error message to avoid any confusion.
Since we already have the same case value in both the error message, adding the line number to the error message might look like too much of duplicate info in error messages.
Let me know your opinion.

test/semantics/select-case.f90
```diff
@@ -0,0 +1,176 @@
+! Test SELECT CASE Constraints: C1145, C1146, C1147, C1148, C1149
```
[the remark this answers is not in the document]
kiranktp: I will move the constrains to the corresponding places where it is being tested.

test/semantics/select-case.f90
```diff
@@ -0,0 +1,176 @@
...
+   character (len = 10) :: name = 'test'
+   real :: grade4 = 2.0
+   logical, parameter :: grade5 = .false.
+   CHARACTER(KIND=1), PARAMETER    :: ze = 'a', at='b'
+   CHARACTER(KIND=2), PARAMETER    :: se = 'c'
+   CHARACTER(KIND=4), PARAMETER    :: mh='d'
+
+   select case (grade1)
+      case ('A')
+      print*, "Excellent!"
```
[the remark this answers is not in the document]
kiranktp: I will do this change.

test/semantics/select-case.f90
```diff
@@ -0,0 +1,176 @@
...
+   ! local variable declaration
+   character :: grade1 = 'B'
+   integer :: grade2 = 3
+   logical :: grade3 = .false.
+   character (len = 10) :: name = 'test'
+   real :: grade4 = 2.0
+   logical, parameter :: grade5 = .false.
+   CHARACTER(KIND=1), PARAMETER    :: ze = 'a', at='b'
+   CHARACTER(KIND=2), PARAMETER    :: se = 'c'
+   CHARACTER(KIND=4), PARAMETER    :: mh='d'
```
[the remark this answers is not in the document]
kiranktp: Apologies for giving such bad names for variables. I would make it sensible.

test/semantics/select-case.f90
```diff
@@ -0,0 +1,176 @@
...
+     case ('hj')
+         print*, "Well done now"
+     case default
+         print*, "Invalid grade"
+   end select
+
+  select case (ze)
+     case (at)
+         print*, "Excellent!"
+     !ERROR: CASE value kind is 4, it must be 1
```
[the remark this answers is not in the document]
kiranktp: Thank you, this error message looks much better than the earlier one. I will incorporate this change.

test/semantics/select-case.f90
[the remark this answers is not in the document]
kiranktp: I will add test case for C1144 as well.

lib/semantics/check-select-stmt.cpp
kiranktp: if GetU32StringValue returns std::nullopt, insertion will fail and we simply return. Should we add assert for this?

include/flang/semantics/tools.h
kiranktp: Really I tried my best to combine these 3 functions into one. But I couldn't. This was the reason, I had earlier commented asking for any help in extracting these 3 values without conversion. Do you have any suggestions how I can make this a single function.?

lib/semantics/check-select-stmt.cpp
kiranktp: Derived types are not allowed in select Case:
f18 throws below error as of now: 

test2.f90:25:16: error: SELECT CASE expression must be of type CHARACTER, INTERGER, OR LOGICAL
    select case (score)
                 ^^^^^

lib/semantics/check-select-stmt.cpp
kiranchandramohan: i mean a default else and an assertion/check for kinds not handled.

lib/semantics/check-select-stmt.cpp
kiranchandramohan: Sorry about that. Yes, it does catch that error. But because execution continues it goes forward and hits an assert/check on selectCaseStmtType->kind() at 313.
You might not see this on a release build.

test/semantics/select-case.f90
kiranchandramohan: @kiranktp can you comment here that you have now handled this case and may be mark it as resolved.

test/semantics/select-case.f90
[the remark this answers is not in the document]
kiranchandramohan: Agree, ignore my comment.

include/flang/semantics/tools.h
kiranchandramohan: I will come back to you on this.

lib/semantics/check-select-stmt.cpp
kiranktp: Thanks for the detailed review Kiran. I will add a check for default else case.
We do not enter this code for types other than Integer, real and bool. If there is a mismatch in kind or type. IsValidSelectCaseType function will issue an error and return to handle next select case.

lib/semantics/check-select-stmt.cpp
kiranktp: I will add a check for default else case.

lib/semantics/check-select-stmt.cpp
kiranktp: I didnt notice that. Thanks for pointing this out Kiran. I will avoid extracting the kind if there is a mismatch in type.

test/semantics/select-case.f90
kiranktp: Somehow i was missing to comment on this. This case has been handled now.

include/flang/semantics/tools.h
DavidTruby: You can write the functions the same way as follows:
```c++
template<typename S = std::string, typename T> 
std::optional<S> GetStringValue(const T &x) {
  if (const auto *expr{GetExpr(x)}) {
    if (const auto *charExpr{
            evaluate::UnwrapExpr<evaluate::Expr<evaluate::SomeCharacter>>(
                *expr)}) {
      return ToString(
          std::get<evaluate::Expr<evaluate::Type<TypeCategory::Character, sizeof(typename S::value_type)>>>(
              (*charExpr).u));
    }
  }
  return std::nullopt;
}
```
and then change the calls to the non-std::string ones from 
`GetU16StringValue` -> `GetStringValue<std::u16string>`
and 
`GetU32StringValue` -> `GetStringValue<std::u32string>`

kiranchandramohan: Thanks @DavidTruby.

kiranktp: Thank you @DavidTruby 
I will use your implementation.

kiranktp:
I have incorporated all the review comments. Let me know if anything else needs to be changed.
Since directory names got changed, all the files are appearing as new files with changes. Earlier review comments are missing in review section.

include/flang/evaluate/fold.h
kiranktp: Now I have one function which returns different character strings without any string conversion.

lib/semantics/check-select-stmt.cpp
kiranktp: I have added the default else case with proper error message.

lib/semantics/check-select-stmt.cpp
kiranktp: I have abstracted common code into function templates.

lib/semantics/check-select-stmt.cpp
```diff
@@ -0,0 +1,385 @@
...
+    }
+  }
+  void ReportOverlapCaseValError(parser::CharBlock currSource,
+      parser::CharBlock prevSource, const char *val) {
+    // C1149
+    messages_
+        .Say(currSource,
+            "CASE value %s matches a previous CASE statement"_err_en_US, val)
+        .Attach(
+            prevSource, "Previous CASE statement matching %s"_err_en_US, val);
```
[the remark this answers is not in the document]
kiranktp: I have not modified the diagnostic message as of now. Please let me know if this has to be modified.

lib/Semantics/check-select-stmt.cpp
```diff
@@ -0,0 +1,398 @@
...
+  }
+  if (!selectCaseStmtType) {
+    return;
+  }
+  if ((selectCaseStmtType.value().category() != TypeCategory::Integer) &&
+      (selectCaseStmtType.value().category() != TypeCategory::Character) &&
+      (selectCaseStmtType.value().category() !=
+          TypeCategory::Logical)) {  // C1145
+    context_.Say(parsedExpr.source,
+        "SELECT CASE expression must be of type CHARACTER, INTEGER, OR LOGICAL"_err_en_US);
```
tskeith: The logic of the rest of this function would be simpler if you returned after reporting this error.

lib/Semantics/check-select-stmt.cpp
```diff
@@ -0,0 +1,398 @@
...
+#include "flang/Semantics/tools.h"
+
+namespace Fortran::semantics {
+
+class SelectCaseHelper {
+
+public:
+  SelectCaseHelper(parser::Messages &messages) : messages_{messages} {}
+  void SetKind(int kind) { kind_ = kind; }
+  void SetTypeCategory(TypeCategory category) { category_ = category; }
```
tskeith: The coding standards call for this to be called `set_kind` and `set_category`. But hopefully you can eliminate these and pass the kind and category in to the constructor.

lib/Semantics/check-select-stmt.cpp
```diff
@@ -0,0 +1,398 @@
...
+
+  template<typename T> std::string GetValueAsString(T val) {
+    if constexpr (std::is_same_v<T, bool>) {
+      return std::string(val ? "true" : "false");
```
tskeith: These should be ".TRUE." and ".FALSE.". But even better would be to use the folded `evaluate::Expr` and call `AsFortran` on that. Then this function wouldn't need to duplicate that functionality.

lib/Semantics/check-select-stmt.cpp
```diff
@@ -0,0 +1,398 @@
...
+    } else if constexpr (std::is_same_v<T, std::int64_t>) {
+      return std::string(std::to_string(val));
```
tskeith:
```suggestion
      return std::to_string(val);
```

lib/Semantics/check-select-stmt.cpp
kiranktp: Thank you Tim!
As you mentioned I have modified to return after reporting the error to keep the logic simple.

lib/Semantics/check-select-stmt.cpp
kiranktp: I have modified to initialize these in constructor.

lib/Semantics/check-select-stmt.cpp
kiranktp: While implimenting this, i had tired to use AsFortran interface. For this, i would have to pass the evaluated expression to multiple functions and was able to use AsFortran in just 2 places. I thought current approach would keep the code more readable. 
I have modified true/false to .TRUE./.FALSE.

lib/Semantics/check-select-stmt.cpp
kiranktp: I have incorporated this review comment.

include/flang/Semantics/tools.h
```diff
@@ -269,6 +269,28 @@ template<typename T> std::optional<std::int64_t> GetIntValue(const T &x) {
...
+}
+
+template<typename S = std::string, typename T>
+std::optional<S> GetStringValue(const T &x) {
+  if (const auto *expr{GetExpr(x)}) {
+    if (const auto *charExpr{
+            evaluate::UnwrapExpr<evaluate::Expr<evaluate::SomeCharacter>>(
+                *expr)}) {
+      return ToString(
+          std::get<evaluate::Expr<evaluate::Type<TypeCategory::Character,
```
klausler: This will crash if the expression is a character expression of a distinct kind.  It also assumes that the kind code values for `CHARACTER` are the same as the `sizeof` values for various C++ string elements.

lib/Semantics/check-select-stmt.cpp
```diff
@@ -0,0 +1,393 @@
...
+
+namespace Fortran::semantics {
+
+class SelectCaseHelper {
+
+public:
+  SelectCaseHelper(parser::Messages &messages, int kind, TypeCategory category)
+    : kind_{kind}, category_(category), messages_{messages} {}
+
+  template<typename T> std::string GetValueAsString(T val) {
```
klausler: This function seems like a useful utility; hiding it as a non-static (why?) member function of a helper class in an implementation file will just lead to code duplication later.

A more general function that accepts more types and converts them to expressions and uses AsFortran() would be better.

lib/Semantics/check-select-stmt.cpp
```diff
@@ -0,0 +1,393 @@
...
+    }
+  }
+  template<typename T> std::string GetValuePairAsString(T fromVal, T toVal) {
+    if constexpr (std::is_same_v<T, std::optional<std::int64_t>>) {
+      return std::string((fromVal ? std::to_string(*fromVal) : "") + ":" +
+          (toVal ? std::to_string(*toVal) : ""));
+    } else {
+      return ((fromVal ? parser::QuoteCharacterLiteral(fromVal.value())
+                       : std::string("")) +
+          std::string(":") +
```
klausler: `":"s` would be better, here and elsewhere.

lib/Semantics/check-select-stmt.cpp
```diff
@@ -0,0 +1,393 @@
...
+  template<typename T> std::string GetValueAsString(T val) {
+    if constexpr (std::is_same_v<T, bool>) {
+      return std::string(val ? ".TRUE." : ".FALSE.");
+    } else if constexpr (std::is_same_v<T, std::int64_t>) {
+      return std::to_string(val);
+    } else {
+      return parser::QuoteCharacterLiteral(val);
+    }
+  }
+  template<typename T> std::string GetValuePairAsString(T fromVal, T toVal) {
```
klausler: Should be `static`.

lib/Semantics/check-select-stmt.cpp
```diff
@@ -0,0 +1,393 @@
...
+    return intVal <= upperIntVal;
+  }
+  template<typename T>
+  bool CheckOverlappingValue(
+      const T &val, std::optional<T> lowerVal, std::optional<T> upperVal) {
+    return ((!lowerVal || IsMoreThanLowerVal(val, lowerVal.value())) &&
+        (!upperVal || IsLessThanUpperVal(val, upperVal.value())));
+  }
+  template<typename T>
+  bool CheckOverlappingRangeValue(std::optional<T> memberFromVal,
```
klausler: This algorithm is overly complicated and the implementation is a mess.

Two intervals [a,b] and [c,d] overlap if and only if (c<=b && a<=d).

There's a utility Interval class template that used to be in lib/common/interval.h that might be helpful here.

lib/Semantics/check-select-stmt.cpp
```diff
@@ -0,0 +1,393 @@
...
+    return selectCaseU16StringRangeVals_;
+  }
+  template<typename T>
+  typename std::enable_if<std::is_same<T, std::u32string>::value,
+      std::vector<SelectCaseRangeType<std::u32string>> &>::type
+  GetRangeValHandle() {
+    return selectCaseU32StringRangeVals_;
+  }
+
+  std::map<bool, parser::CharBlock> selectCaseLogicalVals_;
```
klausler: `CASE(x)` with a single value and no colon can be represented as the range `x:x`.

It is not necessary for this class to be able to represent maps of all of these types at the same time; all of the cases in a `SELECT CASE` have to hold constant ranges of the same type.  Abstracting the entire class as a template would be clearer.

test/Semantics/select-case.f90
```diff
@@ -0,0 +1,207 @@
...
+        print *, "Well done"
+     !ERROR: CASE value must be of type INTEGER
+     case (.true. :)
+  end select
+
+  select case (UTF8_var1)
+     case (UTF8_var2)
+         print *, "Excellent!"
+     !ERROR: Character kind type of case construct (=1) mismatches with the kind type of case value (=4)
+     case (UTF32_var)
```
klausler: `CHARACTER` values are not UTF-8 in memory, and there is no such thing as UTF-16 or UTF-32.

test/Semantics/select-case.f90
```diff
@@ -0,0 +1,207 @@
...
+     case ('hello')
+         print *, "Excellent!"
+     case ('hey')
+         print *, "Good"
+     case ('hi':'ho')
+         print *, "Well done"
+     !ERROR: CASE value "hj" matches a previous CASE statement
+     case ('hj')
+         print *, "Well done now"
+     case default
```
klausler: You need test cases here with open lower & upper bounds.

include/flang/Semantics/tools.h
kiranktp: f18 supports 1, 2, and 4 character kinds. My understanding was, these are made available to encode UTF-8, UTF 16, UTF32 strings. But you mentioned there is nothing like UTF-16 and UTF-32.
Could you please elaborate on the supported kinds of character in f18 and its usage?
This will clarify most of our doubts.
I will remove the dependency on C++ string kinds for this function. 
This will again lead to 3 different functions for each kind, this was my earlier implementation.

lib/Semantics/check-select-stmt.cpp
kiranktp: I will try to make it a generic function and move it to formatting.cpp file.

lib/Semantics/check-select-stmt.cpp
kiranktp: I will replace all occurrences.

lib/Semantics/check-select-stmt.cpp
kiranktp: Similar to GetValueAsString, even this function can be made generic and can be moved to formatting.cpp file.

lib/Semantics/check-select-stmt.cpp
kiranktp: I will fix this as per your suggestion.

test/Semantics/select-case.f90
kiranktp: I will add a check for open lower/upper bounds.

lib/Semantics/check-select-stmt.cpp
kiranktp: Thanks Klausler, 
>> CASE(x) with a single value and no colon can be represented as the range x:x.
This is a very good suggestion. I will re-implement with this modification.

kiranktp: I am stuck while implementing this approach:
I tried to abstract the entire class as a template
template<typename T> class SelectCaseHelper {
  ..
  std::vector<SelectCaseRangeType> selectCaseRangeVals_;
  ..
  }

At the time of creating an object, I need the exact type:
{bool, std::int64_t, std::string, std::u16string, std::u32string}

SelectCaseHelper<Type> selectCaseStmts(context_.messages(),
      selectCaseStmtType->kind(), selectCaseStmtType.value().category());

All we have is the kind and the category. 
Is there a way to get the equivalent type by using the kind and the category?
I tried to figure this out, but couldn't.

klausler: I will implement it when I get time.

include/flang/Semantics/tools.h
kiranchandramohan: The following is what I could glean from the code. This might be stating the obvious but just thought will share.

From function SelectedCharKind in lib/Evaluate/type.cpp we can see that,
ascii/default = kind 1
ucs-2 = kind 2
iso_10646/ucs-4 = kind 4

In lib/Parser/characters.cpp we can see functions for encoding and decoding. For e.g. encoding strings,
```
template std::string EncodeString<Encoding::LATIN_1, std::string>(
    const std::string &);
template std::string EncodeString<Encoding::UTF_8, std::u16string>(
    const std::u16string &);
template std::string EncodeString<Encoding::UTF_8, std::u32string>(
    const std::u32string &);
```
 
In functionExpressionAnalyzer::AnalyzeString in lib/Semantics/expression.cpp we can see the relation between kinds and Encoding/Decoding.
```
  switch (kind) {
  case 1:
    return AsGenericExpr(Constant<Type<TypeCategory::Character, 1>>{
        parser::DecodeString<std::string, parser::Encoding::LATIN_1>(
            string, true)});
  case 2:
    return AsGenericExpr(Constant<Type<TypeCategory::Character, 2>>{
        parser::DecodeString<std::u16string, parser::Encoding::UTF_8>(
            string, true)});
  case 4:
    return AsGenericExpr(Constant<Type<TypeCategory::Character, 4>>{
        parser::DecodeString<std::u32string, parser::Encoding::UTF_8>(
            string, true)});
  default: CRASH_NO_CASE;
  }
```

kiranktp: Thanks for sharing the details Kiran.

lib/Semantics/check-select-stmt.cpp
kiranktp: Hi klausler, Tim,
Converting the value to a expression and using AsFortran() would print the kind value as well:
select-case.f90:98:7: Previous CASE statement matching .true._1
select-case.f90:118:7: Previous CASE statement matching 81_8:90_8

I plan to use string replace to remove _*. I saw the usage of replace in couple of places.
What is the usual way to avoid kind getting printed along with the value?

kiranktp: Here is the current implementation of GetValueAsString and GetValuePairAsString functions:
template<typename T> std::string GetValueAsString(T val) {
  if constexpr (std::is_same_v<T, bool>) {
    return Expr<Type<TypeCategory::Logical, 1>>{
        Constant<Type<TypeCategory::Logical, 1>>{val}}
        .AsFortran();
  } else if constexpr (std::is_same_v<T, std::int64_t>) {
    return Expr<Type<TypeCategory::Integer, 8>>{
        Constant<Type<TypeCategory::Integer, 8>>{val}}
        .AsFortran();
  } else if constexpr (std::is_same_v<T, std::string> ||
      std::is_same_v<T, std::u16string> || std::is_same_v<T, std::u32string>) {
    return parser::QuoteCharacterLiteral(val);
  } else {
    // other unsupported types, Return emtpy string
    return std::string();
  }
}

template<typename T> std::string GetValuePairAsString(T fromVal, T toVal) {
  return ((fromVal ? GetValueAsString(fromVal.value()) : ""s) + ":"s +
      (toVal ? GetValueAsString(toVal.value()) : ""s));
}
Please suggest any better alternative ways if any.

tskeith: > Hi klausler, Tim,
> Converting the value to a expression and using AsFortran() would print the kind value as well:
> select-case.f90:98:7: Previous CASE statement matching .true._1
> select-case.f90:118:7: Previous CASE statement matching 81_8:90_8
> 
> I plan to use string replace to remove _*. I saw the usage of replace in couple of places.
> What is the usual way to avoid kind getting printed along with the value?

I suggest leaving the kind there, at least for now.

kiranktp: Thanks for the clarification Tim. I would retain the kind value.

lib/Semantics/check-select-stmt.cpp
kiranktp: I tried to use factory method to create the object with specific types, but this didn't work because of member function templates. So I gave up here and I have added a TODO for this comment.

kiranktp:
Hi @tskeith @klausler,
I have incorporated the review comments. For one review comment, I have left a TODO with the explanation.

lib/Semantics/check-select-stmt.cpp
```diff
@@ -0,0 +1,329 @@
...
+    return selectCaseU16StringRangeVals_;
+  }
+  template<typename T>
+  typename std::enable_if<std::is_same<T, std::u32string>::value,
+      std::vector<SelectCaseRangeType<std::u32string>> &>::type
+  GetRangeValHandle() {
+    return selectCaseU32StringRangeVals_;
+  }
+
+  std::vector<SelectCaseRangeType<bool>> selectCaseLogicalRangeVals_;
```
klausler: This representation allows a `SELECT CASE` construct to have cases of all of these distinct types, but that's more complicated than Fortran requires.  All of the cases in a `SELECT CASE` construct will have the same type, which is the type on which `SelectCaseHelper` should be templatized.

kiranktp: Yes, the right approach would be to make the class SelectCaseHelper templatized. Like below:

template<typename T>
class SelectCaseHelper {
public:
  SelectCaseHelper(parser::Messages &messages, int kind, TypeCategory category)
    : kind_{kind}, category_(category), messages_{messages} {}
  ...
private:
  struct SelectCaseRangeType { ... };
  ...
  std::vector<SelectCaseRangeType> selectCaseRangeVals_;
  int kind_;
  TypeCategory category_;
  parser::Messages &messages_;
};

This way only one vector data member is needed and we do not need helper functions
to get the handle as well.

Then while creating the object, we should instantiate with the exact type. Like below:
SelectCaseHelper<bool> selectCaseStmts
or
SelectCaseHelper<std::int64_t> selectCaseStmts
or
SelectCaseHelper<std::string> selectCaseStmts
or so on

All we have is the KIND and CATEGORY of the expression type while creating the object.
This is where I got stuck for a long time. I tried to use factory method by creating a dummy base class
for SelectCaseHelper and tried to instantiate the SelectCaseHelper class with specific types based on the 
KIND and CATEGORY values. This method does not work if we have a member function template. 

With current implementation, vector object of all types co-exists, but only one will be used based on
the SELECT CASE expression type, if there is a mismatch of type or kind, we anyway issue errors for type
and kind mismatch.

This is the working soulution I could get, hence I added a TODO to make this class templatized as
you had mentioned earlier that you will implement a way to extract the exact type given the KIND and CATEGORY.
Once that is in place I can make the class templatized and remove redundant vector data members,
corresponding helper functions.

Kindly suggest if you know a way to extract the exact type to instantiate the class SelectCaseHelper.

kiranktp: > This representation allows a SELECT CASE construct to have cases of all of these distinct types, but that's more complicated than Fortran requires.  All of the cases in a SELECT CASE construct will have the same type, which is the type on which SelectCaseHelper should be templatized.
Hi @klausler , 
I have converted SelectCaseHelper to a template class, there by redundant data member and supporting members functions have been removed.

lib/Semantics/check-select-stmt.cpp
```diff
@@ -0,0 +1,311 @@
...
+    parser::CharBlock overlappingSource;
+    if (!IsRangeValueOverlapping(rangeValue, overlappingSource)) {
+      selectCaseRangeVals_.push_back(rangeValue);
+    } else {
+      ReportOverlapCaseValError(source, overlappingSource,
+          Fortran::evaluate::GetValuePairAsString(lowerVal, upperVal).c_str());
+    }
+  }
+  template<typename U> void Insert(const U &x, parser::CharBlock source) {
+    if constexpr (std::is_same_v<T, bool>) {
```
klausler: Long chains of `if constexpr` code defeat the goal of generic programming.

I have a short polymorphic implementation of `SELECT CASE` semantic checking that I will push out for review later today that demonstrates what I think this code should look like.

kiranktp:
> Please see #1089 for a polymorphic approach to the problem that relies on the Constant<> template class to avoid code duplication.
Thanks @klausler. My approach was entirely different than what was expected.

klausler:
Thank you for working so long on this.  I hope that you learned something and will continue contributing to f18.

kiranktp:
> Thank you for working so long on this.  I hope that you learned something and will continue contributing to f18.
Thank you! It was a very nice exposure to f18 internals. I will continue to contribute to f18.

PR #891, opened 2019-12-23
Some fixes

klausler:
More fixes to problems encountered running SPEC CPU codes through f18 with semantics enabled.

The significant bug fixed here is the case of `FOO(x,y,z)` where `FOO` is both the name of a generic interface and the name of a derived type.  The compiler was always treating the reference as a structure constructor, but we need to ensure first that it doesn't match any of the specific procedures in the generic interface.

### DIFF
--- a/lib/semantics/expression.cc
+++ b/lib/semantics/expression.cc
@@ -625,7 +625,7 @@ MaybeExpr ExpressionAnalyzer::Analyze(const parser::BOZLiteralConstant &x) {
     Say("BOZ literal '%s' too large"_err_en_US, x.v);
     return std::nullopt;
   }
-  return {AsGenericExpr(std::move(value.value))};
+  return AsGenericExpr(std::move(value.value));
 }
 
 // For use with SearchTypes to create a TypeParamInquiry with the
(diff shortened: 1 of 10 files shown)


lib/semantics/expression.cc
```diff
@@ -1762,21 +1773,26 @@ auto ExpressionAnalyzer::GetCalleeAndArguments(
...
-      return std::nullopt;
+      if (symbol->GetUltimate().has<semantics::DerivedTypeDetails>()) {
+        if (mightBeStructureConstructor) {
+          return CalleeAndArguments{
+              semantics::SymbolRef{*symbol}, std::move(arguments)};
+        }
+      } else {
+        return CalleeAndArguments{
+            ProcedureDesignator{*symbol}, std::move(arguments)};
+      }
```
tskeith: You could write this as:
```
if (mightBeStructureConstructor || !symbol->GetUltimate().has<semantics::DerivedTypeDetails>()) {
  return CalleeAndArguments{semantics::SymbolRef{*symbol}, std::move(arguments)};
}
```
This avoids the duplicate constructor call but does make the condition more complicated.

klausler: The constructor calls are actually not identical -- one is a `SymbolRef` (signifying a structure constructor) and the other is a `ProcedureDesignator`.

PR #895, opened 2019-12-25
Semantic checks for Select-rank construct(#R1148)

Sameeranjoshi:
Semantic checks for R1148


lib/semantics/check-select-stmt.cc
```diff
@@ -0,0 +1,112 @@
...
+  if (std::holds_alternative<parser::Expr>(selectRankStmtSel.u)) {
+    // TODO:
+    // 1.Checks if selector is an expression ex.Rank(a+b)
+    //
+  } else if (const auto *variable{
+                 std::get_if<parser::Variable>(&selectRankStmtSel.u)}) {
+    if (const Symbol * entity{GetLastName(*variable).symbol}) {
+      if (!IsAssumedRankArray(*entity)) {  // C1150
+        context_.Say(variable->GetSource(),
+            "Selector is not an Assumed-rank array variable"_err_en_US);
```
klausler: Please don't capitalize adjectives unless they start sentences.

lib/semantics/check-select-stmt.cc
```diff
@@ -0,0 +1,112 @@
...
+    const auto &rank{
+        std::get<parser::SelectRankCaseStmt::Rank>(rankCaseStmt.statement.t)};
+
+    // C1153 At most one RANK DEFAULT select-rank-case-stmt allowed
+    if (std::holds_alternative<parser::Default>(rank.u)) {
+      if (!defaultRankFound) {
+        defaultRankFound = true;
+      } else {
+        context_.Say(rankCaseStmt.source,
+            "Not more than one of the selectors of select rank statement "
```
klausler: Please capitalize the keywords `SELECT RANK`.

lib/semantics/check-select-stmt.cc
```diff
@@ -0,0 +1,112 @@
...
+        starRankFound = true;
+      } else {
+        context_.Say(rankCaseStmt.source,
+            "Not more than one of the selectors of select rank statement "
+            "may be '*'"_err_en_US);
+      }
+      if (std::holds_alternative<parser::Expr>(selectRankStmtSel.u)) {
+        // TODO:1.Checks if selector is an expression ex.Rank(a+b)
+      }
+      // C1155 RANK (*) not allowed if the selector has the ALLOCATABLE or
```
klausler: For better formatting, pull this block comment into the following `if`.

lib/semantics/check-select-stmt.cc
```diff
@@ -0,0 +1,112 @@
...
+      // POINTER attribute
+      else if (const auto *variable{
+                   std::get_if<parser::Variable>(&selectRankStmtSel.u)}) {
+        if (const Symbol * entity{GetLastName(*variable).symbol}) {
+          if (IsAllocatableOrPointer(*entity)) {
+            context_.Say(variable->GetSource(),
+                "RANK (*) cannot be used when selector has "
+                "pointer or allocatable"_err_en_US);
```
klausler: `is POINTER or ALLOCATABLE`

lib/semantics/check-select-stmt.cc
```diff
@@ -0,0 +1,112 @@
...
+          }
+        }
+      }
+    } else if (const auto &init{
+                   std::get_if<parser::ScalarIntConstantExpr>(&rank.u)}) {
+      if (const auto val{GetIntValue(*init)}) {
```
klausler: Why `const` here?

lib/semantics/check-select-stmt.cc
```diff
@@ -0,0 +1,112 @@
...
+        if ((*val < 0) || (*val > CFI_MAX_RANK)) {  // C1151
```
klausler: `Fortran::common::maxRank` from `lib/common/Fortran.h` is probably better than `CFI_MAX_RANK` here.

klausler: The needless parentheses make this range testing expression less idiomatic.

lib/semantics/check-select-stmt.cc
```diff
@@ -0,0 +1,112 @@
...
+      }
+    }
+  }
+
+  // R1150 select-rank-case-stmt checks
+  auto &rankCaseList{std::get<std::list<parser::SelectRankConstruct::RankCase>>(
+      selectRankConstruct.t)};
+  bool defaultRankFound{false};
+  bool starRankFound{false};
+  std::vector<std::optional<std::int64_t>> matches;
```
klausler: Why are you using a `vector` to represent a set?  Use `std::set` or `std::bitset`.

lib/semantics/check-select-stmt.cc
```diff
@@ -0,0 +1,112 @@
...
+              "Same rank values not allowed more than once"_err_en_US);
+        } else {
+          matches.push_back(val);
+        }
+      }
+    }
+  }
+}
+
+}  // namespace Fortran::semantics
```
klausler: You don't need to comment the end of a namespace if you don't want to.

lib/semantics/check-select-stmt.cc
```diff
@@ -0,0 +1,112 @@
+//==check-select-stmt.cc - Checker for select-case, select-rank
```
klausler: The comment says you're checking `SELECT CASE` and `SELECT RANK` but this file only contains code for `SELECT RANK`.

There's a [pull request](https://github.com/flang-compiler/f18/pull/889) outstanding for `SELECT CASE` that uses the same file name `check-select-stmt.cc`; are you two coordinating your work?

lib/semantics/check-select-stmt.cc
```diff
@@ -0,0 +1,112 @@
...
+//
+// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===--------------------------------------------------------------------------------------===
+
+#include "check-select-stmt.h"
+#include "tools.h"
+#include "../include/flang/ISO_Fortran_binding.h"
```
klausler: See below; I recommend that you use `../common/Fortran.h` instead.

lib/semantics/check-select-stmt.cc
Sameeranjoshi: Yes :)

lib/semantics/check-select-stmt.cc
```diff
@@ -0,0 +1,110 @@
...
+      }
+    }
+  }
+
+  // R1150 select-rank-case-stmt checks
+  auto &rankCaseList{std::get<std::list<parser::SelectRankConstruct::RankCase>>(
+      selectRankConstruct.t)};
+  bool defaultRankFound{false};
+  bool starRankFound{false};
+  std::set<std::optional<std::int64_t>> matches;
```
DavidTruby: Do you need the `std::optional` indirection here? I think you can just only insert the number if it has a value, since the `nullopt` values aren't going to be useful.

Sameeranjoshi: Removing the `std::optional` indirection I get error 
error: no matching function for call to ‘std::set<long int>::insert(std::optional<long int>&)’
as `if (auto val{GetIntValue(*init)}){ ..... }` the value(`val`) to be inserted is optional.
Is that the correct?

lib/semantics/check-select-stmt.cc
```diff
@@ -0,0 +1,110 @@
...
+    } else if (const auto &init{
+                   std::get_if<parser::ScalarIntConstantExpr>(&rank.u)}) {
+      if (auto val{GetIntValue(*init)}) {
+        if (*val < 0 || *val > Fortran::common::maxRank) {  // C1151
+          context_.Say(rankCaseStmt.source,
+              "The value of the selector must be "
+              "between zero and %d"_err_en_US,
+              Fortran::common::maxRank);
+        }
+        auto result_pair{matches.insert(val)};
```
Sameeranjoshi: @DavidTruby matches is filled inside `if (auto val{GetIntValue(*init)}) ` 
AFAIU it won't fill `nullopt` in `matches`

DavidTruby: You can dereference it when inserting it (e.g. `matches.insert(*val)`) since you don't need to store it as an optional. Then you shouldn't get the type error you mentioned above I believe.
It's a minor thing, it just avoids you having to store an extra pointer that you don't need.

Sameeranjoshi: done

Sameeranjoshi:
> 
> 
> A good start.

Done

lib/semantics/check-select-stmt.cc
```diff
@@ -0,0 +1,110 @@
...
+              "between zero and %d"_err_en_US,
+              Fortran::common::maxRank);
+        }
+        auto result_pair{matches.insert(*val)};
+        if (!result_pair.second) {  // C1152
+          context_.Say(rankCaseStmt.source,
+              "Same rank values not allowed more than once"_err_en_US);
+        }
+      }
+    }
```
tskeith: If you check the alternatives of `rank.u` this way, you should add an `else` clause that has a `DIE` in it so it is clear that there are no other alternatives besides Star, Default, and integer expression.
Even better is to write is as an `std::visit` so that it the compiler checks there are no other alternatives.

lib/semantics/check-select-stmt.cc
```diff
@@ -0,0 +1,110 @@
...
+        }
+      }
+    } else if (const auto &init{
+                   std::get_if<parser::ScalarIntConstantExpr>(&rank.u)}) {
+      if (auto val{GetIntValue(*init)}) {
+        if (*val < 0 || *val > Fortran::common::maxRank) {  // C1151
+          context_.Say(rankCaseStmt.source,
+              "The value of the selector must be "
+              "between zero and %d"_err_en_US,
+              Fortran::common::maxRank);
```
tskeith: The `Fortran::` part is redundant. All of this code is nested under the `Fortran` namespace.

lib/semantics/check-select-stmt.cc
```diff
@@ -0,0 +1,110 @@
...
+        }
+        auto result_pair{matches.insert(*val)};
+        if (!result_pair.second) {  // C1152
```
tskeith: We don't use underscores in variable names.

test/semantics/select-rank.f90
```diff
@@ -0,0 +1,121 @@
+!===--- select-rank.f90 - Test select rank constraints -------------------===
+!
+! Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+! See https://llvm.org/LICENSE.txt for license information.
+! SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+!
+!===------------------------------------------------------------------------===
```
tskeith: You don't need to add this notice on test files.

Sameeranjoshi:
The specification doesn't speak on what happens when selector is an expression.
For example SELECT RANK(a+b) or SELECT RANK(0+1) ....
Is `if (std::holds_alternative<parser::Expr>(selectRankStmtSel.u)) ` always expected to throw a semantic error?

lib/semantics/check-select-stmt.cc
klausler: optionals are pointers?  I'm not sure that's the case.

lib/semantics/check-select-stmt.cc
klausler: Please use `std::visit()` here.

klausler:
> The specification doesn't speak on what happens when selector is an expression.
> For example SELECT RANK(a+b) or SELECT RANK(0+1) ....
> Is `if (std::holds_alternative<parser::Expr>(selectRankStmtSel.u)) ` always expected to throw a semantic error?

R1150 requires that the expression be a scalar integer constant (which will be packaged as an `Expr<SomeType>`).

lib/semantics/check-select-stmt.cc
Sameeranjoshi: Thanks, I have Added the change please review.

lib/semantics/check-select-stmt.cc
Sameeranjoshi: I didn't understand.

Sameeranjoshi:
> 
> 
> > The specification doesn't speak on what happens when selector is an expression.
> > For example SELECT RANK(a+b) or SELECT RANK(0+1) ....
> > Is `if (std::holds_alternative<parser::Expr>(selectRankStmtSel.u)) ` always expected to throw a semantic error?
> 
> R1150 requires that the expression be a scalar integer constant (which will be packaged as an `Expr<SomeType>`).

I am unable to figure out how to handle to TODO case.
`If selector is an expression` 
Can you please point how to move ahead?
Thanks

lib/semantics/check-select-stmt.cc
```diff
@@ -0,0 +1,135 @@
...
+              if (!IsAssumedRankArray(*entity)) {  // C1150
+                context_.Say(variable.GetSource(),
+                    "Selector is not an assumed-rank array variable"_err_en_US);
+              }
+            }
+          },
+          [](auto &) {
+            common::die("Illegal value of selector in SELECT RANK "
+                        "statement");
+          },
```
tskeith: You don't need this lambda: `Expr` and `Variable` are the only alternatives in `Selector`.

Sameeranjoshi: Thank you @tskeith I have removed it.
When I run `make test ` on local
The following tests FAILED:
        **`191 - call12.f90 (Failed)`**
Whereas committing a change upstream the drone.ci bots show results as passing.
Why does that happen, am I missing something while testing them on local?

tskeith: I doubt this change is causing a test to fail. If you put the code back it does it pass?

lib/semantics/check-select-stmt.cc
```diff
@@ -0,0 +1,127 @@
...
+                if (!resultPair.second) {  // C1152
+                  context_.Say(rankCaseStmt.source,
+                      "Same rank values not allowed more than once"_err_en_US);
+                }
+              }
+            },
+            [](auto &) {
+              common::die("Illegal value of selector in SELECT RANK case "
+                          "statement");
+            },
```
tskeith: This lambda is unnecessary too.

lib/semantics/check-select-stmt.cc
```diff
@@ -0,0 +1,127 @@
...
+    std::visit(
+        common::visitors{
+            [&](const parser::Default &) {  // C1153
+              if (!defaultRankFound) {
+                defaultRankFound = true;
+              } else {
+                context_.Say(rankCaseStmt.source,
+                    "Not more than one of the selectors of SELECT RANK "
+                    "statement "
+                    "may be default"_err_en_US);
```
tskeith: This message doesn't need to be split across three lines.

lib/semantics/check-select-stmt.cc
DavidTruby: Sorry, my mistake, I didn't mean an extra pointer I meant an extra tag (since optional must store a tag to indicate if the value is present or not)

lib/semantics/check-select-stmt.cc
```diff
@@ -0,0 +1,121 @@
+//==check-select-stmt.cc - Checker for select-rank
```
klausler: The first and last lines of this license block are wrong.  Please see other files and conform to their usage.

lib/semantics/check-select-stmt.cc
```diff
@@ -0,0 +1,121 @@
...
+
+void SelectStmtChecker::Leave(
+    const parser::SelectRankConstruct &selectRankConstruct) {
+  const auto &selectRankStmt{
+      std::get<parser::Statement<parser::SelectRankStmt>>(
+          selectRankConstruct.t)};
+  const auto &selectRankStmtSel{
+      std::get<parser::Selector>(selectRankStmt.statement.t)};
+  // R1149 select-rank-stmt checks
+  std::visit(
```
klausler: C1150 requires the *selector* to be the name of an assumed-rank dummy argument, so this code can be simplified and completed.  Use `UnwrapWholeSymbolDataRef()` from `lib/evaluate/tools.h` to determine whether the selector is a simple name reference; if it is, verify that it's an assumed-rank dummy argument, and if it isn't, emit an error.

lib/semantics/check-select-stmt.cc
```diff
@@ -0,0 +1,121 @@
...
+            },
+            [&](const parser::Star &) {  // C1153
+              if (!starRankFound) {
+                starRankFound = true;
+              } else {
+                context_.Say(rankCaseStmt.source,
+                    "Not more than one of the selectors of SELECT RANK "
+                    "statement may be '*'"_err_en_US);
+              }
+              std::visit(
```
klausler: See the comment above.  When the selector is confirmed to be an assumed-rank dummy argument, save its symbol for easier checking here.

lib/semantics/check-select-stmt.cc
```diff
@@ -0,0 +1,121 @@
...
+                            context_.Say(variable.GetSource(),
+                                "RANK (*) cannot be used when selector is "
+                                "POINTER or ALLOCATABLE"_err_en_US);
+                          }
+                        }
+                      },
+                  },
+                  selectRankStmtSel.u);
+            },
+            [&](const parser::ScalarIntConstantExpr &init) {
```
klausler: Within the scope of the `RANK` case, the *associate-name* in the symbol table needs to have the corresponding rank.  Is the name resolution phase already handling this task?

test/semantics/select-rank.f90
```diff
@@ -0,0 +1,115 @@
...
+
+   !Error expected
+   subroutine CALL_ME2(x)
+    implicit none
+    integer :: x(..)
+    integer :: y(3)
+    !ERROR: Selector is not an assumed-rank array variable
+    SELECT RANK(y)
+    RANK (0)
+      print *, "PRINT RANK 0"
```
klausler: It would be a better test here if `y` were to be used in some context that required it to be a scalar variable.  Similarly, in the other `RANK` cases, use the *associate-name* in some context that will elicit an error if the required rank were incorrect.

test/semantics/select-rank.f90
```diff
@@ -0,0 +1,115 @@
...
+    implicit none
+    integer :: x(..)
+    SELECT RANK(x)
+    Rank(2)
+      print *, "Merry Christmas"
+    RANK (*)
+      print *, "Happy New Year"
+    !ERROR: Not more than one of the selectors of SELECT RANK statement may be '*'
+    RANK (*)
+      print *, "Opps!! Wrong"
```
klausler: `Opps`?

test/semantics/select-rank.f90
Sameeranjoshi: I have added some tests are they as expected ones ?

lib/semantics/check-select-stmt.cpp
```diff
@@ -0,0 +1,114 @@
...
+  const auto &selectRankStmtSel{
+      std::get<parser::Selector>(selectRankStmt.statement.t)};
+
+  // R1149 select-rank-stmt checks
+  const Symbol *saveSelSymbol{nullptr};
+  if (const auto resolvedSel{ResolveSelector(selectRankStmtSel)}) {
+    if (const Symbol * sel{evaluate::UnwrapWholeSymbolDataRef(*resolvedSel)}) {
+      if (!IsAssumedRankArray(*sel)) {
+        context_.Say(parser::FindSourceLocation(selectRankStmtSel),
+            "Selector is not an assumed-rank array variable"_err_en_US);
```
klausler: Can you emit the name of the symbol as part of the message, here and at line 42?

lib/semantics/check-select-stmt.cpp
```diff
@@ -0,0 +1,114 @@
...
+          "Selector is not an assumed-rank array variable"_err_en_US);
+    }
+  }
+
+  // R1150 select-rank-case-stmt checks
+  auto &rankCaseList{std::get<std::list<parser::SelectRankConstruct::RankCase>>(
+      selectRankConstruct.t)};
+  bool defaultRankFound{false};
+  bool starRankFound{false};
+  std::set<std::int64_t> matches;
```
klausler: Since this set only has to hold the values in [0..maxRank], a bitset would work and might be more clear.  But even better would be a simple array of optional source locations (`std::optional<CharBlock> caseForRank[maxRank + 1];`), so that your error message about duplicate rank cases could point to the preceding duplicate case when it exists.

lib/semantics/check-select-stmt.h
```diff
@@ -0,0 +1,27 @@
+//===-- lib/semantics/check-select-stmt.h --------------------------------*- C++
```
klausler: The weird license header punctuation is garbled on the first two lines here.

lib/semantics/semantics.cpp
```diff
@@ -110,10 +111,11 @@ template<typename... C> class SemanticsVisitor : public virtual C... {
...
 using StatementSemanticsPass1 = ExprChecker;
-using StatementSemanticsPass2 = SemanticsVisitor<AllocateChecker,
-    ArithmeticIfStmtChecker, AssignmentChecker, CoarrayChecker,
-    DeallocateChecker, DoChecker, IfStmtChecker, IoChecker, NullifyChecker,
-    OmpStructureChecker, PurityChecker, ReturnStmtChecker, StopChecker>;
+using StatementSemanticsPass2 =
+    SemanticsVisitor<AllocateChecker, ArithmeticIfStmtChecker,
+        AssignmentChecker, CoarrayChecker, DeallocateChecker, DoChecker,
+        IfStmtChecker, IoChecker, NullifyChecker, OmpStructureChecker,
+        PurityChecker, ReturnStmtChecker, SelectStmtChecker, StopChecker>;
```
klausler: `SelectConstructChecker` would be a more accurate name, right?  You're not just checking the initial `SELECT` statements.

test/semantics/select-rank.f90
klausler: No.  Instead of just having a `PRINT` statement, try using the `SELECT RANK` construct entity in the test in some way that would work *only when the construct entity has the expected rank*, using something with a compile-time test of the rank.

```
SELECT RANK(x)
RANK(0)
  j = INT(0, KIND=MERGE(KIND(0), -1, RANK(x) == 0)) ! will fail when RANK(x) is not zero here
END SELECT
```

test/semantics/select-rank.f90
```diff
@@ -0,0 +1,202 @@
...
+     END SELECT
+   end
+
+   subroutine CALL_ME3(x)
+    implicit none
+    integer :: x(..)
+    SELECT RANK(x)
+    !ERROR: The value of the selector must be between zero and 15
+    RANK (16)
+      print *, "PRINT RANK 0"
```
klausler: This `PRINT` statement isn't needed, and its text is wrong anyway.

lib/semantics/check-select-stmt.cc
Sameeranjoshi: Added `std::optional<parser::CharBlock> caseForRank[common::maxRank + 1];`

test/semantics/select-rank.f90
Sameeranjoshi: For the patch submitted(currently passed a non zero value to pass the drone.io build) , test Fails with below error, when the `rank` is `0`, which is supposed to fail instead when `rank` is `not zero`.
`_../f18/test/semantics/select-rank.f90:39:11: error: 'kind=' argument must be a constant scalar integer whose value is a supported kind for the intrinsic result type
`

`
        j = INT(0, KIND=MERGE(KIND(0), -1, RANK(x) == 0)) ! will fail when RANK(x) is not zero 
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
_`
Is this an issue with the `RANK` construct or `select rank` ?

klausler: That seems to indicate a problem with the semantics or name resolution of  x` in the `RANK`.  If the rank of `x` in that context were zero, as it should be, the `KIND=` argument would have had a valid value.  You can test that hypothesis by changing `==` to `/=` in that expression and seeing whether it (incorrectly) does not emit an error.

Sameeranjoshi: Conformed for above discussed test with suggested changes,
`j = INT(0, KIND=MERGE(KIND(0), -1, RANK(x) /= 0))   ` `!Is evaluating to true, `which means `RANK` is not resolving properly the value of `x`.

Sameeranjoshi:
For the review comments(https://github.com/flang-compiler/f18/pull/895#discussion_r377212295).
I went to fix it and ended up creating some other PR(https://github.com/flang-compiler/f18/pull/1010)
Now talking to @tskeith and also posting to slack channel an RFC(https://flang-compiler.slack.com/archives/C5C58TT32/p1583259635025300), which doesn't have any answer yet probably as no one has any clear idea on it and needs bit research.
This requires to figure out what the appropriate representation is for the symbol within the select rank case so that the rest of semantics (and lowering) does the right thing with it, and in itself would be another PR for name resolution.
So @klausler @sscalpone  can the select rank PR be merged with marking it as TODO as all other constraints have been resolved except the compile time test(https://github.com/flang-compiler/f18/pull/895#discussion_r375974864)? also given the migration to LLVM, which could cause me again to reinvent the wheel on Phabricator(which I haven't used).

Sameeranjoshi:
> 
> 
> For the review comments([#895 (comment)](https://github.com/flang-compiler/f18/pull/895#discussion_r377212295)).
> I went to fix it and ended up creating some other PR(#1010)
> Now talking to @tskeith and also posting to slack channel an RFC(https://flang-compiler.slack.com/archives/C5C58TT32/p1583259635025300), which doesn't have any answer yet probably as no one has any clear idea on it and needs bit research.
> This requires to figure out what the appropriate representation is for the symbol within the select rank case so that the rest of semantics (and lowering) does the right thing with it, and in itself would be another PR for name resolution.
> So @klausler @sscalpone can the select rank PR be merged with marking it as TODO as all other constraints have been resolved except the compile time test([#895 (comment)](https://github.com/flang-compiler/f18/pull/895#discussion_r375974864))? also given the migration to LLVM, which could cause me again to reinvent the wheel on Phabricator(which I haven't used).

Have created a new symbol in each scope, also added the respective 'rank: ' for a symbol.
@klausler I haven't used the ArraySpec method of retrieving the rank, as there is no array-spec in a SelectRankCaseStmt, so it was empty. Not sure if it goes according to F18's standard way, but the blocker test seems to pass.
Thanks @tskeith and @jeanPerier for valuable suggestions.

Sameeranjoshi:
Not sure why is drone.ci failing, am I missing something from my end?

Sameeranjoshi:
Not sure why is drone.ci failing, am I missing something from my end?

test/Semantics/select-rank.f90
```diff
@@ -0,0 +1,216 @@
...
+    SELECT RANK(y)
+    RANK (0)
+      print *, "PRINT RANK 0"
+    RANK (1)
+      print *, "PRINT RANK 1"
+     END SELECT
+
+    SELECT RANK(x)
+    RANK(0)
+      j = INT(0, KIND=MERGE(KIND(0), -1, RANK(x) == 0)) ! will fail when RANK(x) is not zero here
```
klausler: You should use this compilation-time testing technique (or another) in place of (or in addition to) the `PRINT` statements elsewhere so that the tests confirm that the right rank is visible in each case.  You should also test that the other attributes of the `SELECT RANK` construct entities are preserved (e.g., `ALLOCATABLE` entities remain allocatable, `POINTER` assumed-rank objects can be associated with a pointer assignment, and so on.)

lib/Semantics/resolve-names.cpp
```diff
@@ -4933,6 +4977,13 @@ void ConstructVisitor::SetTypeFromAssociation(Symbol &symbol) {
   }
 }
 
+// Set the rank of symbol based on the current rankCase value extracted from
+// evaluating the parser node.
+void ConstructVisitor::SetRankFromParserNode(Symbol &symbol, int actualRank) {
+  const int dummyRank{actualRank};
```
klausler: Why is `actualRank` copied to this new variable?

include/flang/Semantics/symbol.h
```diff
@@ -557,6 +567,7 @@ class Symbol {
   }
 
   void SetType(const DeclTypeSpec &);
+  void SetRank(const int rank);
```
klausler: Why is this argument `const`?  It's not a reference or a pointer.

include/flang/Semantics/symbol.h
```diff
@@ -131,8 +131,18 @@ class AssocEntityDetails : public EntityDetails {
   AssocEntityDetails &operator=(AssocEntityDetails &&) = default;
   const MaybeExpr &expr() const { return expr_; }
 
+  void set_rank(int rank);
+  std::optional<int> associationRank() const {
+    if (associationRank_.has_value()) {
```
klausler: Why not just `return associationRank_;`?

include/flang/Semantics/symbol.h
```diff
@@ -131,8 +131,18 @@ class AssocEntityDetails : public EntityDetails {
...
+      return associationRank_.value();
+    } else {
+      return {};
+    }
+  }
+
 private:
   MaybeExpr expr_;
+  std::optional<int> associationRank_;
```
klausler: Is this new field relevant to the shape analysis code in `lib/Evaluate/shape.cpp`?  I think that it must be, so that code should be extended to use the specific case ranks so that operations that depend on the shape of an array will be compiled correctly.

test/Semantics/select-rank.f90
klausler: Test cases that depend on the shape of the array should also be present.

test/Semantics/select-rank.f90
```diff
@@ -0,0 +1,216 @@
...
+
+    SELECT RANK(ptr=>x)
+    RANK (3)
+      PRINT *, "PRINT RANK 3"
+    RANK (1)
+      PRINT *, "PRINT RANK 1"
+    END SELECT
+   end subroutine
+
+!end program selectRankProg
```
klausler: These last couple lines look like they're obsolete.

include/flang/Semantics/symbol.h
Sameeranjoshi: ### Can you please explain more on what do you mean by 

> new field relevant to the shape analysis code


The tests other than select rank which depend on shape of array seem to be successful for  `ninja test check-all` 

What I can understand from comments is that, 
` MaybeExpr expr_` private member function in `class AssocEntityDetails` is being used in `lib/Evaluate/shape.cpp` inside function `GetShape() `
and my patch doesn't seem to handle the `expr_` inside `SetRankFromParserNode `but is handled in `SetTypeFromAssociation`

include/flang/Semantics/symbol.h
Sameeranjoshi: I was thinking if in case the value of `std::optional `wouldn't be set checking the `has_value() ` would have been saving us from undefined behavior, but I found c++ compiler catches it with `std::bad_optional_access`. So anyways I have removed the guard.

test/Semantics/select-rank.f90
Sameeranjoshi: Added shape related tests, also used compile-time checking test wherever necessary. I have skipped some redundant locations where we could have added the compile-time check.

include/flang/Semantics/symbol.h
klausler: You can copy an `optional` whether it has a value or not without throwing an exception.

include/flang/Semantics/symbol.h
klausler: Make sure that your tests for `SELECT RANK` include operations in the rank cases that will force the compiler to analyze the shape of the array at the specific rank.  Use `FORALL` statements, the `SHAPE` intrinsic, array assignments, &c.  There is probably code in the shape analysis framework that will need to be updated to handle this new representation of construct entities correctly, using the rank of the construct entity and the descriptor of the assumed-rank dummy argument for inquiries.

Sameeranjoshi:
@klausler I have tried to formulate some tests from my understanding of Fortran language.
Pasting the godbolt.org links for keeping the comments clean.
Can you please suggest if they need more changes? 
https://godbolt.org/z/UiRcRg       = SHAPE intrinsics
https://godbolt.org/z/oW2Zyl      = SHAPE intrinsics
https://godbolt.org/z/N86jMD    = using FORALL statement
https://godbolt.org/z/YKWY9A   = using array assignment statement
https://godbolt.org/z/raN-5f      = using LOC for address of array

Sameeranjoshi:
> @klausler I have tried to formulate some tests from my understanding of Fortran language.
> Pasting the godbolt.org links for keeping the comments clean.
> Can you please suggest if they need more changes?
> https://godbolt.org/z/UiRcRg       = SHAPE intrinsics
> https://godbolt.org/z/oW2Zyl      = SHAPE intrinsics
> https://godbolt.org/z/N86jMD    = using FORALL statement
> https://godbolt.org/z/YKWY9A   = using array assignment statement
> https://godbolt.org/z/raN-5f      = using LOC for address of array

If the tests seem to be written as expected, I see all passing except the following condition.

```
 select rank (arr) 
            rank(2)
                j = INT(0, KIND=MERGE(KIND(0), -1, RANK(arr) == SIZE(SHAPE(arr)))) !here aarr is dummy 
        end select

```
Checking 
`lib/Evaluate/shape.cpp `at [this](https://github.com/flang-compiler/f18/blob/5e26830620e5e6d647fba04ae81077d7148e409f/lib/Evaluate/shape.cpp#L421) function
Seems like `assoc.expr() `might not have rank information.Is that the issue you see?
@klausler can you please point what do you expect this to return ? as we don't have `shape_ `in `AssocEntityDetails.` to fill in the `shape vector` to return.

klausler:
> Seems like `assoc.expr() `might not have rank information.Is that the issue you see?
> @klausler can you please point what do you expect this to return ? as we don't have `shape_ `in `AssocEntityDetails.` to fill in the `shape vector` to return.

`assoc.expr()` is the assumed-rank dummy argument itself.

Within a `RANK` case, shape analysis should return a shape vector of the appropriate rank populated with descriptor inquiries into the extents of the leading dimensions of the assumed-rank dummy argument.

Sameeranjoshi:
@klausler thanks for explaining, but there seems to be regression for other tests and the select-rank cases, not sure why are they regressing(They tend to Abort the compiler).
Did I create a wrong patch?

lib/Evaluate/shape.cpp
```diff
@@ -454,7 +460,15 @@ auto GetShapeHelper::operator()(const Component &component) const -> Result {
       shape.emplace_back(GetExtent(context_, base, dimension));
     }
     return shape;
-  } else {
+  } else if (symbol.has<semantics::AssocEntityDetails>()) { //here
```
klausler: `// here`?

lib/Semantics/check-select-stmt.cpp
```diff
@@ -0,0 +1,126 @@
+//===-- lib/semantics/check-select-stmt.cpp -------------------------------===//
```
klausler: This is not a good name for this file.  Fortran has three different `SELECT` statements and this file only checks one of the three.

lib/Semantics/check-select-stmt.cpp
```diff
@@ -0,0 +1,126 @@
...
+      if (!IsAssumedRankArray(*sel)) {  // C1150
+        context_.Say(parser::FindSourceLocation(selectRankStmtSel),
+            "Selector '%s' is not an assumed-rank array variable"_err_en_US,
+            sel->name().ToString());
+      } else {
+        saveSelSymbol = sel;
+      }
+    } else {
+      context_.Say(parser::FindSourceLocation(selectRankStmtSel),
+          "Selector '%s' is not an assumed-rank array variable"_err_en_US,
```
klausler: Is this the right error message for selectors that are not whole variables?

lib/Semantics/check-select-stmt.cpp
```diff
@@ -0,0 +1,126 @@
...
+            },
+            [&](const parser::Star &) {  // C1153
+              if (!starRankFound) {
+                starRankFound = true;
+              } else {
+                context_.Say(rankCaseStmt.source,
+                    "Not more than one of the selectors of SELECT RANK "
+                    "statement may be '*'"_err_en_US);
+              }
+              if (saveSelSymbol) {
```
klausler: These nested `if` statements do not need to be separate.

lib/Semantics/check-select-stmt.cpp
```diff
@@ -0,0 +1,126 @@
...
+#include "flang/Semantics/tools.h"
+#include <list>
+#include <optional>
+#include <set>
+#include <tuple>
+#include <variant>
+
+namespace Fortran::semantics {
+
+void SelectConstructChecker::Leave(
```
klausler: Should be a more specific name.

lib/Semantics/resolve-names.cpp
```diff
@@ -985,11 +985,18 @@ class ConstructVisitor : public virtual DeclarationVisitor {
   void Post(const parser::EndAssociateStmt &);
   void Post(const parser::Association &);
   void Post(const parser::SelectTypeStmt &);
+
+  void Post(const parser::SelectRankStmt &);
   bool Pre(const parser::SelectTypeConstruct &);
   void Post(const parser::SelectTypeConstruct &);
   bool Pre(const parser::SelectTypeConstruct::TypeCase &);
   void Post(const parser::SelectTypeConstruct::TypeCase &);
+  // gives blank scopes
```
klausler: I do not understand this comment; please make it more clear.

lib/Semantics/resolve-names.cpp
```diff
@@ -5094,6 +5119,25 @@ void ConstructVisitor::Post(const parser::TypeGuardStmt::Guard &x) {
   }
 }
 
+void ConstructVisitor::Post(const parser::SelectRankCaseStmt::Rank &x) {
+  if (auto *symbol{MakeAssocEntity()}) {
+    SetTypeFromAssociation(*symbol);
+    SetAttrsFromAssociation(*symbol);
+    std::visit(
```
klausler: When testing for only one alternative, do not use `visit()`.  `std::get_if<T>(&x.u)` is much better.

lib/Semantics/resolve-names.cpp
```diff
@@ -5094,6 +5119,25 @@ void ConstructVisitor::Post(const parser::TypeGuardStmt::Guard &x) {
...
+        common::visitors{
+            [&](const parser::ScalarIntConstantExpr &init) {
+              Walk(init);
+              MaybeIntExpr expr{EvaluateIntExpr(init)};
+              if (auto val{evaluate::ToInt64(expr)}) {
+                SetRankFromParserNode(*symbol, *val);
+              }
```
klausler: What happens if the expression is not a constant?  Is this the place where an error message should be emitted?

Should you validate the value of the `RANK` case here instead?

lib/Semantics/resolve-names.cpp
```diff
@@ -5169,6 +5213,12 @@ void ConstructVisitor::SetTypeFromAssociation(Symbol &symbol) {
   }
 }
 
+// Set the rank of symbol based on the current rankCase value extracted from
```
klausler: What is the purpose of this function?  Why not call `Symbol::SetRank()` directly?

lib/Semantics/symbol.cpp
```diff
@@ -287,6 +288,14 @@ void Symbol::SetType(const DeclTypeSpec &type) {
       details_);
 }
 
+void Symbol::SetRank(int rank) {
+  std::visit(
+      common::visitors{
+          [&](AssocEntityDetails &x) { x.set_rank(rank); },
+          [](auto &) {},
```
klausler: Maybe you should crash instead of ignoring the internal error.

You should use `std::get_if<>()` instead of `visit()` when dealing with a single case.

test/Semantics/select-rank.f90
```diff
@@ -0,0 +1,266 @@
...
+      print *, "PRINT RANK 3"
+      j = INT(0, KIND=MERGE(KIND(0), -1, RANK(ptr) == 1))
+    !ERROR: Must be a constant value
+    RANK(nullptr)
+      print *, "PRINT RANK 3"
+    END SELECT
+
+    !ERROR: Selector 'x(1) + x(2)' is not an assumed-rank array variable
+    SELECT RANK (x(1) + x(2))
+    RANK(1)
```
klausler: These `RANK` cases are not useful in these tests for errors on the `SELECT RANK` statement.

test/Semantics/select-rank.f90
```diff
@@ -0,0 +1,266 @@
...
+   end subroutine
+
+   subroutine CALL_ME3(x)
+    implicit none
+    integer :: x(..),j
+    SELECT RANK(x)
+    !ERROR: The value of the selector must be between zero and 15
+    RANK (16)
+      j = INT(0, KIND=MERGE(KIND(0), -1, RANK(x) == 16))
+    END SELECT
```
klausler: What about errors in type (`RANK(.TRUE.)` or `RANK(1.0)`) or rank (`RANK(array)`) or variability (`RANK(x)` with non-constant x)?

test/Semantics/select-rank02.f90
```diff
@@ -0,0 +1,66 @@
...
+   call check(arr_pass)
+
+contains
+    subroutine check(arr)
+        implicit none
+        integer :: arr(..)
+        INTEGER :: j
+        select rank (arr)
+            rank(2)
+                j = INT(0, KIND=MERGE(KIND(0), -1, RANK(arr) == SIZE(SHAPE(arr)))) !arr is dummy
```
klausler: Don't test `RANK(arr)` and `SIZE(SHAPE(arr))` against each other when you know what their values should be (2).  They could both be wrong.

test/Semantics/select-rank02.f90
```diff
@@ -0,0 +1,66 @@
...
+    end subroutine
+    subroutine check2(arr)
+        implicit none
+        integer :: arr(..)
+        INTEGER :: j
+        integer,dimension(-1:10, 20:30) :: brr
+
+        select rank (arr)
+            rank(2)
+                j = INT(0, KIND=MERGE(KIND(0), -1, RANK(arr) == SIZE(SHAPE(brr)))) !brr is local to subroutine
```
klausler: I don't understand what's being tested here.

test/Semantics/select-rank02.f90
```diff
@@ -0,0 +1,66 @@
...
+    end subroutine
+    subroutine check5(arr)
+        implicit none
+        integer :: arr(..)
+        INTEGER :: j
+        integer, dimension(2) :: k,l
+        integer,dimension(-1:10, 20:30) :: brr
+        select rank (arr)
+            rank(2)
+                j = LOC(brr(0,25))
```
klausler: This test is just working with a local array and doesn't depend on the effective shape of the assumed-rank array at all.

lib/Semantics/check-select-stmt.cpp
Sameeranjoshi: How about `check-select-rank.[cpp|h]`?

lib/Evaluate/shape.cpp
```diff
@@ -417,9 +417,15 @@ auto GetShapeHelper::operator()(const Symbol &symbol) const -> Result {
             } else {
               return Scalar();
             }
-          },
+          },//here
           [&](const semantics::AssocEntityDetails &assoc) {
-            return (*this)(assoc.expr());
+	    Shape shape;
```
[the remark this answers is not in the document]
Sameeranjoshi: Thanks, I have added a guard using `if (!assoc.associationRank())`, the regression pass now. 
Is there any related function you know apart from above usage, or suggest some other design pattern you think, like adding something in class `AssocEntityDetails `like `IsAssumedRank() ` function similar to `ObjectEentityDetails`? I think the only way we can identify select rank is the `-fdebug-dump-symbols `will have a new `rank:` field.

lib/Evaluate/shape.cpp
Sameeranjoshi: Opps!! Debugging comments left. My bad

lib/Semantics/resolve-names.cpp
Sameeranjoshi: How about this?
// Creates Block scopes with no details nor symbol name.
I meant that printing the symbols using `-fdebug-dump-symbols `with this prints something like
```
Block scope:
Block scope:
Block scope:
Block scope:
```
I shall remove if that is misleading..

lib/Semantics/resolve-names.cpp
Sameeranjoshi: Ok, I was trying to create something more readable similar to `SetTypeFromAssociation` and `SetAttrsFromAssociation`,

lib/Semantics/check-select-stmt.cpp
Sameeranjoshi: If I remember, I saw somwthing similar on gfortran front, not sure though.
But can you suggest a better one?

test/Semantics/select-rank02.f90
Sameeranjoshi: Ok, I was trying to test with a local copy and the dummy arguments.
Remove this test?

lib/Semantics/symbol.cpp
Sameeranjoshi: How do we crash? Is it the DIE macro?

lib/Semantics/resolve-names.cpp
Sameeranjoshi: Do you mean "not a constant" as other values of below variant?
`std::variant<ScalarIntConstantExpr, Star, Default> u;`
If so currently `Star` and `Default` rank cases values won't see any `rank:` in printing symbols and the respective details.
This is how it would look for 
`RANK DEFAULT` and `RANK(*)` test cases
`arr: AssocEntity type: INTEGER(4) expr:arr`

Is `rank: 1` a better value for `*` and `DEFAULT`?

test/Semantics/select-rank02.f90
Sameeranjoshi: Ok, changing it to arr.

test/Semantics/select-rank.f90
Sameeranjoshi: They are already caught by `C713`.
Thanks @klausler I have added tests.

lib/Semantics/check-select-stmt.cpp
Sameeranjoshi: Done `SelectRankConstructChecker`

Sameeranjoshi:
@klausler I have incorporated most of the changes.
Some queries I have asked in for the review comments.
Thank you.

Sameeranjoshi:
Ping..

Sameeranjoshi:
Merged in llvm-project/master.

PR #896, opened 2019-12-25
C_F_POINTER

klausler:
Argument checking will be added before merging, if the general approach is found to be sound.

### DIFF
--- a/module/__fortran_builtins.f90
+++ b/module/__fortran_builtins.f90
@@ -1,0 +1,38 @@
+!===-- module/__fortran_builtins.f90 ---------------------------------------===!
+!
+! Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+! See https://llvm.org/LICENSE.txt for license information.
+! SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+!
+!------------------------------------------------------------------------------!
+
+! These naming shenanigans prevent names from Fortran intrinsic modules
+! from being usable on INTRINSIC statements, and force the program
+! to USE the standard intrinsic modules in order to access the
+! standard names of the procedures.
+module __Fortran_builtins
+
+  integer, parameter, private :: int64 = selected_int_kind(18)
+
+  intrinsic :: __builtin_c_f_pointer
+
+  type :: __builtin_c_ptr
+    integer(kind=int64) :: __address = 0
+  end type
+
+  type :: __builtin_c_funptr
+    integer(kind=int64) :: __address = 0
+  end type
+
+  type :: __builtin_event_type
+    integer(kind=int64) :: __count = 0
+  end type
+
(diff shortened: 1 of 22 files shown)


module/__fortran_builtins.f90
```diff
@@ -0,0 +1,15 @@
...
+! See https://llvm.org/LICENSE.txt for license information.
+! SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+!
+!------------------------------------------------------------------------------!
+
+! These naming shenanigans prevent names from Fortran intrinsic modules
+! from being usable on INTRINSIC statements, and force the program
+! to USE the standard intrinsic modules in order to access the
+! standard names of the procedures.
+module __Fortran_builtins
```
tskeith: I'm not sure I understand the comment. Is there anything that prevents user code from using `__Fortran_builtins` and calling `__builtin_c_f_pointer`? Should there be?

klausler: No, there isn't, and I'm not sure that there needs to be one -- it should be obvious that the intrinsics in this module are implementation artifacts.

sscalpone:
I think there needs to be a dependency to make sure the _ file is compiled before the others.  Sorry for the quick note with no supporting evidence.

klausler:
> I think there needs to be a dependency to make sure the _ file is compiled before the others. Sorry for the quick note with no supporting evidence.

I have no idea how to bring this about in the CMakeLists.txt file; I've added a comment that the module of builtins has to appear first.

klausler:
Please take another look now that error checking and tests have been added.  (But not the dependence in CMakeLists.txt that @sscalpone points out is needed.)

test/semantics/c_f_pointer.f90
```diff
@@ -0,0 +1,30 @@
+! Enforce 18.2.3.3
+
+include '../../module/__fortran_builtins.f90'
+include '../../module/iso_c_binding.f90'
```
tskeith: These includes aren't necessary. The error tests are run with a `-I` option that points to the predefined modules.

test/semantics/canondo16.f90
```diff
@@ -15,13 +15,11 @@
 
 ! CHECK: A DO loop should terminate with an END DO or CONTINUE
 
-module iso_fortran_env
-  type :: team_type
-  end type
-end
+include '../../module/__fortran_builtins.f90'
+include '../../module/iso_c_binding.f90'
```
tskeith: Unfortunately, the same isn't true here. We should probably convert all of these to be run with `test_errors.sh`.

tskeith:
Looks good to me.

test/semantics/c_f_pointer.f90
klausler: Ok.

PR #897, opened 2019-12-26
Semantic Checks for critical construct

kiranchandramohan:
This PR includes checks for the critical construct. Implemented checks include the following,
-> C1119: A branch within a CRITICAL construct shall not have a branch target that is outside the construct. Implemented by refactoring the LabelEnforce class in check-do.cc and putting it in tools.h and using it for CRITICAL construct.
-> C1118: The block of a critical-construct shall not contain a RETURN statement or an image control statement. Implemented directly and using the isimagecontrol function in tools.h.

Handled by existing code
C1117: Matching names for critical and end critical statements

Not implemented (TODO)
11.1.6.5: It is permissible to branch to an end-critical-stmt only from within its CRITICAL construct.

### DIFF
--- a/lib/semantics/check-coarray.cc
+++ b/lib/semantics/check-coarray.cc
@@ -17,6 +17,51 @@
 
 namespace Fortran::semantics {
 
+class CriticalBodyEnforce {
+public:
+  CriticalBodyEnforce(
+      SemanticsContext &context, parser::CharBlock criticalSourcePosition)
+    : context_{context}, criticalSourcePosition_{criticalSourcePosition} {}
+  std::set<parser::Label> labels() { return labels_; }
+  template<typename T> bool Pre(const T &) { return true; }
+  template<typename T> void Post(const T &) {}
+
+  template<typename T> bool Pre(const parser::Statement<T> &statement) {
+    currentStatementSourcePosition_ = statement.source;
+    if (statement.label.has_value()) {
+      labels_.insert(*statement.label);
+    }
+    return true;
+  }
+
+  // C1118
+  void Post(const parser::ReturnStmt &) {
+    context_
+        .Say(currentStatementSourcePosition_,
+            "RETURN statement is not allowed in a CRITICAL construct"_err_en_US)
+        .Attach(criticalSourcePosition_, GetEnclosingMsg());
+  }
+  void Post(const parser::ExecutableConstruct &construct) {
+    if (IsImageControlStmt(construct)) {
+      context_
+          .Say(currentStatementSourcePosition_,
+              "An image control statement is not allowed in a CRITICAL"
+              " construct"_err_en_US)
+          .Attach(criticalSourcePosition_, GetEnclosingMsg());
+    }
+  }
+
+private:
+  parser::MessageFixedText GetEnclosingMsg() {
+    return "Enclosing CRITICAL statement"_en_US;
+  }
+
+  SemanticsContext &context_;
+  std::set<parser::Label> labels_;
+  parser::CharBlock currentStatementSourcePosition_;
+  parser::CharBlock criticalSourcePosition_;
+};
+
 template<typename T>
 static void CheckTeamType(SemanticsContext &context, const T &x) {
   if (const auto *expr{GetExpr(x)}) {
@@ -44,6 +89,19 @@ void CoarrayChecker::Leave(const parser::ImageSelectorSpec &x) {
 
 void CoarrayChecker::Leave(const parser::FormTeamStmt &x) {
   CheckTeamType(context_, std::get<parser::TeamVariable>(x.t));
+}
+
+void CoarrayChecker::Enter(const parser::CriticalConstruct &x) {
+  auto &criticalStmt{std::get<parser::Statement<parser::CriticalStmt>>(x.t)};
+
+  const parser::Block &block{std::get<parser::Block>(x.t)};
+  CriticalBodyEnforce criticalBodyEnforce{context_, criticalStmt.source};
+  parser::Walk(block, criticalBodyEnforce);
+
+  // C1119
+  LabelEnforce criticalLabelEnforce{
+      context_, criticalBodyEnforce.labels(), criticalStmt.source, "CRITICAL"};
+  parser::Walk(block, criticalLabelEnforce);
 }
 
 // Check that coarray names and selector names are all distinct.
@@ -85,5 +143,4 @@ void CoarrayChecker::Say2(const parser::CharBlock &name1,
   context_.Say(name1, std::move(msg1), name1)
       .Attach(name2, std::move(msg2), name2);
 }
-
 }
--- a/lib/semantics/check-coarray.h
+++ b/lib/semantics/check-coarray.h
@@ -33,6 +33,8 @@ public:
   void Leave(const parser::ImageSelectorSpec &);
   void Leave(const parser::FormTeamStmt &);
 
+  void Enter(const parser::CriticalConstruct &);
+
 private:
   SemanticsContext &context_;
 
@@ -40,6 +42,5 @@ private:
   void Say2(const parser::CharBlock &, parser::MessageFixedText &&,
       const parser::CharBlock &, parser::MessageFixedText &&);
 };
-
 }
 #endif  // FORTRAN_SEMANTICS_CHECK_COARRAY_H_
--- a/lib/semantics/check-do.cc
+++ b/lib/semantics/check-do.cc
@@ -75,7 +75,6 @@ public:
     : context_{context}, doConcurrentSourcePosition_{
                              doConcurrentSourcePosition} {}
   std::set<parser::Label> labels() { return labels_; }
-  std::set<SourceName> names() { return names_; }
   template<typename T> bool Pre(const T &) { return true; }
   template<typename T> void Post(const T &) {}
 
@@ -213,57 +212,6 @@ public:
     }
   }
 
-  // C1167 -- EXIT statements can't exit a DO CONCURRENT
-  bool Pre(const parser::WhereConstruct &s) {
-    AddName(MaybeGetConstructName(s));
-    return true;
-  }
-
-  bool Pre(const parser::ForallConstruct &s) {
-    AddName(MaybeGetConstructName(s));
-    return true;
-  }
-
-  bool Pre(const parser::ChangeTeamConstruct &s) {
-    AddName(MaybeGetConstructName(s));
-    return true;
-  }
-
-  bool Pre(const parser::CriticalConstruct &s) {
-    AddName(MaybeGetConstructName(s));
-    return true;
-  }
-
-  bool Pre(const parser::LabelDoStmt &s) {
-    AddName(MaybeGetStmtName(s));
-    return true;
-  }
-
-  bool Pre(const parser::NonLabelDoStmt &s) {
-    AddName(MaybeGetStmtName(s));
-    return true;
-  }
-
-  bool Pre(const parser::IfThenStmt &s) {
-    AddName(MaybeGetStmtName(s));
-    return true;
-  }
-
-  bool Pre(const parser::SelectCaseStmt &s) {
-    AddName(MaybeGetStmtName(s));
-    return true;
-  }
-
-  bool Pre(const parser::SelectRankStmt &s) {
-    AddName(MaybeGetStmtName(s));
-    return true;
-  }
-
-  bool Pre(const parser::SelectTypeStmt &s) {
-    AddName(MaybeGetStmtName(s));
-    return true;
-  }
-
   // C1136 -- No RETURN statements in a DO CONCURRENT
   void Post(const parser::ReturnStmt &) {
     context_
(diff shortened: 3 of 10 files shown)


test/semantics/critical03.f90
```diff
@@ -0,0 +1,35 @@
...
+
+subroutine test3(i)
+  integer i
+  critical
+    !ERROR: Control flow escapes from CRITICAL
+    goto (10, 10, 20) i
+    10 i = i + 1
+  end critical
+20 i = i - 1
+end subroutine test3
```
DavidTruby: Maybe add some tests with internal branches that don't error, just for completeness.

lib/semantics/check-coarray.cc
```diff
@@ -46,6 +46,21 @@ void CoarrayChecker::Leave(const parser::FormTeamStmt &x) {
...
 
+void CoarrayChecker::Enter(const parser::CriticalConstruct &x) {
+  auto &criticalStmt{
+      std::get<parser::Statement<parser::CriticalStmt>>(x.t)};
+
+  const parser::Block &block{std::get<parser::Block>(x.t)};
+  CriticalBodyEnforce criticalBodyEnforce{context_, criticalStmt.source};
+  parser::Walk(block, criticalBodyEnforce);
+
+  // C1119: A branch within a CRITICAL construct shall not have a branch target
```
psteinfeld: We usually just name the constraint (C1119) rather than including its full text.

lib/semantics/check-coarray.h
```diff
@@ -10,6 +10,7 @@
 #define FORTRAN_SEMANTICS_CHECK_COARRAY_H_
 
 #include "semantics.h"
+#include "tools.h"
```
psteinfeld: It looks like this file needs to have clang-format run on it.

lib/semantics/check-do.cc
```diff
@@ -75,7 +75,6 @@ class DoConcurrentBodyEnforce {
     : context_{context}, doConcurrentSourcePosition_{
                              doConcurrentSourcePosition} {}
   std::set<parser::Label> labels() { return labels_; }
-  std::set<SourceName> names() { return names_; }
```
psteinfeld: It looks like this file needs to have clang-format run on it.

lib/semantics/check-coarray.cc
kiranchandramohan: I will change this in the next set of changes.

kiranchandramohan:
Thanks @DavidTruby, @psteinfeld. I have made the changes you suggested.

lib/semantics/check-coarray.h
```diff
@@ -41,5 +44,40 @@ class CoarrayChecker : public virtual BaseChecker {
...
+
+  // C1118
+  void Post(const parser::ReturnStmt &) {
+    context_.Say(currentStatementSourcePosition_,
+        "RETURN statement is not allowed in a CRITICAL construct"_err_en_US);
+  }
+  void Post(const parser::ExecutableConstruct &construct) {
+    if (IsImageControlStmt(construct)) {
+      context_.Say(currentStatementSourcePosition_,
+          "An image control statement is not allowed in a CRITICAL construct"_err_en_US);
```
psteinfeld: Can you split this so that the lines fit into 80 characters?  Unfortunately, clang-format doesn't do this automatically.

lib/semantics/check-coarray.h
```diff
@@ -41,5 +44,40 @@ class CoarrayChecker : public virtual BaseChecker {
       const parser::CharBlock &, parser::MessageFixedText &&);
 };
 
+class CriticalBodyEnforce {
```
psteinfeld: I'd guess that the class ```CriticalBodyEnforce``` will only be used in check-coarray.cc.  If so, please move it there.

lib/semantics/tools.h
```diff
@@ -471,5 +471,79 @@ FindPolymorphicAllocatableUltimateComponent(const DerivedTypeSpec &);
...
+  template<typename T> bool Pre(const T &) { return true; }
+  template<typename T> bool Pre(const parser::Statement<T> &statement) {
+    currentStatementSourcePosition_ = statement.source;
+    return true;
+  }
+
+  template<typename T> void Post(const T &) {}
+
+  void Post(const parser::GotoStmt &gotoStmt) { checkLabelUse(gotoStmt.v); }
+  void Post(const parser::ComputedGotoStmt &computedGotoStmt) {
```
psteinfeld: Member function definitions should go into .cc files when possible.  Please move the definition to tools.cc

lib/semantics/tools.h
```diff
@@ -471,5 +471,79 @@ FindPolymorphicAllocatableUltimateComponent(const DerivedTypeSpec &);
...
+  void Post(const parser::EorLabel &eorLabel) { checkLabelUse(eorLabel.v); }
+
+  void checkLabelUse(const parser::Label &labelUsed) {
+    if (labels_.find(labelUsed) == labels_.end()) {
+      SayWithConstruct(context_, currentStatementSourcePosition_,
+          parser::MessageFormattedText{
+              "Control flow escapes from %s"_err_en_US, construct_},
+          constructSourcePosition_);
+    }
+  }
```
psteinfeld: More definitions to be moved.

lib/semantics/tools.h
```diff
@@ -471,5 +471,79 @@ FindPolymorphicAllocatableUltimateComponent(const DerivedTypeSpec &);
...
+    return parser::MessageFormattedText{
+        "Enclosing %s statement"_en_US, construct_};
+  }
+
+  void SayWithConstruct(SemanticsContext &context,
+      parser::CharBlock stmtLocation, parser::MessageFormattedText &&message,
+      parser::CharBlock constructLocation) {
+    context.Say(stmtLocation, message)
+        .Attach(constructLocation, GetEnclosingConstructMsg());
+  }
```
psteinfeld: More definitions to be moved.

lib/semantics/check-coarray.h
```diff
@@ -41,5 +44,40 @@ class CoarrayChecker : public virtual BaseChecker {
...
+      context_.Say(currentStatementSourcePosition_,
+          "An image control statement is not allowed in a CRITICAL construct"_err_en_US);
+    }
+  }
+
+private:
+  SemanticsContext &context_;
+  std::set<parser::Label> labels_;
+  parser::CharBlock currentStatementSourcePosition_;
+  parser::CharBlock criticalSourcePosition_;
```
psteinfeld: I couldn't find where this data member is used.  Should it be removed?

kiranchandramohan: Have used it now to improve the error message by showing the location of the critical statement.

lib/semantics/tools.h
kiranchandramohan: Done.

lib/semantics/tools.h
kiranchandramohan: Done.

lib/semantics/tools.h
kiranchandramohan: Done.

lib/semantics/check-coarray.h
kiranchandramohan: Done.

kiranchandramohan:
> Thank you for doing this. In particular, it's good to have the more general `LabelEnforce` class.
> 
> It would be easier for me to review these changes if you branched directly off of the main GitHub repository rather than forking first.

I do not have write access to the repo hence I think I cannot do what you are suggesting here.

lib/semantics/check-coarray.h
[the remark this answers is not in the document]
kiranchandramohan: Good spot. Removed.

kiranchandramohan:
Thanks @psteinfeld for the review. Let me know if I should squash the commits.

lib/semantics/check-coarray.cc
```diff
@@ -17,6 +17,51 @@
 
 namespace Fortran::semantics {
 
+class CriticalBodyEnforce {
```
klausler: Consider subclassing semantics::BaseChecker here and using Enter and Leave member functions rather than Pre and Post; the results would require less boilerplate and might be more readable.

lib/semantics/tools.cc
```diff
@@ -1174,4 +1174,65 @@ bool IsFunctionResultWithSameNameAsFunction(const Symbol &symbol) {
...
+  if (labels_.find(labelUsed) == labels_.end()) {
+    SayWithConstruct(context_, currentStatementSourcePosition_,
+        parser::MessageFormattedText{
+            "Control flow escapes from %s"_err_en_US, construct_},
+        constructSourcePosition_);
+  }
+}
+
+parser::MessageFormattedText LabelEnforce::GetEnclosingConstructMsg() {
+  return parser::MessageFormattedText{
```
klausler: You could omit the name of the type here -- it's on the previous line, so it's obvious -- and probably save a line.

lib/semantics/tools.h
```diff
@@ -469,5 +469,43 @@ FindPolymorphicAllocatableUltimateComponent(const DerivedTypeSpec &);
...
+  std::set<parser::Label> labels_;
+  parser::CharBlock currentStatementSourcePosition_{nullptr};
+  parser::CharBlock constructSourcePosition_{nullptr};
+  const char *construct_{nullptr};
+
+  parser::MessageFormattedText GetEnclosingConstructMsg();
+  void SayWithConstruct(SemanticsContext &context,
+      parser::CharBlock stmtLocation, parser::MessageFormattedText &&message,
+      parser::CharBlock constructLocation);
+};  // class LabelEnforce
```
klausler: We typically don't comment closing braces of classes or namespaces or functions.

lib/semantics/tools.h
```diff
@@ -469,5 +469,43 @@ FindPolymorphicAllocatableUltimateComponent(const DerivedTypeSpec &);
 UltimateComponentIterator::const_iterator
 FindPolymorphicAllocatableNonCoarrayUltimateComponent(const DerivedTypeSpec &);
 
+class LabelEnforce {
```
klausler: If this class will serve for the implementation of additional semantic checks, please add a line or two of commentary here describing what it does.  But if it's probably only going to be useful for `CRITICAL`, then it should not be in tools.h, since it uses the parse tree walking framework and that's somewhat expensive (in term of build time) to instantiate in all clients of tools.h.

psteinfeld:
> 
> 
> Thanks @psteinfeld for the review. Let me know if I should squash the commits.

Yes, please squash them.  We like having a single commit per change request in the main repository's history.

lib/semantics/tools.cc
kiranchandramohan: I will change this.

lib/semantics/tools.h
kiranchandramohan: OK, I will remove this.

lib/semantics/tools.h
kiranchandramohan: I have slightly modified the DoConcurrentLabelEnforce class (https://github.com/flang-compiler/f18/blob/master/lib/semantics/check-do.cc#L350) in check-do.cc and renamed it as LabelEnforce and moved to tools.h. The LabelEnforce class now has two customers do-concurrent checks in check-do.cc and critical checks in check-coarray.cc. This can also be used for the change-team construct.

The class (given a set of labels) provides an error message if there is a branch to label which is not in the given set. I can add this as a comment. Is this OK?

lib/semantics/check-coarray.cc
kiranchandramohan: My understanding is probably poor here.
This class and its usage follow the DoconcurrentBodyEnforce class and its usage in check-do.cc. The parser::Walk function is used to walk the Body of the constructs (Doconcurrent/critical). The parser::Walk function requires Pre and Post function definitions.
The ones with Enter and Leave seem to be called by the SemanticsVisitor class. Are you suggesting to use the SemanticsVisitor class here?
I was initially thinking of creating a new base class, for these kinds of classes, which houses the base Pre and Post functions. This new class can be a base for similar classes like SymbolDumpVisitor in unparse-with-symbols.cc,
DoConcurrentBodyEnforce in check-do.cc etc.

lib/semantics/tools.h
klausler: That's a great one-sentence description, and it's not something I would have understood immediately just from the name of the class.

lib/semantics/check-coarray.cc
klausler: That's fine.

kiranchandramohan: If you meant fine for the new base class which houses the boiler plate pre and post functions, I can do that in a separate PR.

klausler: No, I mean just leave it as is.

kiranchandramohan:
I have addressed @klausler's comments and squashed the commits as agreed with @psteinfeld.

klausler:
It looks good to me.  Let me know if you'd like me to merge the changes for you.

kiranchandramohan:
Please merge.

PR #900, opened 2019-12-31
Partition fold.cc to reduce build-time memory requirements

klausler:
Breaks up `lib/evaluate/fold.cc` into a header of common definitions and a collection of per-type-category source files, in the hope that less memory will be required by a C++ compiler to build any of them.


klausler:
> Everything builds, tests, and looks good to me.
> 
> Have you measured the effect of this change?

Previously, it took 3m20s wall-clock time to build `fold.cc` alone on my machine, with a max resident set of 2885052 kbytes.  Now it takes 1m36 seconds (parallel builds) and 1887096 kbytes max RSS.

PR #901, opened 2020-01-02
Fix type-bound user-defined operators

klausler:
Type-bound generics of intrinsic operators (e.g., `operator(+)`) work, and non-type-bound generic user-defined operators (e.g., `operator(.foo.)`) work, but type-bound generic user-defined operators didn't.  This PR pulls the pieces together in `semantics/expression.cc` and gets a few more correctness tests to pass.


klausler:
Thanks for the quick reviews!

PR #902, opened 2020-01-02
Check for passing DO variables to OUT arguments in a CALL statement

psteinfeld:
I added code to save the INTENT of a dummy argument in the checked expression
of the actual argument.  When processing a CallStmt, I then retrieve the
ProcedureRef, which contains a list of the checked ActualArguments.  I then
traverse this list looking for actual arguments that are active DO variable
that are being passed to dummy arguments whose INTENT is either OUT or INOUT.
For OUT dummies, I put out an error message and warn for INOUT dummies.

### DIFF
--- a/lib/evaluate/call.h
+++ b/lib/evaluate/call.h
@@ -111,6 +111,11 @@ public:
   void set_isPassedObject(bool yes = true) { isPassedObject_ = yes; }
 
   bool Matches(const characteristics::DummyArgument &) const;
+  common::Intent dummyIntent() const { return dummyIntent_; }
+  ActualArgument &set_dummyIntent(common::Intent intent) {
+    dummyIntent_ = intent;
+    return *this;
+  }
 
   // Wrap this argument in parentheses
   void Parenthesize();
@@ -127,6 +132,7 @@ private:
   std::optional<parser::CharBlock> keyword_;
   bool isAlternateReturn_{false};  // whether expr is a "*label" number
   bool isPassedObject_{false};
+  common::Intent dummyIntent_{common::Intent::Default};
 };
 
 using ActualArguments = std::vector<std::optional<ActualArgument>>;
--- a/lib/semantics/check-do.cc
+++ b/lib/semantics/check-do.cc
@@ -902,6 +902,52 @@ void DoChecker::Leave(const parser::AssignmentStmt &stmt) {
   context_.CheckDoVarRedefine(variable);
 }
 
+// Check to see if a DO variable is being passed as an actual argument to a
+// dummy argument whose intent is OUT or INOUT.  To do this, we need to find
+// the expressions for actual arguments which contain DO variables.  We get the
+// intents of the dummy arguments from the ProcedureRef in the "typedCall"
+// field of the CallStmt which was filled in during expression checking.  At
+// the same time, we need to iterate over the parser::Expr versions of the
+// actual arguments to get their source locations of the arguments for the
+// messages.
+void DoChecker::Leave(const parser::CallStmt &callStmt) {
+  if (const auto &typedCall{callStmt.typedCall}) {
+    const auto &parsedArgs{
+        std::get<std::list<parser::ActualArgSpec>>(callStmt.v.t)};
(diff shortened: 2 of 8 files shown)


lib/evaluate/call.h
```diff
@@ -111,6 +111,11 @@ class ActualArgument {
   void set_isPassedObject(bool yes = true) { isPassedObject_ = yes; }
 
   bool Matches(const characteristics::DummyArgument &) const;
+  common::Intent dummyIntent() const { return dummyIntent_; }
+  ActualArgument &set_dummyIntent(common::Intent intent) {
```
vjayathirtha-nv: why mix camelCase and underscores?

vjayathirtha-nv: I see other instances of this. Is this done intentionally?

klausler: `set_...` signifies a mutator.

psteinfeld: The data member is called "dummyIntent_" and I wanted a mutator for it.  Our style guide says that mutators should start with "set_" followed by the name of the data member without the underscore.  The result is this mix of camelCase and underscores.

vjayathirtha-nv: Got it.

lib/semantics/check-do.cc
```diff
@@ -902,6 +902,52 @@ void DoChecker::Leave(const parser::AssignmentStmt &stmt) {
...
+    for (const auto &checkedOptionalArg : checkedArgs) {
+      if (parsedArgIter == parsedArgs.end()) {
+        break;  // No more parsed arguments, we're done.
+      }
+      const auto &parsedArg{std::get<parser::ActualArg>(parsedArgIter->t)};
+      ++parsedArgIter;
+      if (checkedOptionalArg) {
+        const evaluate::ActualArgument &checkedArg{*checkedOptionalArg};
+        const SomeExpr *checkedExpr{checkedArg.UnwrapExpr()};
+        if (evaluate::IsVariable(checkedExpr)) {
```
klausler: `evaluate::UnwrapWholeSymbolDataRef()` would be better, if you just need a tool that determines whether a typed Expr is nothing more that a reference to a single Symbol.

psteinfeld: Thanks, Peter.  I'll change it.

lib/semantics/semantics.cc
```diff
@@ -203,34 +203,41 @@ void SemanticsContext::PopConstruct() {
...
-  const parser::CharBlock doLoc{GetDoVariableLocation(variable)};
-  CHECK(doLoc != parser::CharBlock{});
-  Say(location, "Cannot redefine DO variable '%s'"_err_en_US, variable.name())
-      .Attach(doLoc, "Enclosing DO construct"_en_US);
-}
-
-void SemanticsContext::CheckDoVarRedefine(
-    const Symbol &variable, const parser::CharBlock &location) {
+void SemanticsContext::CheckDoVarRedefine(const parser::CharBlock &location,
+    const Symbol &variable, const parser::MessageFixedText message) {
```
tskeith: Slightly better to use `parser::MessageFixedText &&message` and then `std::move` it where it is used.

lib/semantics/semantics.cc
```diff
@@ -203,34 +203,41 @@ void SemanticsContext::PopConstruct() {
...
   if (const Symbol * root{GetAssociationRoot(variable)}) {
     if (IsActiveDoVariable(*root)) {
-      SayDoVarRedefine(location, *root);
+      const parser::CharBlock doLoc{GetDoVariableLocation(*root)};
```
tskeith: We don't normally declare local variables `const`.

lib/semantics/semantics.cc
psteinfeld: Thanks, Tim.  I'll change this.

lib/semantics/semantics.cc
psteinfeld: Thanks, Tim.  I'll change my recent declarations of local variables.

PR #904, opened 2020-01-03
Add analyzed form of pointer assignment

tskeith:
Add `typedAssignment` to `PointerAssignmentStmt` parse tree node and
extend `evaluate::Assignment` to include pointer assignment, including
analyzed bounds. Analyze pointer assignments and fill those in.
Emit them in unparsed output and parse tree dump when present.

Change assignment checking to use analyzed expressions and assignments
rather than calling AnalyzeExpr. Check bounds in pointer assignments
for impure function calls in FORALL context.

Add `Fold` convenience function to `ExpressionAnalyzer`.

Fix type resolution in `ConcurrentHeader`.

### DIFF
--- a/lib/evaluate/expression.h
+++ b/lib/evaluate/expression.h
@@ -811,18 +811,27 @@ public:
 };
 
 // An assignment is either intrinsic (with lhs and rhs) or user-defined,
-// represented as a ProcedureRef.
+// represented as a ProcedureRef. A pointer assignment optionally also has
+// a bounds-spec or bounds-remapping.
 class Assignment {
 public:
-  explicit Assignment(ProcedureRef &&x) : u{std::move(x)} {}
-  Assignment(Expr<SomeType> &&lhs, Expr<SomeType> &&rhs)
-    : u{IntrinsicAssignment{std::move(lhs), std::move(rhs)}} {}
+  UNION_CONSTRUCTORS(Assignment)
   struct IntrinsicAssignment {
     Expr<SomeType> lhs;
     Expr<SomeType> rhs;
   };
-
-  std::variant<IntrinsicAssignment, ProcedureRef> u;
+  struct PointerAssignment {
+    using BoundsSpec = std::vector<Expr<SubscriptInteger>>;
+    using BoundsRemapping =
+        std::vector<std::pair<Expr<SubscriptInteger>, Expr<SubscriptInteger>>>;
+    PointerAssignment(Expr<SomeType> &&lhs, Expr<SomeType> &&rhs)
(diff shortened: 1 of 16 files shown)


lib/evaluate/expression.h
```diff
@@ -811,18 +811,32 @@ template<> class Expr<SomeType> : public ExpressionBase<SomeType> {
...
-    : u{IntrinsicAssignment{std::move(lhs), std::move(rhs)}} {}
+  UNION_CONSTRUCTORS(Assignment)
   struct IntrinsicAssignment {
     Expr<SomeType> lhs;
     Expr<SomeType> rhs;
   };
-
-  std::variant<IntrinsicAssignment, ProcedureRef> u;
+  struct PointerAssignment {
+    // all upper bounds are present or all are absent
```
klausler: The representation doesn't enforce this constraint, but it could, by splitting the data member `bounds` into `lowerBounds` and an optional `upperBounds`.

lib/semantics/assignment.cc
```diff
@@ -375,23 +374,43 @@ void AssignmentContext::Analyze(const parser::AssignmentStmt &stmt) {
...
-    // contexts does not appear on LHS
-  }
-  if (lhsExpr && rhsExpr) {
-    CheckForPureContext(*lhsExpr, *rhsExpr, rhs.source, true /* => */);
+  if (const evaluate::Assignment * asst{GetAssignment(stmt)}) {
+    const SomeExpr *lhs;
+    const SomeExpr *rhs;
+    std::visit(
+        common::visitors{
+            [&](const evaluate::Assignment::IntrinsicAssignment &x) {
```
klausler: These lambdas should return pairs of pointers, I think.  Then there's no question of `lhs` & `rhs` being left uninitialized.

lib/evaluate/expression.h
tskeith: That wouldn't enforce the constraint that if upperBounds is present it must have the same size as lower bounds.

Originally I had a variant of: nothing, a list of lower bounds, or a list of pairs. This enforces the constraints but made using it seem too complicated.

tskeith:
> The constraints and "shalls" in 10.2.2 remain to still be checked, right?

Right

lib/evaluate/expression.h
klausler: How about a similar variant but without the empty case, since an empty vector of lower bounds could serve it?

tskeith: OK, my latest commit implements it that way. Let me know what you think.

klausler: I like it, and it should be a good build-time defense against future problems.

PR #905, opened 2020-01-03
Constant folding for ANINT, NINT, FLOOR, and CEILING intrinsic functions

klausler:
Implemented by generalizing previous `AINT` evaluation to accept a rounding mode, and changing previous `ToInteger` member function template  to use the new generalized 'ToIntegerInPlace`.


klausler:
That's the inexact flag.  It probably shouldn't be getting set; will fix.

Issue #877, opened 2019-12-18
Characters given with their unicode number 


coti:
Hello,

I have tried to run the gfortran testsuite with f18.

I have recompiled f18 this morning, using llvm 9.0
```
xxxx$ git rev-parse HEAD
80c27052055fccd20b18130b510169549ff38fea

xxxx$ $F18_FC --version
GNU Fortran (GCC) 8.1.0
Copyright (C) 2018 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
```
A few tests fail. Here is the list:
```
achar_2.f90
coarray/alloc_comp_1.f90
coarray/alloc_comp_4.f90
coarray/alloc_comp_5.f90
coarray/allocate_errgmsg.f90
coarray/atomic_1.f90
coarray/atomic_2.f90
coarray/codimension.f90
coarray/coindexed_1.f90
coarray/collectives_1.f90
coarray/collectives_2.f90
coarray/collectives_3.f90
coarray/collectives_4.f90
coarray/cosubscript_1.f90
coarray/dummy_1.f90
coarray/event_1.f90
coarray/event_2.f90
coarray/event_3.f08
coarray/event_4.f08
coarray/failed_images_2.f08
coarray/get_array.f90
coarray/image_index_1.f90
coarray/image_index_2.f90
coarray/image_status_2.f08
coarray/lib_realloc_1.f90
coarray/lock_1.f90
coarray/lock_2.f90
coarray/move_alloc_1.f90
coarray/poly_run_1.f90
coarray/poly_run_2.f90
coarray/poly_run_3.f90
coarray/ptr_comp_1.f08
coarray/ptr_comp_2.f08
coarray/ptr_comp_3.f08
coarray/ptr_comp_4.f08
coarray/registering_1.f90
coarray/scalar_alloc_1.f90
coarray/scalar_alloc_2.f90
coarray/send_array.f90
coarray/send_char_array_1.f90
coarray/sendget_array.f90
coarray/stopped_images_2.f08
coarray/subobject_1.f90
coarray/sync_1.f90
coarray/this_image_1.f90
coarray/this_image_2.f90
g77/cpp5.F
literal_character_constant_1_x.F
```

[...]

PR #906, opened 2020-01-03
Another attempt to fix bug #877

klausler:
Use hexadecimal escape sequences rather than octal if the underlying Fortran compiler is not PGF90.

Avoid using escape sequences for bytes of quoted characters unless explicitly enabled or the character is a newline.

### DIFF
--- a/lib/parser/characters.cc
+++ b/lib/parser/characters.cc
@@ -15,6 +15,8 @@
 
 namespace Fortran::parser {
 
+bool useHexadecimalEscapeSequences{false};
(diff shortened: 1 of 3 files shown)


klausler:
> What about compilers other than pgi/flang and gnu? Is a command line option needed here (eventually), or is this too minor (and/or short-lived) for additional effort?

The original bug report was using gfortran, and I didn't think it worth the effort to get it right for ifort, xlf, &c. until we hear otherwise.

PR #907, opened 2020-01-04
Enable semantic analysis by default

klausler:
This change makes `-fdebug-semantics` the default setting, and adds `-fdebug-no-semantics` to turn it off.  This means that if one is running f18 with no options today to just check parsing and unparsing, you might start to see messages from semantics.

And revert the default result kind of SIZE() and related intrinsics to be default INTEGER by default, adding a new `-flarge-sizes` flag to change them to be 64-bit.  So `-Mnolargearray` will no longer be needed when testing applications that were broken by the previous nonstandard behavior.  Fix C_F_POINTER(SHAPE=) argument type checking, too.


klausler:
I will defer merging until Monday at the earliest.

kiranchandramohan:
1) With this change there is no need to explicitly enable -fdebug-semantics in tools/f18/flang.sh
2) -fdefault-integer-8 is the flag for large integers supported by flang and gfortran.

klausler:
> 1. With this change there is no need to explicitly enable -fdebug-semantics in tools/f18/flang.sh
> 2. -fdefault-integer-8 is the flag for large integers supported by flang and gfortran.

We already support `-fdefault-integer-8` in the driver.  The new `-flarge-sizes` option is more limited, and only affects the default result types of size-related intrinsic functions like `SIZE()`.  `-fdefault-integer-8` implies `-flarge-sizes`.

PR #909, opened 2020-01-07
Adjust README for upstreaming to llvm.

DavidTruby:


DavidTruby:
> Also, have you checked that it works when invoked as flang?

I don't think there's actually any instructions in the README about invoking the compiler. The binary will still be called f18 unless/until we change that in the CMake files.

kiranchandramohan:
What is the executable that we want users to invoke flang? Ideally, it should be flang and not f18.

Also note that @sscalpone created a wrapper script (tools/f18/flang.sh) which seems to be installed in build/tools/f18/bin/flang. So it is possible to invoke as flang.

kiranchandramohan:
Note: The invocation is there in Overview.md in the documentation directory.
Overview.md:**Command:** `f18 -E src.f90` dumps the cooked character stream
Overview.md:  - `f18 -fdebug-dump-parse-tree -fparse-only src.f90` dumps the parse tree
Overview.md:  - `f18 -funparse src.f90` converts the parse tree to normalized Fortran
Overview.md:**Command:** `f18 -fdebug-dump-symbols -fparse-only src.f90` dumps the

I guess the invocation as f18/flang can be a separate PR if needed.

kiranchandramohan:
@klausler @DavidTruby : Would something like the following be OK?
Flang is a ground-up implementation of a Fortran front end written in modern C++. It started off as the f18 project (https://github.com/flang-compiler/f18) at Nvidia with an aim to replace the old flang project (https://github.com/flang-compiler/flang) and address its various deficiencies.  F18 was subsequently accepted as the Fortran frontend of LLVM and rechristened as Flang.

DavidTruby:
@kiranchandramohan that text sounds fine to me, if @klausler is ok with it I can change the PR to include that.

README.md
```diff
@@ -47,6 +45,8 @@ The code has been compiled and tested with
 clang version 7.0 and 8.0
 using either GNU's libstdc++ or LLVM's libc++.
 
+
+## Building Flang out of tree
```
RichBarton-Arm, line 49: I think I would put a little intro here saying what these instructions are for and why you might need to follow them. Probably also recommend building in-tree as far easier.

README.md
```diff
@@ -1,30 +1,22 @@
...
-
-F18 is a ground-up implementation of a Fortran front end written in modern C++.
-F18, when combined with LLVM, is intended to replace the Flang compiler.
-
-Flang is a Fortran compiler targeting LLVM.
-Visit the [Flang wiki](https://github.com/flang-compiler/flang/wiki)
-for more information about Flang.
+Flang is a ground-up implementation of a Fortran front end written in modern
+C++. It started off as the f18 project (https://github.com/flang-compiler/f18)
+at Nvidia with an aim to replace the old flang project
```
RichBarton-Arm, line 5: I would say "previous flang project" :-)

README.md
```diff
@@ -62,21 +45,26 @@ The code has been compiled and tested with
 clang version 7.0 and 8.0
 using either GNU's libstdc++ or LLVM's libc++.
 
-### LLVM dependency
 
-F18 uses components from LLVM.
+## Building Flang out of tree
+These instructions are for building Flang separately from LLVM; if you are
+building Flang alongside LLVM then follow the standard LLVM build instructions
```
RichBarton-Arm, line 51: I think you need this statement in the section above about Building Flang in-tree too.

DavidTruby: I think in general if you're building in-tree you're not going to read this README and will just follow the normal build instructions? I can change this though, where would you put it?

README.md
sscalpone: No need to call out NVIDIA, but if you do, please also include the NNSA.

README.md
```diff
@@ -1,30 +1,22 @@
...
-F18, when combined with LLVM, is intended to replace the Flang compiler.
-
-Flang is a Fortran compiler targeting LLVM.
-Visit the [Flang wiki](https://github.com/flang-compiler/flang/wiki)
-for more information about Flang.
+Flang is a ground-up implementation of a Fortran front end written in modern
+C++. It started off as the f18 project (https://github.com/flang-compiler/f18)
+at Nvidia with an aim to replace the old flang project
+(https://github.com/flang-compiler/flang) and address its various deficiencies.
+F18 was subsequently accepted as the Fortran frontend of LLVM and rechristened
```
sscalpone, line 7: I'd say "...accepted into the LLVM project and rechristened..."  because there's some blur when you get to FIR.

README.md
```diff
@@ -194,29 +128,28 @@ add
 to the cmake command.
 Release builds execute quickly.
 
-### Build F18
+### Build Flang
```
sscalpone, line 131: Do we want to note that this is the out-of-tree build?

README.md
DavidTruby: Done!

README.md
DavidTruby: I've changed it to not call out anyone in particular, let me know if the new wording is ok!

README.md
DavidTruby: Done!

RichBarton-Arm:
Just had a thought while looking over the merging checklist. We should put a few words in here about where to submit issues. During previous emails about this I proposed we keep F18 repository github 'issues' open for this purpose. Can we add a line or two near the top of this README to that effect, including the URL, etc?

klausler:
> Just had a thought while looking over the merging checklist. We should put a few words in here about where to submit issues. During previous emails about this I proposed we keep F18 repository github 'issues' open for this purpose. Can we add a line or two near the top of this README to that effect, including the URL, etc?

The file should link to a bug tracker, but I don't think that the f18 github repository is the right place unless the fixes to the "issues" are pull requests to the same repository.  Why can't the normal llvm bug tracking site be used for bugs being reported and fixed by the community?

RichBarton-Arm:
I don’t think there is any reason why it can’t be used.

My reasoning behind proposing F18 bugzilla issues is

  *   llvm are discussing a move from bugzilla to github issues<https://mail.google.com/mail/u/0/?zx=wwm6yj2noi7o#search/label%3Adevlist+github+issues%22/FMfcgxwDrtxxqzZcnsgKvSCfGpKfhsvf> so we would be aligning with the direction of travel in that regard (no ETA on that move though.)
  *   It would also be 0 cost to set up.
  *   I think the LLVM Bugzilla is not as nice to use as github issues, but that is personal preference.

All weak reasons though – I don’t think there is a compelling reason to chose either approach. Do you have one?

klausler:
> I don’t think there is any reason why it can’t be used. My reasoning behind proposing F18 issues is * llvm are discussing a move from bugzilla to github issues<https://mail.google.com/mail/u/0/?zx=wwm6yj2noi7o#search/label%3Adevlist+github+issues%22/FMfcgxwDrtxxqzZcnsgKvSCfGpKfhsvf> so we would be aligning with the direction of travel in that regard (no ETA on that move though.) * It would also be 0 cost to set up. * I think the LLVM Bugzilla is not as nice to use as github issues, but that is personal preference. All weak reasons though – I don’t think there is a compelling reason to chose either approach. Do you have one?

Yes, and it's in my original comment.  It would be weird and confusing to have bug reporting and tracking activity proceeding on a github repository while fixes and reviews are taking place elsewhere.  LLVM has bug tracking and code review systems for LLVM.  Use them.

DavidTruby:
I think there's some merit in being ahead of the curve here; LLVM is planning on moving to github issues imminently, and it seems to me odd that we would set up and use bugzilla for 2 weeks and then switch back to github issues.
However, I wouldn't suggest using this repository for issues, but rather the llvm-project one. That is already open for github issues and is what MLIR uses I believe (MLIR never made the move to bugzilla for the same reason as I'm suggesting).
Since LLVM _currently_ has two ways to submit issues and is planning on closing one down in the imminent future, I don't see the harm in suggesting that people submit issues on the other one (i.e. that they should submit issues against github.com/llvm/llvm-project).

kiranchandramohan:
@DavidTruby  Github issues in llvm-project is temporarily disabled until the switch from Bugzilla to github (soon?).
http://lists.llvm.org/pipermail/llvm-dev/2020-March/140129.html

DavidTruby:
Oh ok, my mistake! In that case I agree that we should do what the rest of the project does and use Bugzilla until the rest of the project moves. I have no idea how to set that up on the bugzilla side though.

DavidTruby:
I've sent an email to the bugzilla admins asking if we can be added on there and what the plan is for the github issues transition; I'll report back on that and update the readme as appropriate.

klausler:
> I've sent an email to the bugzilla admins asking if we can be added on there and what the plan is for the github issues transition; I'll report back on that and update the readme as appropriate.

Thank you.

RichBarton-Arm:
> I've sent an email to the bugzilla admins asking if we can be added on there and what the plan is for the github issues transition; I'll report back on that and update the readme as appropriate.

SGTM

DavidTruby:
I've posted about the bugzilla to flang-dev as I think it is relevant to more people than will see this PR, you can read the mail here: http://lists.llvm.org/pipermail/flang-dev/2020-April/000269.html

DavidTruby:
Resolved during upstream merge.

PR #910, opened 2020-01-07
Fold more intrinsic functions

klausler:
`INDEX`, `SCAN`, `VERIFY`, `LEN_TRIM`, and `REPEAT`.


klausler:
> Everything builds and tests OK for me. I don't understand most of the changes.
> 
> One thing I noticed is that both GitHub and my system think that .../test/evaluate/folding05.f90 is a binary file rather than a text file. This is different from the other folding*.f90 files in the directory. You might want to change that.

That's not new.  That test file contains non-ASCII characters encoded in UTF-8.

PR #914, opened 2020-01-08
Start, stop, and error routines for runtime

klausler:
Define basic runtime routines for main program start, `STOP`, `ERROR STOP`, and `FAIL IMAGE`, and runtime internal image termination.

### DIFF
--- a/runtime/main.cc
+++ b/runtime/main.cc
@@ -1,0 +1,35 @@
+//===-- runtime/main.cc -----------------------------------------*- C++ -*-===//
+//
+// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//----------------------------------------------------------------------------//
+
+#include "main.h"
+#include "terminator.h"
+#include <cfenv>
+#include <cstdlib>
+
+namespace Fortran::runtime {
+int argc;
+const char **argv;
+const char **envp;
+}
+
+extern "C" {
+
+void __FortranProgram();  // PROGRAM statement
+
+int main(int argc, const char *argv[], const char *envp[]) {
+  Fortran::runtime::argc = argc;
+  Fortran::runtime::argv = argv;
+  Fortran::runtime::envp = envp;
+  std::feclearexcept(FE_ALL_EXCEPT);
+  std::fesetround(FE_TONEAREST);
+  std::atexit(Fortran::runtime::NotifyOtherImagesOfNormalEnd);
+  // TODO: Runtime configuration settings from environment
+  __FortranProgram();
+  return EXIT_SUCCESS;
+}
+}
--- a/runtime/main.h
+++ b/runtime/main.h
@@ -1,0 +1,18 @@
+//===-- runtime/main.cc -----------------------------------------*- C++ -*-===//
+//
+// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//----------------------------------------------------------------------------//
+
+#ifndef FORTRAN_RUNTIME_MAIN_H_
+#define FORTRAN_RUNTIME_MAIN_H_
+
+namespace Fortran::runtime {
+extern int argc;
+extern const char **argv;
+extern const char **envp;
+}
+
+#endif  // FORTRAN_RUNTIME_MAIN_H_
--- a/runtime/stop.h
+++ b/runtime/stop.h
@@ -1,0 +1,27 @@
+//===-- runtime/stop.h ------------------------------------------*- C++ -*-===//
+//
+// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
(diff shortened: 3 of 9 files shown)


runtime/terminator.h
```diff
@@ -0,0 +1,42 @@
...
+#define FORTRAN_RUNTIME_TERMINATOR_H_
+
+#include "entry-names.h"
+#include <cstdarg>
+
+namespace Fortran::runtime {
+
+// A mixin class for statement-specific image error termination
+// for errors detected in the runtime library
+class Terminator {
```
jeanPerier: Who will own/create terminator objects, the runtime or the user program ? For I/O for instance, will it be stored in the `Cookie` in an opaque way to the user program ?

runtime/stop.h
```diff
@@ -0,0 +1,28 @@
...
+#include "c-or-cpp.h"
+#include "entry-names.h"
+#include <stdlib.h>
+
+EXTERN_C_BEGIN
+
+// Program-initiated image stop
+NORETURN void RTNAME(ProgramEndStatement)(NO_ARGUMENTS);
+NORETURN void RTNAME(StopStatement)(int code DEFAULT_VALUE(EXIT_SUCCESS),
+    bool isErrorStop DEFAULT_VALUE(false), bool quiet DEFAULT_VALUE(false));
```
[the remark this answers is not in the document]
klausler: Thanks

runtime/start.cc
```diff
@@ -0,0 +1,20 @@
...
+#include "terminator.h"
+#include <cfenv>
+#include <cstdlib>
+
+extern "C" {
+void RTNAME(ProgramStart)() {
+  std::feclearexcept(FE_ALL_EXCEPT);
+  std::fesetround(FE_TONEAREST);
+  std::atexit(Fortran::runtime::NotifyOtherImagesOfNormalEnd);
+}
```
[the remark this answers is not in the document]
klausler: I think that `main` has to be in the runtime in order to capture arguments and environment.

runtime/terminator.h
klausler: This is just for the runtime.  Instances will be part of per-I/O-statement state.

runtime/format.h
```diff
@@ -0,0 +1,83 @@
...
+  // the FORMAT string.
+  static int GetMaxParenthesisNesting(const char *format, std::size_t formatLength);
+
+  // For attempting to allocate in a user-supplied stack area
+  static std::size_t GetNeededSize(int maxHeight) {
+    return sizeof(Format) - sizeof(Iteration) * (maxMaxHeight - maxHeight);
+  }
+
+  void SetFlags(enum Flags);
+  void SetRounding(Rounding rounding) { rounding_ = rounding; }
```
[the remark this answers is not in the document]
klausler: `format.h` and `format.cc` are works in progress and were added to this PR by accident; they'll be removed before merging.

runtime/main.h
```diff
@@ -0,0 +1,18 @@
...
+//
+//----------------------------------------------------------------------------//
+
+#ifndef FORTRAN_RUNTIME_MAIN_H_
+#define FORTRAN_RUNTIME_MAIN_H_
+
+namespace Fortran::runtime {
+extern int argc;
+extern const char **argv;
+extern const char **envp;
```
jeanPerier: Since these global are `extern` here, think you need to declare them in `main.cc` so that they live somewhere. Also, maybe this file could be called "globals.h" instead of "main.h", though maybe that would be an incentive to create more globals...

runtime/main.cc
```diff
@@ -0,0 +1,29 @@
...
+int main(int argc, const char *argv[], const char *envp[]) {
+  Fortran::runtime::argc = argc;
+  Fortran::runtime::argv = argv;
+  Fortran::runtime::envp = envp;
+  std::feclearexcept(FE_ALL_EXCEPT);
+  std::fesetround(FE_TONEAREST);
+  std::atexit(Fortran::runtime::NotifyOtherImagesOfNormalEnd);
+  // TODO: Runtime configuration settings from environment
+  __FortranProgram();
+  return EXIT_SUCCESS;
```
jeanPerier: Technically this return should never be hit because `ProgramEndStatement` should be called by `__FortranProgram` right ?

klausler: It might be more convenient to have the Fortran program return to `main`, I think.

vdonaldson: Is a user PROGRAM main, SUBROUTINE main, bind(C) main, or other use of "main" mangled or otherwise accounted for to avoid conflict?

klausler: > Is a user PROGRAM main, SUBROUTINE main, bind(C) main, or other use of "main" mangled or otherwise accounted for to avoid conflict?

Yes, yes, no, and yes, I think.

vdonaldson:
Is there any issue with the use of varargs here in the context of gpu code?

klausler:
> Is there any issue with the use of varargs here in the context of gpu code?

I'll confirm that it's ok.

klausler:
> Is there any issue with the use of varargs here in the context of gpu code?

I'm told that varargs doesn't work, but then there's https://stackoverflow.com/questions/59192971/can-i-use-vararg-functions-in-cuda-device-side-code
so I guess I don't know.

I think I'll leave it as is for now, avoid the use of varargs for any other routines, and reengineer `Crash()` later if it turns out to be a problem in the future.

PR #916, opened 2020-01-09
Outline operator<< for CharBlock.

DavidTruby:
This fixes an issue where the Dump function definitions in dump.cc
were relying on the forward declaration of operator<< for CharBlock
which was marked inline and only present in char-block.h.

This is not allowed under section 6.2.10 of the C++17 standard, and
caused a compilation failure when building with clang 9 as this was
(validly) inlining every use of the function and therefore not generating 
an outlined definition for linking.

### DIFF
--- a/lib/parser/char-block.cc
+++ b/lib/parser/char-block.cc
@@ -1,0 +1,18 @@
+//===-- lib/parser/char-block.cc --------------------------------*- C++ -*-===//
+//
+// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//----------------------------------------------------------------------------//
+
+#include "char-block.h"
+#include <ostream>
+
+namespace Fortran::parser {
+
+std::ostream &operator<<(std::ostream &os, const CharBlock &x) {
+  return os << x.ToString();
+}
+
+}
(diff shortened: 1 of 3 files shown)


lib/parser/char-block.cc
```diff
@@ -0,0 +1,18 @@
+//===-- lib/parser/char-block.cc --------------------------------*- C++ -*-===//
+//
+// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//----------------------------------------------------------------------------//
+
+#include "char-block.h"
+#include <ostream>
```
kiranchandramohan: Is ostream required here? Isn't it available through iosfwd in char-block.h? Alternatively, iosfwd can be removed from char-block.h.

DavidTruby: Oddly, ostream _is_ necessary here. I wasn't able to work out why, but it doesn't work with just isofwd which also surprised me. It seems that isofwd is missing the forward declaration of operator<<(std::string)... at least in libstdc++.
I can remove isofwd from char-block.h though, I don't think that's necessary anymore.

kiranchandramohan: OK. if it is not needed anymore in char-block.h then remove isofwd.

DavidTruby:
@kiranchandramohan on reflection iosfwd is still needed in the header, as the header still has to have a forward declaration of operator<< and needs those symbols there

kiranchandramohan:
> @kiranchandramohan on reflection iosfwd is still needed in the header, as the header still has to have a forward declaration of operator<< and needs those symbols there

Thanks for looking into this. No change needed then.

Issue #912, opened 2020-01-08
Misleading explanation of -flarge-sizes


jme52:
The current Extensions.md file states that https://github.com/flang-compiler/f18/blob/4f737d4cf684fe84abaa3b44c70187f1f8b53072/documentation/Extensions.md#L118-L126

  1. This is incomplete, at best.  According to the standard, all the intrinsic functions named in the paragraph (`SIZE`, `LBOUND`, `UBOUND`, `SHAPE`,  `FINDLOC`, `MAXLOC`, and `MINLOC`) accept an optional, integer `KIND` argument: if `KIND` is present, the kind type parameter of the result is that specified by the value of `KIND`; otherwise the kind type parameter of the result is that of default integer type.

      It should also be noted that all these functions are related to array sizes and indexes of existing arrays, so they can only return values bigger than 2<sup>31</sup>-1 if integers of more than 32 bits have been used somewhere else in the code.

      This paragraph should be rewritten to be more accurate.

  2. The paragraph is in the "Extensions, deletions, and legacy features supported by default" section, so it seems like the "extension" is enabled by default, but then the explanation makes it conditional to the use of the `-flarge-sizes` flag.  If `-flarge-sizes` is not enabled by default, this paragraph should be in the next section ("Extensions supported when enabled by options").  If `-flarge-sizes` is enabled by default, this paragraph should say so clearly, and inform that it can be disabled with the `-fno-large-sizes` option.

klausler:
Does pull request #917 satisfy your concerns?

jme52:
I think that's indeed better, thank you.

jme52:
By the way, as far as I can see in 4f737d4c the `-flarge-sizes` and `-fno-large-sizes` flags are not explicitly listed in the output of `f18 --help`, maybe they should also be added there (with their explanation)?

klausler:
The f18 driver is throwaway code.  Arm is working on the production driver (see #759 & #762).

PR #917, opened 2020-01-09
Address complaints in bug #912

klausler:


klausler:
The bug submitter is satisfied so I just need an approval.

PR #918, opened 2020-01-09
Checks for DO variables passed to INTENT(OUT) dummies in functions

psteinfeld:
I added a traveral framework to find actual arguments in expressions.
For arguments that are DO variables being passed to dummy arguments with
INTENT(OUT), I emit an error message.  For INTENT(INOUT), I emit a
warning.

### DIFF
--- a/lib/semantics/check-do.cc
+++ b/lib/semantics/check-do.cc
@@ -14,11 +14,20 @@
 #include "tools.h"
 #include "type.h"
 #include "../common/template.h"
+#include "../evaluate/call.h"
 #include "../evaluate/expression.h"
 #include "../evaluate/tools.h"
 #include "../parser/message.h"
 #include "../parser/parse-tree-visitor.h"
 #include "../parser/tools.h"
+
+namespace Fortran::evaluate {
+using ActualArgumentRef = common::Reference<const ActualArgument>;
+
+inline bool operator<(ActualArgumentRef x, ActualArgumentRef y) {
+  return &*x < &*y;
+}
+}
 
 namespace Fortran::semantics {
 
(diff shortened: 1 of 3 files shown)


lib/semantics/check-do.cc
```diff
@@ -833,6 +834,50 @@ void DoChecker::Leave(const parser::ConnectSpec &connectSpec) {
...
+};
+
+template<typename A> ActualArgumentSet CollectActualArguments(const A &x) {
+  return CollectActualArgumentsHelper{}(x);
+}
+
+template ActualArgumentSet CollectActualArguments(const SomeExpr &);
+
+void DoChecker::Leave(const parser::Expr &parserExpr) {
+  if (parserExpr.typedExpr) {
```
tskeith: This check should be unnecessary after PR #921.

lib/semantics/check-do.cc
```diff
@@ -833,6 +834,50 @@ void DoChecker::Leave(const parser::ConnectSpec &connectSpec) {
...
+    if (const SomeExpr * expr{GetExpr(parserExpr)}) {
+      ActualArgumentSet argSet{CollectActualArguments(*expr)};
+      for (const evaluate::ActualArgumentRef &argRef : argSet) {
+        if (const SomeExpr * argExpr{argRef->UnwrapExpr()}) {
+          if (evaluate::IsVariable(*argExpr)) {
+            if (const Symbol * symbol{evaluate::GetFirstSymbol(*argExpr)}) {
```
kiranchandramohan: Can this be written as a helper function/lambda returning a symbol for an argExpr which satisfies some conditions?

lib/semantics/check-do.cc
```diff
@@ -833,6 +834,50 @@ void DoChecker::Leave(const parser::ConnectSpec &connectSpec) {
...
+              switch (argRef->dummyIntent()) {
```
kiranchandramohan: Would this be better as an if statement?

Since argRef is available before the three preceding if statements can we consider hoisting this up or putting a continue for the negation (INTENT(IN) or default intent)?

lib/evaluate/call.h
```diff
@@ -137,6 +137,11 @@ class ActualArgument {
 
 using ActualArguments = std::vector<std::optional<ActualArgument>>;
 
+using ActualArgumentRef = common::Reference<const ActualArgument>;
+inline bool operator<(ActualArgumentRef x, ActualArgumentRef y) {
+  return &*x < &*y;
+}
```
kiranchandramohan: for std::set<evaluate::ActualArgumentRef>? If there is only one use should we move this to the place where the set is defined?

lib/semantics/check-do.cc
psteinfeld: Thanks, Kiran.  In fact, there's already a function (UnwrapWholeSymbolDataRef) that returns a Symbol for an expression that's a variable.  I'll use that function here.

psteinfeld: In fact, it looks like I can share a lot of the code between the checking for CallStmt and Expr nodes.

lib/semantics/check-do.cc
psteinfeld: Good idea.  That'll make the code a little faster for most cases.

lib/evaluate/call.h
psteinfeld: Good idea.  I had to flail a little to get the namespace for the operator correct, but I finally figured it out.

PR #919, opened 2020-01-09
Dodge build problem in some Power environments

klausler:
Looks like some APIs from `#include <cfenv>` are macros, not `std::` entry points, in some Power build environments, so make their usage conditional in some new code that I merged earlier today.


sscalpone:
You could try, before the first #include, 
```
#if defined __GNUC__
#define __NO_MATH_INLINES
#endif
```

PR #921, opened 2020-01-09
Perform expression analysis on assumed-type dummies

### LINKED ISSUE #915: CHECK fails on call to Fortran::semantics::GetExpr
Here's the Fortran source:
subroutine s1()

  type(*)  :: x
  y = x

end subroutine s1

The compiler generates the following parse tree for the expression "x" in the assignment statement:
| | | Expr -> Designator -> DataRef -> Name = 'x'

I added the following code to the compiler:

void DoChecker::Leave(const parser::Expr &parserExpr) {
    if (const auto *expr{GetExpr(parserExpr)}) {

The call to "GetExpr()" fails in the following CHECK in lib/semantics/tools.h:

struct GetExprHelper {
  const SomeExpr *Get(const parser::Expr::TypedExpr &x) {
    CHECK(x);
    return x && x->v ? &*x->v : nullptr;
  }
I can avoid this problem by testing parserExpr.typedExpr before making the call to GetExpr(), but I'm concerned that there's some other potential problem that I'll be covering over by doing this.

tskeith:
When we recognized an expression as an assumed-type dummy argument
we did not otherwise analyze it, so `typedExpr` was not filled in.
When expression check is complete we expect every `typedExpr` data
member to have a value.

The fix is always to analyze the expression and then detect the
assumed-type dummy argument.

Fixes #915.

### DIFF
--- a/lib/semantics/expression.cc
+++ b/lib/semantics/expression.cc
@@ -2061,19 +2061,21 @@ MaybeExpr ExpressionAnalyzer::Analyze(const parser::Expr::PercentLoc &x) {
   // Represent %LOC() exactly as if it had been a call to the LOC() extension
   // intrinsic function.
   // Use the actual source for the name of the call for error reporting.
-  std::optional<ActualArgument> arg;
-  if (const Symbol * assumedTypeDummy{AssumedTypeDummy(x.v.value())}) {
-    arg = ActualArgument{ActualArgument::AssumedType{*assumedTypeDummy}};
-  } else if (MaybeExpr argExpr{Analyze(x.v.value())}) {
-    arg = ActualArgument{std::move(*argExpr)};
-  } else {
+  MaybeExpr argExpr{Analyze(x.v.value())};
+  if (!argExpr) {
     return std::nullopt;
+  }
+  ActualArguments args;
+  if (const Symbol * assumedTypeDummy{AssumedTypeDummy(x.v.value())}) {
+    args.emplace_back(ActualArgument::AssumedType{*assumedTypeDummy});
+  } else {
+    args.emplace_back(std::move(*argExpr));
   }
   parser::CharBlock at{GetContextualMessages().at()};
   CHECK(at.size() >= 4);
   parser::CharBlock loc{at.begin() + 1, 3};
   CHECK(loc == "loc");
-  return MakeFunctionRef(loc, ActualArguments{std::move(*arg)});
+  return MakeFunctionRef(loc, std::move(args));
 }
 
 MaybeExpr ExpressionAnalyzer::Analyze(const parser::Expr::DefinedUnary &x) {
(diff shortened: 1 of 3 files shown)


tskeith:
> do you know what's coming back from these calls to `Analyze`?

It's a Designator containing a Symbol for the dummy argument.

klausler:
> > do you know what's coming back from these calls to `Analyze`?
> 
> It's a Designator containing a Symbol for the dummy argument.

It has to be a `std::optional<Expr<SomeType>>`, I think, wrapping (eventually) a `Designator<T>` for some Fortran type `T`, and what concerns me is that there's no Fortran type `T` that can stand for an assumed-type dummy argument.

tskeith:
> It has to be a `std::optional<Expr<SomeType>>`, I think, wrapping (eventually) a `Designator<T>` for some Fortran type `T`, and what concerns me is that there's no Fortran type `T` that can stand for an assumed-type dummy argument.

The `Expr<SomeType>` contains an `Expr<SomeDerived>` which contains a `Designator<SomeDerived>` which contains a `SymbolRef`. It's `DynamicType` is a representation of `TYPE(*)`.

If we don't analyze these expressions we need to decide when it means when `typedExpr` is not set. Currently it means that expression checking hasn't happened yet or it's the kind of expression that doesn't get analyzed (i.e. an assumed-type dummy).

psteinfeld:
> 
> 
> These changes look OK. I built a compiler with these changes and with the contents of PR #918 with the check for the null typedExpr. Everything built successfully, which was an improvement. But tests resolve40.f90 and resolve49.f90 both failed at the same CHECK in GetExprHelper when calling GetExpr() in lib/semantics/tools.h. I haven't yet isolated test cases for them.

It looks to me like the cases that are failing involve expressions that are erroneous, and the typedExpr isn't being set for them because expression analysis found errors.

klausler:
> > It has to be a `std::optional<Expr<SomeType>>`, I think, wrapping (eventually) a `Designator<T>` for some Fortran type `T`, and what concerns me is that there's no Fortran type `T` that can stand for an assumed-type dummy argument.
> 
> The `Expr<SomeType>` contains an `Expr<SomeDerived>` which contains a `Designator<SomeDerived>` which contains a `SymbolRef`. It's `DynamicType` is a representation of `TYPE(*)`.
> 
> If we don't analyze these expressions we need to decide when it means when `typedExpr` is not set. Currently it means that expression checking hasn't happened yet or it's the kind of expression that doesn't get analyzed (i.e. an assumed-type dummy).

Or that expression semantic analysis failed due to an error, too, I think.

We could extend the typed expression representation to add another item to the set of untyped entities that `Expr<SomeType>` can hold that can show up in expressions but that don't really have types (e.g., procedure pointers, BOZ literals).  An untyped `SymbolRef` would serve.  And then of course any code that deals with a top-level `Expr<SomeType>` will have to be extended to cope with an assumed-type dummy argument.  That's what I was hoping to avoid, since the only context where an assumed-type dummy can appear is as an actual argument, so I special-cased that situation in the `ActualArgument` structure.  Or we could leave the `typedExpr` pointer null here, and adjust the code that got to the failing `CHECK` to not do so.  I'm not sure which is most clear and/or robust; do you have a preference?

I don't know why the current expression analyzer returned a `Designator<SomeDerived>`; it probably should have crashed instead.

tskeith:
My concern is that if we don't have an analyzed expression for every non-erroneous `parser::Expr`, we can't count on using analyzed expressions for semantic analysis.

For example:
```
subroutine s(x)
  type(*) :: x
  y = x
end
```
Suppose we get to the AssignmentStmt and look at the `typedExpr` for the RHS. It's not set so we assume there was an expression error and don't worry about checking the assignment any further.

klausler:
> My concern is that if we don't have an analyzed expression for every non-erroneous `parser::Expr`, we can't count on using analyzed expressions for semantic analysis.
> 
> For example:
> 
> ```
> subroutine s(x)
>   type(*) :: x
>   y = x
> end
> ```
> 
> Suppose we get to the AssignmentStmt and look at the `typedExpr` for the RHS. It's not set so we assume there was an expression error and don't worry about checking the assignment any further.

In this case, expression analysis *should* have failed with a fatal error.

tskeith:
> In this case, expression analysis _should_ have failed with a fatal error.

OK, I'll try to implement a different fix that does that.

PR #923, opened 2020-01-10
Detect incorrect use of assumed-type dummy arguments

### LINKED ISSUE #915: CHECK fails on call to Fortran::semantics::GetExpr

tskeith:
Assumed-type dummy arguments can only be used as actual arguments. If
they are used in other contexts it is an error. Change argument analysis
to handle these differently depending on the context. `allowAssumedType`
is set when the argument can be assumed-type. These expressions now all
get `typedExpr` set to `nullopt`.

Change `AnalyzeSectionSubscripts` to analyze all of the subscripts
even if one has an error. This ensures they all get analyzed expressions
(or `nullopt` in case of error).

Fix a bug analyzing `BoundsRemapping`: the lower bound was analyzed
twice and the upper bound not at all.

These change mean that `typedExpr` is set in all known cases.
Fixes #915.

### DIFF
--- a/lib/semantics/expression.cc
+++ b/lib/semantics/expression.cc
@@ -130,9 +130,11 @@ common::IfNoLvalue<MaybeExpr, WRAPPED> TypedWrapper(
 
 class ArgumentAnalyzer {
 public:
-  explicit ArgumentAnalyzer(ExpressionAnalyzer &context) : context_{context} {}
-  ArgumentAnalyzer(ExpressionAnalyzer &context, parser::CharBlock source)
-    : context_{context}, source_{source} {}
+  explicit ArgumentAnalyzer(ExpressionAnalyzer &context)
+    : context_{context}, allowAssumedType_{false} {}
+  ArgumentAnalyzer(ExpressionAnalyzer &context, parser::CharBlock source,
+      bool allowAssumedType = false)
+    : context_{context}, source_{source}, allowAssumedType_{allowAssumedType} {}
   bool fatalErrors() const { return fatalErrors_; }
   ActualArguments &&GetActuals() {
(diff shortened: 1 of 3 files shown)


lib/semantics/expression.cc
```diff
@@ -134,6 +134,7 @@ class ArgumentAnalyzer {
   ArgumentAnalyzer(ExpressionAnalyzer &context, parser::CharBlock source)
     : context_{context}, source_{source} {}
   bool fatalErrors() const { return fatalErrors_; }
+  void set_allowAssumedType(bool x = true) { allowAssumedType_ = x; }
```
klausler: This flag appears to be only set after construction in the two cases of function and subroutine references, and never cleared, so it could be specified with an additional Boolean argument to the second constructor, and also made a `const` data member.

tskeith: OK, good idea.

PR #925, opened 2020-01-13
RFC: example implementation of combined constructs

DavidTruby:
This is a possible way of implementing combined constructs without repetition, by outlining the set of accepted clauses for each construct as a static constexpr variable that can be reused.
If other people think this is an acceptable way to move forward then I will implement all the constructs this way and push that as one PR with comments explaining what this does, so don't merge this PR.

PS. ignore the char-block changes, those are in a separate PR but I need them to be able to build on my system.

### DIFF
--- a/lib/semantics/check-omp-structure.cc
+++ b/lib/semantics/check-omp-structure.cc
@@ -56,6 +56,17 @@ static constexpr OmpDirectiveSet targetSet{OmpDirective::TARGET,
     OmpDirective::TARGET_TEAMS_DISTRIBUTE_PARALLEL_DO_SIMD,
     OmpDirective::TARGET_TEAMS_DISTRIBUTE_SIMD};
 
+static constexpr inline OmpClauseSet targetAllowedClauses{OmpClause::IF,
+    OmpClause::PRIVATE, OmpClause::FIRSTPRIVATE, OmpClause::MAP,
+    OmpClause::IS_DEVICE_PTR, OmpClause::DEPEND};
+static constexpr inline OmpClauseSet targetAllowedOnceClauses{
+    OmpClause::DEVICE, OmpClause::DEFAULTMAP, OmpClause::NOWAIT};
+
+static constexpr inline OmpClauseSet teamsAllowedClauses{OmpClause::PRIVATE,
+    OmpClause::FIRSTPRIVATE, OmpClause::SHARED, OmpClause::REDUCTION};
+static constexpr inline OmpClauseSet teamsAllowedOnceClauses{
+    OmpClause::NUM_TEAMS, OmpClause::THREAD_LIMIT, OmpClause::DEFAULT};
+
 std::string OmpStructureChecker::ContextDirectiveAsFortran() {
   auto dir{EnumToString(GetContext().directive)};
   std::replace(dir.begin(), dir.end(), '_', ' ');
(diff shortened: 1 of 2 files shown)


DavidTruby:
I don't know enough about tablegen to know whether that'd be a good approach to be honest. I did it this way as it doesn't require a lot of change to the current code.

As regards additional restrictions (e.g. forbid branching out) we can also outline those into functions that are called in both places too. 

One can't necessarily just add together the restrictions for combined constructs, as I believe some restrictions contradict each other (and the spec says which ones you should allow for each combined construct).

lib/semantics/check-omp-structure.cc
```diff
@@ -56,6 +56,17 @@ static constexpr OmpDirectiveSet targetSet{OmpDirective::TARGET,
     OmpDirective::TARGET_TEAMS_DISTRIBUTE_PARALLEL_DO_SIMD,
     OmpDirective::TARGET_TEAMS_DISTRIBUTE_SIMD};
 
+static constexpr inline OmpClauseSet targetAllowedClauses{OmpClause::IF,
```
ichoyjx: They're already sets of clauses, I think we can drop the `clauses` from variable name: `targetAllowed`.

ichoyjx:
David and I discussed this method and we think it's handy to utilize the feature of `EnumSet`, we could do something like `parallelAllowedOnce + doAllowedOnce - NoWait` for `parallel do` directive.

lib/semantics/check-omp-structure.cc
sscalpone: I prefer the name to says what it is, so I prefer targetAllowedClauses.

sscalpone:
Did you consider organizing the information by clause, where the entries form the list of constructs where the clause is required, allowed once, or allowed multiple times.  Such a table could also encode information about the clause parameters as well.

DavidTruby:
The reason I chose to do it this way rather than using a table (which I assume would be a std::map or similar?) is that these static inline constexpr variables can be resolved at compile time and without writing a separate constructor.
In order to construct a table presumably we would need to either build it in the constructor of this class, which would mean that it would run each time a new instance is created, or it would need to be a static variable with a non-trivial constructor, something which is forbidden by LLVM.
I agree that a table/map would be nicer conceptually and would avoid introducing so many new variables, but for these reasons I thought doing it this way would be better.
I hope that explains the rationale?

klausler:
> The reason I chose to do it this way rather than using a table (which I assume would be a std::map or similar?) is that these static inline constexpr variables can be resolved at compile time and without writing a separate constructor.
> In order to construct a table presumably we would need to either build it in the constructor of this class, which would mean that it would run each time a new instance is created, or it would need to be a static variable with a non-trivial constructor, something which is forbidden by LLVM.
> I agree that a table/map would be nicer conceptually and would avoid introducing so many new variables, but for these reasons I thought doing it this way would be better.
> I hope that explains the rationale?

If you need to, you can create a data structure on its first reference.  But I prefer the constexprs.

DavidTruby:
@sscalpone are you happy for me to go ahead with implementing these this way?

sscalpone:
You are closer to the implementation than I am, so my advice is fairly generic:  Organizing the data into one or two tables may be easier to understand, modify, use, and maintain. The model I have in mind is the tables and code used for intrinsics.
```
static const IntrinsicInterface genericIntrinsicFunction[]
```
and related in `lib/evaluate/intrinsics.cc`.

I just want to make sure you've consider that kind of organization.  It is proving to be *very* handy for intrinsics -- concise, clear, and easy to maintain.

DavidTruby:
The difference between here and the intrinsics tables is that the way we need to access the clause sets is quite different here. In this case if we used an array such as is used there, we would end up with a lot of magic numbers accessing the fields since we aren't iterating over the fields but rather need random access of them. So in this particular case I think just having separate variables is a little clearer.

lib/semantics/check-omp-structure.cc
klausler: Why are these variables declared `inline`?

DavidTruby: I wanted to signal that I didn't really want these variables to be present in the generated binary, however after checking the standard I realise a) that's not how inline variables work and b) constexpr implies inline anyway. I'll remove this when I do the actual implementation.

PR #926, opened 2020-01-13
BLOCK DATA semantics

klausler:
Check semantics on `BLOCK DATA` subprograms.

### DIFF
--- a/lib/semantics/check-declarations.cc
+++ b/lib/semantics/check-declarations.cc
@@ -17,7 +17,6 @@
 #include "../evaluate/check-expression.h"
 #include "../evaluate/fold.h"
 #include "../evaluate/tools.h"
-#include <algorithm>
 
 namespace Fortran::semantics {
 
@@ -925,6 +924,10 @@ void CheckHelper::CheckEquivalenceSet(const EquivalenceSet &) {
 
 void CheckHelper::CheckBlockData(const Scope &scope) {
   // BLOCK DATA subprograms should contain only named common blocks.
+  // C1415 presents a list of statements that shouldn't appear in
+  // BLOCK DATA, but so long as the subprogram contains no executable
+  // code and allocates no storage outside named COMMON, we're happy
+  // (e.g., an ENUM is strictly not allowed).
   for (const auto &pair : scope) {
     const Symbol &symbol{*pair.second};
     if (!(symbol.has<CommonBlockDetails>() || symbol.has<UseDetails>() ||
--- a/lib/semantics/check-declarations.cc
+++ b/lib/semantics/check-declarations.cc
@@ -17,6 +17,7 @@
 #include "../evaluate/check-expression.h"
 #include "../evaluate/fold.h"
 #include "../evaluate/tools.h"
+#include <algorithm>
 
 namespace Fortran::semantics {
 
@@ -69,6 +70,9 @@ private:
   bool CheckDefinedAssignmentArg(const Symbol &, const DummyArgument &, int);
   void CheckSpecificsAreDistinguishable(
       const Symbol &, const GenericDetails &, const std::vector<Procedure> &);
+  void CheckEquivalenceSet(const EquivalenceSet &);
+  void CheckBlockData(const Scope &);
+
   void SayNotDistinguishable(
       const SourceName &, GenericKind, const Symbol &, const Symbol &);
   bool InPure() const {
@@ -341,6 +345,18 @@ void CheckHelper::CheckObjectEntity(
         messages_.Say(
             "non-POINTER dummy argument of pure subroutine must have INTENT() or VALUE attribute"_err_en_US);
       }
+    }
+  }
+  if (symbol.owner().kind() != Scope::Kind::DerivedType &&
+      IsInitialized(symbol)) {
+    if (details.commonBlock()) {
+      if (details.commonBlock()->name().empty()) {
+        messages_.Say(
+            "A variable in blank COMMON should not be initialized"_en_US);
+      }
+    } else if (symbol.owner().kind() == Scope::Kind::BlockData) {
+      messages_.Say(
+          "An initialized variable in BLOCK DATA must be in a COMMON block"_err_en_US);
     }
   }
 }
@@ -889,11 +905,38 @@ void CheckHelper::Check(const Scope &scope) {
   } else if (scope.IsDerivedType()) {
     return;  // PDT instantiations have null symbol()
   }
+  for (const auto &set : scope.equivalenceSets()) {
+    CheckEquivalenceSet(set);
+  }
   for (const auto &pair : scope) {
     Check(*pair.second);
   }
   for (const Scope &child : scope.children()) {
     Check(child);
+  }
+  if (scope.kind() == Scope::Kind::BlockData) {
+    CheckBlockData(scope);
+  }
+}
+
+void CheckHelper::CheckEquivalenceSet(const EquivalenceSet &) {
+  // TODO: Move C8106 (&al.) checks here from resolve-names-utils.cc
+}
+
+void CheckHelper::CheckBlockData(const Scope &scope) {
+  // BLOCK DATA subprograms should contain only named common blocks.
+  for (const auto &pair : scope) {
+    const Symbol &symbol{*pair.second};
+    if (!(symbol.has<CommonBlockDetails>() || symbol.has<UseDetails>() ||
+            symbol.has<UseErrorDetails>() || symbol.has<DerivedTypeDetails>() ||
+            symbol.has<SubprogramDetails>() ||
+            symbol.has<ObjectEntityDetails>() ||
+            (symbol.has<ProcEntityDetails>() &&
+                !symbol.attrs().test(Attr::POINTER)))) {
+      messages_.Say(symbol.name(),
+          "'%s' may not appear in a BLOCK DATA subprogram"_err_en_US,
+          symbol.name());
+    }
   }
 }
 
--- a/lib/semantics/check-declarations.cc
+++ b/lib/semantics/check-declarations.cc
@@ -935,7 +935,7 @@ void CheckHelper::CheckEquivalenceSet(const EquivalenceSet &set) {
       if (&object != &*iter) {
         if (auto *details{object.symbol.detailsIf<ObjectEntityDetails>()}) {
           if (details->commonBlock()) {
-            if (details->commonBlock() != &commonBlock) {
+            if (details->commonBlock() != &commonBlock) {  // 8.10.3 paragraph 1
               if (auto *msg{messages_.Say(object.symbol.name(),
                       "Two objects in the same EQUIVALENCE set may not be members of distinct COMMON blocks"_err_en_US)}) {
                 msg->Attach(iter->symbol.name(),
@@ -950,7 +950,8 @@ void CheckHelper::CheckEquivalenceSet(const EquivalenceSet &set) {
             }
           } else {
             // Mark all symbols in the equivalence set with the same COMMON
-            // block
+            // block to prevent spurious error messages about initialization
+            // in BLOCK DATA outside COMMON
             details->set_commonBlock(commonBlock);
           }
         }
--- a/lib/semantics/check-declarations.cc
+++ b/lib/semantics/check-declarations.cc
@@ -354,8 +354,13 @@ void CheckHelper::CheckObjectEntity(
             "A variable in blank COMMON should not be initialized"_en_US);
       }
     } else if (symbol.owner().kind() == Scope::Kind::BlockData) {
-      messages_.Say(
-          "An initialized variable in BLOCK DATA must be in a COMMON block"_err_en_US);
+      if (IsAllocatable(symbol)) {
+        messages_.Say(
+            "An ALLOCATABLE variable may not appear in a BLOCK DATA subprogram"_err_en_US);
+      } else {
+        messages_.Say(
+            "An initialized variable in BLOCK DATA must be in a COMMON block"_err_en_US);
+      }
     }
   }
 }
--- a/lib/semantics/check-declarations.cc
+++ b/lib/semantics/check-declarations.cc
@@ -17,6 +17,7 @@
 #include "../evaluate/check-expression.h"
 #include "../evaluate/fold.h"
 #include "../evaluate/tools.h"
+#include <algorithm>
 
 namespace Fortran::semantics {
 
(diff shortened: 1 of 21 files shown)


lib/semantics/check-declarations.cc
```diff
@@ -889,12 +905,39 @@ void CheckHelper::Check(const Scope &scope) {
...
+    const Symbol &symbol{*pair.second};
+    if (!(symbol.has<CommonBlockDetails>() || symbol.has<UseDetails>() ||
+            symbol.has<UseErrorDetails>() || symbol.has<DerivedTypeDetails>() ||
+            symbol.has<SubprogramDetails>() ||
+            symbol.has<ObjectEntityDetails>() ||
+            (symbol.has<ProcEntityDetails>() &&
+                !symbol.attrs().test(Attr::POINTER)))) {
+      messages_.Say(symbol.name(),
+          "'%s' may not appear in a BLOCK DATA subprogram"_err_en_US,
+          symbol.name());
```
tskeith: There are a bunch of declaration-constructs that are prohibited by C1415 and I don't think they can all be detected by looking at the symbols. import-stmt and access-stmt are caught in `resolve-names.cc` but there are more like that.

lib/semantics/check-return.cc
```diff
@@ -33,18 +30,17 @@ void ReturnStmtChecker::Leave(const parser::ReturnStmt &returnStmt) {
...
-  if (!subprogramScope) {
-    context_.Say(
-        "RETURN must in the inclusive scope of a SUBPROGRAM"_err_en_US);
-    return;
-  }
-  if (returnStmt.v && subprogramScope->kind() == Scope::Kind::Subprogram) {
-    if (IsFunction(*subprogramScope->GetSymbol())) {
-      context_.Say(
-          "RETURN with expression is only allowed in SUBROUTINE subprogram"_err_en_US);
+  if (const auto *subprogramScope{FindContainingSubprogram(scope)}) {
```
CarolineConcatto: Which Block Data clause is this?

lib/semantics/resolve-names.cc
```diff
@@ -2790,6 +2791,22 @@ Symbol &SubprogramVisitor::PushSubprogramScope(
   return *symbol;
 }
```
CarolineConcatto: It would be nice if we could know what clause of the standard are you implementing

test/semantics/block-data01.f90
```diff
@@ -0,0 +1,13 @@
+block data foo
+  real :: pi = asin(-1.0) ! ok
```
CarolineConcatto: Would be nice again to write the clauses you are testing here.

lib/semantics/check-declarations.cc
```diff
@@ -17,6 +17,7 @@
 #include "../evaluate/check-expression.h"
 #include "../evaluate/fold.h"
 #include "../evaluate/tools.h"
+#include <algorithm>
```
tskeith: Is this needed?

lib/semantics/resolve-names.cc
```diff
@@ -5044,6 +5070,9 @@ bool ResolveNamesVisitor::Pre(const parser::ImportStmt &x) {
       return false;
     }
     break;
```
CarolineConcatto: It would be nice if you add which part of the standard are you implementing!
It is not clause C1614.

lib/semantics/check-return.cc
```diff
@@ -33,18 +30,17 @@ void ReturnStmtChecker::Leave(const parser::ReturnStmt &returnStmt) {
...
+    if (returnStmt.v && subprogramScope->kind() == Scope::Kind::Subprogram) {
+      if (IsFunction(*subprogramScope->GetSymbol())) {
+        context_.Say(
+            "RETURN with expression is only allowed in SUBROUTINE subprogram"_err_en_US);
+      }
     }
+  } else {
+    context_.Say(
+        "RETURN must be in the inclusive scope of a subprogram"_err_en_US);
```
psteinfeld: I didn't see a test case for this condition.

lib/semantics/check-declarations.cc
klausler: Not any more; thanks.  I was using `std::find_if()` for a while.

lib/semantics/check-declarations.cc
klausler: So long as nothing in the `BLOCK DATA` generates any code or any allocates any storage not in `COMMON`, I don't think that we should care.

lib/semantics/resolve-names.cc
```diff
@@ -2790,6 +2791,22 @@ Symbol &SubprogramVisitor::PushSubprogramScope(
   return *symbol;
 }
 
+Symbol &SubprogramVisitor::PushBlockDataScope(const parser::Name &name) {
```
psteinfeld: This function is only called from one place that doesn't use its return value.  Should it just return void?

lib/semantics/check-declarations.cc
tskeith: I think we want, at least optionally, a strict standard-checking mode. But maybe it's not very high priority.

lib/semantics/resolve-names.cc
```diff
@@ -5044,6 +5070,9 @@ bool ResolveNamesVisitor::Pre(const parser::ImportStmt &x) {
       return false;
     }
     break;
+  case Scope::Kind::BlockData:
+    Say("IMPORT is not allowed in a BLOCK DATA subprogram"_err_en_US);
```
psteinfeld: I didn't see a test that produces this message.

lib/semantics/check-return.cc
klausler: It is not a `BLOCK DATA` clause.  It is cleaning up some code that I noticed could be improved after implementing changes to `FindProgramUnitContaining()` and searching for all of the sites in the code that should use it.  The `FindContainingSubprogram()` function in `lib/semantics/check-return.cc` could be written in terms of `FindProgramUnitContaining()` and reduce its tight coupling with internals of the scope tree, so I did so.

lib/semantics/resolve-names.cc
klausler: `BLOCK DATA` is in clause 14.3.

test/semantics/block-data01.f90
klausler: Done.

lib/semantics/resolve-names.cc
klausler: There is no constraint C1614.  The exclusion of `IMPORT` from `BLOCK DATA` is implied by C1415.  I added a comment.

lib/semantics/check-return.cc
CarolineConcatto: Ok

lib/semantics/resolve-names.cc
```diff
@@ -2790,6 +2791,22 @@ Symbol &SubprogramVisitor::PushSubprogramScope(
...
 
+Symbol &SubprogramVisitor::PushBlockDataScope(const parser::Name &name) {
+  if (auto *prev{FindSymbol(name)}) {
+    if (prev->attrs().test(Attr::EXTERNAL) && prev->has<ProcEntityDetails>()) {
+      if (prev->test(Symbol::Flag::Subroutine) ||
+          prev->test(Symbol::Flag::Function)) {
+        Say2(name, "BLOCK DATA '%s' has been called"_err_en_US, *prev,
+            "Previous call of '%s'"_en_US);
+      }
+      EraseSymbol(name);
```
CarolineConcatto: Are you erasing the symbol because EXTERNAL cannot be in BlockData? If yes, are we printing the error message?

lib/semantics/resolve-names.cc
CarolineConcatto: Ops! Sorry. I mean C1416.
A type declaration statement in  ablock-data specification-part shall not contain ALLOCATABLE, EXTERNAL,or BIND attribute specifiers.
But, yes I remember that IMPORT cannot be inside BLOCK DATA. If I am correct it is in another part of the standard

lib/semantics/check-return.cc
klausler: There should have already been one, but I couldn't find it, so I'll add one.

klausler: Or I'll delete the message, since it could only be produced in a program already eliciting parsing errors about an executable part in a `MODULE`, `SUBMODULE`, or `BLOCK DATA` program unit.

lib/semantics/resolve-names.cc
klausler: Sure.

lib/semantics/resolve-names.cc
CarolineConcatto: Ok, I am wrong, this is the implementation of:
There shall not be more than one unnamed block data program unit in a program.

klausler:
> I believe that would be nice if you could point which parts of the standard were implemented.
> I cannot see C1414, being implemented. Should this be implemented in resolve-names.cc?
> Am I wrong or the clause C1415 was not fully implemented?
> What about the same common block name not being called in a another Block Data?

C1414 was already implemented in `resolve-labels.cc`.
C1415 is implemented indirectly in `check-declarations.cc`.
The requirement that the same common block not be initialized in multiple subprograms must be checked at link time.

lib/semantics/check-return.cc
tskeith: What about in a `PROGRAM`? Are we allowing that as an extension?

klausler: GNU Fortran, PGI, and XLF accept it.  Intel Fortran and NAG do not.  So we probably should.

DavidTruby:
It looks from the CI like this is regressing a unit test (`altreturn05.f90`)
I haven't looked in to why this is, is it an expected regression?

lib/semantics/check-declarations.cc
sscalpone: If there's a todo here, please add a TODO.

CarolineConcatto: Yes, I believe it would be nice to have a TODO of what is not yet implemented

I have tried this small code with block data and it is passing. Is that correct? Shouldn't the compiler print an error when using External and BIND?

block data  bd1_c1416
    integer, external :: beta
end block data bd1_c1416

block data bd2_c1416
  integer, bind(C) :: x
end block data bd2_c1416

In my mind this could be implemented inside resolve-names.cc in by checking for 
(Attr::EXTERNAL) and (Attr::BIND_C) when inside BlockData no?

lib/semantics/check-declarations.cc
```diff
@@ -343,6 +347,18 @@ void CheckHelper::CheckObjectEntity(
...
     }
   }
+  if (symbol.owner().kind() != Scope::Kind::DerivedType &&
+      IsInitialized(symbol)) {
+    if (details.commonBlock()) {
+      if (details.commonBlock()->name().empty()) {
+        messages_.Say(
+            "A variable in blank COMMON should not be initialized"_en_US);
+      }
+    } else if (symbol.owner().kind() == Scope::Kind::BlockData) {
```
CarolineConcatto: Is the error message that the compiler prints important?

When I run this test:
block data test
  integer, allocatable :: t1
end block data
It prints the message:
An initialized variable in BLOCK DATA must be in a COMMON block

The compiler is behaving correctly by printing an error message. 
But is this an initialization of the variable? I guess you could say yes because it dynamically allocates space for the array.
But allocatable is also in the clause C1416. 
Gfortran prints:
ALLOCATABLE attribute not allowed in BLOCK DATA program unit at
Is that a way to make the message more clear?

With the current message the user would try to solve by adding in the common, and it will not work because allocatable is not allowed in common.

lib/semantics/check-return.cc
```diff
@@ -33,18 +30,16 @@ void ReturnStmtChecker::Leave(const parser::ReturnStmt &returnStmt) {
...
-  if (!subprogramScope) {
-    context_.Say(
-        "RETURN must in the inclusive scope of a SUBPROGRAM"_err_en_US);
-    return;
-  }
-  if (returnStmt.v && subprogramScope->kind() == Scope::Kind::Subprogram) {
-    if (IsFunction(*subprogramScope->GetSymbol())) {
+  if (const auto *subprogramScope{FindContainingSubprogram(scope)}) {
+    if (returnStmt.v &&
+        (subprogramScope->kind() == Scope::Kind::MainProgram ||
```
CarolineConcatto: Could you add here also for Block Data the same message.
Is it possible to test  Scope::Kind::BlockData?
When I run this test
block data bdpointer
  common /test/ t1
  return t1
end block data
It shows the standard message that I did something wrong, but nothing specific as saying return only in a subroutine.

lib/semantics/resolve-names.cc
```diff
@@ -2790,6 +2791,25 @@ Symbol &SubprogramVisitor::PushSubprogramScope(
...
+    }
+  }
+  if (name.source.empty()) {
+    // Don't let unnamed BLOCK DATA conflict with unnamed PROGRAM
+    PushScope(Scope::Kind::BlockData, nullptr);
```
CarolineConcatto: According to the standard:
There shall not be more than one unnamed block data program unit in a program.
But this test is working:
block data
  common /test/ t1
end block data

block data
  integer x
end block data

lib/semantics/tools.cc
```diff
@@ -617,6 +618,28 @@ bool CanBeTypeBoundProc(const Symbol *symbol) {
   }
 }
 
+bool IsInitialized(const Symbol &symbol) {
```
CarolineConcatto: I don't know if IsInitialized that is not getting when DATA is being used or if is the check in check-declarations.cc.
But this test:
BLOCK DATA DBAXES
      !BLOCK DATA element ‘ifail’ at (1) must be in COMMON
      Integer ifail
      COMMON /CBAXES/I,J,K
      DATA I,J,K,ifail /1,2,3,4/
END block data DBAXES
Is passing, when it shouldn't, correct?

klausler: We haven't implemented `DATA` statement semantics yet, but I can catch this particular case.

lib/semantics/resolve-names.cc
klausler: This "shall" requirement in the standard is something that can't be checked until link time.

lib/semantics/check-declarations.cc
klausler: `ALLOCATABLE` variables are initialized to the unallocated state.

I will improve the message.

klausler:
> It looks from the CI like this is regressing a unit test (`altreturn05.f90`)
> I haven't looked in to why this is, is it an expected regression?
> It seems like that test got added to the default run a couple of commits ago.

It is a new test, and it passes for my builds here.

lib/semantics/check-return.cc
klausler: A `BLOCK DATA` subprogram cannot contain any executable statements.  This restriction is implemented in the parser.

tskeith:
> > It looks from the CI like this is regressing a unit test (`altreturn05.f90`)
> > I haven't looked in to why this is, is it an expected regression?
> > It seems like that test got added to the default run a couple of commits ago.
> 
> It is a new test, and it passes for my builds here.

I see the same problem. `altreturn05.f90` was added to the cmake file but the file doesn't exist.

klausler:
> > > It looks from the CI like this is regressing a unit test (`altreturn05.f90`)
> > > I haven't looked in to why this is, is it an expected regression?
> > > It seems like that test got added to the default run a couple of commits ago.
> > 
> > 
> > It is a new test, and it passes for my builds here.
> 
> I see the same problem. `altreturn05.f90` was added to the cmake file but the file doesn't exist.

I neglected to `git add` it.

lib/semantics/tools.cc
CarolineConcatto: Nice!

lib/semantics/resolve-names.cc
CarolineConcatto: Even if they are in the same file?

lib/semantics/check-declarations.cc
CarolineConcatto: Thanks.

lib/semantics/resolve-names.cc
klausler: `SayAlreadyDeclared`  emits an error message; it doesn't check anything.

PR #927, opened 2020-01-13
Runtime FORMAT scanning

klausler:
To be driven by each data item in the io-list of a formatted I/O statement.

This new code locates the next data edit descriptor to be processed from a FORMAT string.  It maintains the state necessary to implement repeated data edit descriptors and repeated parenthesized lists of FORMAT edit descriptors.  It avoids error checking that the common FORMAT string validator performs.  As it seeks the next data edit descriptor, it recognizes and handles control edit descriptors by adjusting mutable modes (e.g. rounding) and invoking callbacks.

Still rough; I'm looking for comments on the approach more than on the implementation.

### DIFF
--- a/runtime/format.cc
+++ b/runtime/format.cc
@@ -1,0 +1,365 @@
+//===-- runtime/format.cc ---------------------------------------*- C++ -*-===//
+//
+// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
+
+#include "format.h"
+#include "../lib/common/format.h"
+#include "../lib/decimal/decimal.h"
+#include <limits>
+
+namespace Fortran::runtime {
+
+template<typename CHAR>
+FormatControl<CHAR>::FormatControl(FormatContext &context, const CHAR *format,
+    std::size_t formatLength, const MutableModes &modes, int maxHeight)
+  : context_{context}, modes_{modes}, maxHeight_{static_cast<std::uint8_t>(
+                                          maxHeight)},
+    format_{format}, formatLength_{static_cast<int>(formatLength)} {
+  // The additional two items are for the whole string and a
+  // repeated non-parenthesized edit descriptor.
+  if (maxHeight > std::numeric_limits<std::int8_t>::max()) {
+    context_.terminator.Crash(
+        "internal Fortran runtime error: maxHeight %d", maxHeight);
+  }
+  stack_[0].start = offset_;
+  stack_[0].remaining = Iteration::unlimited;  // 13.4(8)
+}
+
+template<typename CHAR>
+int FormatControl<CHAR>::GetMaxParenthesisNesting(
+    Terminator &terminator, const CHAR *format, std::size_t formatLength) {
+  using Validator = common::FormatValidator<CHAR>;
+  typename Validator::Reporter reporter{
+      [&](const common::FormatMessage &message) {
+        terminator.Crash(message.text, message.arg);
+        return false;  // crashes on error above
+      }};
+  Validator validator{format, formatLength, reporter};
+  validator.Check();
+  return validator.maxNesting();
+}
+
+static void HandleCharacterLiteral(
+    FormatContext &context, const char *str, std::size_t chars) {
+  if (context.handleCharacterLiteral1) {
+    context.handleCharacterLiteral1(str, chars);
+  }
+}
+
+static void HandleCharacterLiteral(
+    FormatContext &context, const char16_t *str, std::size_t chars) {
+  if (context.handleCharacterLiteral2) {
+    context.handleCharacterLiteral2(str, chars);
+  }
+}
+
+static void HandleCharacterLiteral(
+    FormatContext &context, const char32_t *str, std::size_t chars) {
+  if (context.handleCharacterLiteral4) {
+    context.handleCharacterLiteral4(str, chars);
+  }
+}
+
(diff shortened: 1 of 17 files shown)


runtime/format.cc
```diff
@@ -0,0 +1,330 @@
...
+//===----------------------------------------------------------------------===//
+
+#include "format.h"
+#include "../lib/common/format.h"
+#include "../lib/decimal/decimal.h"
+#include <limits>
+
+namespace Fortran::runtime {
+
+FormatControl::FormatControl(Context &context, const char *format,
```
vdonaldson: Do some of these functions need to be templatized to handle char(2) and char(4) cases?  If yes, I suppose that could come later.

runtime/format.cc
```diff
@@ -0,0 +1,330 @@
...
+
+void FormatControl::SkipBlanks() {
+  while (offset_ < formatLength_ && format_[offset_] == ' ') {
+    ++offset_;
+  }
+}
+
+char FormatControl::GetNextChar() {
+  if (offset_ >= formatLength_) {
+    context_.terminator.Crash("FORMAT missing at least one ')'");
```
vdonaldson: Might this message look strange if it ever gets generated?  (Or is that the expectation, that it's a fallback that should probably never be generated?)

runtime/format.h
```diff
@@ -0,0 +1,108 @@
...
+private:
+  static constexpr std::uint8_t maxMaxHeight{100};
+
+  struct Iteration {
+    static constexpr int unlimited{-1};
+    int start{0};  // offset in format_ of '(' or a repeated edit descriptor
+    int remaining{0};  // while >0, decrement and iterate
+  };
+
+  char PeekNext() { return offset_ < formatLength_ ? format_[offset_] : '\0'; }
```
vdonaldson: Does PeekNext need to account for intervening white space?

runtime/format.cc
klausler: It's what will come out if format control crashes out of the end of the string, which it should do only if there's an insufficiency of `)` characters to close out the format and any nested parenthesized groups.

runtime/format.h
klausler: No, that's a responsibility of its callers.

runtime/format.cc
klausler: I should do it now to ensure that it can be done.

runtime/format.h
vdonaldson: Does the call in GetIntField do that?  At least a couple of cases have SkipBlanks predecessors that wouldn't be needed if PeekNext were able to skip.  (Assuming that's not problematic.)

runtime/format.h
```diff
@@ -0,0 +1,108 @@
...
+};
+
+// A single edit descriptor extracted from a FORMAT
+struct DataEdit {
+  char descriptor;  // capitalized: one of A, I, B, O, Z, F, E(N/S/X), D, G
+  char variation{'\0'};  // N, S, or X for EN, ES, EX
+  int width;  // the 'w' field
+  std::optional<int> digits;  // the 'm' or 'd' field
+  std::optional<int> expoDigits;  // 'Ee' field
+  MutableModes modes;
```
jeanPerier: While I agree keeping `MutableModes` inside `DataEdit`  makes passing data structure easier, I think it gives a "live" aspect to `DataEdit` that would make storing `DataEdit` in an attempt to prepocess formats weird. I would keep them separate.
That being said, I think storing `DataEdit` does not really make sens with the current approach because bigger changes would be needed to expose the iterations to the user, so that may just be fine to keep it there if you think it is better.

runtime/format.h
klausler: No, it doesn't.  Do you think that it should?  I should either completely ignore white space outside character literals and Hollerith, or just allow for it between descriptors and punctuation.  Not sure what the best policy would be.

runtime/format.h
klausler: These modes are the ones that can be changed within a FORMAT string, so either each data edit descriptor has to hold a copy of the current modes, or the FORMAT processor will have to interact with the context to ensure that the context always has the right modes.  Since the representation is small, associating the modes with the data edit descriptor makes the most sense to me.

klausler:
> ```
> SUBROUTINE process(stream, array, n)
>   CHARACTER(len=n) :: stream
>   CHARACTER(len=5) :: array(n/5)
>   READ(UNIT=stream, FMT='(*(A5))') array
> END SUBROUTINE
> ```
> 
> The most optimal handling of this example statement would probably be a single memcpy between stream and array, but with the proposed `FormatControl` (as I understand it), `GetNext` would have to be called `n/5` times.

That's right.

> I agree with starting with a "simple" utility as you did, but do you think it will be possible to extend your design to optimize this later or do you think it will never be a goal to do so ?

Special cases could be checked for, as in your example.  And maybe the API to the FORMAT control should be extended with a "repeat count", so if the caller needs 5 data edits for an array in the I/O list, FORMAT control could return a data edit descriptor with a repeat count and save some calls.

runtime/format.h
vdonaldson: Since 1 2  34 5  is just as valid as 12345, even if numeric values aren't likely to appear that way very often, that does need to be accepted.  In that case, the code needs to either make sure all PeekNext calls have an associated SkipBlanks predecessor or some other way to skip blanks.  It may be more robust to have PeekNext do that itself, but of course that isn't required.  Another option would be for FormatValidator  to squeeze blanks (could even be done in situ), but that would complicate error message generation.  My suggestion is to pick a simple, but correct solution.  I was under the impression that we weren't too concerned about repeated format scanning, even for the case Jean noticed.  For that case, a star repetition count could be represented by an appropriate MAXINT.

klausler: Sounds good to me, will do.  Thanks!

runtime/format.cc
klausler: Done!

PR #928, opened 2020-01-14
Pointer assignment semantic checks

tskeith:
See individual commits.


lib/evaluate/characteristics.cc
```diff
@@ -636,7 +638,7 @@ std::optional<Procedure> Procedure::Characterize(
           [&](const semantics::ProcEntityDetails &proc)
               -> std::optional<Procedure> {
             if (symbol.attrs().test(semantics::Attr::INTRINSIC)) {
-              return intrinsics.IsUnrestrictedSpecificIntrinsicFunction(
+              return intrinsics.IsSpecificIntrinsicFunction(
```
klausler: Could you summarize for me what's changed here?  Do you permit procedure pointers to point to restricted specific intrinsic functions?

lib/semantics/check-declarations.cc
```diff
@@ -659,6 +673,17 @@ bool CheckHelper::CheckDefinedAssignmentArg(
   return true;
 }
 
+// Report a conflicting attribute error if symbols has both of these attributes
```
klausler: s/has/have/

lib/evaluate/characteristics.cc
tskeith: The change is to be able to detect specific intrinsic functions, both restricted and unrestricted. `IsSpecificIntrinsicFunction` returns a struct that has a flag to distinguish restricted and unrestricted. That flag is propagated to `SpecificIntrinsic` in `ProcedureDesignator` by `ExpressionAnalyzer::Designate`.

The new tests in `resolve46.f90` verify that procedure pointers can't point to restricted specific intrinsic functions.

lib/semantics/check-declarations.cc
tskeith: Thanks

PR #934, opened 2020-01-15
Porting F18 tests to use LLVM lit

LukeIreland1:
These commits contain:

- Scripts that will port error tests
- Verified ported error tests
- Remaining ported error tests

The remaining code for porting the remaining generic, symbol, and modfile tests is mostly done, but needs adjusting to this format.


LukeIreland1:
Merging removed my last commit, just trying to fix this now

sscalpone:
Why not change the tests to use the standard prefix?

LukeIreland1:
> Why not change the tests to use the standard prefix?

My aim was to make minimal changes to existing tests, but that's doable, and I don't mind strongly either way.

Keeping the original prefixes may be beneficial to others, and you may be able to identify the test and/or it's objective more easily.

test-lit/lit_port.py
```diff
@@ -0,0 +1,256 @@
...
+                if token.startswith("!"):
+                    prefix = token[1:]
+                    break
+            if prefix and prefix.endswith(":") and prefix not in non_prefixes:
+                prefixes.add(prefix[:-1])
+    return prefixes
+
+
+def run(command):
+    command = "PATH=~/f18/build/install/bin:$PATH;{}".format(command)
```
kiranchandramohan: There seem to be assumptions about where the compiler and associated binaries will be. If this script is designed to be run by others then it might make sense to define this is a global variable somewhere or passed as an argument.

test-lit/lit_port.py
```diff
@@ -0,0 +1,256 @@
...
+import re
+import shlex
+import sys
+import subprocess
+from pathlib import Path
+
+from test_splitter import TestSplitter
+
+ROOT = Path.cwd()
+while ROOT.name != "f18":
```
kiranchandramohan: can this f18 name be a global variable since in future this could also be flang. 
Ignore if this is only meant to be a run once script.

test-lit/lit_port.py
```diff
@@ -0,0 +1,256 @@
...
+                            path.unlink()
+
+            port_single_test(test, output)
+            cleanup()
+    if FAILS:
+        for fail in FAILS:
+            print("Test {} failed".format(fail))
+    else:
+        print("No fails detected")
+
```
kiranchandramohan: You have used 2 new lines in between functions everywhere else.

test-lit/semantics/allocate10.f90
```diff
@@ -3,6 +3,9 @@
 !TODO: mixing expr and source-expr?
 !TODO: using subcomponent in source expressions
 
+! RUN: not %flang -fdebug-resolve-names -fparse-only %s 2>&1 | FileCheck --check-prefixes=C942a,ERROR %s
```
kiranchandramohan: what is the use of C942a prefix?

test-lit/semantics/omp-atomic.f90
```diff
@@ -2,6 +2,9 @@
 
 ! Check OpenMP 2.13.6 atomic Construct
 
+! RUN: %flang -fdebug-resolve-names -fparse-only OPTIONS: -fopenmp %s 2>&1
```
kiranchandramohan: Does this work? Should OPTIONS appear like this?

test-lit/semantics/if_stmt02.f90
```diff
@@ -0,0 +1,6 @@
+!ERROR: IF statement is not allowed in IF statement
```
kiranchandramohan: Why did this come as the first line here (and not the run line)?
Why doesn't git show this file as a renaming/movement of file (test/semantics/if_stmt02.f90)? At the end of this patch the file is shown as deleted.

test-lit/semantics/implicit05.f90
```diff
@@ -0,0 +1,7 @@
+! RUN: not %flang -fdebug-resolve-names -fparse-only %s 2>&1 | FileCheck --check-prefixes=ERROR %s
```
kiranchandramohan: why is the original file deleted and a new file created?

test-lit/semantics/resolve03.f90
```diff
@@ -0,0 +1,8 @@
+! RUN: not %flang -fdebug-resolve-names -fparse-only %s 2>&1 | FileCheck --check-prefixes=ERROR %s
```
kiranchandramohan: why is the original file deleted and a new file created?

LukeIreland1:
> Does this handle only tests from the semantics directory? If so, can that be mentioned in the commit message. What about tests in other directories?

Only error tests from the semantics directory, so I was explicit about that part, but not about it also being a semantics test. Given there are no other error tests (as far as I remember), do you think it will be okay to leave it like this? I will be explicit about the other directories, so I think it's unlikely to cause confusion.

test-lit/lit_port.py
```diff
@@ -0,0 +1,256 @@
...
+import subprocess
+from pathlib import Path
+
+from test_splitter import TestSplitter
+
+ROOT = Path.cwd()
+while ROOT.name != "f18":
+    ROOT = ROOT.parent
+DEFAULT_TEMPLATE = "! RUN: %flang {} %s 2>&1\n"
+ERROR_TEMPLATE = "! RUN: not %flang {} %s 2>&1 | FileCheck --check-prefixes={} %s\n"
```
kiranchandramohan: Is the redirection because the driver does not support "-o -" ?

test-lit/lit_port.py
```diff
@@ -0,0 +1,256 @@
...
+import sys
+import subprocess
+from pathlib import Path
+
+from test_splitter import TestSplitter
+
+ROOT = Path.cwd()
+while ROOT.name != "f18":
+    ROOT = ROOT.parent
+DEFAULT_TEMPLATE = "! RUN: %flang {} %s 2>&1\n"
```
kiranchandramohan: Is the redirection "2>&1" necessary here?

test-lit/semantics/resolve17.f90
```diff
@@ -1,3 +1,6 @@
+! RUN: not %flang -fdebug-resolve-names -fparse-only %s 2>&1 | FileCheck --check-prefixes=Note,ERROR %s
```
kiranchandramohan: Note is a comment and not a prefix to check I guess.

test-lit/semantics/resolve18.f90
```diff
@@ -1,3 +1,6 @@
+! RUN: not %flang -fdebug-resolve-names -fparse-only %s 2>&1 | FileCheck --check-prefixes=Note,ERROR %s
```
kiranchandramohan: Note is a comment and not a prefix to check I guess.

test-lit/semantics/resolve32.f90
```diff
@@ -1,3 +1,6 @@
+! RUN: not %flang -fdebug-resolve-names -fparse-only %s 2>&1 | FileCheck --check-prefixes=Note,ERROR %s
```
kiranchandramohan: Note is a comment and not a prefix to check I guess.

test-lit/semantics/equivalence01.f90
```diff
@@ -1,3 +1,6 @@
+! RUN: not %flang -fdebug-resolve-names -fparse-only %s 2>&1 | FileCheck --check-prefixes=OK,ERROR %s
```
kiranchandramohan: !OK is probably a comment and not a prefix to check?

test-lit/semantics/resolve09.f90
```diff
@@ -1,3 +1,6 @@
+! RUN: not %flang -fdebug-resolve-names -fparse-only %s 2>&1 | FileCheck --check-prefixes=OK,ERROR %s
```
kiranchandramohan: !OK is probably a comment and not a prefix to check?

test-lit/semantics/resolve22.f90
```diff
@@ -1,3 +1,6 @@
+! RUN: not %flang -fdebug-resolve-names -fparse-only %s 2>&1 | FileCheck --check-prefixes=OK,ERROR %s
```
kiranchandramohan: !OK is probably a comment and not a prefix to check?

test-lit/lit_port.py
```diff
@@ -0,0 +1,256 @@
+""" lit_port.py
```
CarolineConcatto: I imagine that as we want to merge with LLVM. This header needs to be changed to:
# -*- Python -*-
I believe you can remove the licence part. I don't see that in the python files inside LLVM. I can be wrong, so double check.

test-lit/lit_port.py
```diff
@@ -0,0 +1,256 @@
...
+
+    Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+    See https://llvm.org/LICENSE.txt for license information.
+    SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+"""
+
+import argparse
```
CarolineConcatto: Do we need all this import

test-lit/lit_port.py
```diff
@@ -0,0 +1,256 @@
...
+        command,
+        stdout=subprocess.PIPE,
+        stderr=subprocess.PIPE,
+        shell=True).stdout.decode('utf-8')
+
+
+def test_single_port(new_path, old_path):
+    test = new_path.name
+    new_path = ROOT.parent.joinpath(
+        "build-f18/test-lit", new_path.parent.name, new_path.name
```
CarolineConcatto: Same comment as Kiran, you are assuming that the user will always use build-f18, what is not always true. Can you use the CMAKE_BINARY_DIR here? Or something that is not that specific?

CarolineConcatto: Have a look in the lit.site.cfg, maybe you could use "@FLANG_BINARY_DIR@".

test-lit/semantics/altreturn01.f90
```diff
@@ -1,5 +1,8 @@
 ! Check calls with alt returns
 
+! RUN: %flang -fdebug-resolve-names -fparse-only %s 2>&1
```
CarolineConcatto: What is the reason we don't have FileCheck here?

test-lit/semantics/computed-goto01.f90
```diff
@@ -1,5 +1,8 @@
 ! Check that a basic computed goto compiles
 
+! RUN: %flang -fdebug-resolve-names -fparse-only %s 2>&1
```
CarolineConcatto: Another one without FileCheck, what is the reason?

test-lit/semantics/if_arith01.f90
```diff
@@ -1,5 +1,8 @@
 ! Check that a basic arithmetic if compiles.
 
+! RUN: %flang -fdebug-resolve-names -fparse-only %s 2>&1
```
CarolineConcatto: So these tests without FileCheck. I imagine they will always pass even if they are not doing the right thing correct? 
I don't understand them quite well, the reason they are this way.

test-lit/semantics/separate-module-procs.f90
```diff
@@ -7,6 +7,9 @@
 !===------------------------------------------------------------------------===
```
CarolineConcatto: What is this line? It comes from the old file, but any way very odd, no?

test-lit/test_splitter.py
```diff
@@ -0,0 +1,126 @@
...
+class TestSplitter:
+    def __init__(self):
+        self.error_tests = []
+        self.modfile_tests = []
+        self.symbol_tests = []
+        self.generic_tests = []
+        self.unsupported_tests = []
+        self.test_types = ['ERROR', 'MODFILE', 'SYMBOL', 'GENERIC']
+        root = os.getcwd()
+        while not root.endswith("f18"):
```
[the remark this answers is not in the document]
CarolineConcatto: I believe you could use FLANG_SRC_DIR maybe!

test-lit/test_splitter.py
```diff
@@ -0,0 +1,126 @@
+""" test_splitter.py
+
+    Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+    See https://llvm.org/LICENSE.txt for license information.
+    SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+    
+"""
+
```
CarolineConcatto: I would like if these 2 new python files had a little explanation of Why they exists and what they do.
The header I believe the way it is, is not needed

test-lit/test_splitter.py
```diff
@@ -0,0 +1,126 @@
...
+                if match:
+                    if match.group(1) == "ERROR_TESTS":
+                        current_test_set = self.error_tests
+                    elif match.group(1) == "MODFILE_TESTS":
+                        current_test_set = self.modfile_tests
+                    elif match.group(1) == "SYMBOL_TESTS":
+                        current_test_set = self.symbol_tests
+                    else:
+                        current_test_set = self.generic_tests
+                elif re.match(r".*\..*f[90|\]| \n]", line):
```
CarolineConcatto: I believe that we could have other files, no? Will this only work if the files are f.90?

test-lit/lit_port.py
LukeIreland1: After all tests are lit compatible, this script will be obsolete. The question is how likely is F18 to be renamed between now and then. I imagine while flang-compiler/flang is active, this won't happen, so we're probably okay for now.

test-lit/semantics/allocate10.f90
LukeIreland1: None, unfortunately the prefix finder picks up anything prefix-like, and these standard references (I assume), are sometimes formatted like prefixes, and not like comments (like they should be)

test-lit/semantics/allocate10.f90
```diff
@@ -3,6 +3,9 @@
...
+
+
 subroutine C939_C942a_C945b(xsrc1a, xsrc1c, xsrc0, xsrc2a, xsrc2c, pos)
 ! C939: If an allocate-object is an array, either allocate-shape-spec-list shall
```
LukeIreland1: @kiranchandramohan Notice the space between the comment character and the standard reference.

test-lit/semantics/allocate10.f90
LukeIreland1: So, I'm going to take a fairly big assumption that prefix-like string that starts with a C, followed by a number, is not a prefix. From the prefixes I've seen, I don't think this is a problem, but could need further refinement later on.

test-lit/semantics/omp-atomic.f90
LukeIreland1: I assume it is a difference in formatting with spaces again. Originally I had "only add opt if '-' in opt" , but Rich said it's too general of an assumption. Changed it to not include the opt if it has a comment character or OPTIONS in it.

test-lit/lit_port.py
LukeIreland1: Sadly, all needed. Shlex is hardly used, but is preferable to str.split().

test-lit/lit_port.py
LukeIreland1: @kiranchandramohan asked me to add the license I think, and I'm not 100% sure, but I remember Rich confirming this. I will double check though.

test-lit/lit_port.py
LukeIreland1: Trying to emulate the shell scripts used (i.e test_errors.sh) suggests yes, but I'm not 100% sure. I suspect yes, since error tests will need piping from stderr under normal circumstances, but I'm unsure if lit is able to automatically capture all output without this pipe.

test-lit/lit_port.py
LukeIreland1: I'm believe `-o` is supported, but isn't needed for these types of tests. Probably will be needed for semantics modfile tests.

test-lit/lit_port.py
```diff
@@ -0,0 +1,256 @@
...
+                    try:
+                        output.mkdir(parents=True)
+                    except PermissionError:
+                        print(
+                            "Invalid permission to create directory at given path",
+                            file=sys.stderr)
+                        exit(1)
+
+            if args.clean:
+                if list(output.glob('**/*.f*')) != 0:
```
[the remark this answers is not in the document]
LukeIreland1: Nope, thanks for finding this :)

test-lit/lit_port.py
kiranchandramohan: Sorry, my original comment is wrong since these are error tests and errors go to the error stream and needs to be redirected to the output stream before piping into filecheck so that the error message can be compared.

test-lit/lit_port.py
LukeIreland1: The other one is for binary, this is for the test files, but I have them in two different folders. I'll make it a global variable though.

test-lit/semantics/equivalence01.f90
LukeIreland1: Another prefix-like comment...I'll just have to add this one as a non-prefix.

test-lit/semantics/if_stmt02.f90
LukeIreland1: My script expects a license boilerplate or other comments at the beginning, I'll change it to detect if the insert line is after prefixes.

It has been renamed/moved, git just doesn't recognise it. It's happened to a few files, and I verified all of them. It has the same effect so it doesn't matter too much.

test-lit/semantics/implicit05.f90
LukeIreland1: See [here](https://github.com/flang-compiler/f18/pull/934#discussion_r367391638)

test-lit/test_splitter.py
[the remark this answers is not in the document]
LukeIreland1: I could use this, and the source is hard to differ between person to person, but at the moment, since I'm not sure I can use the other lit variables for binaries and test files, I think I'll keep it as minimalistic as possible. I have created a global though.

test-lit/semantics/if_arith01.f90
LukeIreland1: These tests just need to not throw any errors when ran.

test-lit/semantics/separate-module-procs.f90
LukeIreland1: It is very odd...

test-lit/test_splitter.py
LukeIreland1: I agree with the explanation, I'm not 100% sure about the license, since I was told to add it, but I'll verify it's needed, since the tests no longer have any license in them.

test-lit/test_splitter.py
LukeIreland1: The regex matches every entry in the CMakeLists.txt which is its purpose. It includes things like `  canondo*.[Ff]90`, but is globbed to get all files after matching.

CarolineConcatto:
I think the TODO can still be there, but maybe in another line, not in the same as the ERROR!

LukeIreland1:
> I think the TODO can still be there, but maybe in another line, not in the same as the ERROR!

Well the TODO invalidates the ERROR line as it is incomplete, and FileCheck will try to look for an invalid line.

CarolineConcatto:
Another solution is to add the tests that are failing in the list of Expected Failures.
But the way it is I believe is not the best.

LukeIreland1:
> Another solution is to add the tests that are failing in the list of Expected Failures.
> But the way it is I believe is not the best.

This is what I was doing originally, but Rich said it was unfair on the rest of the test that was working basically, we'll get clarification later in scrum so I'll just wait till then.

test-lit/lit_port.py
kiranchandramohan: I don't remember asking you to add license. But i can be forgetful. @RichBarton-Arm probably asked.
I don't see the license in lit.cfg.py and other python files in the test directory but i see the license in several other python files. See the grep below.
~/test_build/organic-clang$ grep Apache `find -name *.py` | wc -l
62

sscalpone:
I think the TODO lines are more about an incomplete compiler checks and less about an incomplete test. But let's not argue semantics. Those TODO lines are probably important reminders of work that needs to be completed.  We do not want to lose that information.  It's reasonable to ask the original author about their intent.

You might poll about changing ERROR to CHECK.  That'll be inline with most lit tests in llvm & leave the TODO lines unchanged.

You might also ask to eliminate the ERROR on the TODO lines.  That might meet a positive reception.

LukeIreland1:
> You might poll about changing ERROR to CHECK. That'll be inline with most lit tests in llvm & leave the TODO lines unchanged.
> 
> You might also ask to eliminate the ERROR on the TODO lines. That might meet a positive reception.

We discussed the second options in the scrum just now, but I should have brought up standard prefixes again. Another option was to rename ERROR: to simply error: or ERR: when after a TODO:

LukeIreland1:
I'm going to do a poll on the prefix situation, but I will mention it may be easier to leave them as they are, because while tests would work by replacing everything with check, the custom prefixes are useful in displaying their purpose clearly.

👍 for keeping prefixes as they are. 👎 for changing all prefixes to CHECK

LukeIreland1:
> The current "ERROR:" prefix means the message must come out as an error. How would that work if it is changed to "CHECK:"?

So the run line is "not"ed so it will only pass if the code throws an error. The prefix doesn't matter at all, you could call it whatever you want. I have tested before, but just verified again, swapping ERROR with CHECK works fine. But again, for the same reason TODO is useful to developers, so could be ERROR (easily distinguish tests from one another)

```
! RUN: not %flang -fdebug-resolve-names -fparse-only %s 2>&1 | FileCheck %s


subroutine s1
  implicit none
  !CHECK: More than one IMPLICIT NONE statement
  implicit none(type)
end subroutine

subroutine s2
  implicit none(external)
  !CHECK: More than one IMPLICIT NONE statement
  implicit none
end subroutine

```

klausler:
> Annoyingly, there's not actually much I can do here. It's attempted to port an incomplete test. I need to check with Rich if I should bother, if there are "TODO"s in there, as I previously intentionally didn't touch them, but he said it should still be able to port the rest of the test. I suppose the answer here is to remove all TODO lines from tests that aren't commented out, but really, if they are incomplete, they should be commented out.

Both of those ideas are bad.  It's not the test that incomplete, it's the feature.  Once the feature is complete, additional errors will be emitted, and the comment will help update the test once the anticipated error begins to appear.

tskeith:
> So the run line is "not"ed so it will only pass if the code throws an error. The prefix doesn't matter at all, you could call it whatever you want. I have tested before, but just verified again, swapping ERROR with CHECK works fine. But again, for the same reason TODO is useful to developers, so could be ERROR (easily distinguish tests from one another)

"ERROR:" isn't just useful for developers, it means something in the test: the message must come out as an error, not a warning or anything else. If you are proposing to change that behavior you should make that explicit so that it can be discussed.

LukeIreland1:
> If unexpected errors are reported by the compiler or if they are at the wrong line, this does not catch it.

I can't see which file this is referring to, could you help me with that?

LukeIreland1:
> > Annoyingly, there's not actually much I can do here. It's attempted to port an incomplete test. I need to check with Rich if I should bother, if there are "TODO"s in there, as I previously intentionally didn't touch them, but he said it should still be able to port the rest of the test. I suppose the answer here is to remove all TODO lines from tests that aren't commented out, but really, if they are incomplete, they should be commented out.
> 
> Both of those ideas are bad. It's not the test that incomplete, it's the feature. Once the feature is complete, additional errors will be emitted, and the comment will help update the test once the anticipated error begins to appear.

What do you suggest? At the moment I'm just changing the case of the prefix in the TODO line so it isn't checked during testing.

LukeIreland1:
> > So the run line is "not"ed so it will only pass if the code throws an error. The prefix doesn't matter at all, you could call it whatever you want. I have tested before, but just verified again, swapping ERROR with CHECK works fine. But again, for the same reason TODO is useful to developers, so could be ERROR (easily distinguish tests from one another)
> 
> "ERROR:" isn't just useful for developers, it means something in the test: the message must come out as an error, not a warning or anything else. If you are proposing to change that behavior you should make that explicit so that it can be discussed.

No, I agree, that was in response to Steve's comment about changing/standardising it. I made a poll and hopefully it's explicit enough.

kiranchandramohan:
> > If unexpected errors are reported by the compiler or if they are at the wrong line, this does not catch it.
> 
> I can't see which file this is referring to, could you help me with that?

There is an implicit line number check in test_errors.sh

sscalpone:
> "ERROR:" isn't just useful for developers

This issue solves itself once flang starts to prefix error messages with error: (along with warning:, note:, etc.).

sscalpone:
> If unexpected errors are reported by the compiler or if they are at the wrong line, this does not catch it.

If FileCheck is really unable to catch extra messages or spurious text, that's a fatal flaw.  However, flang's existing test-checking tools ought to work with lit, or be easily changed to be compatible with lit.  Lit is just a driver.

klausler:
> > "ERROR:" isn't just useful for developers
> 
> This issue solves itself once flang starts to prefix error messages with error: (along with warning:, note:, etc.).

f18 emits `error:` on error messages.

sscalpone:
> f18 emits `error:` on error messages.

Cool.  I don't know why I didn't know that.

I think this would be ok:
`  !CHECK: error: More than one IMPLICIT NONE statement`
and
`  !TODO: error: More than one IMPLICIT NONE statement`

tskeith:
> > If unexpected errors are reported by the compiler or if they are at the wrong line, this does not catch it.
> 
> I can't see which file this is referring to, could you help me with that?

Edit `test-lit/semantics/allocate01.f90` (which currently passes) and remove or move one of the "ERROR:" lines. The test should then fail but it doesn't.

tskeith:
> > If unexpected errors are reported by the compiler or if they are at the wrong line, this does not catch it.
> 
> If FileCheck is really unable to catch extra messages or spurious text, that's a fatal flaw. However, flang's existing test-checking tools ought to work with lit, or be easily changed to be compatible with lit. Lit is just a driver.

Right, `test_errors.sh` could be adapted to work with lit driving it. Probably that's what will have to happen with `test_modfile.sh` and `test_symbols.sh` anyway.

kiranchandramohan:
> > If unexpected errors are reported by the compiler or if they are at the wrong line, this does not catch it.
> 
> If FileCheck is really unable to catch extra messages or spurious text, that's a fatal flaw. However, flang's existing test-checking tools ought to work with lit, or be easily changed to be compatible with lit. Lit is just a driver.

FileCheck can do this by,
1) Having a check with count, which will fail if the number of error messages differs.
2) Having a line number along with the check.

klausler:
> > > If unexpected errors are reported by the compiler or if they are at the wrong line, this does not catch it.
> > 
> > 
> > If FileCheck is really unable to catch extra messages or spurious text, that's a fatal flaw. However, flang's existing test-checking tools ought to work with lit, or be easily changed to be compatible with lit. Lit is just a driver.
> 
> FileCheck can do this by,
> 
> 1. Having a check with count, which will fail if the number of error messages differs.
>    https://github.com/llvm-mirror/llvm/blob/master/test/FileCheck/check-count.txt
> 2. Having a line number along with the check.
>    https://github.com/llvm-mirror/llvm/blob/master/test/FileCheck/line-count-2.txt

The capability that is essential to me is being told the exact source line number associated with a missing error message *or* with an unexpected error message.

LukeIreland1:
> > > > If unexpected errors are reported by the compiler or if they are at the wrong line, this does not catch it.
> > > 
> > > 
> > > If FileCheck is really unable to catch extra messages or spurious text, that's a fatal flaw. However, flang's existing test-checking tools ought to work with lit, or be easily changed to be compatible with lit. Lit is just a driver.
> > 
> > 
> > FileCheck can do this by,
> > 
> > 1. Having a check with count, which will fail if the number of error messages differs.
> >    https://github.com/llvm-mirror/llvm/blob/master/test/FileCheck/check-count.txt
> > 2. Having a line number along with the check.
> >    https://github.com/llvm-mirror/llvm/blob/master/test/FileCheck/line-count-2.txt
> 
> The capability that is essential to me is being told the exact source line number associated with a missing error message _or_ with an unexpected error message.

Lit already provides the source line number for unexpected error messages.

sscalpone:
> Lit already provides the source line number for unexpected error messages.

How does lit know that the line is an unexpected error message vs. an uninteresting line of output?

tskeith:
I suggest that for the purposes of reviewing this you add some tests that fail so that we can see what the output looks like. For example, missing error, extra error, message at wrong line, expected error comes out as non-error.

You would remove these before merging once the  PR is approved.

LukeIreland1:
> > Lit already provides the source line number for unexpected error messages.
> 
> How does lit know that the line is an unexpected error message vs. an uninteresting line of output?

The compiler gives all the output to lit, specifies which parts are errors, and if it isn't expected via a prefix, it fails and displays this uncaught error.

kiranchandramohan:
> > > > If unexpected errors are reported by the compiler or if they are at the wrong line, this does not catch it.
> > > 
> > > 
> > > If FileCheck is really unable to catch extra messages or spurious text, that's a fatal flaw. However, flang's existing test-checking tools ought to work with lit, or be easily changed to be compatible with lit. Lit is just a driver.
> > 
> > 
> > FileCheck can do this by,
> > 
> > 1. Having a check with count, which will fail if the number of error messages differs.
> >    https://github.com/llvm-mirror/llvm/blob/master/test/FileCheck/check-count.txt
> > 2. Having a line number along with the check.
> >    https://github.com/llvm-mirror/llvm/blob/master/test/FileCheck/line-count-2.txt
> 
> The capability that is essential to me is being told the exact source line number associated with a missing error message _or_ with an unexpected error message.

We can also add check-not in between the error messages to capture any unexpected error messages.

sscalpone:
> exact source line number

No please.  That make tests hard to maintain.  The FileCheck doesn't require line numbers.  Having line numbers or counts does not help catch extra message because, as far as I know, there's no way to tell FileCheck to complain or count an arbitrary line.

>  check-not in between the error messages to capture any unexpected error messages.

Wouldn't that require a check not for all possible strings on all non-error lines?

Wait, is FileCheck smart enough to parse line numbers out of error messages?  If not, how does it match up error messages with the precise line number?

LukeIreland1:
> > > > > If unexpected errors are reported by the compiler or if they are at the wrong line, this does not catch it.
> > > > 
> > > > 
> > > > If FileCheck is really unable to catch extra messages or spurious text, that's a fatal flaw. However, flang's existing test-checking tools ought to work with lit, or be easily changed to be compatible with lit. Lit is just a driver.
> > > 
> > > 
> > > FileCheck can do this by,
> > > 
> > > 1. Having a check with count, which will fail if the number of error messages differs.
> > >    https://github.com/llvm-mirror/llvm/blob/master/test/FileCheck/check-count.txt
> > > 2. Having a line number along with the check.
> > >    https://github.com/llvm-mirror/llvm/blob/master/test/FileCheck/line-count-2.txt
> > 
> > 
> > The capability that is essential to me is being told the exact source line number associated with a missing error message _or_ with an unexpected error message.
> 
> We can also add check-not in between the error messages to capture any unexpected error messages.

I suppose if they're unexpected, then it's hard to know what to put in the check-not lines. My understanding of lit, is there should be no errors thrown in order to pass, and if there is, it has to explicitly be matched correctly to pass. That means adding code that throws a different error causes a fail, and anything but the exact right test will cause a fail, so there's no need to check for what's not there.

LukeIreland1:
> > exact source line number
> 
> No please. That make tests hard to maintain. The FileCheck doesn't require line numbers. Having line numbers or counts does not help catch extra message because, as far as I know, there's no way to tell FileCheck to complain or count an arbitrary line.

Annoyingly, I found a way to make it complain about specific lines! I can remove it if it's unnecessary, but that was in order to make @tskeith happy really. 

> > check-not in between the error messages to capture any unexpected error messages.
> 
> Wouldn't that require a check not for all possible strings on all non-error lines?
> 
> Wait, is FileCheck smart enough to parse line numbers out of error messages? If not, how does it match up error messages with the precise line number?

I'll give you a snippet of my changes.

```
  !WARNING: [[@LINE+3]]:{{[0-9]+}}:{{.*}} 'X' edit descriptor must have a positive position value
  !WARNING: [[@LINE+2]]:{{[0-9]+}}:{{.*}} Expected ',' or ')' in format expression
  !WARNING: [[@LINE+1]]:{{[0-9]+}}:{{.*}} 'X' edit descriptor must have a positive position value
  write(*,'(XX)')
```

This is from io10.f90. Yes, it checks the same warning message twice. I didn't write the original tests so no comment on that really.

klausler:
> > We can also add check-not in between the error messages to capture any unexpected error messages.
> 
> I suppose if they're unexpected, then it's hard to know what to put in the check-not lines. My understanding of lit, is there should be no errors thrown in order to pass, and if there is, it has to explicitly be matched correctly to pass. That means adding code that throws a different error causes a fail, and anything but the exact right test will cause a fail, so there's no need to check for what's not there.

That is correct.  These tests must fail if the error messages produced by the compiler do not exactly match the expected errors, and the output should clearly distinguish both the false positives and false negative cases.  (Requiring an explicit marker to catch unexpected errors would be error-prone, verbose, and also impossible; how would one specify that a line with two expected errors shouldn't elicit three or more?)

LukeIreland1:
> I suggest that for the purposes of reviewing this you add some tests that fail so that we can see what the output looks like. For example, missing error, extra error, message at wrong line, expected error comes out as non-error.
> 
> You would remove these before merging once the PR is approved.

With these changes I'll add now before I go home, it should cover everything

- Missing error: Missing ERROR: checks will cause the test to fail as is
- Extra error: Some tests actually already have duplicate error/warning checks for the same line of code. The same as the old style test, there's no way to remove them besides manually finding them and removing them. If the old test doesn't add any extra, then it won't fail. If other people add them in future, the test will fail. 
-Message at wrong line: My changes should 🤞 fix that
-Expected error comes out as non-error: My changes look at both the line number and if it's an error test, it has to have "error" in the appropriate output line

klausler:
>   !WARNING: [[@LINE+3]]:{{[0-9]+}}:{{.*}} 'X' edit descriptor must have a positive position value
>   !WARNING: [[@LINE+2]]:{{[0-9]+}}:{{.*}} Expected ',' or ')' in format expression
>   !WARNING: [[@LINE+1]]:{{[0-9]+}}:{{.*}} 'X' edit descriptor must have a positive position value
>   write(*,'(XX)')
> ```
> 
> This is from io10.f90. Yes, it checks the same warning message twice. I didn't write the original tests so no comment on that really.

The line in question deserves both warnings, since it violates the same constraint twice.

LukeIreland1:
> > !WARNING: [[@line+3]]:{{[0-9]+}}:{{._}} 'X' edit descriptor must have a positive position value
> > !WARNING: [[@line+2]]:{{[0-9]+}}:{{._}} Expected ',' or ')' in format expression
> > !WARNING: [[@line+1]]:{{[0-9]+}}:{{._}} 'X' edit descriptor must have a positive position value
> > write(_,'(XX)')
> > ```
> > 
> > This is from io10.f90. Yes, it checks the same warning message twice. I didn't write the original tests so no comment on that really.
> > ```
> 
> The line in question deserves both warnings, since it violates the same constraint twice.

My bad, didn't really look at the code, but fair enough. At least that part of my script works!

klausler:
> > I suggest that for the purposes of reviewing this you add some tests that fail so that we can see what the output looks like. For example, missing error, extra error, message at wrong line, expected error comes out as non-error.
> > You would remove these before merging once the PR is approved.
> 
> With these changes I'll add now before I go home, it should cover everything
> 
> * Missing error: Missing ERROR: checks will cause the test to fail as is
> * Extra error: Some tests actually already have duplicate error/warning checks for the same line of code. The same as the old style test, there's no way to remove them besides manually finding them and removing them. If the old test doesn't add any extra, then it won't fail. If other people add them in future, the test will fail.
>   -Message at wrong line: My changes should 🤞 fix that
>   -Expected error comes out as non-error: My changes look at both the line number and if it's an error test, it has to have "error" in the appropriate output line

It is possible for one line of Fortran to legitimately elicit the same message more than once.

LukeIreland1:
> > > I suggest that for the purposes of reviewing this you add some tests that fail so that we can see what the output looks like. For example, missing error, extra error, message at wrong line, expected error comes out as non-error.
> > > You would remove these before merging once the PR is approved.
> > 
> > 
> > With these changes I'll add now before I go home, it should cover everything
> > 
> > * Missing error: Missing ERROR: checks will cause the test to fail as is
> > * Extra error: Some tests actually already have duplicate error/warning checks for the same line of code. The same as the old style test, there's no way to remove them besides manually finding them and removing them. If the old test doesn't add any extra, then it won't fail. If other people add them in future, the test will fail.
> >   -Message at wrong line: My changes should 🤞 fix that
> >   -Expected error comes out as non-error: My changes look at both the line number and if it's an error test, it has to have "error" in the appropriate output line
> 
> It is possible for one line of Fortran to legitimately elicit the same message more than once.

Yep, misunderstanding. Extra errors shouldn't be a problem, but I still need to do more testing regarding breaking the test after porting to see if it correctly fails.

LukeIreland1:
allocate01.f90 isn't fixed, because lit seems to be funny with check lines containing identical(bar line number) strings consecutively. Will look further into this tomorrow.

tskeith:
Please don't force push while the PR is being reviewed -- your reviewers can't tell what you have changed.

After it is approved and you are ready to merge you can squash together commits that should be.

kiranchandramohan:
> > exact source line number
> 
> No please. That make tests hard to maintain. The FileCheck doesn't require line numbers. Having line numbers or counts does not help catch extra message because, as far as I know, there's no way to tell FileCheck to complain or count an arbitrary line.

All the lines with errors have the pattern ":<regex_number>:<regex_number>: error" in them. FileCheck can count the lines with that pattern. 
CHECK-COUNT-3: :<regex_number>:<regex_number>: error
> 
> > check-not in between the error messages to capture any unexpected error messages.
> 
> Wouldn't that require a check not for all possible strings on all non-error lines?

I am hoping that a single CHECK-NOT ":<regex_number>:<regex_number>: error" in between the expected error messages will catch that. It will probably bail out at the first unexpected error message. 
 
> 
> Wait, is FileCheck smart enough to parse line numbers out of error messages? If not, how does it match up error messages with the precise line number?

Yes, FileCheck can parse regular expressions and a line number can be matched with [0-9]+. As shown in @LukeIreland1's example you can also specify expected line numbers without hardcoding it. For e.g with  [[@line+1]] you are specifying that the error should be on the next line.

klausler:
> > Wouldn't that require a check not for all possible strings on all non-error lines?
> 
> I am hoping that a single CHECK-NOT ":<regex_number>:<regex_number>: error" in between the expected error messages will catch that. It will probably bail out at the first unexpected error message.

Would this directive have to be inserted into the test source files?

kiranchandramohan:
> > > Wouldn't that require a check not for all possible strings on all non-error lines?
> > 
> > 
> > I am hoping that a single CHECK-NOT ":<regex_number>:<regex_number>: error" in between the expected error messages will catch that. It will probably bail out at the first unexpected error message.
> 
> Would this directive have to be inserted into the test source files?

Yes, this will have to be in the test source file. I am looking at your other point

> > > > > > If unexpected errors are reported by the compiler or if they are at the wrong line, this does not catch it.
> > > > > 
> > > > > 
> > > > > If FileCheck is really unable to catch extra messages or spurious text, that's a fatal flaw. However, flang's existing test-checking tools ought to work with lit, or be easily changed to be compatible with lit. Lit is just a driver.
> > > > 
> > > > 
> > > > FileCheck can do this by,
> > > > 
> > > > 1. Having a check with count, which will fail if the number of error messages differs.
> > > >    https://github.com/llvm-mirror/llvm/blob/master/test/FileCheck/check-count.txt
> > > > 2. Having a line number along with the check.
> > > >    https://github.com/llvm-mirror/llvm/blob/master/test/FileCheck/line-count-2.txt
> > > 
> > > 
> > > The capability that is essential to me is being told the exact source line number associated with a missing error message _or_ with an unexpected error message.
> > 
> > 
> > We can also add check-not in between the error messages to capture any unexpected error messages.
> 
> I suppose if they're unexpected, then it's hard to know what to put in the check-not lines. My understanding of lit, is there should be no errors thrown in order to pass, and if there is, it has to explicitly be matched correctly to pass. That means adding code that throws a different error causes a fail, and anything but the exact right test will cause a fail, so there's no need to check for what's not there.

@LukeIreland1 I agree with the first part that there should be no errors to pass. But I am not sure about the second part that all errors should be explicitly matched correctly to pass. As far as FileCheck is concerned, it is getting a set of strings as input, it does not know what is an error message and what is not. I was thinking that FileCheck will match all the CHECK's and will be happy with that, if there are extra lines in the input it is not going to complain/error.

> > > Wouldn't that require a check not for all possible strings on all non-error lines?
> > 
> > 
> > I am hoping that a single CHECK-NOT ":<regex_number>:<regex_number>: error" in between the expected error messages will catch that. It will probably bail out at the first unexpected error message.
> 
> Would this directive have to be inserted into the test source files?

Yes. There are also implicit check-nots, see below.
--implicit-check-not check-pattern
Adds implicit negative checks for the specified patterns between positive checks. The option allows writing stricter tests without stuffing them with CHECK-NOTs.

LukeIreland1:
> Yes. There are also implicit check-nots, see below.
> --implicit-check-not check-pattern
> Adds implicit negative checks for the specified patterns between positive checks. The option allows writing stricter tests without stuffing them with CHECK-NOTs.

That seems to have worked for me. `--implicit-check-not error:` will mean any error not captured (even duplicates of captured ones, will cause the test to fail. Need to test more, but thanks for highlighting this!

LukeIreland1:
Having to force push, because:
```
 $ git push myfork lit     
To git@github.com:LukeIreland1/f18.git
 ! [rejected]        lit -> lit (non-fast-forward)
error: failed to push some refs to 'git@github.com:LukeIreland1/f18.git'
hint: Updates were rejected because the tip of your current branch is behind
hint: its remote counterpart. Integrate the remote changes (e.g.
hint: 'git pull ...') before pushing again.
hint: See the 'Note about fast-forwards' in 'git push --help' for details.
```

Doing as it says seems to result in rebasing my commits on top of themselves (only solution is to completely reset the fork [which I'm not sure how to do without messing up the pull request])

LukeIreland1:
Closing to change approach and ease in lit (ideally in time for monorepo merge) rather than adding FileCheck in at the same time. Will also hopefully allow me to push without `-f`

PR #936, opened 2020-01-15
Checklist to precede pull requests

psteinfeld:
I added a checklist to the C++ Style document for things to check before submitting a pull request.

### DIFF
--- a/documentation/PullRequestChecklist.md
+++ b/documentation/PullRequestChecklist.md
@@ -1,0 +1,44 @@
+<!--===- documentation/PullRequestChecklist.md 
+
+   Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+   See https://llvm.org/LICENSE.txt for license information.
+   SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+-->
+
+# Pull request checklist
+Please review the following items before submitting a pull request.  This list
+can also be used when reviewing pull requests.
+*  Verify that new files have a license with correct file name.
+*  Run `git diff` on all modified files to look for spurious changes such as
+   `#include <iostream>`.
+*  If you added code that causes the compiler to emit a new error message, make
+   sure that you also added a test that causes that error message to appear
+   and verifies its correctness.
+*  Annotate the code and tests with appropriate references to constraint and
+   requirement numbers from the Fortran standard.
+*  Check dereferences of pointers and optionals where necessary.
+*  Ensure that the scopes of all functions and variables are as local as
+   possible.
+*  Try to make all functions fit on a screen (40 lines).
+*  Build and test with both GNU and clang compilers.
+*  When submitting an update to a pull request, review previous pull request
+   comments and make sure that you've actually made all of the changes that
+   were requested.
+
+## Follow the style guide
+The following items are taken from the [C++ style guide](C++style.md).  But
+even though I've read the style guide, they regularly trip me up.
+*  Run clang-format version 7 on all .cc and .h files.
+*  Make sure that all source lines have 80 or fewer characters.  Note that
+   clang-format will do this for most code.  But you may need to break up long
+   strings.
+*  Review declarations for proper use of `constexpr` and `const`.
+*  Follow the C++ naming guidelines.  Ensure that the names evoke their
+   purpose and are consistent with existing code.
+*  Review pointer and reference types to make sure that you're using them
+   appropriately.  Note that the [C++ style guide](C++style.md) contains a
+   section that describes all of the pointer types along with their
+   characteristics.
+*  Declare non-member functions ```static``` when possible.  Prefer
+   ```static``` functions over functions in anonymous namespaces.


documentation/C++style.md
```diff
@@ -296,6 +297,23 @@ like `Http`.
 Consistency is one of many aspects in the pursuit of clarity,
 but not an end in itself.
 
+## <a name="checklist"></a>Pull request checklist
+Before submitting a pull request, review the following items:
+*  Verify that all files have a license.
+*  Make sure that all source lines have 80 or fewer characters.
```
vjayathirtha-nv: Does clang-format not cover this?

documentation/C++style.md
```diff
@@ -296,6 +297,23 @@ like `Http`.
...
+*  Run `git diff` on all files to look for spurious changes such as `include <iostream>`.
+*  Run clang-format version 7 on all .cc and .h files.
+*  Build and test with both GNU and clang compilers.
+*  Make sure that all error messages you added appear in tests.
```
vjayathirtha-nv: This does not seem appropriate for a C++ style guide.

documentation/C++style.md
```diff
@@ -296,6 +297,23 @@ like `Http`.
...
+*  Annotate the code and tests with appropriate references to constraint and requirement numbers in the standard.
```
vjayathirtha-nv: This too.

documentation/C++style.md
```diff
@@ -296,6 +297,23 @@ like `Http`.
...
+*  Review declarations for proper use of `constexpr` and `const`.
+*  Ensure that the names of C++ constructs (functions, classes, etc.) evoke their purpose.
+*  Ensure that you check all dereferences of non-owning pointers.
+*  Ensure that the scopes of all functions and variables are as local as possible.
+*  Try to make all functions fit on a screen (40 lines).
+*  Declare non-member functions static when possible.
+*  Review previous pull request comments and make sure that you've actually made all of the changes that were requested.
```
vjayathirtha-nv: Even this.

documentation/C++style.md
psteinfeld: clang-format does not handle the case where a string causes a source line to exceed 80 characters.  To handle that, you must break up the string yourself.  I will add this information.

documentation/C++style.md
```diff
@@ -19,6 +19,7 @@ is clear on usage, follow it.
   follow it.  [Google's](https://google.github.io/styleguide/cppguide.html)
   is pretty good and comes with lots of justifications for its rules.
 * Reasonable exceptions to these guidelines can be made.
+* Before submitting a pull request, review the [checkist below](#checklist).
```
schweitzpgi: Do you want to matriculate from LLVM 7 at some point?

documentation/C++style.md
psteinfeld: As above, I agree with you.  Should I remove it?

documentation/C++style.md
psteinfeld: Yep.

documentation/C++style.md
schweitzpgi: This just means diagnostic errors to the user about her Fortran code, right?

documentation/C++style.md
schweitzpgi: +1

documentation/C++style.md
```diff
@@ -296,6 +297,23 @@ like `Http`.
...
+*  Run clang-format version 7 on all .cc and .h files.
+*  Build and test with both GNU and clang compilers.
+*  Make sure that all error messages you added appear in tests.
+*  Annotate the code and tests with appropriate references to constraint and requirement numbers in the standard.
+*  Review declarations for proper use of `constexpr` and `const`.
+*  Ensure that the names of C++ constructs (functions, classes, etc.) evoke their purpose.
+*  Ensure that you check all dereferences of non-owning pointers.
+*  Ensure that the scopes of all functions and variables are as local as possible.
+*  Try to make all functions fit on a screen (40 lines).
+*  Declare non-member functions static when possible.
```
schweitzpgi: https://llvm.org/docs/CodingStandards.html#anonymous-namespaces

documentation/C++style.md
vjayathirtha-nv: Yes, thanks.

documentation/C++style.md
psteinfeld: I had a similar thought.  Maybe we should delay adding some (or all) of these items until we merge into LLVM.

documentation/C++style.md
psteinfeld: The point I was trying to make about error messages is as follows.  When you add code to the compiler that produces an error message when processing Fortran source (as happens when you add a semantic check), you should add a test to the test suite that tests for the appearance of that message.

Thanks for the reference to using anonymous namespaces.  I didn't even realize that they existed.  But after reading that section of the LLVM coding standards, they don't seem like something that we should encourage.

kiranchandramohan:
Thanks @psteinfeld for writing this. I think this will be more useful as a separate document so that you can cover more than c++ (like what @vjayathirtha-nv is pointing out).

documentation/C++style.md
tskeith: If this is about the version of clang-format, we should move to a newer one sometime soon. But it's good to specify the version so that everyone formats the same way.

tskeith:
> Thanks @psteinfeld for writing this. I think this will be more useful as a separate document so that you can cover more than c++ (like what @vjayathirtha-nv is pointing out).

I agree. One item on the list should be "follow the C++ style guide" with a link to it. The items on this list that relate to C++ coding style should be there. Perhaps on this list we might call out specific items in the C++ style guide to emphasize (e.g. choose the right reference type from the many available).

documentation/C++style.md
psteinfeld: Exactly.  If, for example, you add a semantic check that results in an error message, you should also add a Fortran test that causes the compiler to produce that message.

psteinfeld:
> > Thanks @psteinfeld for writing this. I think this will be more useful as a separate document so that you can cover more than c++ (like what @vjayathirtha-nv is pointing out).
> 
> I agree. One item on the list should be "follow the C++ style guide" with a link to it. The items on this list that relate to C++ coding style should be there. Perhaps on this list we might call out specific items in the C++ style guide to emphasize (e.g. choose the right reference type from the many available).
Thanks, Tim and Varun.  I'll make this a separate document, reference the style guide from it, and add an item to check that the new code uses appropriate reference/pointer types.

documentation/PullRequestChecklist.md
```diff
@@ -0,0 +1,44 @@
...
+   SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+-->
+
+# Pull request checklist
+Please review the following items before submitting a pull request.  This list
+can also be used when reviewing pull requests.
+*  Verify that new files have a license with correct file name.
+*  Run `git diff` on all modified files to look for spurious changes such as
+   ```#include <iostream>```.
```
tskeith: For inline code fragments like this you can use single backquotes. Triple backquotes are for code blocks.

documentation/PullRequestChecklist.md
```diff
@@ -0,0 +1,44 @@
...
+## Follow the style guide
+The following items are taken from the [C++ style guide](C++style.md).  But
+even though I've read the style guide, they regularly trip me up.
+*  Run clang-format version 7 on all .cc and .h files.
+*  Make sure that all source lines have 80 or fewer characters.  Note that
+   clang-format will do this for most code.  But you may need to break up long
+   strings.
+*  Review declarations for proper use of `constexpr` and `const`.
+*  Follow the C++ naming guidelines.  Ensure that the names evoke their
+   purpose.
```
tskeith: I would add: "Use names that are consistent with names in existing code."

documentation/PullRequestChecklist.md
```diff
@@ -0,0 +1,44 @@
...
+can also be used when reviewing pull requests.
+*  Verify that new files have a license with correct file name.
+*  Run `git diff` on all modified files to look for spurious changes such as
+   ```#include <iostream>```.
+*  If you added code that causes the compiler to emit a new error message, make
+   sure that you also added a test that causes that error message to appear
+   and verifies its correctness.
+*  Annotate the code and tests with appropriate references to constraint and
+   requirement numbers from the Fortran standard.
+*  Check all dereferences of non-owning pointers.
```
tskeith: How about "Check dereferences of pointers and optionals when necessary."? Owning vs. non-owning isn't relevant.

documentation/PullRequestChecklist.md
psteinfeld: Thanks, Tim.  I'll fix this.

documentation/PullRequestChecklist.md
psteinfeld: Thanks, Tim.  I'll add that.

documentation/PullRequestChecklist.md
psteinfeld: Much better.  I'll change it.

PR #939, opened 2020-01-17
Explanation of how to implement a semantic check

psteinfeld:
This is the story of implementing semantic checks for passing DO
variables to functions with dummy arguments with INTENT(OUT) or
INTENT(INOUT).  I wrote it as an example of how to make changes
to the compiler for a newcomer.

### DIFF
--- a/documentation/ImplementingASemanticCheck.md
+++ b/documentation/ImplementingASemanticCheck.md
@@ -1,0 +1,833 @@
+<!--===- documentation/ImplementingASemanticCheck.md 
+  
+   Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+   See https://llvm.org/LICENSE.txt for license information.
+   SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+  
+-->
+# Introduction
+I recently added a semantic check to the f18 compiler front end.  This document
+describes my thought process and the resulting implementation.
+
+For more information about the compiler, start with the 
+[compiler overview](Overview.md).
+
+# Problem definition
+
+In the 2018 Fortran standard, section 11.1.7.4.3, paragraph 2, states that:
+
+```
+Except for the incrementation of the DO variable that occurs in step (3), the DO variable 
+shall neither be redefined nor become undefined while the DO construct is active.
+```
+One of the ways that DO variables might be redefined is if they are passed to
+functions with dummy arguments whose `INTENT` is `INTENT(OUT)` or
+`INTENT(INOUT)`.  I implemented this semantic check.  Specifically, I changed
+the compiler to emit an error message if an active DO variable was passed to a
+dummy argument of a FUNCTION with INTENT(OUT).  Similarly, I had the compiler
+emit a warning if an active DO variable was passed to a dummy argument with
+INTENT(INOUT).  Previously, I had implemented similar checks for SUBROUTINE
+calls.
+
+# Creating a test
+
+My first step was to create a test case to cause the problem.  I called it testfun.f90 and used it to check the behavior of other Fortran compilers.  Here's the initial version:
+
+```fortran
+  subroutine s()
+    Integer :: ivar, jvar
+
+    do ivar = 1, 10
+      jvar = intentOutFunc(ivar) ! Error since ivar is a DO variable
+    end do
+
+  contains
+    function intentOutFunc(dummyArg)
+      integer, intent(out) :: dummyArg
+      integer  :: intentOutFunc
+
+      dummyArg = 216
+    end function intentOutFunc
+  end subroutine s
+```
+
+I verified that other Fortran compilers produced an error message at the point
+of the call to `intentOutFunc()`:
+
+```fortran
+      jvar = intentOutFunc(ivar) ! Error since ivar is a DO variable
+```
+
+
+I also used this program to produce a parse tree for the program using the command:
+```bash
+  f18 -fdebug-dump-parse-tree -fparse-only testfun.f90
+```
+
+Here's the relevant fragment of the parse tree produced by the compiler:
+
+```
+| | ExecutionPartConstruct -> ExecutableConstruct -> DoConstruct
+| | | NonLabelDoStmt
+| | | | LoopControl -> LoopBounds
+| | | | | Scalar -> Name = 'ivar'
+| | | | | Scalar -> Expr = '1_4'
+| | | | | | LiteralConstant -> IntLiteralConstant = '1'
+| | | | | Scalar -> Expr = '10_4'
+| | | | | | LiteralConstant -> IntLiteralConstant = '10'
+| | | Block
+| | | | ExecutionPartConstruct -> ExecutableConstruct -> ActionStmt -> AssignmentStmt = 'jvar=intentoutfunc(ivar)'
+| | | | | Variable -> Designator -> DataRef -> Name = 'jvar'
+| | | | | Expr = 'intentoutfunc(ivar)'
+| | | | | | FunctionReference -> Call
+| | | | | | | ProcedureDesignator -> Name = 'intentoutfunc'
+| | | | | | | ActualArgSpec
+| | | | | | | | ActualArg -> Expr = 'ivar'
+| | | | | | | | | Designator -> DataRef -> Name = 'ivar'
+| | | EndDoStmt -> 
+```
+
+Note that this fragment of the tree only shows four `parser::Expr` nodes,
+but the full parse tree also contained a fifth `parser::Expr` node for the
+constant 216 in the statement:
+
+```fortran
+      dummyArg = 216
+```
+# Analysis and implementation planning
+
+I then considered what I needed to do.  I needed to detect situations where an
+active DO variable was passed to a dummy argument with `INTENT(OUT)` or
+`INTENT(INOUT)`.  Once I detected such a situation, I needed to produce a
+message that highlighted the erroneous source code.  
+
+## Deciding where to add the code to the compiler
+This new semantic check would depend on several types of information -- the
+parse tree, source code location information, symbols, and expressions.  Thus I
+needed to put my new code in a place in the compiler after the parse tree had
+been created, name resolution had already happened, and expression semantic
+checking had already taken place.
+
+Most semantic checks for statements are implemented by walking the parse tree
+and performing analysis on the nodes they visit.  My plan was to use this
+method.  The infrastructure for walking the parse tree for statement semantic
+checking is implemented in the files `lib/semantics/semantics.cpp`.
+Here's a fragment of the declaration of the framework's parse tree visitor from
+`lib/semantics/semantics.cpp`:
+
+```C++
+  // A parse tree visitor that calls Enter/Leave functions from each checker
+  // class C supplied as template parameters. Enter is called before the node's
+  // children are visited, Leave is called after. No two checkers may have the
+  // same Enter or Leave function. Each checker must be constructible from
+  // SemanticsContext and have BaseChecker as a virtual base class.
+  template<typename... C> class SemanticsVisitor : public virtual C... {
+  public:
+    using C::Enter...;
+    using C::Leave...;
+    using BaseChecker::Enter;
+    using BaseChecker::Leave;
+    SemanticsVisitor(SemanticsContext &context)
+      : C{context}..., context_{context} {}
+      ...
+
+```
+
+Since FUNCTION calls are a kind of expression, I was planning to base my
+implementation on the contents of `parser::Expr` nodes.  I would need to define
+either an `Enter()` or `Leave()` function whose parameter was a `parser::Expr`
+node.  Here's the declaration I put into `lib/semantics/check-do.h`:
+
+```C++
+  void Leave(const parser::Expr &);
+```
+The `Enter()` functions get called at the time the node is first visited --
+that is, before its children.  The `Leave()` function gets called after the
+children are visited.  For my check the visitation order didn't matter, so I
+arbitrarily chose to implement the `Leave()` function to visit the parse tree
+node.
+
+Since my semantic check was focused on DO CONCURRENT statements, I added it to
+the file `lib/semantics/check-do.cpp` where most of the semantic checking for
+DO statements already lived.
+
+## Taking advantage of prior work
+When implementing a similar check for SUBROUTINE calls, I created a utility
+functions in `lib/semantics/semantics.cpp` to emit messages if
+a symbol corresponding to an active DO variable was being potentially modified:
+
+```C++
+  void WarnDoVarRedefine(const parser::CharBlock &location, const Symbol &var);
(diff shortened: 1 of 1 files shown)


documentation/FrontEndTutorial.md
```diff
@@ -0,0 +1,747 @@
...
+# Problem definition
+
+In the 2018 Fortran standard, section 11.1.7.4.3, paragraph 2, states that:
+
+```
+Except for the incrementation of the DO variable that occurs in step (3), the DO variable 
+shall neither be redefined nor become undefined while the DO construct is active.
+```
+One of the ways that DO variables might be redefined is if they are passed to
+functions with dummy arguments whose ```INTENT``` is ```INTENT(IN)``` or
```
kiranchandramohan: INTENT(OUT)?

documentation/FrontEndTutorial.md
```diff
@@ -0,0 +1,747 @@
...
+come from the parse tree.  I looked in the file .../lib/parser/parse-tree.h and
+determined that a ```struct Expr``` contained source location information since
+it had the field ```CharBlock source```.  Thus, if I visited a
+```parser::Expr``` node, I could get the source location information for the
+associated expression.
+
+## Determining the ```INTENT```
+I knew from previous experience that I could find the
+```INTENT``` of the dummy argument associated with the actual argument from the
+a function called ```dummyIntent()``` in the class
```
kiranchandramohan: typo "the a function called"

documentation/FrontEndTutorial.md
```diff
@@ -0,0 +1,747 @@
...
+to traverse an ``` evaluate::Expr```  node to find all of the ```
+evaluate::ActualArgument``` nodes.  
+
+So my plan was to start with the ```parser::Expr``` node and extract its
+associated ```evaluate::Expr``` field.  I would then traverse the
+```evaluate::Expr``` tree collecting all of the ```evaluate::ActualArgument```
+nodes.  I would look at each of these nodes to determine the ```INTENT``` of
+the associated dummy argument.
+
+This combination of the traveral framework and ```dummyIntent()``` would give
```
kiranchandramohan: typo traversal

psteinfeld:
> 
> 
> Thanks @psteinfeld for writing this document based on your experience of implementing the do index variable redefinition check. Should be helpful for new people writing semantic checks.
> 
> I like the title of this PR (Explanation of how to implement a semantic check) but the filename probably has to be modified to look something like the title of the PR.
> 
> What information would you have missed if you had the visitor at the ActualArgSpec level? Is it just the source location?
> 
> Have noted some minor nits inline.

Thanks for reviewing this, @kiranchandramohan.

I wasn't sure what to call the document.  How about "ImplementingASemanticCheck.md"?

I probably should have discussed this in the document, but at one point I had a visitor that triggered on the type `parser::ActualArg`.  As you mention, this node does not contain source information.  But more importantly, the expression it contains does not contain any `evaluate::ActualArgument` nodes.  These are needed to discover the intent of the dummy arguments.  To get at these `evaluate::ActualArgument` nodes, I had to start with the expression that contained the function call.

documentation/FrontEndTutorial.md
psteinfeld: Good catch.

documentation/FrontEndTutorial.md
psteinfeld: Thanks!

documentation/FrontEndTutorial.md
klausler: You can use single backquotes (grave accents) for inline code.

psteinfeld: Thanks!

kiranchandramohan:
> > Thanks @psteinfeld for writing this document based on your experience of implementing the do index variable redefinition check. Should be helpful for new people writing semantic checks.
> > I like the title of this PR (Explanation of how to implement a semantic check) but the filename probably has to be modified to look something like the title of the PR.
> > What information would you have missed if you had the visitor at the ActualArgSpec level? Is it just the source location?
> > Have noted some minor nits inline.
> 
> Thanks for reviewing this, @kiranchandramohan.
> 
> I wasn't sure what to call the document. How about "ImplementingASemanticCheck.md"?
> 
> I probably should have discussed this in the document, but at one point I had a visitor that triggered on the type `parser::ActualArg`. As you mention, this node does not contain source information. 

Since you have written in a conversational style it will be good to have this info that you went the parser::ActualArg path and found that the information that you need is not available. A word about why source information is there in some nodes and not there in others would also be useful.

Also, some information about how an expression models a function call might be useful. In the case of subroutines, you have a call statement so it is obvious there.

> But more importantly, the expression it contains does not contain any `evaluate::ActualArgument` nodes. These are needed to discover the intent of the dummy arguments. To get at these `evaluate::ActualArgument` nodes, I had to start with the expression that contained the function call.

Are you hinting here that the Subroutine's arguments have the Actual Argument info but a Function's arguments do not have? 
A simplistic first thought would be that once you get the argument list of the Function/Subroutine call the processing should be the same.

kiranchandramohan:
A bit of time spent on where a semantic check should be added would also be good. That is probably the first thing that hits a first-time contributor. Some high-level guidance like if it is an Execution Construct it should be here or if it is a Declaration Construct it should be there. When should a person write his own checker and when not. Should a check be added during name or label resolution? Also in some cases, it is obvious like If it is something related to a do loop, OpenMP or co-array. 

Feel free to ignore if this is outside the scope of this document.

psteinfeld:
> Since you have written in a conversational style it will be good to have this info that you went the parser::ActualArg path and found that the information that you need is not available. A word about why source information is there in some nodes and not there in others would also be useful.
> 

Thanks, Kiran.

The very fact that you're wondering about this means that I should have included it in the document.  I'll do that now.  I plan to explain the difference between this case (for function calls), and the case for subroutine calls along with their relevant differences that affected the implementation.  Please review again after I add this information.

> Also, some information about how an expression models a function call might be useful. In the case of subroutines, you have a call statement so it is obvious there.

Will do.
 
> > But more importantly, the expression it contains does not contain any `evaluate::ActualArgument` nodes. These are needed to discover the intent of the dummy arguments. To get at these `evaluate::ActualArgument` nodes, I had to start with the expression that contained the function call.
> 
> Are you hinting here that the Subroutine's arguments have the Actual Argument info but a Function's arguments do not have?
> A simplistic first thought would be that once you get the argument list of the Function/Subroutine call the processing should be the same.

I plan to explain this further in the addition I describe above, but here's a preview.  In both cases, the most difficult piece of information I need is the INTENT of the dummy argument.  This is accessible through the `evaluate::ActualArgument`.  In the case of a subroutine call, the `evaluate::ActualArgument` nodes are accessible through the `typedCall` field of a `parser::CallStmt`:
```
struct CallStmt {
  WRAPPER_CLASS_BOILERPLATE(CallStmt, Call);
  mutable std::unique_ptr<evaluate::ProcedureRef,
      common::Deleter<evaluate::ProcedureRef>>
      typedCall;  // filled by semantics
};
```
The `evaluate::ProcedureRef``` contains a list of `evaluate::ActualArgument` nodes.

For a function call, though, there is no way to map from a parse tree node to an `evaluate::ProcedureRef` node.  But I can take advantage of the fact that all function calls are embedded in expressions, either directly or indirectly.  Once I have an expression, I can traverse it to gather all of the `evaluate::ActualArgument` nodes as described in the document.

For a `parser::CallStmt` node, there's no way (other than through the `typedCall` field) to get to the `evaluate::ActualArgument` node.

psteinfeld:
> 
> 
> A bit of time spent on where a semantic check should be added would also be good. That is probably the first thing that hits a first-time contributor. Some high-level guidance like if it is an Execution Construct it should be here or if it is a Declaration Construct it should be there. When should a person write his own checker and when not. Should a check be added during name or label resolution? Also in some cases, it is obvious like If it is something related to a do loop, OpenMP or co-array.
> 
> Feel free to ignore if this is outside the scope of this document.

I can certainly add a little more information about how I decided where to implement this check.  I think it is out of the scope of this document to cover the topic generally.

documentation/ImplementingASemanticCheck.md
```diff
@@ -0,0 +1,813 @@
...
+of the call to `intentOutFunc()`:
+
+```fortran
+      jvar = intentOutFunc(ivar) ! Error since ivar is a DO variable
+```
+
+
+I also used this program to produce a parse tree for the program using the command:
+```bash
+  ./bin/f18 -fdebug-dump-parse-tree -fparse-only testfun.f90
```
ichoyjx: maybe just `f18` instead of `./bin/f18`?

documentation/ImplementingASemanticCheck.md
```diff
@@ -0,0 +1,813 @@
...
+needed to put my new code in a place in the compiler after the parse tree had
+been created, name resolution had already happened, and expression semantic
+checking had already taken place.
+
+Most semantic checks for statements are implemented by walking the parse tree
+and performing analysis on the nodes they visit.  My plan was to use this
+method.  The infrastructure for walking the parse tree for statement semantic
+checking is implemented in the files .../lib/semantics/semantics.[cc,h].
+Here's a fragment of the declaration of the framework's parse tree visitor from
+.../lib/semantics/semantics.cc:
```
ichoyjx: I think you could simply put `lib/semantics/semantics.[cpp,h]` & `lib/semantics/semantics.cpp`. (They are in code format)

documentation/ImplementingASemanticCheck.md
```diff
@@ -0,0 +1,813 @@
...
+  void Leave(const parser::Expr &);
+```
+The `Enter()` functions get called at the time the node is first visited --
+that is, before its children.  The `Leave()` function gets called after the
+children are visited.  For my check the visitation order didn't matter, so I
+arbitrarily chose to implement the `Leave()` function to visit the parse tree
+node.
+
+Since my semantic check was focused on DO CONCURRENT statements, I added it to
+the file .../lib/semantics/check-do.cc where most of the semantic checking for
```
ichoyjx: same here: `lib/semantics/check-do.cpp`.

documentation/ImplementingASemanticCheck.md
```diff
@@ -0,0 +1,813 @@
...
+DO statements already lived.
+
+## Taking advantage of prior work
+When implementing a similar check for SUBROUTINE calls, I created a utility
+functions in .../lib/semantics/semantics.[cc.h] to emit messages if
```
ichoyjx: same here & the rest of this doc. (note that `.cc` was changed to `.cpp` too)

documentation/ImplementingASemanticCheck.md
```diff
@@ -0,0 +1,813 @@
...
+
+## Collecting the actual arguments
+To get the `INTENT` of the dummy arguments and the `semantics::Symbol` associated with the
+actual argument, I needed to find all of the actual arguments embedded in an
+expression that contained a FUNCTION call.  So my next step was to write the
+framework to walk the `evaluate::Expr` to gather all of the
+`evaluate::ActualArgument` nodes.  The code that I planned to model it on
+was the existing infrastructure that collected all of the `semantics::Symbol` nodes from an
+`evaluate::Expr`.  I found this implementation in
+.../lib/evaluate/tools.cc:
```
ichoyjx: I strongly recommend getting rid of all the `.../`.

documentation/ImplementingASemanticCheck.md
```diff
@@ -0,0 +1,813 @@
+<!--===- documentation/FrontEndTutorial.md
```
ichoyjx: haven't decided the final name yet?

documentation/ImplementingASemanticCheck.md
psteinfeld: Yes.  I think your choice is better.

documentation/ImplementingASemanticCheck.md
psteinfeld: Good catch.  Given the rearrangement of the header files, I think it's best not to mention them at all.  I'll remove them and change all of the references to ".cc" to ".cpp".

psteinfeld: Actually, it looks like there are several header files I need to mention.  I'll fix up their path names.

documentation/ImplementingASemanticCheck.md
psteinfeld: Yep.

documentation/ImplementingASemanticCheck.md
psteinfeld: Will do.

documentation/ImplementingASemanticCheck.md
psteinfeld: Good catch!

And thanks for reading through this and providing feedback.

documentation/ImplementingASemanticCheck.md
```diff
@@ -0,0 +1,816 @@
...
+All of the declarations associated with both FUNCTION and SUBROUTINE calls are
+in `include/flang/evaluate/call.h`.  An `evaluate::FunctionRef` inherits from
+an `evaluate::ProcedureRef` which contains the list of
+`evaluate::ActualArgument` nodes.  But the relationship between an
+`evaluate::FunctionRef` node and its associated arguments is not relevant.  I
+only needed to find the `evaluate::ActualArgument` nodes in an expression.
+They hold all of the information I needed.
+
+So my plan was to start with the `parser::Expr` node and extract its
+associated `evaluate::Expr` field.  I would then traverse the
```
ichoyjx: Do you think a little bit explanation between `parser::Expr` and `evaluate::Expr` would be better?

psteinfeld: Good point.  I was certainly confused when I first learned that there were two different types named `Expr` in the compiler.  I'll add a short explanation.

documentation/ImplementingASemanticCheck.md
```diff
@@ -0,0 +1,833 @@
...
+The second `Expr` type is in the `evaluate` namespace.   The `evaluate`
+namespace contains many types associated with semantic checking of expressions.
+`evaluate::Expr` is defined in the file `include/flang/evaluate/expression.h`.
+It represents an expression after it has undergone semantic checking and
+contains information that is only available after semantic analysis.  This
+information includes the Fortran type of the expression, whether it's a
+reference to a function, whether it's an actual argument, etc.  After an
+expression has undergone semantic analysis, the field `typedExpr` in the
+`parser::Expr` node is filled in with a pointer to the analyzed expression in
+`evaluate::Expr`.
```
ichoyjx: Very nice, thanks!

klausler: The `evaluate::Expr<T>` template classes, which is parameterized over the various types of Fortran, constitute a suite of strongly-typed representations of valid Fortran expressions of type `T` that has been fully elaborated with conversion operations and subjected to constant folding.

PR #940, opened 2020-01-18
[OpenMP] Name Resolution for OpenMP constructs

ichoyjx:
This is an extended framework based on the previous work that addresses
the NR on OpenMP directives/clauses (b2ea520). In this change:
  * New `OmpVisitor` is created (ResolveNamesVisitor derives from it) to
    create necessary scopes for certain OpenMP constructs. This is along
    with the regular Fortran NR process.
  * Old `OmpVisitor` is adjusted and converted to a standalone visitor--
    `OmpAttributeVisitor`. This is used to walk through the OpenMP constructs
    and do the NR for variables on the OpenMP directives or data references
    within the OpenMP constructs. "Do the NR" here means that based on the NR
    results of the regular Fortran NR, fix the symbols of `Names` related
    to the OpenMP constructs. Note that there is an `OmpContext` in this
    visitor (similar to the one in `OmpStructureChecker`), this is necessary
    when dealing with the nested OpenMP constructs in the future.

Given an OpenMP code:
```
real*8 a, b
  a = 1.
  b = 2.
  !$omp parallel private(a)
  a = 3.
  b = 4.
  !$omp end parallel
  print *, a, b
end
```

w/o -fopenmp:
```
real*8 a, b
 !REF: /MainProgram1/a
 a = 1.
 !REF: /MainProgram1/b
 b = 2.
 !!!! OMP parallel
 !REF: /MainProgram1/a
 a = 3.
 !REF: /MainProgram1/b
 b = 4.
 !!!! OMP end parallel
 !REF: /MainProgram1/a
 !REF: /MainProgram1/b
 print *, a, b
end
```

w/ -fopenmp:
```
real*8 a, b
 !REF: /MainProgram1/a
 a = 1.
 !REF: /MainProgram1/b
 b = 2.
!$omp parallel  private(a)   <-- new Symbol for 'a' created
 !DEF: /MainProgram1/Block1/a (OmpPrivate) HostAssoc REAL(8)
 a = 3.   <-- fix the old symbol with new Symbol in parallel scope
 !REF: /MainProgram1/b
 b = 4.   <-- do nothing because by default it is shared in this scope
!$omp end parallel
 !REF: /MainProgram1/a
 !REF: /MainProgram1/b
 print *, a, b
end
```

Please note that this is a framework update, there are still many
things on the TODO list for finishing the NR for OpenMP (based on
the `OpenMP-semantics.md` design doc), which will be on top of this
framework. Existing tests are `omp-resolve*.f90` and `omp-symbol*.f90`.

Some TODO items:
- Create a generic function to go through all the rules for deciding
  `predetermined`, `explicitly determined`, and `implicitly determined`
  data-sharing attributes. (This is the next biggest part)
- Handle `Array Sections` and `Array or Structure Element`.
- Take association into consideration for example Pointer association,
  `ASSOCIATE` construct, and etc.
- Handle all the name resolution for directives/clauses that have
  `parser::Name`.

### DIFF
--- a/include/flang/parser/parse-tree.h
+++ b/include/flang/parser/parse-tree.h
@@ -3498,10 +3498,12 @@ struct OmpSectionsDirective {
 struct OmpBeginSectionsDirective {
   TUPLE_CLASS_BOILERPLATE(OmpBeginSectionsDirective);
   std::tuple<OmpSectionsDirective, OmpClauseList> t;
+  CharBlock source;
 };
 struct OmpEndSectionsDirective {
   TUPLE_CLASS_BOILERPLATE(OmpEndSectionsDirective);
   std::tuple<OmpSectionsDirective, OmpClauseList> t;
+  CharBlock source;
 };
 
 // [!$omp section]
@@ -3742,21 +3744,25 @@ struct OpenMPStandaloneConstruct {
 struct OmpBeginLoopDirective {
   TUPLE_CLASS_BOILERPLATE(OmpBeginLoopDirective);
   std::tuple<OmpLoopDirective, OmpClauseList> t;
+  CharBlock source;
 };
 
 struct OmpEndLoopDirective {
   TUPLE_CLASS_BOILERPLATE(OmpEndLoopDirective);
   std::tuple<OmpLoopDirective, OmpClauseList> t;
+  CharBlock source;
 };
 
 struct OmpBeginBlockDirective {
   TUPLE_CLASS_BOILERPLATE(OmpBeginBlockDirective);
   std::tuple<OmpBlockDirective, OmpClauseList> t;
+  CharBlock source;
 };
 
 struct OmpEndBlockDirective {
   TUPLE_CLASS_BOILERPLATE(OmpEndBlockDirective);
   std::tuple<OmpBlockDirective, OmpClauseList> t;
+  CharBlock source;
 };
 
 struct OpenMPBlockConstruct {
--- a/include/flang/semantics/symbol.h
+++ b/include/flang/semantics/symbol.h
@@ -467,7 +467,7 @@ public:
       // OpenMP miscellaneous flags
       OmpCommonBlock, OmpReduction, OmpDeclareSimd, OmpDeclareTarget,
       OmpThreadprivate, OmpDeclareReduction, OmpFlushed, OmpCriticalLock,
-      OmpIfSpecified);
+      OmpIfSpecified, OmpNone);
   using Flags = common::EnumSet<Flag, Flag_enumSize>;
 
   const Scope &owner() const { return *owner_; }
--- a/lib/parser/openmp-parsers.cpp
+++ b/lib/parser/openmp-parsers.cpp
@@ -280,8 +280,8 @@ TYPE_PARSER(sourced(construct<OmpLoopDirective>(first(
         pure(OmpLoopDirective::Directive::TeamsDistributeSimd),
     "TEAMS DISTRIBUTE" >> pure(OmpLoopDirective::Directive::TeamsDistribute)))))
 
-TYPE_PARSER(construct<OmpBeginLoopDirective>(
-    sourced(Parser<OmpLoopDirective>{}), Parser<OmpClauseList>{}))
+TYPE_PARSER(sourced(construct<OmpBeginLoopDirective>(
+    sourced(Parser<OmpLoopDirective>{}), Parser<OmpClauseList>{})))
 
 // 2.14.1 construct-type-clause -> PARALLEL | SECTIONS | DO | TASKGROUP
 TYPE_PARSER(sourced(construct<OmpCancelType>(
@@ -345,8 +345,8 @@ TYPE_PARSER(construct<OmpBlockDirective>(
         "TEAMS" >> pure(OmpBlockDirective::Directive::Teams),
         "WORKSHARE" >> pure(OmpBlockDirective::Directive::Workshare))))
 
-TYPE_PARSER(construct<OmpBeginBlockDirective>(
-    sourced(Parser<OmpBlockDirective>{}), Parser<OmpClauseList>{}))
+TYPE_PARSER(sourced(construct<OmpBeginBlockDirective>(
+    sourced(Parser<OmpBlockDirective>{}), Parser<OmpClauseList>{})))
 
 TYPE_PARSER(construct<OmpReductionInitializerClause>(
     "INITIALIZER" >> parenthesized("OMP_PRIV =" >> expr)))
@@ -482,12 +482,12 @@ TYPE_PARSER(construct<OmpSectionsDirective>(
             pure(OmpSectionsDirective::Directive::ParallelSections))))
 
 // OMP BEGIN and END SECTIONS Directive
-TYPE_PARSER(construct<OmpBeginSectionsDirective>(
-    sourced(Parser<OmpSectionsDirective>{}), Parser<OmpClauseList>{}))
+TYPE_PARSER(sourced(construct<OmpBeginSectionsDirective>(
+    sourced(Parser<OmpSectionsDirective>{}), Parser<OmpClauseList>{})))
 TYPE_PARSER(
-    startOmpLine >> construct<OmpEndSectionsDirective>(
+    startOmpLine >> sourced(construct<OmpEndSectionsDirective>(
                         sourced("END"_tok >> Parser<OmpSectionsDirective>{}),
-                        Parser<OmpClauseList>{}))
+                        Parser<OmpClauseList>{})))
 
 // OMP SECTION-BLOCK
 TYPE_PARSER(maybe(startOmpLine >> "SECTION"_tok / endOmpLine) >>
@@ -512,15 +512,15 @@ TYPE_CONTEXT_PARSER("OpenMP construct"_en_US,
 
 // END OMP Block directives
 TYPE_PARSER(
-    startOmpLine >> construct<OmpEndBlockDirective>(
+    startOmpLine >> sourced(construct<OmpEndBlockDirective>(
                         sourced("END"_tok >> Parser<OmpBlockDirective>{}),
-                        Parser<OmpClauseList>{}))
+                        Parser<OmpClauseList>{})))
 
 // END OMP Loop directives
 TYPE_PARSER(
-    startOmpLine >> construct<OmpEndLoopDirective>(
+    startOmpLine >> sourced(construct<OmpEndLoopDirective>(
                         sourced("END"_tok >> Parser<OmpLoopDirective>{}),
-                        Parser<OmpClauseList>{}))
+                        Parser<OmpClauseList>{})))
 
 TYPE_PARSER(construct<OpenMPLoopConstruct>(
     Parser<OmpBeginLoopDirective>{} / endOmpLine))
--- a/lib/semantics/resolve-names.cpp
+++ b/lib/semantics/resolve-names.cpp
@@ -6046,15 +6046,14 @@ void OmpAttributeVisitor::Post(const parser::Name &name) {
       // TODO: create a separate function to go through the rules for
       //       predetermined, explicitly determined, and implicitly
       //       determined data-sharing attributes (2.15.1.1).
-      if (GetContext().defaultDSA == Symbol::Flag::OmpNone) {
-        context_.Say(name.source,
-            "The DEFAULT(NONE) clause requires '%s' must be listed in "
-            "a data-sharing attribute clause"_err_en_US,
-            symbol->name());
-      }
       if (Symbol * found{currScope().FindSymbol(name.source)}) {
         if (HasObject(found)) {
           name.symbol = found;  // adjust the symbol within region
+        } else if (GetContext().defaultDSA == Symbol::Flag::OmpNone) {
+          context_.Say(name.source,
+              "The DEFAULT(NONE) clause requires that '%s' must be listed in "
+              "a data-sharing attribute clause"_err_en_US,
+              symbol->name());
         }
       }
     }
--- a/lib/semantics/resolve-names.cpp
+++ b/lib/semantics/resolve-names.cpp
@@ -1286,13 +1286,13 @@ private:
 
   void ResolveOmpObjectList(const parser::OmpObjectList &, Symbol::Flag);
   void ResolveOmpObject(const parser::OmpObject &, Symbol::Flag);
-  Symbol &ResolveOmp(const parser::Name &, Symbol::Flag);
-  Symbol &ResolveOmp(Symbol &, Symbol::Flag);
+  Symbol *ResolveOmp(const parser::Name &, Symbol::Flag);
+  Symbol *ResolveOmp(Symbol &, Symbol::Flag);
   Symbol *ResolveOmpCommonBlockName(const parser::Name *);
-  Symbol &DeclarePrivateAccessEntity(const parser::Name &, Symbol::Flag);
-  Symbol &DeclarePrivateAccessEntity(Symbol &, Symbol::Flag);
-  Symbol &DeclareOrMarkOtherAccessEntity(const parser::Name &, Symbol::Flag);
-  Symbol &DeclareOrMarkOtherAccessEntity(Symbol &, Symbol::Flag);
+  Symbol *DeclarePrivateAccessEntity(const parser::Name &, Symbol::Flag);
+  Symbol *DeclarePrivateAccessEntity(Symbol &, Symbol::Flag);
+  Symbol *DeclareOrMarkOtherAccessEntity(const parser::Name &, Symbol::Flag);
+  Symbol *DeclareOrMarkOtherAccessEntity(Symbol &, Symbol::Flag);
   void CheckMultipleAppearances(
       const parser::Name &, const Symbol &, Symbol::Flag);
   SymbolSet dataSharingAttributeObjects_;  // on one directive
@@ -6039,8 +6039,7 @@ void OmpAttributeVisitor::Post(const parser::OmpDefaultClause &x) {
 // For OpenMP constructs, check all the data-refs within the constructs
 // and adjust the symbol for each Name if necessary
 void OmpAttributeVisitor::Post(const parser::Name &name) {
-  if (!ompContext_.empty() && GetContext().withinConstruct) {
-    CHECK(name.symbol);  // all the Names should be resolved at this point
+  if (name.symbol && !ompContext_.empty() && GetContext().withinConstruct) {
     auto *symbol{name.symbol};
     if (!symbol->owner().IsDerivedType() && !HasObject(symbol)) {
       // TODO: create a separate function to go through the rules for
@@ -6090,10 +6089,11 @@ void OmpAttributeVisitor::ResolveOmpObject(
       common::visitors{
           [&](const parser::Designator &designator) {
             if (const auto *name{GetDesignatorNameIfDataRef(designator)}) {
-              auto &symbol{ResolveOmp(*name, ompFlag)};
-              SetContextObjectMap(&symbol, ompFlag);
-              if (dataSharingAttributeFlags.test(ompFlag)) {
-                CheckMultipleAppearances(*name, symbol, ompFlag);
+              if (auto *symbol{ResolveOmp(*name, ompFlag)}) {
+                SetContextObjectMap(symbol, ompFlag);
+                if (dataSharingAttributeFlags.test(ompFlag)) {
+                  CheckMultipleAppearances(*name, *symbol, ompFlag);
+                }
               }
             } else if (const auto *designatorName{
                            resolver_.ResolveDesignator(designator)};
@@ -6127,8 +6127,9 @@ void OmpAttributeVisitor::ResolveOmpObject(
               for (const Symbol &object :
                   symbol->get<CommonBlockDetails>().objects()) {
                 Symbol &mutableObject{const_cast<Symbol &>(object)};
-                auto &resolvedObject{ResolveOmp(mutableObject, ompFlag)};
-                SetContextObjectMap(&resolvedObject, ompFlag);
+                if (auto *resolvedObject{ResolveOmp(mutableObject, ompFlag)}) {
+                  SetContextObjectMap(resolvedObject, ompFlag);
+                }
               }
             } else {
               context_.Say(name.source,  // 2.15.3
@@ -6140,7 +6141,7 @@ void OmpAttributeVisitor::ResolveOmpObject(
       ompObject.u);
 }
 
-Symbol &OmpAttributeVisitor::ResolveOmp(
+Symbol *OmpAttributeVisitor::ResolveOmp(
     const parser::Name &name, Symbol::Flag ompFlag) {
   if (ompFlagsRequireNewSymbol.test(ompFlag)) {
     return DeclarePrivateAccessEntity(name, ompFlag);
@@ -6149,7 +6150,7 @@ Symbol &OmpAttributeVisitor::ResolveOmp(
   }
 }
 
-Symbol &OmpAttributeVisitor::ResolveOmp(Symbol &symbol, Symbol::Flag ompFlag) {
+Symbol *OmpAttributeVisitor::ResolveOmp(Symbol &symbol, Symbol::Flag ompFlag) {
   if (ompFlagsRequireNewSymbol.test(ompFlag)) {
     return DeclarePrivateAccessEntity(symbol, ompFlag);
   } else {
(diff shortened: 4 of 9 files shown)


lib/semantics/resolve-names.cc
```diff
@@ -5915,6 +5917,319 @@ class DeferredCheckVisitor {
...
+                if (std::holds_alternative<parser::Substring>(designator.u)) {
+                  context_.Say(designator.source,
+                      "Substrings are not allowed on OpenMP "
+                      "directives or clauses"_err_en_US);
+                }
+              }
+              // other checks, more TBD
+              if (const auto *details{designatorName->symbol
+                                          ->detailsIf<ObjectEntityDetails>()}) {
+                if (details->IsArray()) {
```
psteinfeld: Does ```details``` need to be checked for null before dereferencing?

lib/semantics/resolve-names.cc
```diff
@@ -5915,6 +5917,319 @@ class DeferredCheckVisitor {
...
+  if (ompFlagsRequireNewSymbol.test(ompFlag)) {
+    return DeclarePrivateAccessEntity(symbol, ompFlag);
+  } else {
+    return DeclareOrMarkOtherAccessEntity(symbol, ompFlag);
+  }
+}
+
+Symbol &OmpAttributeVisitor::DeclarePrivateAccessEntity(
+    const parser::Name &name, Symbol::Flag ompFlag) {
+  if (name.symbol->owner() != currScope()) {
```
psteinfeld: Are we guaranteed that ```name.symbol``` is not null?

lib/semantics/resolve-names.cc
```diff
@@ -5915,6 +5917,319 @@ class DeferredCheckVisitor {
...
+}
+
+Symbol &OmpAttributeVisitor::DeclareOrMarkOtherAccessEntity(
+    const parser::Name &name, Symbol::Flag ompFlag) {
+  Symbol *prev{currScope().FindSymbol(name.source)};
+  if (prev != name.symbol) {
+    name.symbol = prev;
+  }
+  if (ompFlagsRequireMark.test(ompFlag)) {
+    prev->set(ompFlag);
```
psteinfeld: Do we know that ```prev``` is not null?

lib/semantics/resolve-names.cc
```diff
@@ -5915,6 +5917,319 @@ class DeferredCheckVisitor {
...
+  if (!ompContext_.empty() && GetContext().withinConstruct) {
+    CHECK(name.symbol);  // all the Names should be resolved at this point
+    auto *symbol{name.symbol};
+    if (!symbol->owner().IsDerivedType() && !HasObject(symbol)) {
+      // TODO: create a separate function to go through the rules for
+      //       predetermined, explicitly determined, and implicitly
+      //       determined data-sharing attributes (2.15.1.1).
+      if (GetContext().defaultDSA == Symbol::Flag::OmpNone) {
+        context_.Say(name.source,
+            "The DEFAULT(NONE) clause requires '%s' must be listed in "
```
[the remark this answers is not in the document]
ichoyjx: Thanks, I will add a test case! how did i miss that..

lib/semantics/resolve-names.cc
```diff
@@ -1099,7 +1207,62 @@ class OmpVisitor : public virtual DeclarationVisitor {
...
+  const Scope &currScope() { return *GetContext().scope; }
+  void SetContextDefaultDSA(Symbol::Flag flag) {
+    GetContext().defaultDSA = flag;
+  }
+  void SetContextObjectMap(const Symbol *symbol, Symbol::Flag flag) {
+    GetContext().objectMap.emplace(symbol, flag);
+  }
+  bool HasObject(const Symbol *symbol) {
+    auto it{GetContext().objectMap.find(symbol)};
+    if (it != GetContext().objectMap.end()) {
```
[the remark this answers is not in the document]
ichoyjx: Nice!!

lib/semantics/resolve-names.cc
```diff
@@ -1099,7 +1207,62 @@ class OmpVisitor : public virtual DeclarationVisitor {
...
+  OmpContext &GetContext() {
+    CHECK(!ompContext_.empty());
+    return ompContext_.back();
+  }
+  void PushContext(const parser::CharBlock &source, OmpDirective dir) {
+    ompContext_.emplace_back(source, dir);
+    SetContextScope(source);
+  }
+  void PopContext() { ompContext_.pop_back(); }
+  void SetContextDirectiveSource(parser::CharBlock &dir) {
```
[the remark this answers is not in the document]
ichoyjx: Yes, they shall be used soon for error msg & nested constructs check.

lib/semantics/resolve-names.cc
```diff
@@ -1099,7 +1207,62 @@ class OmpVisitor : public virtual DeclarationVisitor {
...
+    OmpContext(const parser::CharBlock &source, OmpDirective d)
+      : directiveSource{source}, directive{d} {}
+    parser::CharBlock directiveSource{nullptr};
+    OmpDirective directive;
+    Scope *scope{nullptr};
+    // TODO: default DSA is implicitly determined in different ways
+    Symbol::Flag defaultDSA{Symbol::Flag::OmpShared};
+    // variables on Data-sharing attribute clauses
+    std::map<const Symbol *, Symbol::Flag> objectMap;
+    bool withinConstruct{false};
```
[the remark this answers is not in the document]
ichoyjx: I added a few constructs only for passing the existing `omp-clause-validity01.f90` and this is one of them. Sorry, I should've mentioned that. `withinConstruct` needs to be set as long as the corresponding OpenMP construct can wrap a `Block` of code. I need this `withinConstruct` to distinguish the `Names` on the directives/clauses and within the constructs.

lib/semantics/resolve-names.cc
```diff
@@ -5915,6 +5917,319 @@ class DeferredCheckVisitor {
...
+    case parser::OmpDefaultClause::Type::None:
+      SetContextDefaultDSA(Symbol::Flag::OmpNone);
+      break;
+    }
+  }
+}
+
+// For OpenMP constructs, check all the data-refs within the constructs
+// and adjust the symbol for each Name if necessary
+void OmpAttributeVisitor::Post(const parser::Name &name) {
```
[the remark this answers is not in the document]
ichoyjx: Thanks for pointing it out. This case actually falls into the TODO category, similar to the one few lines below: `!symbol->owner().IsDerivedType()`, which skip the `Name` fix for 'b' in `a%b'. They all will be wrapped into a function.

lib/semantics/resolve-names.cc
```diff
@@ -5915,6 +5917,318 @@ class DeferredCheckVisitor {
...
+  if (!ompContext_.empty() && GetContext().withinConstruct) {
```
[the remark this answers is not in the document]
ichoyjx: It is intended for now because it falls into the TODO category for the `implicitly determined` data-sharing attributes, which is listed in the `OpenMP-semantics.md` doc. In this case when processing `sections`, we need to locate the upper `parallel` context and see what the `default` DSA is. I am currently working on the nested constructs example.

lib/semantics/resolve-names.cc
```diff
@@ -5915,6 +5917,319 @@ class DeferredCheckVisitor {
...
+    CHECK(name.symbol);  // all the Names should be resolved at this point
```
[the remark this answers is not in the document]
ichoyjx: I agree. As we discussed, I will change the prototypes of those `ResolveOmp` functions to return `Symbol *` and check whether there are any fatal errors before return.

lib/semantics/resolve-names.cc
```diff
@@ -5915,6 +5917,318 @@ class DeferredCheckVisitor {
...
+    auto *symbol{name.symbol};
+    if (!symbol->owner().IsDerivedType() && !HasObject(symbol)) {
```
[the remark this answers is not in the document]
ichoyjx: Same answer as the one above about `implicitly determined`.

lib/semantics/resolve-names.cc
```diff
@@ -5915,6 +5917,319 @@ class DeferredCheckVisitor {
...
+}
+
+void OmpAttributeVisitor::ResolveOmpObject(
+    const parser::OmpObject &ompObject, Symbol::Flag ompFlag) {
+  std::visit(
+      common::visitors{
+          [&](const parser::Designator &designator) {
+            if (const auto *name{GetDesignatorNameIfDataRef(designator)}) {
+              auto &symbol{ResolveOmp(*name, ompFlag)};
+              SetContextObjectMap(&symbol, ompFlag);
```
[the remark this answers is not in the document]
ichoyjx: `omp-symbol06.f90` shows an example of two flags:
```
!DEF: /MainProgram1/a (Implicit) ObjectEntity REAL(4)
  a = 1.
  !$omp parallel do  firstprivate(a) lastprivate(a)
  !DEF: /MainProgram1/i (Implicit) ObjectEntity INTEGER(4)
  do i=1,10
     !DEF: /MainProgram1/Block1/a (OmpFirstPrivate, OmpLastPrivate) HostAssoc REAL(4)
     a = 2.
  end do
end program
```

lib/semantics/resolve-names.cc
```diff
@@ -5977,6 +6292,7 @@ void ResolveNamesVisitor::ResolveExecutionParts(const ProgramTree &node) {
   SetScope(*node.scope());
   if (const auto *exec{node.exec()}) {
     Walk(*exec);
+    OmpAttributeVisitor{context(), *this}.Walk(*exec);
```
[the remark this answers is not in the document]
ichoyjx: I will try.

lib/semantics/resolve-names.cc
```diff
@@ -1045,43 +1047,149 @@ class ConstructVisitor : public virtual DeclarationVisitor {
...
+
+  static const parser::Name *GetDesignatorNameIfDataRef(
+      const parser::Designator &designator) {
+    const auto *dataRef{std::get_if<parser::DataRef>(&designator.u)};
+    return dataRef ? std::get_if<parser::Name>(&dataRef->u) : nullptr;
+  }
+
+  bool Pre(const parser::OpenMPBlockConstruct &);
+  void Post(const parser::OpenMPBlockConstruct &) {
+    GetContext().withinConstruct = false;
```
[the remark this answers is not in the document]
ichoyjx: Good point.

lib/semantics/resolve-names.cc
```diff
@@ -1045,43 +1047,149 @@ class ConstructVisitor : public virtual DeclarationVisitor {
...
+  void Post(const parser::OmpEndSectionsDirective &) {
+    messageHandler().set_currStmtSource(std::nullopt);
+  }
+};
+
+bool OmpVisitor::NeedsScope(const parser::OpenMPBlockConstruct &x) {
+  const auto &beginBlockDir{std::get<parser::OmpBeginBlockDirective>(x.t)};
+  const auto &beginDir{std::get<parser::OmpBlockDirective>(beginBlockDir.t)};
+  switch (beginDir.v) {
+  case parser::OmpBlockDirective::Directive::TargetData:
```
schweitzpgi: Maybe add [[fallthrough]]; attributes for these. https://en.cppreference.com/w/cpp/language/attributes/fallthrough

lib/semantics/resolve-names.cc
```diff
@@ -1045,43 +1047,149 @@ class ConstructVisitor : public virtual DeclarationVisitor {
...
+  case parser::OmpBlockDirective::Directive::Master:
+  case parser::OmpBlockDirective::Directive::Ordered: return false;
+  default: return true;
+  }
+}
 
-  bool Pre(const parser::OpenMPThreadprivate &x) {
+void OmpVisitor::AddOmpSourceRange(const parser::CharBlock &source) {
+  messageHandler().set_currStmtSource(source);
+  for (auto *scope = &currScope(); scope; scope = &scope->parent()) {
```
tskeith: `scope` can't be null so there is no need to check it.

lib/semantics/resolve-names.cc
```diff
@@ -1045,43 +1047,149 @@ class ConstructVisitor : public virtual DeclarationVisitor {
...
+    scope->AddSourceRange(source);
+    if (scope->IsGlobal()) {
+      break;
+    }
+  }
```
tskeith: Maybe this loop should be part of the implementation of `Scope::AddSourceRange` so that it doesn't have to be duplicated here and in `ScopeHandler`.

lib/semantics/resolve-names.cc
```diff
@@ -1045,43 +1047,149 @@ class ConstructVisitor : public virtual DeclarationVisitor {
...
+  template<typename A> void Walk(const A &x) { parser::Walk(x, *this); }
+
+  template<typename A> bool Pre(const A &) { return true; }
+  template<typename A> void Post(const A &) {}
+
+  static const parser::Name *GetDesignatorNameIfDataRef(
+      const parser::Designator &designator) {
+    const auto *dataRef{std::get_if<parser::DataRef>(&designator.u)};
+    return dataRef ? std::get_if<parser::Name>(&dataRef->u) : nullptr;
+  }
```
tskeith: This function should be private.

lib/semantics/resolve-names.cc
```diff
@@ -1099,7 +1207,62 @@ class OmpVisitor : public virtual DeclarationVisitor {
...
+  void SetContextScope(const parser::CharBlock &source) {
+    auto &currScope{context_.FindScope(source)};
+    GetContext().scope = &currScope;
+  }
+  const Scope &currScope() { return *GetContext().scope; }
+  void SetContextDefaultDSA(Symbol::Flag flag) {
+    GetContext().defaultDSA = flag;
+  }
+  void SetContextObjectMap(const Symbol *symbol, Symbol::Flag flag) {
```
tskeith: This function should take a `const Symbol &` because nullptr is never passed in.

lib/semantics/resolve-names.cc
```diff
@@ -1099,7 +1207,62 @@ class OmpVisitor : public virtual DeclarationVisitor {
...
+    GetContext().objectMap.emplace(symbol, flag);
+  }
+  bool HasObject(const Symbol *symbol) {
```
tskeith: Same here: `const Symbol &`

lib/semantics/resolve-names.cc
```diff
@@ -1099,7 +1207,62 @@ class OmpVisitor : public virtual DeclarationVisitor {
...
+    auto it{GetContext().objectMap.find(symbol)};
+    if (it != GetContext().objectMap.end()) {
+      return true;
+    }
+    return false;
+  }
+
+  Symbol &MakeAssocSymbol(const SourceName &name, Symbol &prev) {
+    const auto pair{(*GetContext().scope)
+                        .try_emplace(name, Attrs{}, HostAssocDetails{prev})};
```
tskeith: This should be written `GetContext().scope->try_emplace(`

lib/semantics/resolve-names.cc
```diff
@@ -1099,7 +1207,62 @@ class OmpVisitor : public virtual DeclarationVisitor {
...
+    bool withinConstruct{false};
+  };
+  // back() is the top of the stack
+  OmpContext &GetContext() {
+    CHECK(!ompContext_.empty());
+    return ompContext_.back();
+  }
+  void PushContext(const parser::CharBlock &source, OmpDirective dir) {
+    ompContext_.emplace_back(source, dir);
+    SetContextScope(source);
```
tskeith: If you made this `ompContext_.emplace_back(source, dir, context_.FindScope(source));` then `scope` in `OmpContext` could be a reference so you wouldn't have to worry about whether it is null. And you wouldn't need `SetContextScope`.

lib/semantics/resolve-names.cc
```diff
@@ -1099,7 +1207,62 @@ class OmpVisitor : public virtual DeclarationVisitor {
     return false;
   }
 
+  void Post(const parser::Name &);
+
 private:
+  struct OmpContext {
+    OmpContext(const parser::CharBlock &source, OmpDirective d)
+      : directiveSource{source}, directive{d} {}
+    parser::CharBlock directiveSource{nullptr};
```
tskeith: You don't need to initialize this because it is set by the only constructor.

lib/semantics/resolve-names.cc
```diff
@@ -1099,7 +1207,62 @@ class OmpVisitor : public virtual DeclarationVisitor {
...
+    OmpDirective directive;
+    Scope *scope{nullptr};
+    // TODO: default DSA is implicitly determined in different ways
+    Symbol::Flag defaultDSA{Symbol::Flag::OmpShared};
+    // variables on Data-sharing attribute clauses
+    std::map<const Symbol *, Symbol::Flag> objectMap;
```
tskeith: The name `objectMap` doesn't say much about what it is for. How about `dataSharingAttributes` or `objectToDSA`? Then rename `SetContextObjectMap` and `HasObject` to reflect the new name.

lib/semantics/resolve-names.cc
ichoyjx: I discussed with Tim and Steve, we decided that in the case of `Names` are not successfully resolved, we do nothing on them. I changed all the `ResolveOmp` related functions to return `Symbol *` and function will return `nullptr` in case of errors.

lib/semantics/resolve-names.cc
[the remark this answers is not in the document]
ichoyjx: Actions have been made.

lib/semantics/resolve-names.cc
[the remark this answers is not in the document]
ichoyjx: I am adding a test case for subroutine/function calls within OpenMP constructs and compiler should avoid fixing the `Names` for them.

lib/semantics/resolve-names.cc
ichoyjx: Very cool!

ichoyjx: I've changed the `scope` to reference.

lib/semantics/resolve-names.cc
ichoyjx: I renamed it to `objectWithDSA`.

lib/semantics/resolve-names.cc
```diff
@@ -5776,6 +5777,7 @@ void ResolveNamesVisitor::ResolveSpecificationParts(ProgramTree &node) {
       },
       node.stmt());
   Walk(node.spec());
+  OmpAttributeVisitor{context(), *this}.Walk(node.spec());
```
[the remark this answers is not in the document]
ichoyjx: I've created `ResolveOmpParts` to run through the `root` separately after `ResolveSpecificationParts` and `ResolveExecutionParts`.

lib/semantics/resolve-names.cc
```diff
@@ -1116,20 +1116,18 @@ bool OmpVisitor::NeedsScope(const parser::OpenMPBlockConstruct &x) {
   const auto &beginBlockDir{std::get<parser::OmpBeginBlockDirective>(x.t)};
   const auto &beginDir{std::get<parser::OmpBlockDirective>(beginBlockDir.t)};
   switch (beginDir.v) {
-  case parser::OmpBlockDirective::Directive::TargetData:
-  case parser::OmpBlockDirective::Directive::Master:
+  case parser::OmpBlockDirective::Directive::TargetData: [[fallthrough]];
+  case parser::OmpBlockDirective::Directive::Master: [[fallthrough]];
```
tskeith: There are probably hundreds of other case labels like this where this annotation would apply. What makes these two special?

lib/semantics/resolve-names.cc
```diff
@@ -1238,31 +1223,30 @@ class OmpAttributeVisitor {
...
+  void AddToContextObjectWithDSA(const Symbol &symbol, Symbol::Flag flag) {
+    GetContext().objectWithDSA.emplace(&symbol, flag);
   }
-  bool HasObject(const Symbol *symbol) {
-    auto it{GetContext().objectMap.find(symbol)};
-    if (it != GetContext().objectMap.end()) {
-      return true;
-    }
-    return false;
+  bool HasObjectWithDSA(const Symbol &symbol) {
```
tskeith: `IsObjectWithDSA` would be a better name.

lib/semantics/resolve-names.cc
```diff
@@ -6272,14 +6273,30 @@ void ResolveNamesVisitor::ResolveExecutionParts(const ProgramTree &node) {
...
+void ResolveNamesVisitor::ResolveOmpParts(const ProgramTree &node) {
+  if (!node.scope()) {
+    return;  // error occurred creating scope
+  }
+  SetScope(*node.scope());
+  OmpAttributeVisitor{context(), *this}.Walk(node.spec());
+  if (const auto *exec{node.exec()}) {
+    OmpAttributeVisitor{context(), *this}.Walk(*exec);
+  }
+  PopScope();  // converts unclassified entities into objects
```
tskeith: You don't need this, and the comment is inaccurate because that has already happened.

lib/semantics/resolve-names.cc
```diff
@@ -6272,14 +6273,30 @@ void ResolveNamesVisitor::ResolveExecutionParts(const ProgramTree &node) {
...
   }
   PopScope();  // converts unclassified entities into objects
   for (const auto &child : node.children()) {
     ResolveExecutionParts(child);
   }
 }
 
+// Resolve names for OpenMP in the specification and execution part of this node
+// and its children
+void ResolveNamesVisitor::ResolveOmpParts(const ProgramTree &node) {
```
tskeith: If you don't need to do all of the specs before all of the exes, you can probably just call OmpAttributeVisitor on the entire parse tree for a program unit and get rid of this function.

lib/semantics/resolve-names.cc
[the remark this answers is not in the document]
jeanPerier: OK, I see `objectWithDSA` map is actually not used to get back the flag associated to a symbol, so it does not matter indeed that several flags might be concurrently set in the map for a same symbol.

lib/semantics/resolve-names.cc
ichoyjx: I've removed `ResolveOmpParts` function and called `OmpAttributeVisitor{context(), *this}.Walk(x);` in `ResolveNamesVisitor::Pre(const parser::ProgramUnit &x)` instead. Thanks for the suggestions.

lib/semantics/resolve-names.cc
```diff
@@ -5915,6 +5893,315 @@ class DeferredCheckVisitor {
...
+      break;
+    }
+  }
+}
+
+// For OpenMP constructs, check all the data-refs within the constructs
+// and adjust the symbol for each Name if necessary
+void OmpAttributeVisitor::Post(const parser::Name &name) {
+  if (name.symbol && !ompContext_.empty() && GetContext().withinConstruct) {
+    auto *symbol{name.symbol};
```
tskeith: It would be a little clearer to declare symbol outside the if and test symbol in the if condition.
As it is, it looks like you're dereferencing symbol without checking for null.

lib/semantics/resolve-names.cc
```diff
@@ -5915,6 +5893,315 @@ class DeferredCheckVisitor {
...
+    if (!symbol->owner().IsDerivedType() && !symbol->has<ProcEntityDetails>() &&
+        !IsObjectWithDSA(*symbol)) {
```
tskeith: What about other kinds of symbols that aren't object entities? Can they occur here? What about use-associated names or symbols from ASSOCIATE construct?

ichoyjx: If I understand your question correctly, `!symbol->owner().IsDerivedType() && !symbol->has<ProcEntityDetails>()` is actually not a complete list, they're more like an example in this PR. Except for few OpenMP constructs (for example within `!$omp workshare` constructs, only certain constructs/statements are allowed), most OpenMP constructs can enclose any kind of Fortran `ExecutionParts`. So `AssocEntity`, `ProcEntity`, `ObjectEntity`, symbol with `UseDetails` and etc. can all occur within OpenMP constructs. In the next line 6022, it says:
> // TODO: create a separate function to go through the rules for
      //       predetermined, explicitly determined, and implicitly
      //       determined data-sharing attributes (2.15.1.1).

I think eventually `!symbol->owner().IsDerivedType() && !symbol->has<ProcEntityDetails>()` along with other exclusions will be organized in a function which is part of the steps within the `separate function` mentioned in the above quotation.

About the `ASSOCIATE` construct (listed in the PR description), yes, the symbols from the construct are allowed in OpenMP constructs but it's only legal when `associate-name` and `selector` have the same data-sharing attribute.

PR #941, opened 2020-01-20
Porting F18 tests to use LLVM lit without FileCheck

LukeIreland1:
This approach uses @tskeith's suggestion [here](http://lists.llvm.org/pipermail/flang-dev/2020-January/000149.html) to deliver lit without FileCheck in order to get the non-controversial changes out of the way, and to be able to narrow the focus, as discussed by @RichBarton-Arm [here](http://lists.llvm.org/pipermail/flang-dev/2020-January/000151.html)

The commits just look at error tests and are as follows:

- Porting script that adds lit run lines to use test_errors.sh
- Hand picked error tests just for variation's sake.
- The remaining error tests.


sscalpone:
Are the tests unchanged other than adding the lit line?

For future tests, you might first add the unchanged test & then commit the diffs.

test-lit/lit_only_port.py
```diff
@@ -0,0 +1,232 @@
+""" lit_port.py
+
+    Ports old style F18 tests to be compatible with lit
```
sscalpone: You might replace "old" with "ctest" :)

LukeIreland1:
> Are the tests unchanged other than adding the lit line?
> 
> For future tests, you might first add the unchanged test & then commit the diffs.

Yep, just the lit run line added.

I'm not sure how to do that other than renaming all the old tests, then renaming them in the ported patches.

test-lit/lit.cfg.py
```diff
@@ -26,17 +26,22 @@
...
 
 # test_source_root: The root path where tests are located.
 config.test_source_root = os.path.dirname(__file__)
 
-
 # test_exec_root: The root path where tests should be run.
 config.test_exec_root = os.path.join(config.flang_obj_root, 'test-lit')
 
+config.old_test_source_root = os.path.join(
+    config.test_source_root, "..", "test")
```
RichBarton-Arm: Alternative to this approach is to copy over the old test scripts to the new test directory. Then you can use config.test_source_root.

However, would a neater way to be to use lit's %S variable which expands to the test directory. Your script would insert %S/test_errors.sh instead of %test_error. This could be expanded to all the testing tools without need to plumb them all through lit. I could be wrong, but I think the advantages of the lit tool expansion don't apply to a script (rather than executable) that is sitting next to the test inputs (rather than in some other directory).

RichBarton-Arm:
> > For future tests, you might first add the unchanged test & then commit the diffs.
> I'm not sure how to do that other than renaming all the old tests, then renaming them in the ported patches.

Darn - when Luke and I were playing with this idea internally we used Gerrit for review which shows a move + edit as a diff, so we never hit this issue.

You would need to restructure the script. Instead of expanding user arguments then processing them one by one to read the old script, edit the contents, then write to new location. Instead, you would process them in two passes, first pass to move them to their new location with git cp, then git commit your working copy, then do your editing in place and git commit that. I've not found a python library for git that I like yet, so I would just use cmd to call git. Others may have a recommendation.

Alternatively, you could write a separate little script which uses testSplitter to figure out which tests are error tests and moves them to their new location. Then your flow is to call this script; git commit -a; then call your original script but with --output set to the same dir as input. If that was the flow though, you may as well make the default behaviour of the script to edit in place. Or maybe simplify the original script interface to just take input files to edit in place and lose the --output option.

If we took this route though, we'd want to squash the commits after review though so we don't get a commit in the history with tests in the wrong place and duplicated.

test-lit/lit.cfg.py
sscalpone: Might be good to isolate the scripts in their own directory separate from the tests.

sscalpone:
Why is the test directory called semantics_alt?

LukeIreland1:
> Why is the test directory called semantics_alt?

`python lit_only_port.py -e ../test/semantics --output semantics_alt`
Just so I can keep it separate from the other FileCheck output. The name doesn't actually matter though, I can swap them.

LukeIreland1:
Came up with the same issue where git complains "Updates were rejected because the tip of your current branch is behind its remote counterpart.", so I deleted the branch, pulled my fork again, then made my changes again, and it didn't make a difference. Fortunately, the diff is still available, but the patch history is limited to 1 commit at a time.

tskeith:
The changes to get the tests to run with lit look good to me.

I assume that ultimately the tests will be back under `test/semantics` and there won't be any there that are run by ctest, correct? I'm not clear on how we are going to get there from here. Can you explain what the plan is?

Is it possible to configure lit to run just a subset of the tests in `test/semantics`? Then you could change the tests in place and remove them from `CMakeLists.txt` as they are added to lit. That would allow you to merge at any point in the transition.

LukeIreland1:
> I assume that ultimately the tests will be back under `test/semantics` and there won't be any there that are run by ctest, correct? I'm not clear on how we are going to get there from here. Can you explain what the plan is?

Yep, that's right. Once all tests from test are in test-lit, it's just a case of renaming test-lit to test (and removing the old(ctest) test folder)

> Is it possible to configure lit to run just a subset of the tests in `test/semantics`? Then you could change the tests in place and remove them from `CMakeLists.txt` as they are added to lit. That would allow you to merge at any point in the transition.

I'm just double checking if it's possible to run lit on individual tests, but included as one run with the rest of the tests.

LukeIreland1:
> I'm just double checking if it's possible to run lit on individual tests, but included as one run with the rest of the tests.

So it would work if we made a folder in semantics called "error" for example and split them up like that. Otherwise, there isn't really a nice way to do it, so it may be better off keeping the tests in test-lit for now.

tskeith:
> Yep, that's right. Once all tests from test are in test-lit, it's just a case of renaming test-lit to test (and removing the old(ctest) test folder)

That is all happening in this PR, right?

RichBarton-Arm:
> > Yep, that's right. Once all tests from test are in test-lit, it's just a case of renaming test-lit to test (and removing the old(ctest) test folder)
> 
> That is all happening in this PR, right?

When @LukeIreland1 and I were planning this we envisioned a gradual transition attacking groups of tests in different PRs keeping both styles of tests running together. However, now we have changed approach to moving from ctest to lit first then working out the harder problems of how to port off the shell scripts. Given that, I think doing it in one big go under this PR would work really nicely. 

So this PR would have
 * A patch to add the porting script somewhere
 * A patch that applies the output of that script run on all tests in test, adding them to test-lit as lit tests
 * A patch that moves the testing scripts over to new locations and updates the lit config files accordingly
 * A patch that deletes test and removes the ctest apparatus from the top-level build system and renames test-lit to test.

Does that all sound good?

tskeith:
> So this PR would have
> 
>     * A patch to add the porting script somewhere
> 
>     * A patch that applies the output of that script run on all tests in test, adding them to test-lit as lit tests
> 
>     * A patch that moves the testing scripts over to new locations and updates the lit config files accordingly
> 
>     * A patch that deletes test and removes the ctest apparatus from the top-level build system and renames test-lit to test.
> 
> Does that all sound good?

That sounds good to me, though before approving I would like the see the commits in the form that they will be merged. I assume there will be no evidence of the porting script or test-lit. Just removing ctest configuration, adding lit configuration, and 1-line (or so) changes to each test source file. And maybe some changes to the shell scripts.

LukeIreland1:
> > Yep, that's right. Once all tests from test are in test-lit, it's just a case of renaming test-lit to test (and removing the old(ctest) test folder)
> 
> That is all happening in this PR, right?

Once all ctest tests are in test-lit, in lit form, test can be removed/replaced. That's my understanding of our approach at least.

LukeIreland1:
> > > Yep, that's right. Once all tests from test are in test-lit, it's just a case of renaming test-lit to test (and removing the old(ctest) test folder)
> > 
> > 
> > That is all happening in this PR, right?
> 
> When @LukeIreland1 and I were planning this we envisioned a gradual transition attacking groups of tests in different PRs keeping both styles of tests running together. However, now we have changed approach to moving from ctest to lit first then working out the harder problems of how to port off the shell scripts. Given that, I think doing it in one big go under this PR would work really nicely.
> 
> So this PR would have
> 
> * A patch to add the porting script somewhere
> * A patch that applies the output of that script run on all tests in test, adding them to test-lit as lit tests
> * A patch that moves the testing scripts over to new locations and updates the lit config files accordingly
> * A patch that deletes test and removes the ctest apparatus from the top-level build system and renames test-lit to test.
> 
> Does that all sound good?

Ah I missed this...I'll make sure to read all comments next time.

LukeIreland1:
> > > I'm just double checking if it's possible to run lit on individual tests, but included as one run with the rest of the tests.
> > 
> > 
> > So it would work if we made a folder in semantics called "error" for example and split them up like that. Otherwise, there isn't really a nice way to do it, so it may be better off keeping the tests in test-lit for now.
> 
> Isn't there a way to exclude files in the lit configuration?

Yes, but it would look pretty ugly given the amount of tests in semantics that aren't error tests.

schweitzpgi:
Is there a plan for a "flang -verify" mode for diagnostic testing?

sscalpone:
> Is there a plan for a "flang -verify" mode for diagnostic testing?

No, not that I know of.

schweitzpgi:
> > Is there a plan for a "flang -verify" mode for diagnostic testing?
> 
> No, not that I know of.

It'd be a good project for someone.

CarolineConcatto:
@sscalpone and @schweitzpgi we notice the need for --verify in the driver design. I believe that was Steve that asked that question. I am  still at the beginning of the driver design implementation.
But I believe that we have in mind that --verify needs to be implemented, just don't know when or who is going to do yet. 
If I am not completely wrong it is something in the Diagnosis class that implements that command in the driver. Maybe @peterwaller-arm knows better where it is sitting down for clang  driver.

LukeIreland1:
Working on porting the remaining semantics tests. 5 tests from that folder fail, and 2 of them are to do with compiler output

```
actual at 36: TARGET type 'REAL(8)' is not compatible with POINTER type 'REAL(4)'
expect at 36: Target type REAL(8) is not compatible with pointer type REAL(4)
```

I rebuilt last night, and the test history shows they haven't been updated since, but I will rebuild for these ones and see if they change.

Besides that `modfile09-b/c/d.f90` gives me ```/home/lukire01/f18/f18/test-lit/semantics/modfile09-b.f90:3:11: error: Error reading module file for module 'm'```

These three tests pass in `make test`, but `assign02.f90` and `assign03.f90` do not. I will rebuild and hopefully all 5 are fixed, but I'm confused about the inconsistency.

tskeith:
> Working on porting the remaining semantics tests. 5 tests from that folder fail, and 2 of them are to do with compiler output
> 
> ```
> actual at 36: TARGET type 'REAL(8)' is not compatible with POINTER type 'REAL(4)'
> expect at 36: Target type REAL(8) is not compatible with pointer type REAL(4)
> ```
> 
> I rebuilt last night, and the test history shows they haven't been updated since, but I will rebuild for these ones and see if they change.
> 
> Besides that `modfile09-b/c/d.f90` gives me `/home/lukire01/f18/f18/test-lit/semantics/modfile09-b.f90:3:11: error: Error reading module file for module 'm'`
> 
> These three tests pass in `make test`, but `assign02.f90` and `assign03.f90` do not. I will rebuild and hopefully all 5 are fixed, but I'm confused about the inconsistency.

`modfile09` is tricky because each of the files need to be compiled individually, but it's all part of the same test. Compiling `modfile09-a.f90` produces `m.mod` which needs to be available to compile `modfile09-b.f90`. So far that is probably the only test in `test/semantics` like that.

The other problems are due to your compiler being out of sync with your tests. They were both updated in PR #944.

LukeIreland1:
> > Working on porting the remaining semantics tests. 5 tests from that folder fail, and 2 of them are to do with compiler output
> > ```
> > actual at 36: TARGET type 'REAL(8)' is not compatible with POINTER type 'REAL(4)'
> > expect at 36: Target type REAL(8) is not compatible with pointer type REAL(4)
> > ```
> > 
> > 
> > I rebuilt last night, and the test history shows they haven't been updated since, but I will rebuild for these ones and see if they change.
> > Besides that `modfile09-b/c/d.f90` gives me `/home/lukire01/f18/f18/test-lit/semantics/modfile09-b.f90:3:11: error: Error reading module file for module 'm'`
> > These three tests pass in `make test`, but `assign02.f90` and `assign03.f90` do not. I will rebuild and hopefully all 5 are fixed, but I'm confused about the inconsistency.
> 
> `modfile09` is tricky because each of the files need to be compiled individually, but it's all part of the same test. Compiling `modfile09-a.f90` produces `m.mod` which needs to be available to compile `modfile09-b.f90`. So far that is probably the only test in `test/semantics` like that.
> 
> The other problems are due to your compiler being out of sync with your tests. They were both updated in PR #944.

Well what happens with ctest to make those tests pass? Running test_modfile.sh doesn't produce `m.mod`, yet modfile09-a.f90 passes. I've had a look at where things are deleted in the scripts, but I can't find a way to actually keep `m.mod` after running `modfile09-a.f90`. Merging all 4 tests makes all semantics tests pass, but alternatively, it would have the same effect if they were merged like ab abc abcd rather than just abcd.

Either of those will result in all tests passing with lit, meaning we can move forward with this PR.

tskeith:
> Well what happens with ctest to make those tests pass? Running test_modfile.sh doesn't produce `m.mod`, yet modfile09-a.f90 passes. I've had a look at where things are deleted in the scripts, but I can't find a way to actually keep `m.mod` after running `modfile09-a.f90`.

To ctest this is all one test. It passes `modfile09-*.f90` to `test_modfile.sh` which expands that to the 4 source files, performs each compilation, checks the .mod files, and cleans up.

You might be able to make this work by having the RUN command for `modfile09-a.f90` be:
  `%S/test_modfile.sh 'modfile09-*.f90' %flang`
Then for the other modfile09 tests put in a dummy run command (or exclude them).

RichBarton-Arm:
> You might be able to make this work by having the RUN command for `modfile09-a.f90` be:
> `%S/test_modfile.sh 'modfile09-*.f90' %flang`
> Then for the other modfile09 tests put in a dummy run command (or exclude them).

+1, this is the way to write a lit test that uses multiple files. The standard lit pattern for doing this is to put the input files in a directory called "Inputs", which is explicitly excluded from test discovery. This prevents lit from telling you that there are no RUN lines in the other files. I think the test would be basically an empty file called modfile09.f90 with the RUN line like above. 

Were we to later port this test to FileCheck, the file would contain four runlines like:
`RUN: %flang -fdebug-names -fparse-only %S/Inputs/modfile-a.f90 | FileCheck %S/Inputs/modfile-a.f90`

test-lit/mm3.mod
```diff
@@ -0,0 +1,4 @@
+﻿!mod$ v1 sum:8d1e01df5c32946e
```
LukeIreland1: I'm not sure how this managed to end up in this commit, I specifically went out of my way to avoid that...Will remove in next patch set.

test-lit/lit.cfg.py
```diff
@@ -64,6 +64,10 @@
 
 tools = [ToolSubst('%flang', command=FindTool('flang'), unresolved='fatal'),
          ToolSubst('%f18', command=FindTool('f18'), unresolved='fatal'),
-         ToolSubst('%test_error', command=FindTool('test_errors.sh'), unresolved='fatal')]
+         ToolSubst('%test_error', command=FindTool('test_errors.sh'), unresolved='fatal'),
+         ToolSubst('%test_symbol', command=FindTool('test_symbols.sh'), unresolved='fatal'),
+         ToolSubst('%test_modfile', command=FindTool('test_modfile.sh'), unresolved='fatal'),
+         ToolSubst('%test_generic', command=FindTool('test_any.sh'), unresolved='fatal')
```
tskeith: Why are you doing it this way rather than using `%S` to find the scripts, as @RichBarton-Arm suggested?

LukeIreland1: Looks nicer to me, I don't mind doing it the other way, but until the folder movement happens (after all tests are ported) then this is probably a better idea for that reason.

tskeith: It's fine with me to wait until then to change it. The problem with this approach is that it implies the scripts might be somewhere other than in the test directory (i.e. wherever FindTools looks). Using %S tells the reader exactly where to find them.

tskeith:
When this PR is in its final form (i.e. ready to merge) it would make it a lot easier on reviewers if the mechanical changes are in a separate commit from the hand-written changes.

test/evaluate/test_folding.sh
```diff
@@ -104,3 +95,5 @@ else
   echo all $passed tests passed
   echo PASS
 fi
+
+[[ $KEEP ]] || trap "rm -rf $temp" EXIT
```
tskeith: By moving this down here it means the temp directory won't get deleted in case of an early exit. That was the whole point of using trap.

test/evaluate/test_folding.sh
```diff
@@ -52,12 +48,7 @@ actual_warnings=$temp/actwarnings.log
 expected_warnings=$temp/expwarnings.log
 warning_diffs=$temp/warnings.diff
 
-if ! ( cd $temp; $CMD $src ) > $src1 2> $messages # compile, dumping symbols
-then
-  cat $messages
-  echo FAIL compilation
-  exit 1
-fi
+( cd $temp; $CMD $src ) > $src1 2> $messages # compile, dumping symbols
```
tskeith: The test is supposed to fail if there are compilation errors. Does that still happen?

test/evaluate/test_folding.sh
LukeIreland1: I was having an issue getting this to work, as it was complaining about the directory already existing, directly after the `rm -rf`, as well as the logs not existing. I suppose moving this here was part of me debugging it. I'll move it back and see if it still works, I imagine so.

test/evaluate/test_folding.sh
LukeIreland1: This code can also probably return, as per my previous comment, it complained messages didn't exist, so couldn't `cat` it, yet it needed to exist in order to pass the clause.

test-lit/lit.cfg.py
LukeIreland1: If the idea is to start replacing these scripts with FileCheck or `--verify`, why would the reader need to know where they are? Also, `find` and `grep` still exist until then, but eventually the scripts will be in the same folder as the test, so it will be very straightforward.

test/evaluate/test_folding.sh
tskeith: A better way to handle the temp directory is to have the run command pass in `%t` and use that. Then presumably lit takes care of cleaning up so the scripts don't have to worry about it.

test-lit/semantics/getdefinition03-a.f90
```diff
@@ -0,0 +1,15 @@
+! Tests -fget-definition with INCLUDE
```
RichBarton-Arm: I second @tskeith's request to separate these manual changes out from the scripted ones. The code review required for each is very different. For manual changes, we'll be looking at the changes to the test files line by line. For the scripted ones, most of the review will be on the script logic, with ported tests there simply to aid the review on the script. I would not look at each ported test line by line, or at least not very carefully. By combining changes like this together into one big patch you are forcing me to either check each ported one line by line, which is overkill and/or increasing the risk that I would not review the manual changes.

test-lit/mm3.mod
```diff
@@ -1,4 +0,0 @@
-﻿!mod$ v1 sum:8d1e01df5c32946e
```
RichBarton-Arm: It's not clear to me where this file was introduced but it should not have been.

test-lit/test_splitter.py
```diff
@@ -20,22 +19,25 @@ def __init__(self):
...
-        root = os.getcwd()
-        while not root.endswith("f18"):
-            root = os.path.abspath(os.path.dirname(root))
+        self.test_types = ['ERROR', 'MODFILE', 'SYMBOL', 'GENERIC', 'FOLDING']
+        root = Path(__file__).resolve()
+        while root.name != "f18":
+            root = root.parent
         self.root = root
-        self.test_path = os.path.join(self.root, "test")
+        self.test_path = self.root / "test"
```
RichBarton-Arm: It's not the most important thing in the world, but using os.path.join is always better than hard coding path separators like this. I really don't understand why you have gone backwards from this. What's the rationale?

test-lit/semantics/canondo02.f90
```diff
@@ -1,6 +1,6 @@
 ! negative test -- invalid labels, out of range
-
-! RUN: ${F18} -funparse-with-symbols %s 2>&1 | ${FileCheck} %s
+!RUN: %test_generic %s %flang
+! EXEC: ${F18} -funparse-with-symbols %s 2>&1 | ${FileCheck} %s
```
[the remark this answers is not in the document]
tskeith: I think it would make sense to use the real FileCheck. You might even be able to eliminate test_any.sh and have these tests use just FileCheck. I'm not very familiar with these tests. @schweitzpgi might have more insight.

test-lit/semantics/omp-symbol01.f90
```diff
@@ -1,3 +1,4 @@
+!RUN: %test_symbol %s %flang
```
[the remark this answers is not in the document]
tskeith: I think you can eliminate any use of the F18 environment variable.

test-lit/semantics/canondo02.f90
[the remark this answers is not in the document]
schweitzpgi: The intent was that test_any.sh was to be a throwaway stop gap script until such time as someone decided to replace ctest and the homegrown scripts with lit and FileCheck.

It appears that time has arrived.

test-lit/test_splitter.py
LukeIreland1: Unifying the use of Pathlib over os.path across the files. The alternative to this is self.root.joinpath("test") so I'll switch to that if it looks nicer.

test-lit/semantics/omp-symbol01.f90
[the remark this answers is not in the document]
LukeIreland1: All 4 scripts are used in the form described in common.sh:
1. Script
2. Test
3. F18
I'm not sure how else I'm supposed to use these scripts (It fails if I remove it). Also, I can't find your comment for test_any.sh @RichBarton-Arm

test/semantics/test_any.sh
```diff
@@ -48,9 +48,9 @@ function internal_check() {
 }
 
 gr=0
-for input in ${srcdir}/$*; do
+for input in "$@"; do
```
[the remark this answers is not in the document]
LukeIreland1: Found it...GitHub hiding things.

test_any.sh is called the way you said it's called now, so nothing has changed (except the paths work from lit's perspective instead of the script's). I copied this bit from test_errors.sh, trying to make them all run the same way (due to the issue of paths getting mixed up)

```
foreach(test ${LABEL_TESTS} ${CANONDO_TESTS} ${DOCONCURRENT_TESTS}
             ${CRITICAL_TESTS} ${GETSYMBOLS_TESTS} ${GETDEFINITION_TESTS})
  add_test(NAME ${test}
    COMMAND ${CMAKE_CURRENT_SOURCE_DIR}/test_any.sh ${test} ${F18})
endforeach()
```

test-lit/semantics/canondo02.f90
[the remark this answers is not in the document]
LukeIreland1: I'm happy to use lit and FileCheck for generic/any tests. Although, it makes the patch name confusing...

As for @RichBarton-Arm's comment, test_any needs %flang/%f18 to stop common.sh from complaining, but as you point out, it isn't actually used in test_any.sh. But that is yet another reason to just use actual FileCheck instead in this instance.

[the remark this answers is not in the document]
sscalpone: This PR is about enabling lit in the test infrastructure.  A future PR can focus on test content.

test/semantics/test_any.sh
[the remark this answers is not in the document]
RichBarton-Arm: Apologies - I misunderstood the cmake syntax there. Now you have explained I can see that the lists are concatenated and expanded so test takes the each value in this iteration and the multiple file examples are actually globs in the lists themselves.

So I guess what we are seeing here is a mostly harmless bug in the original script in that it will try and treat the compiler passed on the last argument as a test input. When I run a test I see this: 
`$ ./test_any.sh critical04.f90 /home/ricbar01/ossf18/Release/bin/flang

PASS  ./critical04.f90
PASS  /home/ricbar01/ossf18/Release/bin/flang` 
Somehow $input on that second iteration of the loop is the absolute path and not $srcdir/<path> so it passes the check on filename with that absolute path. I guess that is a neat bash trick.

In this case, I think your changes to preserve the original semantics are fine.

test-lit/semantics/canondo02.f90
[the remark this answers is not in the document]
RichBarton-Arm: RE: my original comment - I see now that common.sh sets up the $F18 variable from the input, so that was the missing link I could not see in the original review. Apologies again.

So given my misunderstanding is corrected then we can proceed with the current plan to port to lit only. It is a bit odd that lit is calling test_any.sh which is going to try and ape lit itself and call a FileCheck-like utility but we can swiftly fix that after this PR lands. Agree to stay the course.

test-lit/semantics/omp-symbol01.f90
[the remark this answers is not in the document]
RichBarton-Arm: Agreed - you can ignore my comment as per the comment on test_any.sh

test-lit/semantics/modfile35.f90
```diff
@@ -35,6 +35,7 @@ subroutine test3(x, y, a)
     real :: a(x + y)
   end
 end
+!RUN: %test_modfile %s %f18
```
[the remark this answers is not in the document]
RichBarton-Arm: Clearly you can ignore this one too.

test-lit/test_splitter.py
RichBarton-Arm: Yes please

PR #943, opened 2020-01-21
Moved public headers to include/flang

DavidTruby:
This PR changes the folder structure to closer match that used by other LLVM projects. I have moved any headers that are used by files outside a particular "module" e.g. common, evaluate, semantics etc into include/flang/_module_. This includes any files that are transitively included. This avoids relative include paths other than "thing.h". 

This folder structure and the slightly altered include order that goes with it are detailed in https://llvm.org/docs/CodingStandards.html


tskeith:
I think the `IncludeCategories` in `.clang-format` need to be updated. The "../" includes were sorted in a particular place in the list of includes and that is where "flang/" includes should be located.

As it is, the "flang/" includes are sorted in with the include of files in the same directory.

include/flang/common/Fortran-features.h
```diff
@@ -9,9 +9,9 @@
 #ifndef FORTRAN_COMMON_FORTRAN_FEATURES_H_
 #define FORTRAN_COMMON_FORTRAN_FEATURES_H_
```
kiranchandramohan: Should these guards be also renamed to FLANG_COMMON?

include/flang/evaluate/call.h
```diff
@@ -11,10 +11,10 @@
...
 #include "common.h"
 #include "constant.h"
+#include "flang/common/indirection.h"
+#include "flang/common/reference.h"
 #include "formatting.h"
 #include "type.h"
-#include "../common/indirection.h"
-#include "../common/reference.h"
 #include "../parser/char-block.h"
 #include "../semantics/attr.h"
```
tskeith: Why are these still using "../"?

include/flang/common/format.h
```diff
@@ -9,8 +9,8 @@
 #ifndef FORTRAN_COMMON_FORMAT_H_
 #define FORTRAN_COMMON_FORMAT_H_
 
-#include "Fortran.h"
 #include "enum-set.h"
```
kiranchandramohan: flang/common/enum-set.h ?

tskeith:
I suppose the ridiculous comments at the start of each `.h` file that shows its path have to be updated to reflect their new location.

kiranchandramohan:
A few headers were used in both the same directory and the test directory. Should they be included?
./lib/evaluate/intrinsics-library-templates.h
./lib/evaluate/host.h
./runtime/transformational.h
./runtime/descriptor.h
./runtime/format.h

Also, the following one in runtime and modules
./runtime/magic-numbers.h

DavidTruby:
> A few headers were used in both the same directory and the test directory. Should they be included?
> ./lib/evaluate/intrinsics-library-templates.h
> ./lib/evaluate/host.h
> ./runtime/transformational.h
> ./runtime/descriptor.h
> ./runtime/format.h
I decided that headers only used by tests are not necessarily public, but I'm happy to revisit this if you think they should be.
> 
> Also, the following one in runtime and modules
> ./runtime/magic-numbers.h
This one looks like a mistake that I missed!

include/flang/evaluate/call.h
DavidTruby: I'll fix this mistake

include/flang/common/Fortran-features.h
DavidTruby: They should probably be changed to `LLVM_FLANG_COMMON_FORTRAN_FEATURES_H` to match clang, but I think that should be a separate PR with a separate discussion

include/flang/common/format.h
DavidTruby: it's in the same directory so it doesn't need changing necessarily. Happy to change them all to be absolute if you think that's better though

DavidTruby:
> I suppose the ridiculous comments at the start of each `.h` file that shows its path have to be updated to reflect their new location.

I've fixed this in the latest commit.

tskeith:
> I decided that headers only used by tests are not necessarily public, but I'm happy to revisit this if you think they should be.

I agree. Tests should be able to use non-public headers without having to make them public.

DavidTruby:
> The intended different between the namespace names Fortran and Flang (if it exists), is that Fortran deals with parsing and checking Fortran and Flang is an environment where that code operates, Flang being a compiler. You can imagine that Fortran might be used in a different tool that would have its own namespace and use Fortran, not Flang.

The intention with the include guards at least in the other LLVM projects is that they refer to the path of the header within the repository, rather than anything conceptual about the header itself. Hence my suggestion to change them to `LLVM_FLANG_COMMON_FORTRAN_FEATURES_H` to match the header guards in the other LLVM projects.

I think the discussion of this and a decision whether to change or not is outside the scope of this PR and should be tackled separately though.

klausler:
> > A few headers were used in both the same directory and the test directory. Should they be included?
> > ./lib/evaluate/intrinsics-library-templates.h
> > ./lib/evaluate/host.h
> > ./runtime/transformational.h
> > ./runtime/descriptor.h
> > ./runtime/format.h
> 
> I decided that headers only used by tests are not necessarily public, but I'm happy to revisit this if you think they should be.
> 
> > Also, the following one in runtime and modules
> > ./runtime/magic-numbers.h
> 
> This one looks like a mistake that I missed!

`magic-numbers.h` does not define an API and should not be publicized.  It exists so that some arbitrary constant values used in both Fortran and C++ source files can be defined in one place.

tskeith:
Is `include/` supposed to define the API between components of f18 or between f18 and clients outside of f18? Hopefully not the latter because most of these headers should not be part of that API.

DavidTruby:
It is meant to define both the api between "modules" within f18 (eg semantics and parser) and the api for external users. The coding guidelines don't seem to make a distinction between the two things. So headers should be in there if they are used outside the module that they are in.

klausler:
> It is meant to define both the api between "modules" within f18 (eg semantics and parser) and the api for external users. The coding guidelines don't seem to make a distinction between the two things. So headers should be in there if they are used outside the module that they are in.

Needlessly expanding the "surface area" of the public API is bad for both clients of that API and future maintainers of the implementation.

DavidTruby:
There's no implication that things in the public headers need to have a stable API, for example things in clang's public headers change all the time. 

I believe the rationale for the decision is more that if something from one module is used by another, that thing is likely going to be useful to tools that might want to sit between those two layers. And given that things being in the public header path doesn't imply that they have a stable API there's not really a reason not to expose things that might be useful.

klausler:
> There's no implication that things in the public headers need to have a stable API, for example things in clang's public headers change all the time.
> 
> I believe the rationale for the decision is more that if something from one module is used by another, that thing is likely going to be useful to tools that might want to sit between those two layers. And given that things being in the public header path doesn't imply that they have a stable API there's not really a reason not to expose things that might be useful.

If I modify an internal detail of an f18 component that you have published, and it breaks the build of some future in-tree component that will have been written with a dependence on that internal detail, then I have to also add concomitant changes to that in-tree component in order to pass the build bots, yes?

DavidTruby:
> If I modify an internal detail of an f18 component that you have published, and it breaks the build of some future in-tree component that will have been written with a dependence on that internal detail, then I have to also add concomitant changes to that in-tree component in order to pass the build bots, yes?

Ideally yes, in the same way that you would if your change broke another component of F18. The LLVM project is supposed to be a single project and repository, and that does of course have consequences for us merging into it. One of those consequences is that if we break other components of the project with a change we should fix that.

klausler:
> > If I modify an internal detail of an f18 component that you have published, and it breaks the build of some future in-tree component that will have been written with a dependence on that internal detail, then I have to also add concomitant changes to that in-tree component in order to pass the build bots, yes?
> 
> Ideally yes, in the same way that you would if your change broke another component of F18. The LLVM project is supposed to be a single project and repository, and that does of course have consequences for us merging into it. One of those consequences is that if we break other components of the project with a change we should fix that.

Then we should restrict the published APIs to reduce the risk of accidental in-tree dependence upon implementation details, yes?

DavidTruby:
> Then we should restrict the published APIs to reduce the risk of accidental in-tree dependence upon implementation details, yes?

There's already an in-tree dependence on all of the headers I've moved to the include directory. Any module-private headers I have left in the lib directory.

klausler:
> > Then we should restrict the published APIs to reduce the risk of accidental in-tree dependence upon implementation details, yes?
> 
> There's already an in-tree dependence on all of the headers I've moved to the include directory. Any module-private headers I have left in the lib directory.

I don't believe that we enjoy the same understanding of what should constitute a "module" from the perspective of code outside the f18 front-end.  We think of `lib/parser/parsing.h` and `lib/semantics/semantics.h` as the main APIs of the f18 front-end; specifically, the public interfaces of the classes `Parsing` and `Semantics`.

DavidTruby:
LLVM doesn't make a distinction between things being publicly accessible to other modules in the same subproject and things being publicly accessible in general, mostly due to the tooling case. Rather, the LLVM project doesn't make any guarantee of API stability between releases.

As an example, semantics/check-allocate.cc includes and uses a number of things from the parser and evaluate modules that it then uses to do its semantic checks. Lets imagine you wanted to write a semantic plugin that enforced some extra checks about allocations in your program; presumably you would need those things from parser and evaluate as well, since that's how the existing semantic checks are implemented. Therefore, those things should be public so that everyone can use them.

We could construct a stable API around each module and only expose that, however even in that case the other modules in F18 should use that and not non-public interfaces, if only to prove that the public interface is complete enough. I'm not sure how useful this would be anyway though given that the rest of the llvm project doesn't provide the same guarantee.

sscalpone:
There's an unwritten but implied assumption in llvm guidelines that a directory and module have a 1-to-1 mapping.  In f18, it may be the case that _semantics_ and _evaluate_ together form a single logical module but spread over two directories.  Perhaps the the include files that are shared only between these two directories should be consider private to those two modules and excluded from the top-level include directory.

jdoerfert:
> There's an unwritten but implied assumption in llvm guidelines that a directory and module have a 1-to-1 mapping. In f18, it may be the case that _semantics_ and _evaluate_ together form a single logical module but spread over two directories. Perhaps the the include files that are shared only between these two directories should be consider private to those two modules and excluded from the top-level include directory.

I think you can very well have "public' headers included from various potentially unrelated places. IMHO, "private" headers should be used rarely.

tskeith:
`unparse-with-symbols.h` still has an instance of `#include "../`...

DavidTruby:
The force pushes here are just rebases on master with no code changes, let me know if you want me to squash the other 3 commits before merge. Thanks!

tskeith:
> The force pushes here are just rebases on master with no code changes, let me know if you want me to squash the other 3 commits before merge.

I think they should be squashed. It doesn't add anything to the history to have them separate.

DavidTruby:
> > The force pushes here are just rebases on master with no code changes, let me know if you want me to squash the other 3 commits before merge.
> 
> I think they should be squashed. It doesn't add anything to the history to have them separate.

Done!

sscalpone:
@DavidTruby please fix the conflict & squash again.

DavidTruby:
@sscalpone this should be safe to merge now, thanks!

PR #944, opened 2020-01-22
Check bounds on pointer assignment

tskeith:
Perform checks on bounds-spec and bounds-remapping in a pointer
assignment statement:
- check that the rank of the bounds specified matches the rank of the
  pointer
- for bounds-spec, check that the pointer rank matches the target rank
- for bounds-remapping:
  - check that the target is rank 1 or simply contiguous
  - check that there are sufficient elements on the RHS for the bounds
    specified, when it can be determined at compile time

Move more of the pointer-specific checking from `assignment.cc`
to `pointer-assignment.cc`.

With this PR pointer assignment checks should be complete.

### DIFF
--- a/lib/semantics/pointer-assignment.cc
+++ b/lib/semantics/pointer-assignment.cc
@@ -326,19 +326,15 @@ void CheckPointerAssignment(
     evaluate::FoldingContext &context, const Symbol &lhs, const SomeExpr &rhs) {
   // TODO: Acquire values of deferred type parameters &/or array bounds
   // from the RHS.
-  if (!IsPointer(lhs)) {
-    evaluate::SayWithDeclaration(
-        context.messages(), lhs, "'%s' is not a pointer"_err_en_US, lhs.name());
-  } else {
-    std::string description{"pointer '"s + lhs.name().ToString() + '\''};
-    PointerAssignmentChecker{lhs.name(), description, context}
-        .set_lhsType(TypeAndShape::Characterize(lhs, context))
-        .set_procedure(Procedure::Characterize(lhs, context.intrinsics()))
-        .set_lhs(lhs)
-        .set_isContiguous(lhs.attrs().test(Attr::CONTIGUOUS))
-        .set_isVolatile(lhs.attrs().test(Attr::VOLATILE))
-        .Check(rhs);
-  }
+  CHECK(IsPointer(lhs));
+  std::string description{"pointer '"s + lhs.name().ToString() + '\''};
+  PointerAssignmentChecker{lhs.name(), description, context}
+      .set_lhsType(TypeAndShape::Characterize(lhs, context))
+      .set_procedure(Procedure::Characterize(lhs, context.intrinsics()))
+      .set_lhs(lhs)
+      .set_isContiguous(lhs.attrs().test(Attr::CONTIGUOUS))
+      .set_isVolatile(lhs.attrs().test(Attr::VOLATILE))
+      .Check(rhs);
 }
 
 void CheckPointerAssignment(evaluate::FoldingContext &context,
--- a/lib/semantics/pointer-assignment.cc
+++ b/lib/semantics/pointer-assignment.cc
@@ -33,21 +33,31 @@ using evaluate::characteristics::DummyDataObject;
 using evaluate::characteristics::FunctionResult;
 using evaluate::characteristics::Procedure;
 using evaluate::characteristics::TypeAndShape;
+using parser::MessageFixedText;
+using parser::MessageFormattedText;
+using PointerAssignment = evaluate::Assignment::PointerAssignment;
 
 class PointerAssignmentChecker {
 public:
-  PointerAssignmentChecker(parser::CharBlock source,
-      const std::string &description, evaluate::FoldingContext &context)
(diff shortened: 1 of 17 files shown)


lib/semantics/pointer-assignment.cc
```diff
@@ -322,29 +345,93 @@ parser::Message *PointerAssignmentChecker::Say(A &&... x) {
...
+      },
+      assignment.bounds)};
+  if (numBounds > 0) {
+    if (lhs.Rank() != static_cast<int>(numBounds)) {
+      messages.Say("Pointer '%s' has rank %d but the number of bounds specified"
+                   " is %d"_err_en_US,
+          lhs.AsFortran(), lhs.Rank(), numBounds);  // C1018
+    }
+  }
+  if (isBoundsRemapping && rhs.Rank() > 1 &&
```
jeanPerier: Should it be `rhs.Rank() != 1` instead or is the case `rhs.Rank() == 0` already handled ?

lib/semantics/pointer-assignment.cc
```diff
@@ -322,29 +345,93 @@ parser::Message *PointerAssignmentChecker::Say(A &&... x) {
   return msg;
 }
 
+static bool CheckPointerBounds(
```
jeanPerier: The function may deserve a small comment to explain that the returned `bool` does not indicate error/ok but bounds/bounds remapping.

lib/semantics/pointer-assignment.cc
tskeith: Currently when the rhs is a scalar, `IsSimplyContiguous` returns true. That seems reasonable, though I'm not positive it is correct.

I'll change this to `!= 1` because it more correctly expresses the check. And if we decide a scalar is not simply contiguous and change the function, this will be doing the right thing.

lib/semantics/pointer-assignment.cc
tskeith: Good idea, I've done that.

lib/semantics/pointer-assignment.cc
```diff
@@ -33,17 +33,25 @@ using evaluate::characteristics::DummyDataObject;
...
 public:
-  PointerAssignmentChecker(parser::CharBlock source,
-      const std::string &description, evaluate::FoldingContext &context)
-    : source_{source}, description_{description}, context_{context} {}
-  PointerAssignmentChecker &set_lhs(const Symbol &);
+  PointerAssignmentChecker(evaluate::FoldingContext &context,
+      parser::CharBlock source, const std::string &description)
+    : context_{context}, source_{source}, description_{description} {}
+  PointerAssignmentChecker(evaluate::FoldingContext &context, const Symbol &lhs)
+    : context_{context}, source_{lhs.name()},  //
```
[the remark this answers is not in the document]
tskeith: Good question -- I'll get rid of it.

lib/semantics/pointer-assignment.cc
klausler: The *simply contiguous* property is defined in the standard only as an attribute of arrays, but it seemed to me to be digging a pitfall to not further define it vacuously true for scalars.

lib/semantics/pointer-assignment.cc
```diff
@@ -180,7 +184,9 @@ void PointerAssignmentChecker::Check(const evaluate::Designator<T> &d) {
     context_.messages().Say("Pointer target is not a named entity"_err_en_US);
     return;
   }
-  std::optional<parser::MessageFixedText> msg;
+  std::variant<std::monostate, parser::MessageFixedText,
```
klausler: Might be more clear if `std::optional<std::variant<parser::MessageFixedText, parser::MessageFormattedText>>`, avoiding the use of `std::monostate`.

test/semantics/assign02.f90
```diff
@@ -30,18 +30,18 @@ subroutine s1
     logical, target :: l
     real, pointer :: p
     p => r
-    !ERROR: TARGET type 'REAL(8)' is not compatible with POINTER type 'REAL(4)'
+    !ERROR: Target type REAL(8) is not compatible with pointer type REAL(4)
```
klausler: Why no longer capitalizing the keywords?  We capitalize `REAL` and `INTEGER`.

lib/semantics/pointer-assignment.cc
tskeith: OK

test/semantics/assign02.f90
tskeith: I have been capitalizing keywords when they refer directly to elements of Fortran source, but here "target" and "pointer" are just describing the RHS and LHS of a pointer assignment statement, not the attributes. Another example: "... the target is not an object with POINTER or TARGET attribute".

But this is kind of vague and I would be happy to have more concrete guidelines. Currently in master we have 45 messages with lower-case "pointer" in them. What should we do with "pointer assignment statement" or the RHS of one which we sometimes call the "pointer target" or "POINTER target" but not "POINTER TARGET"?

klausler: That's fine.  There's some capitalized `POINTER` and `TARGET` keywords used in messages in call semantics checking that we should lower for consistency (not necessarily now).

PR #945, opened 2020-01-22
Bump clang-format version and remove some differences from LLVM

DavidTruby:
Hi all,

The purpose of this PR is to start a discussion about whether we can move towards matching the LLVM style by accepting some/all of the formatting changes present here, and whether there are some changes we really cannot accept so that we can start a discussion in the wider community about those.

I changed the F18 clang-format file to match the base LLVM style to see how far from matching we are. The changes are attached in this PR for comparison.

It seems that most of the changes are due to LLVM having spaces after the template keyword, and lining up function arguments differently.

I've also run this with clang-format-9 instead of 7, but the only difference between 7 and 8+ is 8+ has an additional pass related to the indentation of lambdas when used as function arguments.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -3,16 +3,10 @@
 BasedOnStyle: LLVM
 AlignAfterOpenBracket: DontAlign
 AlignEscapedNewlines: DontAlign
-AllowShortCaseLabelsOnASingleLine: true
-AllowShortIfStatementsOnASingleLine: true
 AlignConsecutiveAssignments: false
 AlignConsecutiveDeclarations: false
 AlignOperands: false
 AlignTrailingComments: false
-ConstructorInitializerIndentWidth: 2
-SpaceAfterTemplateKeyword: false
-SpacesBeforeTrailingComments: 2
-FixNamespaceComments: false
 IncludeCategories:
   - Regex:           '^<'
     Priority:        4
--- a/documentation/C++style.md
+++ b/documentation/C++style.md
@@ -88,9 +88,9 @@ well as you do and avoid distracting her by calling out usage of new
 features in comments.
 
 ### Layout
-Always run `clang-format` before committing code.  Other developers should
-be able to run `git pull`, then `clang-format`, and see only their own
-changes. Use `clang-format` from llvm 7.
+Always run `clang-format` on your changes before committing code. LLVM
+has a `git-clang-format` script to facilitate running clang-format only
+on the lines that have changed.
 
 Here's what you can expect to see `clang-format` do:
 1. Indent with two spaces.
--- a/documentation/PullRequestChecklist.md
+++ b/documentation/PullRequestChecklist.md
@@ -31,7 +31,7 @@ can also be used when reviewing pull requests.
 ## Follow the style guide
 The following items are taken from the [C++ style guide](C++style.md).  But
 even though I've read the style guide, they regularly trip me up.
-*  Run clang-format version 7 on all .cpp and .h files.
+*  Run clang-format using the git-clang-format script from LLVM HEAD.
 *  Make sure that all source lines have 80 or fewer characters.  Note that
    clang-format will do this for most code.  But you may need to break up long
    strings.


peterwaller-arm:
By the way -- there is a file in the llvm monorepo where you can put git commit shas of commits that touch lots of things like this, and can be used to prevent `git blame` contamination from changes like this:

See https://github.com/llvm/llvm-project/blob/9dc9f7ca145e7fd5fafbdf071a2e5b5914918c04/.git-blame-ignore-revs

.clang-format
```diff
@@ -1,18 +1,6 @@
...
 # See: https://clang.llvm.org/docs/ClangFormatStyleOptions.html
 BasedOnStyle: LLVM
-AlignAfterOpenBracket: DontAlign
-AlignEscapedNewlines: DontAlign
-AllowShortCaseLabelsOnASingleLine: true
-AllowShortIfStatementsOnASingleLine: true
-AlignConsecutiveAssignments: false
-AlignConsecutiveDeclarations: false
-AlignOperands: false
-AlignTrailingComments: false
```
klausler: Comment alignment needlessly turns small code changes into large code changes.

tskeith: > Comment alignment needlessly turns small code changes into large code changes.

Not just comments. clang-format by default loves to align things, function call arguments, declarations, backslashes at the end of multi-line macros.

The main difference we see between clang-format 7 and 8 is how lambdas in `common::visitors` are now sometimes aligned. I couldn't find a way to turn that one off.

tskeith:
I think this is a bad idea. The default rules for aligning things is a maintenance headache as @klausler points out.

I notice that neither clang nor llvm have their source files formatted in the "LLVM style" either. In llvm only a third of source files are properly formatted and in clang it's only a quarter. So I don't see a strong consistency argument against minor differences between f18 formatting and LLVM style.

klausler:
> I think this is a bad idea. The default rules for aligning things is a maintenance headache as @klausler points out.
> 
> I notice that neither clang nor llvm have their source files formatted in the "LLVM style" either. In llvm only a third of source files are properly formatted and in clang it's only a quarter. So I don't see a strong consistency argument against minor differences between f18 formatting and LLVM style.

I agree that there are technical reasons why this is a bad idea, but this isn't being done for technical reasons.  Maybe we can find a `.clang-format` file elsewhere in llvm that doesn't force alignments (something in MLIR would be best) and use it as a precedent to show that sufficient sacrifices have been made.

schweitzpgi:
MLIR's .clang-format file is
```
BasedOnStyle: LLVM
AlwaysBreakTemplateDeclarations: Yes
```
Proper clang-formatting was/is enforced by build bots.

DavidTruby:
I'm not sure I understand why comment alignment affects the size of a change at a glance, other than of course in this first run. I don't suppose you have an example of where it causes a larger than necessary change?

If we want to have a different style to the rest of the llvm project I think that's a discussion we should have with the llvm community at large; it may be that if we can provide good arguments for the current style then we won't need to change it. However up till now this charge has been requested as part of our move towards the monorepo.

tskeith:
> I'm not sure I understand why comment alignment affects the size of a change at a glance, other than of course in this first run. I don't suppose you have an example of where it causes a larger than necessary change?

Look at RoundingMode in common/Fortran.h. If you were to add a new enumerator whose name is longer than TiesAwayFromZero, five extra lines would change to adjust to the new comment alignment.

> If we want to have a different style to the rest of the llvm project I think that's a discussion we should have with the llvm community at large; it may be that if we can provide good arguments for the current style then we won't need to change it. However up till now this charge has been requested as part of our move towards the monorepo.

What is the standard that we are supposed to meet? Is it "all files must be formatted in the LLVM style"? LLVM and clang are nowhere near meeting that so why is it important that we do? mlir has its own non-standard .clang-format too.

DavidTruby:
I can't comment on clang or llvm; however conforming to the LLVM style is something we've been requested to do as the LLVM committing guidelines do require it for new code.

If any of the style changes are absolute blockers for us, could we narrow down to specifically which ones they are and start a discussion with the LLVM community as to wheather leaving those divergent from the rest of the project is acceptable?

jeanPerier:
Moving clang-format discussion for #959 to centralize clang-format discussions.
Topic: Adding `fir|mlir|mlir-c` to the `IncludeCategories` with priority 4 (same as `llvm`) when `flang` has priorty 3.

Comments from @tskeith :
> Shouldn't fir be at the same priority as flang? It's in this (sub-)project but the rest listed here are not.

Reply from @schweitzpgi : https://github.com/flang-compiler/f18/pull/959/files/c36740a747d01885027c86fead0448d161fb4b4f#r372482133
> The second reason fir is grouped with mlir which is grouped with llvm is that fir is a dialect of mlir and makes no sense without mlir (which transitively makes no sense without llvm). Conversely, the FIR dialect has no dependence on the f18 front-end and could be targeted by other Fortran front-ends.

Any other opinions ?

tskeith:
> Moving clang-format discussion for #959 to centralize clang-format discussions.
> Topic: Adding `fir|mlir|mlir-c` to the `IncludeCategories` with priority 4 (same as `llvm`) when `flang` has priorty 3.
> 
> Comments from @tskeith :
> 
> > Shouldn't fir be at the same priority as flang? It's in this (sub-)project but the rest listed here are not.
> 
> Reply from @schweitzpgi : https://github.com/flang-compiler/f18/pull/959/files/c36740a747d01885027c86fead0448d161fb4b4f#r372482133
> 
> > The second reason fir is grouped with mlir which is grouped with llvm is that fir is a dialect of mlir and makes no sense without mlir (which transitively makes no sense without llvm). Conversely, the FIR dialect has no dependence on the f18 front-end and could be targeted by other Fortran front-ends.
> 
> Any other opinions ?

The relevant standard is here:
https://llvm.org/docs/CodingStandards.html#include-style
In particular, "LLVM project and subproject headers should be grouped from most specific to least specific ...".
Based on that, the include order should be: flang, fir, mlir, clang, llvm.

DavidTruby:
> 
> The relevant standard is here:
> https://llvm.org/docs/CodingStandards.html#include-style
> In particular, "LLVM project and subproject headers should be grouped from most specific to least specific ...".
> Based on that, the include order should be: flang, fir, mlir, clang, llvm.

Totally agree on this one, I rolled this header preference change in to the header file move (and it is in master, as you can see here: https://github.com/flang-compiler/f18/blob/master/.clang-format), I just haven't rebased this PR since we did that change.

I'll rebase this on top of that and edit it so that we have `..`, flang, fir, mlir, clang, llvm in that order as the priority for header includes.

tskeith:
> I'm not sure I understand why comment alignment affects the size of a change at a glance, other than of course in this first run. I don't suppose you have an example of where it causes a larger than necessary change?

Here's a real-life example from the FIR PR:
https://github.com/flang-compiler/f18/pull/959/commits/02f8aa9046ff3f18c345bf354bd7555744ed6b31#r378929852

One declaration was replaced by two, causing two extra lines to change. It took me a long time to determine there wasn't a real change in those two lines. GitHub could help with better diffing tools.

DavidTruby:
> One declaration was replaced by two, causing two extra lines to change. It took me a long time to determine there wasn't a real change in those two lines. GitHub could help with better diffing tools.

Is it possible that Phabricator might display this better? Would that alleviate the issue if so?

PS. People have requested that we continue this discussion on the mailing list as they want greater community involvement in the discussion. I also think that they might know the answer to this question about Phabricator. Would you mind if I forwarded this comment on to there to see if anyone has any input?

tskeith:
> Is it possible that Phabricator might display this better? Would that alleviate the issue if so?

That would help reviews (until llvm moves to github PRs). There is also git diff (which has options for whitespace) and git log and git annotate (which don't, as far as I can tell).

> PS. People have requested that we continue this discussion on the mailing list as they want greater community involvement in the discussion. I also think that they might know the answer to this question about Phabricator. Would you mind if I forwarded this comment on to there to see if anyone has any input?

Not at all.

.clang-format
Meinersbur: With `BasedOnStyle=LLVM`, `AlignConsecutiveAssignments` and `AlignConsecutiveDeclarations` are already false.

Regarding `AlignTrailingComments`, I don't recommend using trailing comments at all.

For the others, I don't think it's a big issue. In cases such as
```
  int ASomewhatLengthVariableName = aaaaaaaa +
                                                                 bbbbbbb;
```
changing to 
```
  int ShorterName = aaaaaaaa +
                                                                 bbbbbbb;
```
where the `bbbbbbb` would change as well. If the merge-tool ignores whitespace, it would no even show up.

tskeith:
In the interests of making progress, here is a set of changes we might be able to get agreement on:
- keep `AlignAfterOpenBracket`, `AlignOperands`, `AlignTrailingComments`
- update `IncludeCategories` as discussed above
- remove the rest from .clang-format
- update `documentation/C++style.md` with the change in clang-format version

Apparently `AlignConsecutiveAssignments` and `AlignConsecutiveDeclarations` default to false.

The default `AlignEscapedNewlines` is annoying (IMO) but doesn't suffer from the problem of introducing extraneous changes.

Meinersbur:
LLVM defaults are here: https://github.com/llvm/llvm-project/blob/master/clang/lib/Format/Format.cpp#L740

RichBarton-Arm:
We discussed this on the flang community call on March 9 and agreed:

- We will de-scope this patch to just change the clang-format rules we all agree on.
- We will highlight that we are not all aligned as a community on the best story is for code readability on comment alignment, which explains the final lines of divergence in the file. @RichBarton-Arm to do this mid-week.
-  We need to escalate the header-file order issue that occurs with mlir and flang to llvm-dev
-  We will update the clang-format version to align with the version used by master LLVM

DavidTruby:
@tskeith @sscalpone I've updated this PR to reflect that we are not currently making the changes to alignment. Please check it over and make sure there aren't other issues.

Thanks!

tskeith:
What is the required version of clang-format? Is it documented?

DavidTruby:
I ran it with the clang-format I built from LLVM HEAD this morning, commit hash 2c36c23f3476baf3b9283ea387c579419a70b112. Once we merge I think it makes the most sense to just use the clang-format that is built next to f18.

tskeith:
> I ran it with the clang-format I built from LLVM HEAD this morning, commit hash 2c36c23f3476baf3b9283ea387c579419a70b112. Once we merge I think it makes the most sense to just use the clang-format that is built next to f18.

We can't keep the code properly formatted if different developers are using different versions of clang-format. That's why the documentation currently specifies version 7. I think we should continue to do that.

DavidTruby:
If we're going to specify a version I think we should at the least specify the latest release of llvm, not a release from 2 years ago.
However, I believed the consensus on the call on Monday was that we should use the version of clang-format distributed alongside flang, and I personally think that doing anything else is confusing. There's a git-clang-format script in the llvm repository to facilitate exactly this

RichBarton-Arm:
If we go with clang-format 7 and the rest of LLVM is on clang format (latest) then different developers will be using different versions of clang format. I agree that would be bad so we should align with the rest of the project here.

tskeith:
Can you update the documentation with your proposed process for formatting?

Meinersbur:
The version of clang-format is only relevant if flang intends to globally enforce formatting on each commit. Polly does this: `make check-polly` will fail if not formatted correctly. There is also `make polly-update-format` to fix formatting. Since it uses the repository's version of clang, Polly has to be re-formatted occasionally if clang-format changes. clang-format changes are not that frequently, the last one on Polly was: https://github.com/llvm/llvm-project/commit/67dc08bd4c58c7571cf2c10d50ade0f2a3e0f5da . 
More frequently. `make check-polly-format` breaks because of changes in LLVM that also require changes in Polly, and the author does not run clang-format themselves. For instance, https://github.com/llvm/llvm-project/commit/109bfb7b3293cdee1794da866596e68705afc9a2 .

Note that it is NOT feasible to require every LLVM contributor to use a specific version of clang-format (and know about it) that is not the branch's HEAD.

Instead, consider the LLVM policy to only apply clang-format on changes lines using `git clang-format`. Even if using different versions of clang-format, the difference is not noticeable when reviewing. Occasionally, when making larger changes to a file, authors commit a NFC to re-format the entire file separately using the latest version: e.g. https://reviews.llvm.org/rL313744

DavidTruby:
I'm happy to go with either the llvm policy of using git clang-format or to add a check for correct subproject-wide formatting and a target to reformat subproject-wide. If that check is part of check-flang then we will get notified by our build bots when that test fails and can do a reformat commit at that stage. That seems to me to satisfy both consistency across the whole flang codebase and consistency with the version of clang-format. 

I agree that having a separate version of clang-format from the one that gets built alongside flang in the monorepo isn't really feasible.

tskeith:
> I'm happy to go with either the llvm policy of using git clang-format or to add a check for correct subproject-wide formatting and a target to reformat subproject-wide. If that check is part of check-flang then we will get notified by our build bots when that test fails and can do a reformat commit at that stage. That seems to me to satisfy both consistency across the whole flang codebase and consistency with the version of clang-format.

Adding the check sounds good to me. I don't mind if it doesn't happen in this PR but the docs should be updated to reflect what developers should be doing.

tskeith:
> > I'm happy to go with either the llvm policy of using git clang-format or to add a check for correct subproject-wide formatting and a target to reformat subproject-wide. If that check is part of check-flang then we will get notified by our build bots when that test fails and can do a reformat commit at that stage. That seems to me to satisfy both consistency across the whole flang codebase and consistency with the version of clang-format.
> 
> Adding the check sounds good to me. I don't mind if it doesn't happen in this PR but the docs should be updated to reflect what developers should be doing.

Apparently @sscalpone and @klausler don't like the idea of an automatic check of formatting. Perhaps they will comment here.

DavidTruby:
In order to get this moved forward for the merge, can we document in the README that clang-format from LLVM HEAD is used and discuss automation or otherwise later?

klausler:
> Apparently @sscalpone and @klausler don't like the idea of an automatic check of formatting. Perhaps they will comment here.

I don't mind an automatic *check* of formatting, but I would prefer that reformatting be a manual process; we don't want changes to clang-format itself to initiate automatic reformatting over f18.  Best to me seems to be occasional reformatting sweeps when necessary.

Gratuitous differences between repositories due solely to formatting can complicate up/down-streaming and merging, & the harder it is to roll up work and push it into llvm, the less frequently I'll be likely to do it.

DavidTruby:
Ah, maybe I was unclear what the intended workflow will be. I wasn't trying to suggest doing anything automatically.

I was proposing we have a lit test that checks whether clang-format will change any formatting of the sources. This would then email us through the build bots when formatting changes.

Separately, and optionally, we could have a "clang-format-flang" (or whatever name) target that isn't run by default (so not part of "all") that runs clang-format on every cpp and h file in flang, just to make the process of doing a formatting sweep. You would need to run this target manually; I completely agree that this shouldn't happen automatically.

Does that make sense?

Meinersbur:
> Separately, and optionally, we could have a "clang-format-flang" (or whatever name) target that isn't run by default (so not part of "all") that runs clang-format on every cpp and h file in flang, just to make the process of doing a formatting sweep simpler. You would need to run this target manually; I completely agree that this shouldn't happen automatically.

As mentioned before, Polly has a `check-polly-format` target which does exactly this. I'd prefer some consistent with the naming. `check-polly` depends on `check-polly-format` which results in the buildbot failures on formatting inconsistencies.

DavidTruby:
I have changed the documentation to refer to the latest clang-format instead.
I don't propose to merge this on github; in order to minimise disruption I will perform the reformat at the same time as the LLVM upstream merge, so it will be effectively the last commit before the merge commit to LLVM. Does that sound like a sensible approach?

tskeith:
I'm looking for a documented process for dealing with clang-format as a moving target.
C++stye.md currently says this:

> Always run `clang-format` before committing code.  Other developers should
be able to run `git pull`, then `clang-format`, and see only their own
changes. Use `clang-format` from llvm 7.

This will not necessarily be true any more, right? How should a developer use clang-format after this change? How do they avoid extraneous formatting changes?

klausler:
> I have changed the documentation to refer to the latest clang-format instead.
> I don't propose to merge this on github; in order to minimise disruption I will perform the reformat at the same time as the LLVM upstream merge, so it will be effectively the last commit before the merge commit to LLVM. Does that sound like a sensible approach?

That means that the LLVM copy of f18 isn't going to match the original, right?

sscalpone:
My understanding is that files should be reformatted on demand when the file  is checked in.  That seems to be reasonable yet in conflict with the process that is proposed here.

tskeith:
> My understanding is that files should be reformatted on demand when the file is checked in. That seems to be reasonable yet in conflict with the process that is proposed here.

The whole file or just the changes?

sscalpone:
Oh gosh, I can't imagine not reformatting the whole file.

From https://llvm.org/docs/CodingStandards.html:

> Our long term goal is for the entire codebase to follow the convention, but we explicitly do not want patches that do large-scale reformatting of existing code.

RichBarton-Arm:
To summarise what I said earlier today on the Flang community call:
 * We've agreed on the clang-format file rules we'll use
 * We've agreed to use the latest clang-format version from LLVM as our default version (in line with the rest of the project)
 * We want to do a one-off wholesale reformatting of F18 before we upstream to LLVM so that the codebase matches up to the new clang-format.
 * After this point we'll require that all new code is formatted to the new clang-format file using the git clang-format tool and check this via code review (in line with the rest of the project)

Any make targets to help auto-formatting or doing this as part of a builder, etc. is out of scope for this PR. We can think about how to improve in that area later.

David's latest patch should line up the documentation to make this all clear and hopefully this patch can be approved.

After that, David will need to squash and rebase it. Then we have two options:
1. Apply to F18 master before we upstream and start applying the new policy of clang-format to all new patches in F18, then in llvm-project/flang.
2. Apply this on the top of the F18 history that we'll push into llvm-project/flang so it happens that way.

We prefer option 2 as it seems less disruptive to ongoing development on F18 and saves another point of synchronisation between David and Steve when upstreaming to LLVM. Weakly held view though. Ultimately it doesn't matter too much as the resulting llvm-project/flang is the same with either approach and the code in the F18 repo will not exist after merge.

@tskeith  - @sscalpone tells me it's your call - are you happy with the above?

tskeith:
That all sounds good to me. I think the documentation of the process should also say what happens when clang-format changes in a way that causes formatting changes. When that happens (hopefully rarely) we should have a reformat-only change.

As for when to do it, it seems to me that option 1 would be better for ongoing f18 development. If you have a change in progress you can rebase to that version and then make a patch that is guaranteed to apply cleanly to the version in llvm-project (modulo other changes made in the llvm version).

tskeith:
Thinking about it more, I think the best way to proceed would be to separate out the part of the change involving moving to the latest version of clang-format. That could be merged now and would allow us to adapt to the new process (e.g. building the latest clang-format) independent of the move to llvm-project. It also brings the rest of the project in sync with Lower and Optimizer.

The downside is that there would be two reformatting events. A small one due to the change in clang-format and a big one with the move to llvm.

RichBarton-Arm:
Please can you merge it?


________________________________

PR #946, opened 2020-01-23
12HHELLO, WORLD

klausler:
Drill down to a working implementation of the APIs for an internal formatted WRITE with no data list items.

- Improve argument names in io-api.h
- Bump up error number to not conflict with errno values
- Use Fortran::runtime::io namespace
- Add wrapper around malloc/free, allow use of unique_ptr with wrapper
- Define IoErrorHandler
- Revamp FormatContext, use virtual member functions
- Update comment syntax, allow for old C
- Demonstrate 12HHELLO, WORLD
- Remove files not yet ready for review
- Use std::forward
- Fix GCC build warnings in io-api.h

### DIFF
--- a/runtime/format.cc
+++ b/runtime/format.cc
@@ -7,23 +7,43 @@
 //===----------------------------------------------------------------------===//
 
 #include "format.h"
+#include "io-stmt.h"
 #include "../lib/common/format.h"
 #include "../lib/decimal/decimal.h"
 #include <limits>
 
-namespace Fortran::runtime {
+namespace Fortran::runtime::io {
+
+// Default FormatContext virtual member functions
+void FormatContext::Emit(const char *, std::size_t) {
+  Crash("Cannot emit data from this FORMAT string");
+}
+void FormatContext::Emit(const char16_t *, std::size_t) {
+  Crash("Cannot emit data from this FORMAT string");
+}
+void FormatContext::Emit(const char32_t *, std::size_t) {
+  Crash("Cannot emit data from this FORMAT string");
+}
+void FormatContext::HandleSlash(int) {
+  Crash("A / control edit descriptor may not appear in this FORMAT string");
+}
+void FormatContext::HandleAbsolutePosition(int) {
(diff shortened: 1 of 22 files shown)


runtime/main.h
```diff
@@ -12,9 +12,15 @@
 #include "entry-names.h"
 
 namespace Fortran::runtime {
-extern int argc;
-extern const char **argv;
-extern const char **envp;
+struct ExecutionEnvironment {
```
psteinfeld: I noticed that this file has the wrong name in the license at the beginning of the file.

test/runtime/hello.cc
```diff
@@ -0,0 +1,33 @@
+// Basic tests of I/O API
```
psteinfeld: This is very cool.  What causes the compiled version of this to be executed?

klausler: It's run from `test/runtime/CMakeLists.txt`.

runtime/format.cc
```diff
@@ -7,23 +7,43 @@
...
 #include "../lib/decimal/decimal.h"
 #include <limits>
 
-namespace Fortran::runtime {
+namespace Fortran::runtime::io {
+
+// Default FormatContext virtual member functions
+void FormatContext::Emit(const char *, std::size_t) {
+  Crash("Cannot emit data from this FORMAT string");
+}
```
jeanPerier: I am a bit lost as regarding to when these messages should be emitted.
Are the internal error (i.e the runtime engineer should have overridden this), or are they also intended to handle user error ?
In the first case, I would suggest adding "Internal error" or something similar.
In the second case, the `FormatContext` does not have access to an `IoErrorHandler`, so these crashes cannot be prevented if the user used "ERR", right ?

klausler: `READ(unit,"(`text')") x` would elicit this error; since the format string could be a run-time variable, we can't always validate it until execution.  These errors cannot be caught with `ERR=/IOSTAT=` (although that would be a reasonable extension).

PR #950, opened 2020-01-25
Formatted output editing & initial buffer framing code

klausler:
Defines a class that represents a buffer in memory for an open external file with the special property that a "frame" can be requested that is an interval of file positions that is also a contiguous range of memory.  Within this frame, formatted I/O can modify the current position within the current record arbitrarily to support Tn, TLn, & TRn positioning control edit descriptors.

This code isn't used or tested yet.


jeanPerier:
Is the plan to always load entire files into these frame buffers ? If not, how will Tn outside of the buffer be handled ?

runtime/buffer.h
```diff
@@ -0,0 +1,179 @@
...
+// This CRTP mixin maintains a circular buffer of valid data that has been
+// read from or is yet to be written to a file.  Within that interval of
+// valid data, it supports a current frame that is guaranteed to be
+// contiguous.  The API comprises routines that establish a given interval
+// of the file as the current frame for reading or writing,
+template<typename STORE> class FileFrame {
+public:
+  using FileOffset = std::uint64_t;
+
+  ~FileFrame() { FreeMemoryAndNullify(buffer_); }
```
psteinfeld: When I build using GNU 8.3.0, I get an error message for this line:
../runtime/buffer.h:40:39: error: cannot bind non-const lvalue reference of type â€˜void*&â€™ to an rvalue of type 'void*'
It builds OK for me with Clang.

runtime/buffer.h
```diff
@@ -0,0 +1,179 @@
...
+      Flush(handler);
+      fileOffset_ = at;
+      Reallocate(bytes, handler);
+    }
+    dirty_ = true;
+    frame_ = at - fileOffset_;
+    length_ = std::max(length_, frame_ + bytes);
+  }
+
+  bool Flush(IoErrorHandler &handler) {
```
psteinfeld: This function is declared to return a "bool", but there's no "return" statement.  This seems to be OK with Clang, but not with GNU 8.3.0.

runtime/buffer.h
```diff
@@ -0,0 +1,179 @@
...
+
+private:
+  STORE &Store() { return static_cast<STORE &>(*this); }
+
+  void Reallocate(std::size_t bytes, Terminator &terminator) {
+    if (bytes > size_) {
+      char *old{buffer_};
+      auto oldSize{size_};
+      size_ = std::max(bytes, minBuffer);
+      buffer_ = AllocateMemoryOrCrash(size_, terminator);
```
psteinfeld: GNU 8.3.0 gives an error on this line:
../runtime/buffer.h:134:39: error: invalid initialization of reference of type 'Fortran::runtime::Terminator&' from expression of type 'std::size_t' {aka 'long unsigned int'}

klausler:
> Is the plan to always load entire files into these frame buffers ? If not, how will Tn outside of the buffer be handled ?

No.  The frame should only contain the current record.  Positioning edit descriptors like Tn can operate only between the beginning of the current record (at the beginning of the current I/O data transfer statement in the case of non-advancing I/O) and its end.

runtime/buffer.h
klausler: Fixed.

runtime/buffer.h
klausler: Fixed.

runtime/buffer.h
klausler: Fixed.

DavidTruby:
Is this substantially different to the llvm::MemoryBuffer class? (https://llvm.org/doxygen/classllvm_1_1MemoryBuffer.html)
If not, we could possibly leverage that to do some of what this class does. There may be something different about the semantics that I am missing though.

tskeith:
> Is this substantially different to the llvm::MemoryBuffer class? (https://llvm.org/doxygen/classllvm_1_1MemoryBuffer.html)
> If not, we could possibly leverage that to do some of what this class does. There may be something different about the semantics that I am missing though.

That would make the runtime depend on LLVM libraries, and presumably also the C++ runtime library, right?. That's an approach we shouldn't take lightly.

klausler:
> Is this substantially different to the llvm::MemoryBuffer class? (https://llvm.org/doxygen/classllvm_1_1MemoryBuffer.html)
> If not, we could possibly leverage that to do some of what this class does. There may be something different about the semantics that I am missing though.

It is substantially different in its behavior, interface, use case, and underlying dependences.

DavidTruby:
> That would make the runtime depend on LLVM libraries, and presumably also the C++ runtime library, right?. That's an approach we shouldn't take lightly.

I hadn't considered that, however the LLVM libraries are statically linked by default and these files already depend on the C++ runtime library as they use various headers from the standard library (e.g. <algorithm>, <memory> and <optional>).

> It is substantially different in its behavior, interface, use case, and underlying dependences.

Could you please elaborate? At a cursory glance it isn't obvious which is why I asked, and the description of the two classes is almost identical.

klausler:
> > That would make the runtime depend on LLVM libraries, and presumably also the C++ runtime library, right?. That's an approach we shouldn't take lightly.
> 
> I hadn't considered that, however the LLVM libraries are statically linked by default and these files already depend on the C++ runtime library as they use various headers from the standard library (e.g. , and ).
> 
> > It is substantially different in its behavior, interface, use case, and underlying dependences.
> 
> Could you please elaborate? At a cursory glance it isn't obvious which is why I asked, and the description of the two classes is almost identical.

I will try to explain.  The C++ standard library is implemented with header files and binary libraries.  The header files contain data structures, inline functions, macros, template classes, and template functions.  The binary libraries contain static data and executable code.

Usage of macros, template classes, inline functions, and template functions of the C++ standard library can be done in ways that do not introduce a link-time or run-time dependence on the binary library.  This is our goal; we would prefer that the Fortran runtime library have minimal dependences on the surrounding software environment.  This will ease portability of the Fortran runtime library and Fortran applications.  We specifically do not want to have to build and test and ship multiple versions of the Fortran runtime library that each depend on distinct C++ runtime library binaries &/or LLVM versions.

So we are trying to construct a Fortran runtime that depends only on standard interfaces that we can reasonably expect to link from C (not C++) runtime support libraries.  We can still use headers from the C++ library when using capabilities that are fully elaborated at build time, such as macros and templates classes and template functions.  For a specific example, `std::optional` can be configured to use a custom object deleter, and we are using that class with a deleter that calls `free()` rather than `operator delete`.  Many other STL facilities can also be safely used if we need them.

DavidTruby:
> 
> 
> > > That would make the runtime depend on LLVM libraries, and presumably also the C++ runtime library, right?. That's an approach we shouldn't take lightly.
> > 
> > 
> > I hadn't considered that, however the LLVM libraries are statically linked by default and these files already depend on the C++ runtime library as they use various headers from the standard library (e.g. , and ).
> > > It is substantially different in its behavior, interface, use case, and underlying dependences.
> > 
> > 
> > Could you please elaborate? At a cursory glance it isn't obvious which is why I asked, and the description of the two classes is almost identical.
> 
> I will try to explain. The C++ standard library is implemented with header files and binary libraries. The header files contain data structures, inline functions, macros, template classes, and template functions. The binary libraries contain static data and executable code.
> 
> Usage of macros, template classes, inline functions, and template functions of the C++ standard library can be done in ways that do not introduce a link-time or run-time dependence on the binary library. This is our goal; we would prefer that the Fortran runtime library have minimal dependences on the surrounding software environment. This will ease portability of the Fortran runtime library and Fortran applications. We specifically do not want to have to build and test and ship multiple versions of the Fortran runtime library that each depend on distinct C++ runtime library binaries &/or LLVM versions.

Once we are in the LLVM repo, there will be an implicit dependency on a specific LLVM version anyway as we will be part of that LLVM version.

> 
> So we are trying to construct a Fortran runtime that depends only on standard interfaces that we can reasonably expect to link from C (not C++) runtime support libraries. We can still use headers from the C++ library when using capabilities that are fully elaborated at build time, such as macros and templates classes and template functions. For a specific example, `std::optional` can be configured to use a custom object deleter, and we are using that class with a deleter that calls `free()` rather than `operator delete`. Many other STL facilities can also be safely used if we need them.

I don't believe there's any guarantee that any of the STL facilities won't require linking to a runtime library; it's certainly never mentioned in the C++ standard. Most of the functions in `<algorithm>` do require linking to the runtime library, at least when using `libc++` (I'm less familiar with `libstdc++`). 
There's also no guarantee that things that are used from the STL that don't currently require linking won't require linking in the future. I would suggest that if there is a hard requirement not to link to the C++ runtime library we should avoid any use of C++ library headers and functions.

On my system libFortranRuntime.a already contains references to C++ runtime library functions (and hence requires that you link the C++ runtime with any binary that uses it) before this PR.

klausler:
> > > > That would make the runtime depend on LLVM libraries, and presumably also the C++ runtime library, right?. That's an approach we shouldn't take lightly.
> > > 
> > > 
> > > I hadn't considered that, however the LLVM libraries are statically linked by default and these files already depend on the C++ runtime library as they use various headers from the standard library (e.g. , and ).
> > > > It is substantially different in its behavior, interface, use case, and underlying dependences.
> > > 
> > > 
> > > Could you please elaborate? At a cursory glance it isn't obvious which is why I asked, and the description of the two classes is almost identical.
> > 
> > 
> > I will try to explain. The C++ standard library is implemented with header files and binary libraries. The header files contain data structures, inline functions, macros, template classes, and template functions. The binary libraries contain static data and executable code.
> > Usage of macros, template classes, inline functions, and template functions of the C++ standard library can be done in ways that do not introduce a link-time or run-time dependence on the binary library. This is our goal; we would prefer that the Fortran runtime library have minimal dependences on the surrounding software environment. This will ease portability of the Fortran runtime library and Fortran applications. We specifically do not want to have to build and test and ship multiple versions of the Fortran runtime library that each depend on distinct C++ runtime library binaries &/or LLVM versions.
> 
> Once we are in the LLVM repo, there will be an implicit dependency on a specific LLVM version anyway as we will be part of that LLVM version.

A Fortran product that requires a Fortran application vendor to ship or link LLVM with/into their product would be a hard sell to our commercial customers.

> 
> > So we are trying to construct a Fortran runtime that depends only on standard interfaces that we can reasonably expect to link from C (not C++) runtime support libraries. We can still use headers from the C++ library when using capabilities that are fully elaborated at build time, such as macros and templates classes and template functions. For a specific example, `std::optional` can be configured to use a custom object deleter, and we are using that class with a deleter that calls `free()` rather than `operator delete`. Many other STL facilities can also be safely used if we need them.
> 
> I don't believe there's any guarantee that any of the STL facilities won't require linking to a runtime library; it's certainly never mentioned in the C++ standard. Most of the functions in `<algorithm>` do require linking to the runtime library, at least when using `libc++` (I'm less familiar with `libstdc++`).
> There's also no guarantee that things that are used from the STL that don't currently require linking won't require linking in the future. I would suggest that if there is a hard requirement not to link to the C++ runtime library we should avoid any use of C++ library headers and functions.
> 
> On my system libFortranRuntime.a already contains references to C++ runtime library functions (and hence requires that you link the C++ runtime with any binary that uses it) before this PR.

This is a work in progress.  At the moment, I am attempting to compose a basic I/O runtime to support a development milestone by next month.  The few external references to C++ library externals will be replaced at that time, and tests will ensure that they do not recrudesce.

klausler:
While awaiting approval from reviewers, I've pushed ahead on the runtime implementation and added support for formatted output editing descriptors (except `EX`).

lib/evaluate/check-expression.cpp
```diff
@@ -11,6 +11,7 @@
 #include "flang/evaluate/type.h"
 #include "flang/semantics/symbol.h"
 #include "flang/semantics/tools.h"
+#include <iostream>  // TODO pmk rm
```
jeanPerier: Left over ?

klausler: Yes, thanks!

PR #952, opened 2020-01-26
Fix bug detecting simply contiguous component

tskeith:
We were always return false when testing a component for simple
contiguity. Change to check that the component is an array that is
simply continguous.

A pointer with bounds remapping to a complex part is a similar case
so add a test for that too.

### DIFF
--- a/lib/evaluate/check-expression.cc
+++ b/lib/evaluate/check-expression.cc
@@ -283,7 +283,9 @@ public:
   Result operator()(const CoarrayRef &x) const {
     return CheckSubscripts(x.subscript());
   }
-  Result operator()(const Component &) const { return false; }
+  Result operator()(const Component &x) const {
+    return x.base().Rank() == 0 && (*this)(x.GetLastSymbol());
+  }
   Result operator()(const ComplexPart &) const { return false; }
   Result operator()(const Substring &) const { return false; }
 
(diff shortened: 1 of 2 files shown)


lib/evaluate/check-expression.cc
```diff
@@ -283,7 +283,10 @@ class IsSimplyContiguousHelper
   Result operator()(const CoarrayRef &x) const {
     return CheckSubscripts(x.subscript());
   }
-  Result operator()(const Component &) const { return false; }
+  Result operator()(const Component &x) const {
```
klausler: Should a scalar component of a scalar base be simply contiguous, if simple scalar are so?

tskeith: I considered that, but it seemed silly to add an extra check for a scalar base when it isn't technically simply contiguous anyway. And then what about the case where there is an array base but the derived type has only one component?

I would be more inclined to follow the standard and treat a simple scalar as not "simply contiguous" if we want to be consistent.

klausler: So long as things are internally consistent with the treatment of a simple scalar, and a comment to the effect that "the standard defines simple contiguity only as an attribute of arrays, so we arbitrarily return false for scalars" appears, I don't mind.

tskeith: Looking at it further, it seems the simplest way to be consistent is to take your original suggestion and treat a scalar component of a scalar as simply contiguous.

PR #953, opened 2020-01-27
changed *.cc file extensions to *.cpp

AlexisPerry:


tskeith:
There are some new `.cc` files now in runtime.

AlexisPerry:
> There are some new `.cc` files now in runtime.

@tskeith I noticed that too. I addressed it in my most recent commit.

AlexisPerry:
> The first lines of the license headers in the source files have a useless copy of the file name. Those are also going to have to change.
> 
> Some headers and source files reference others by name in comments. They're trivial to find with grep.
> Those will also have to change.

@klausler Thank you for noticing that.  I believe I have fixed the issues.  Would you please take another look?

klausler:
> > The first lines of the license headers in the source files have a useless copy of the file name. Those are also going to have to change.
> > Some headers and source files reference others by name in comments. They're trivial to find with grep.
> > Those will also have to change.
> 
> @klausler Thank you for noticing that. I believe I have fixed the issues. Would you please take another look?

You may have changed the license headers without looking at the results of the changes.

sscalpone:
@DavidTruby  -- Will this change cause hard-to-fix merge issues with your re-structuring of header files?  If so, I'll hold off merging until your change get merged.  Otherwise, @AlexisPerry , if you're otherwise done, please squash this prior to the merge.

runtime/CMakeLists.txt
```diff
@@ -7,21 +7,20 @@
...
-  memory.cc
-  stop.cc
-  terminator.cc
-  tools.cc
-  transformational.cc
-  type-code.cc
+  ISO_Fortran_binding.cpp
+  derived-type.cpp
+  descriptor.cpp
+  format.cpp
```
klausler: You deleted file.cc.

DavidTruby:
I think if the header change gets merged first then this should be easy to rebase on top of that, given that git is rename-aware.

runtime/CMakeLists.txt
AlexisPerry: Thank you.  I had been unable to find the extra deletion myself.  I believe [https://github.com/flang-compiler/f18/pull/953/commits/dafa1a95db8c02014bafa40f8c0e89b3610cebed](url) fixes this.

DavidTruby:
If this does turn out to be a complicated rebase given my patch, I am happy to take a look at it

DavidTruby:
This doesn't actually conflict with my PR, but rather the runtime PR that went in before mine. I can have a further look tomorrow as it seems like editing someone else's PR is not as simple as I thought...  I am happy to rebase this and squash the commits if that needs doing but I think I'll need to open another PR to do it so let me know if you'd like me to do that :)

AlexisPerry:
I am closing this PR in favor of #957 , which should be much cleaner.

PR #955, opened 2020-01-27
Fix another bug checking simple contiguity

tskeith:
My last change was not complete: when the RHS of a pointer bounds
remapping was an array section rather than a whole array we weren't
correctly determine if it was simply contiguous.

### DIFF
--- a/lib/evaluate/check-expression.cc
+++ b/lib/evaluate/check-expression.cc
@@ -277,8 +277,7 @@ public:
   }
 
   Result operator()(const ArrayRef &x) const {
-    return (x.base().IsSymbol() || x.base().Rank() == 0) &&
-        CheckSubscripts(x.subscript()) && (*this)(x.base());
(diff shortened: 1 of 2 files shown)


lib/evaluate/check-expression.cc
```diff
@@ -277,8 +277,7 @@ class IsSimplyContiguousHelper
   }
 
   Result operator()(const ArrayRef &x) const {
-    return (x.base().IsSymbol() || x.base().Rank() == 0) &&
-        CheckSubscripts(x.subscript()) && (*this)(x.base());
+    return CheckSubscripts(x.subscript()) && (*this)(x.base());
```
klausler: The reason for the current predicate is to ensure that `A(:)%B(1)` is not a false positive for simple contiguity.  It might be broken, but we need to ensure that this fix keeps that case negative.

PR #956, opened 2020-01-27
More checklist items

psteinfeld:
I added more items when reviewing some actual pull request comments.


ichoyjx:
This might not be able to 100% apply to OpenMP stuff, which does not have restriction number for restrictions...

psteinfeld:
> 
> 
> This might not be able to 100% apply to OpenMP stuff, which does not have restriction number for restrictions...

I'm not sure what you mean about "restriction number for restrictions".

PR #959, opened 2020-01-28
RFC: FIR Merge - PR1: Add PFTBuilder structure to help lowering the parse-tree

jeanPerier:
This PR contains the directory structure and .clang-format for FIR and the first piece of code: the ASTBuilder.

The ASTBuilder structure is a transient data structure that
is meant to be built from the parse tree just before lowering to
FIR and that will be deleted just afterwards. It is not meant to perform
optimization analysis and transformations. It only provides temporary
information, such as label target information or parse tree parent nodes,
that is meant to be used to lower the parse tree structure into
FIR operations.
A pretty printer is available to visualize this data structure.

This code depends on LLVM LLVMSupport library but does not require MLIR (it can build without changes to the current main CMakeLists.txt).

Note that the commits does not reflect an actual work log, it is a feature based split of the
changes done in the FIR experimental branch. The related work log can be found in the commits between: 8c320e3 and 9b9ea05.

### DIFF
--- a/include/flang/lower/PFTBuilder.h
+++ b/include/flang/lower/PFTBuilder.h
@@ -1,0 +1,394 @@
+//===-- include/flang/lower/PFTBuilder.h ------------------------*- C++ -*-===//
+//
+// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
+
+#ifndef FORTRAN_LOWER_PFT_BUILDER_H_
+#define FORTRAN_LOWER_PFT_BUILDER_H_
+
+#include "flang/common/template.h"
+#include "flang/parser/parse-tree.h"
+#include "llvm/Support/raw_ostream.h"
+#include <memory>
+
+/// Build a light-weight tree over the parse-tree to help with lowering to FIR.
+/// It is named Pre-FIR Tree (PFT) to underline it has no other usage than
+/// helping lowering to FIR.
+/// The PFT will capture pointers back into the parse tree, so the parse tree
+/// data structure may <em>not</em> be changed between the construction of the
+/// PFT and all of its uses.
+///
+/// The PFT captures a structured view of the program.  The program is a list of
+/// units.  Function like units will contain lists of evaluations.  Evaluations
+/// are either statements or constructs, where a construct contains a list of
+/// evaluations. The resulting PFT structure can then be used to create FIR.
+
+namespace Fortran::lower {
+namespace pft {
+
+struct Evaluation;
+struct Program;
+struct ModuleLikeUnit;
+struct FunctionLikeUnit;
+
+// TODO: A collection of Evaluations can obviously be any of the container
+// types; leaving this as a std::list _for now_ because we reserve the right to
+// insert PFT nodes in any order in O(1) time.
+using EvaluationCollection = std::list<Evaluation>;
+
+struct ParentType {
+  template <typename A>
+  ParentType(A &parent) : p{&parent} {}
+  const std::variant<Program *, ModuleLikeUnit *, FunctionLikeUnit *,
+                     Evaluation *>
+      p;
+};
+
+/// Flags to describe the impact of parse-trees nodes on the program
+/// control flow. These annotations to parse-tree nodes are later used to
+/// build the control flow graph when lowering to FIR.
+enum class CFGAnnotation {
+  None,            // Node does not impact control flow.
+  Goto,            // Node acts like a goto on the control flow.
+  CondGoto,        // Node acts like a conditional goto on the control flow.
+  IndGoto,         // Node acts like an indirect goto on the control flow.
+  IoSwitch,        // Node is an IO statement with ERR, END, or EOR specifier.
+  Switch,          // Node acts like a switch on the control flow.
+  Iterative,       // Node creates iterations in the control flow.
+  FirStructuredOp, // Node is a structured loop.
+  Return,          // Node triggers a return from the current procedure.
+  Terminate        // Node terminates the program.
+};
+
+/// Compiler-generated jump
+///
+/// This is used to convert implicit control-flow edges to explicit form in the
+/// decorated PFT
+struct CGJump {
+  CGJump(Evaluation &to) : target{to} {}
+  Evaluation &target;
+};
+
+/// Classify the parse-tree nodes from ExecutablePartConstruct
+
+using ActionStmts = std::tuple<
+    parser::AllocateStmt, parser::AssignmentStmt, parser::BackspaceStmt,
+    parser::CallStmt, parser::CloseStmt, parser::ContinueStmt,
+    parser::CycleStmt, parser::DeallocateStmt, parser::EndfileStmt,
+    parser::EventPostStmt, parser::EventWaitStmt, parser::ExitStmt,
+    parser::FailImageStmt, parser::FlushStmt, parser::FormTeamStmt,
+    parser::GotoStmt, parser::IfStmt, parser::InquireStmt, parser::LockStmt,
+    parser::NullifyStmt, parser::OpenStmt, parser::PointerAssignmentStmt,
+    parser::PrintStmt, parser::ReadStmt, parser::ReturnStmt, parser::RewindStmt,
+    parser::StopStmt, parser::SyncAllStmt, parser::SyncImagesStmt,
+    parser::SyncMemoryStmt, parser::SyncTeamStmt, parser::UnlockStmt,
+    parser::WaitStmt, parser::WhereStmt, parser::WriteStmt,
+    parser::ComputedGotoStmt, parser::ForallStmt, parser::ArithmeticIfStmt,
+    parser::AssignStmt, parser::AssignedGotoStmt, parser::PauseStmt>;
+
+using OtherStmts = std::tuple<parser::FormatStmt, parser::EntryStmt,
+                              parser::DataStmt, parser::NamelistStmt>;
+
+using Constructs =
+    std::tuple<parser::AssociateConstruct, parser::BlockConstruct,
+               parser::CaseConstruct, parser::ChangeTeamConstruct,
+               parser::CriticalConstruct, parser::DoConstruct,
+               parser::IfConstruct, parser::SelectRankConstruct,
+               parser::SelectTypeConstruct, parser::WhereConstruct,
+               parser::ForallConstruct, parser::CompilerDirective,
+               parser::OpenMPConstruct, parser::OmpEndLoopDirective>;
+
+using ConstructStmts = std::tuple<
+    parser::AssociateStmt, parser::EndAssociateStmt, parser::BlockStmt,
+    parser::EndBlockStmt, parser::SelectCaseStmt, parser::CaseStmt,
+    parser::EndSelectStmt, parser::ChangeTeamStmt, parser::EndChangeTeamStmt,
+    parser::CriticalStmt, parser::EndCriticalStmt, parser::NonLabelDoStmt,
+    parser::EndDoStmt, parser::IfThenStmt, parser::ElseIfStmt, parser::ElseStmt,
+    parser::EndIfStmt, parser::SelectRankStmt, parser::SelectRankCaseStmt,
+    parser::SelectTypeStmt, parser::TypeGuardStmt, parser::WhereConstructStmt,
+    parser::MaskedElsewhereStmt, parser::ElsewhereStmt, parser::EndWhereStmt,
+    parser::ForallConstructStmt, parser::EndForallStmt>;
+
+template <typename A>
+constexpr static bool isActionStmt{common::HasMember<A, ActionStmts>};
+
+template <typename A>
+constexpr static bool isConstruct{common::HasMember<A, Constructs>};
+
+template <typename A>
+constexpr static bool isConstructStmt{common::HasMember<A, ConstructStmts>};
+
+template <typename A>
+constexpr static bool isOtherStmt{common::HasMember<A, OtherStmts>};
+
+template <typename A>
+constexpr static bool isGenerated{std::is_same_v<A, CGJump>};
+
+template <typename A>
+constexpr static bool isFunctionLike{common::HasMember<
+    A, std::tuple<parser::MainProgram, parser::FunctionSubprogram,
+                  parser::SubroutineSubprogram,
+                  parser::SeparateModuleSubprogram>>};
+
+/// Function-like units can contains lists of evaluations.  These can be
+/// (simple) statements or constructs, where a construct contains its own
+/// evaluations.
+struct Evaluation {
+  using EvalTuple = common::CombineTuples<ActionStmts, OtherStmts, Constructs,
+                                          ConstructStmts>;
+
+  /// Hide non-nullable pointers to the parse-tree node.
+  template <typename A>
+  using MakeRefType = const A *const;
+  using EvalVariant =
+      common::CombineVariants<common::MapTemplate<MakeRefType, EvalTuple>,
+                              std::variant<CGJump>>;
+  template <typename A>
+  constexpr auto visit(A visitor) const {
+    return std::visit(common::visitors{
+                          [&](const auto *p) { return visitor(*p); },
+                          [&](auto &r) { return visitor(r); },
+                      },
+                      u);
+  }
+  template <typename A>
+  constexpr const A *getIf() const {
+    if constexpr (!std::is_same_v<A, CGJump>) {
+      if (auto *ptr{std::get_if<MakeRefType<A>>(&u)}) {
+        return *ptr;
+      }
+    } else {
+      return std::get_if<CGJump>(&u);
+    }
+    return nullptr;
+  }
+  template <typename A>
+  constexpr bool isA() const {
+    if constexpr (!std::is_same_v<A, CGJump>) {
+      return std::holds_alternative<MakeRefType<A>>(u);
+    }
+    return std::holds_alternative<CGJump>(u);
+  }
+
+  Evaluation() = delete;
+  Evaluation(const Evaluation &) = delete;
+  Evaluation(Evaluation &&) = default;
+
+  /// General ctor
+  template <typename A>
+  Evaluation(const A &a, const ParentType &p, const parser::CharBlock &pos,
+             const std::optional<parser::Label> &lab)
+      : u{&a}, parent{p}, pos{pos}, lab{lab} {}
+
+  /// Compiler-generated jump
+  Evaluation(const CGJump &jump, const ParentType &p)
+      : u{jump}, parent{p}, cfg{CFGAnnotation::Goto} {}
+
+  /// Construct ctor
+  template <typename A>
+  Evaluation(const A &a, const ParentType &parent) : u{&a}, parent{parent} {
+    static_assert(pft::isConstruct<A>, "must be a construct");
+  }
+
+  constexpr bool isActionOrGenerated() const {
+    return visit(common::visitors{
+        [](auto &r) {
+          using T = std::decay_t<decltype(r)>;
+          return isActionStmt<T> || isGenerated<T>;
+        },
+    });
+  }
+
+  constexpr bool isStmt() const {
+    return visit(common::visitors{
+        [](auto &r) {
+          using T = std::decay_t<decltype(r)>;
+          static constexpr bool isStmt{isActionStmt<T> || isOtherStmt<T> ||
+                                       isConstructStmt<T>};
+          static_assert(!(isStmt && pft::isConstruct<T>),
+                        "statement classification is inconsistent");
+          return isStmt;
+        },
+    });
+  }
+  constexpr bool isConstruct() const { return !isStmt(); }
+
+  /// Set the type of originating control flow type for this evaluation.
+  void setCFG(CFGAnnotation a, Evaluation *cstr) {
+    cfg = a;
+    setBranches(cstr);
+  }
+
+  /// Is this evaluation a control-flow origin? (The PFT must be annotated)
+  bool isControlOrigin() const { return cfg != CFGAnnotation::None; }
+
+  /// Is this evaluation a control-flow target? (The PFT must be annotated)
+  bool isControlTarget() const { return isTarget; }
+
+  /// Set the containsBranches flag iff this evaluation (a construct) contains
+  /// control flow
+  void setBranches() { containsBranches = true; }
+
+  EvaluationCollection *getConstructEvals() {
+    auto *evals{subs.get()};
+    if (isStmt() && !evals) {
+      return nullptr;
+    }
+    if (isConstruct() && evals) {
+      return evals;
+    }
+    llvm_unreachable("evaluation subs is inconsistent");
+    return nullptr;
+  }
+
+  /// Set that the construct `cstr` (if not a nullptr) has branches.
+  static void setBranches(Evaluation *cstr) {
+    if (cstr)
+      cstr->setBranches();
+  }
+
+  EvalVariant u;
+  ParentType parent;
+  parser::CharBlock pos;
+  std::optional<parser::Label> lab;
+  std::unique_ptr<EvaluationCollection> subs; // construct sub-statements
+  CFGAnnotation cfg{CFGAnnotation::None};
+  bool isTarget{false};         // this evaluation is a control target
+  bool containsBranches{false}; // construct contains branches
+};
+
+/// A program is a list of program units.
+/// These units can be function like, module like, or block data
+struct ProgramUnit {
+  template <typename A>
+  ProgramUnit(const A &ptr, const ParentType &parent)
+      : p{&ptr}, parent{parent} {}
+  ProgramUnit(ProgramUnit &&) = default;
+  ProgramUnit(const ProgramUnit &) = delete;
+
+  const std::variant<
+      const parser::MainProgram *, const parser::FunctionSubprogram *,
+      const parser::SubroutineSubprogram *, const parser::Module *,
+      const parser::Submodule *, const parser::SeparateModuleSubprogram *,
+      const parser::BlockData *>
+      p;
+  ParentType parent;
+};
+
+/// Function-like units have similar structure. They all can contain executable
+/// statements as well as other function-like units (internal procedures and
+/// function statements).
+struct FunctionLikeUnit : public ProgramUnit {
+  // wrapper statements for function-like syntactic structures
+  using FunctionStatement =
+      std::variant<const parser::Statement<parser::ProgramStmt> *,
+                   const parser::Statement<parser::EndProgramStmt> *,
+                   const parser::Statement<parser::FunctionStmt> *,
+                   const parser::Statement<parser::EndFunctionStmt> *,
+                   const parser::Statement<parser::SubroutineStmt> *,
+                   const parser::Statement<parser::EndSubroutineStmt> *,
+                   const parser::Statement<parser::MpSubprogramStmt> *,
+                   const parser::Statement<parser::EndMpSubprogramStmt> *>;
+
+  FunctionLikeUnit(const parser::MainProgram &f, const ParentType &parent);
+  FunctionLikeUnit(const parser::FunctionSubprogram &f,
+                   const ParentType &parent);
+  FunctionLikeUnit(const parser::SubroutineSubprogram &f,
+                   const ParentType &parent);
+  FunctionLikeUnit(const parser::SeparateModuleSubprogram &f,
+                   const ParentType &parent);
+  FunctionLikeUnit(FunctionLikeUnit &&) = default;
+  FunctionLikeUnit(const FunctionLikeUnit &) = delete;
+
+  bool isMainProgram() {
+    return std::holds_alternative<
+        const parser::Statement<parser::EndProgramStmt> *>(endStmt);
+  }
+  const parser::FunctionStmt *getFunction() {
+    return getA<parser::FunctionStmt>();
+  }
+  const parser::SubroutineStmt *getSubroutine() {
+    return getA<parser::SubroutineStmt>();
+  }
+  const parser::MpSubprogramStmt *getMPSubp() {
+    return getA<parser::MpSubprogramStmt>();
+  }
+
+  /// Anonymous programs do not have a begin statement
+  std::optional<FunctionStatement> beginStmt;
+  FunctionStatement endStmt;
+  EvaluationCollection evals;        // statements
+  std::list<FunctionLikeUnit> funcs; // internal procedures
+
+private:
+  template <typename A>
+  const A *getA() {
+    if (beginStmt) {
+      if (auto p =
+              std::get_if<const parser::Statement<A> *>(&beginStmt.value()))
+        return &(*p)->statement;
+    }
+    return nullptr;
+  }
+};
+
+/// Module-like units have similar structure. They all can contain a list of
+/// function-like units.
+struct ModuleLikeUnit : public ProgramUnit {
+  // wrapper statements for module-like syntactic structures
+  using ModuleStatement =
+      std::variant<const parser::Statement<parser::ModuleStmt> *,
+                   const parser::Statement<parser::EndModuleStmt> *,
+                   const parser::Statement<parser::SubmoduleStmt> *,
+                   const parser::Statement<parser::EndSubmoduleStmt> *>;
+
+  ModuleLikeUnit(const parser::Module &m, const ParentType &parent);
+  ModuleLikeUnit(const parser::Submodule &m, const ParentType &parent);
+  ~ModuleLikeUnit() = default;
+  ModuleLikeUnit(ModuleLikeUnit &&) = default;
+  ModuleLikeUnit(const ModuleLikeUnit &) = delete;
+
+  ModuleStatement beginStmt;
+  ModuleStatement endStmt;
+  std::list<FunctionLikeUnit> funcs;
+};
+
+struct BlockDataUnit : public ProgramUnit {
+  BlockDataUnit(const parser::BlockData &bd, const ParentType &parent);
+  BlockDataUnit(BlockDataUnit &&) = default;
+  BlockDataUnit(const BlockDataUnit &) = delete;
+};
+
+/// A Program is the top-level PFT
+struct Program {
+  using Units = std::variant<FunctionLikeUnit, ModuleLikeUnit, BlockDataUnit>;
+
+  Program() = default;
+  Program(Program &&) = default;
+  Program(const Program &) = delete;
+
+  std::list<Units> &getUnits() { return units; }
+
+private:
+  std::list<Units> units;
+};
+
+} // namespace pft
+
+/// Create an PFT from the parse tree
+std::unique_ptr<pft::Program> createPFT(const parser::Program &root);
+
+/// Decorate the PFT with control flow annotations
+///
+/// The PFT must be decorated with control-flow annotations to prepare it for
+/// use in generating a CFG-like structure.
+void annotateControl(pft::Program &);
+
+void dumpPFT(llvm::raw_ostream &o, pft::Program &);
+
+} // namespace Fortran::lower
+
+#endif // FORTRAN_LOWER_PFT_BUILDER_H_
--- a/lib/lower/CMakeLists.txt
+++ b/lib/lower/CMakeLists.txt
@@ -1,0 +1,13 @@
+add_library(FortranLower
+  PFTBuilder.cpp
+)
+
+target_link_libraries(FortranLower
+  LLVMSupport
+)
+
+install (TARGETS FortranLower
+  ARCHIVE DESTINATION lib
+  LIBRARY DESTINATION lib
+  RUNTIME DESTINATION bin
+)
--- a/lib/lower/PFTBuilder.cpp
+++ b/lib/lower/PFTBuilder.cpp
@@ -1,0 +1,697 @@
+//===-- lib/lower/PFTBuilder.cc -------------------------------------------===//
+//
+// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
+
+#include "flang/lower/PFTBuilder.h"
+#include "flang/parser/dump-parse-tree.h"
+#include "flang/parser/parse-tree-visitor.h"
+#include "llvm/ADT/DenseMap.h"
+#include <algorithm>
+#include <cassert>
+#include <utility>
+
+namespace Fortran::lower {
+namespace {
+
+/// Helpers to unveil parser node inside parser::Statement<>,
+/// parser::UnlabeledStatement, and common::Indirection<>
+template <typename A>
+struct RemoveIndirectionHelper {
+  using Type = A;
+  static constexpr const Type &unwrap(const A &a) { return a; }
+};
+template <typename A>
+struct RemoveIndirectionHelper<common::Indirection<A>> {
+  using Type = A;
+  static constexpr const Type &unwrap(const common::Indirection<A> &a) {
+    return a.value();
+  }
+};
+
+template <typename A>
+const auto &removeIndirection(const A &a) {
+  return RemoveIndirectionHelper<A>::unwrap(a);
+}
+
+template <typename A>
+struct UnwrapStmt {
+  static constexpr bool isStmt{false};
+};
+template <typename A>
+struct UnwrapStmt<parser::Statement<A>> {
+  static constexpr bool isStmt{true};
+  using Type = typename RemoveIndirectionHelper<A>::Type;
+  constexpr UnwrapStmt(const parser::Statement<A> &a)
+      : unwrapped{removeIndirection(a.statement)}, pos{a.source}, lab{a.label} {
+  }
+  const Type &unwrapped;
+  parser::CharBlock pos;
+  std::optional<parser::Label> lab;
+};
+template <typename A>
+struct UnwrapStmt<parser::UnlabeledStatement<A>> {
+  static constexpr bool isStmt{true};
+  using Type = typename RemoveIndirectionHelper<A>::Type;
+  constexpr UnwrapStmt(const parser::UnlabeledStatement<A> &a)
+      : unwrapped{removeIndirection(a.statement)}, pos{a.source} {}
+  const Type &unwrapped;
+  parser::CharBlock pos;
+  std::optional<parser::Label> lab;
+};
+
+/// The instantiation of a parse tree visitor (Pre and Post) is extremely
+/// expensive in terms of compile and link time, so one goal here is to limit
+/// the bridge to one such instantiation.
+class PFTBuilder {
+public:
+  PFTBuilder() : pgm{new pft::Program}, parents{*pgm.get()} {}
+
+  /// Get the result
+  std::unique_ptr<pft::Program> result() { return std::move(pgm); }
+
+  template <typename A>
+  constexpr bool Pre(const A &a) {
+    bool visit{true};
+    if constexpr (pft::isFunctionLike<A>) {
+      return enterFunc(a);
+    } else if constexpr (pft::isConstruct<A>) {
+      return enterConstruct(a);
+    } else if constexpr (UnwrapStmt<A>::isStmt) {
+      using T = typename UnwrapStmt<A>::Type;
+      // Node "a" being visited has one of the following types:
+      // Statement<T>, Statement<Indirection<T>, UnlabeledStatement<T>,
+      // or UnlabeledStatement<Indirection<T>>
+      auto stmt{UnwrapStmt<A>(a)};
+      if constexpr (pft::isConstructStmt<T> || pft::isOtherStmt<T>) {
+        addEval(pft::Evaluation{stmt.unwrapped, parents.back(), stmt.pos,
+                                stmt.lab});
+        visit = false;
+      } else if constexpr (std::is_same_v<T, parser::ActionStmt>) {
+        addEval(makeEvalAction(stmt.unwrapped, stmt.pos, stmt.lab));
+        visit = false;
+      }
+    }
+    return visit;
+  }
+
+  template <typename A>
+  constexpr void Post(const A &) {
+    if constexpr (pft::isFunctionLike<A>) {
+      exitFunc();
+    } else if constexpr (pft::isConstruct<A>) {
+      exitConstruct();
+    }
+  }
+
+  // Module like
+  bool Pre(const parser::Module &node) { return enterModule(node); }
+  bool Pre(const parser::Submodule &node) { return enterModule(node); }
+
+  void Post(const parser::Module &) { exitModule(); }
+  void Post(const parser::Submodule &) { exitModule(); }
+
+  // Block data
+  bool Pre(const parser::BlockData &node) {
+    addUnit(pft::BlockDataUnit{node, parents.back()});
+    return false;
+  }
+
+  // Get rid of production wrapper
+  bool Pre(const parser::UnlabeledStatement<parser::ForallAssignmentStmt>
+               &statement) {
+    addEval(std::visit(
+        [&](const auto &x) {
+          return pft::Evaluation{x, parents.back(), statement.source, {}};
+        },
+        statement.statement.u));
+    return false;
+  }
+  bool Pre(const parser::Statement<parser::ForallAssignmentStmt> &statement) {
+    addEval(std::visit(
+        [&](const auto &x) {
+          return pft::Evaluation{x, parents.back(), statement.source,
+                                 statement.label};
+        },
+        statement.statement.u));
+    return false;
+  }
+  bool Pre(const parser::WhereBodyConstruct &whereBody) {
+    return std::visit(
+        common::visitors{
+            [&](const parser::Statement<parser::AssignmentStmt> &stmt) {
+              // Not caught as other AssignmentStmt because it is not
+              // wrapped in a parser::ActionStmt.
+              addEval(pft::Evaluation{stmt.statement, parents.back(),
+                                      stmt.source, stmt.label});
+              return false;
+            },
+            [&](const auto &) { return true; },
+        },
+        whereBody.u);
+  }
+
+private:
+  // ActionStmt has a couple of non-conforming cases, which get handled
+  // explicitly here.  The other cases use an Indirection, which we discard in
+  // the PFT.
+  pft::Evaluation makeEvalAction(const parser::ActionStmt &statement,
+                                 parser::CharBlock pos,
+                                 std::optional<parser::Label> lab) {
+    return std::visit(
+        common::visitors{
+            [&](const auto &x) {
+              return pft::Evaluation{removeIndirection(x), parents.back(), pos,
+                                     lab};
+            },
+        },
+        statement.u);
+  }
+
+  // When we enter a function-like structure, we want to build a new unit and
+  // set the builder's cursors to point to it.
+  template <typename A>
+  bool enterFunc(const A &func) {
+    auto &unit = addFunc(pft::FunctionLikeUnit{func, parents.back()});
+    funclist = &unit.funcs;
+    pushEval(&unit.evals);
+    parents.emplace_back(unit);
+    return true;
+  }
+  /// Make funclist to point to current parent function list if it exists.
+  void setFunctListToParentFuncs() {
+    if (!parents.empty()) {
+      std::visit(common::visitors{
+                     [&](pft::FunctionLikeUnit *p) { funclist = &p->funcs; },
+                     [&](pft::ModuleLikeUnit *p) { funclist = &p->funcs; },
+                     [&](auto *) { funclist = nullptr; },
+                 },
+                 parents.back().p);
+    }
+  }
+
+  void exitFunc() {
+    popEval();
+    parents.pop_back();
+    setFunctListToParentFuncs();
+  }
+
+  // When we enter a construct structure, we want to build a new construct and
+  // set the builder's evaluation cursor to point to it.
+  template <typename A>
+  bool enterConstruct(const A &construct) {
+    auto &con = addEval(pft::Evaluation{construct, parents.back()});
+    con.subs.reset(new pft::EvaluationCollection);
+    pushEval(con.subs.get());
+    parents.emplace_back(con);
+    return true;
+  }
+
+  void exitConstruct() {
+    popEval();
+    parents.pop_back();
+  }
+
+  // When we enter a module structure, we want to build a new module and
+  // set the builder's function cursor to point to it.
+  template <typename A>
+  bool enterModule(const A &func) {
+    auto &unit = addUnit(pft::ModuleLikeUnit{func, parents.back()});
+    funclist = &unit.funcs;
+    parents.emplace_back(unit);
+    return true;
+  }
+
+  void exitModule() {
+    parents.pop_back();
+    setFunctListToParentFuncs();
+  }
+
+  template <typename A>
+  A &addUnit(A &&unit) {
+    pgm->getUnits().emplace_back(std::move(unit));
+    return std::get<A>(pgm->getUnits().back());
+  }
+
+  template <typename A>
+  A &addFunc(A &&func) {
+    if (funclist) {
+      funclist->emplace_back(std::move(func));
+      return funclist->back();
+    }
+    return addUnit(std::move(func));
+  }
+
+  /// move the Evaluation to the end of the current list
+  pft::Evaluation &addEval(pft::Evaluation &&eval) {
+    assert(funclist && "not in a function");
+    assert(evallist.size() > 0);
+    evallist.back()->emplace_back(std::move(eval));
+    return evallist.back()->back();
+  }
+
+  /// push a new list on the stack of Evaluation lists
+  void pushEval(pft::EvaluationCollection *eval) {
+    assert(funclist && "not in a function");
+    assert(eval && eval->empty() && "evaluation list isn't correct");
+    evallist.emplace_back(eval);
+  }
+
+  /// pop the current list and return to the last Evaluation list
+  void popEval() {
+    assert(funclist && "not in a function");
+    evallist.pop_back();
+  }
+
+  std::unique_ptr<pft::Program> pgm;
+  /// funclist points to FunctionLikeUnit::funcs list (resp.
+  /// ModuleLikeUnit::funcs) when building a FunctionLikeUnit (resp.
+  /// ModuleLikeUnit) to store internal procedures (resp. module procedures).
+  /// Otherwise (e.g. when building the top level Program), it is null.
+  std::list<pft::FunctionLikeUnit> *funclist{nullptr};
+  /// evallist is a stack of pointer to FunctionLikeUnit::evals (or
+  /// Evaluation::subs) that are being build.
+  std::vector<pft::EvaluationCollection *> evallist;
+  std::vector<pft::ParentType> parents;
+};
+
+template <typename Label, typename A>
+constexpr bool hasLabel(const A &stmt) {
+  auto isLabel{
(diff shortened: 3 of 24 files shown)


include/flang/lower/ASTBuilder.h
```diff
@@ -0,0 +1,332 @@
+//===-- lib/lower/ast-builder.h ---------------------------------*- C++ -*-===//
```
klausler: The relative path name in the license header does not match the name of the file.

lib/lower/ASTBuilder.cpp
```diff
@@ -0,0 +1,872 @@
+//===-- lib/lower/ast-builder.cc ------------------------------------------===//
```
klausler: The relative path name in the license header does not match the name of the file.

.clang-format
```diff
@@ -16,7 +16,7 @@ FixNamespaceComments: false
 IncludeCategories:
   - Regex:           '^<'
     Priority:        4
-  - Regex:           '^"(llvm|llvm-c|clang|clang-c)/'
+  - Regex:           '^"(llvm|llvm-c|clang|clang-c|fir|mlir|mlir-c)/'
```
tskeith: Shouldn't fir be at the same priority as flang? It's in this (sub-)project but the rest listed here are not.

lib/lower/ASTBuilder.cpp
```diff
@@ -0,0 +1,872 @@
...
+/// Annotate the AST with CFG source decorations (see CFGAnnotation) and mark
+/// potential branch targets
+inline void annotateFuncCFG(AST::FunctionLikeUnit &flu) {
+  annotateEvalListCFG(flu.evals, nullptr);
+}
+
+L::StringRef evalName(AST::Evaluation &e) {
+  return std::visit(
+      Co::visitors{
+          [](const Pa::AllocateStmt *) { return "AllocateStmt"; },
```
tskeith: `dump-parse-tree.h` also maps nodes to names. Maybe code could be shared?

lib/lower/ASTBuilder.cpp
```diff
@@ -0,0 +1,872 @@
...
+///
+/// The AST captures a structured view of the program.  The program is a list of
+/// units.  Function like units will contain lists of evaluations.  Evaluations
+/// are either statements or constructs, where a construct contains a list of
+/// evaluations.  The resulting AST structure can then be used to create FIR.
+
+namespace Br = Fortran::lower;
+namespace Co = Fortran::common;
+namespace L = llvm;
+namespace Pa = Fortran::parser;
```
tskeith: Code is easier to understand if the same names are used everywhere. Is `Pa` really that much easier to type than `parser`?

klausler: This was commented upon during a previous review of this (or similar) code.

Are the reviewers of this pull request expected to perform a full review, or should our comments be limited to new items?

lib/lower/ASTBuilder.cpp
```diff
@@ -0,0 +1,872 @@
...
+
+namespace {
+
+/// The instantiation of a parse tree visitor (Pre and Post) is extremely
+/// expensive in terms of compile and link time, so one goal here is to limit
+/// the bridge to one such instantiation.
+class ASTBuilder {
+public:
+  ASTBuilder() {
+    pgm = new AST::Program;
```
tskeith: Where is this freed? Why not use a smart pointer or store values in parents rather than pointers?

.clang-format
schweitzpgi: Until yesterday, f18 didn't use an include structure like llvm.

lib/lower/ASTBuilder.cpp
schweitzpgi: We're going to use short name aliases to avoid proliferation of using namespaces and typing long namespace identifiers everywhere.

sscalpone: _Originally posted by @DavidTruby in https://github.com/flang-compiler/f18/diffs_

I don't think these aid readability in this file.

include/flang/lower/ASTBuilder.h
```diff
@@ -0,0 +1,332 @@
...
+  ModuleLikeUnit(const parser::Submodule &m, const ParentType &parent);
+  ~ModuleLikeUnit() = default;
+
+  const semantics::Scope *scope{nullptr};
+  std::list<ModuleStatement> modStmts;
+  std::list<FunctionLikeUnit> funcs;
+};
+
+struct BlockDataUnit : public ProgramUnit {
+  BlockDataUnit(const parser::BlockData &db, const ParentType &parent);
```
kiranchandramohan: db -> bd ?

lib/lower/ASTBuilder.cpp
```diff
@@ -0,0 +1,872 @@
...
+}
+
+Br::AST::ModuleLikeUnit::ModuleLikeUnit(const Pa::Submodule &m,
+                                        const AST::ParentType &parent)
+    : ProgramUnit{&m, parent} {
+  modStmts.push_back(&std::get<Pa::Statement<Pa::SubmoduleStmt>>(m.t));
+  modStmts.push_back(&std::get<Pa::Statement<Pa::EndSubmoduleStmt>>(m.t));
+}
+
+Br::AST::BlockDataUnit::BlockDataUnit(const Pa::BlockData &db,
```
kiranchandramohan: bd?

include/flang/lower/ASTBuilder.h
```diff
@@ -0,0 +1,332 @@
...
+  }
+  const parser::FunctionStmt *isFunction() {
+    return isA<parser::FunctionStmt>();
+  }
+  const parser::SubroutineStmt *isSubroutine() {
+    return isA<parser::SubroutineStmt>();
+  }
+  const parser::MpSubprogramStmt *isMPSubp() {
+    return isA<parser::MpSubprogramStmt>();
+  }
```
kiranchandramohan: Are these useful/used anywhere?

include/flang/lower/ASTBuilder.h
jeanPerier: Fixed, thanks

.clang-format
DavidTruby: Could we roll a discussion about these clang-format files into #945? It would be nice to have a cohesive clang-format file for the whole F18 project.

lib/lower/ASTBuilder.cpp
DavidTruby: Just to elaborate on what I meant:
I think that when diving into the middle of these files (imagine a bug-fix PR or similar review) it's unclear when immediately reading what the shortened namespaces actually mean. 
I think e.g. `namespace parser = Fortran::parser;` is a lot clearer and not a significantly longer namespace identifier.

include/flang/lower/ASTBuilder.h
```diff
@@ -0,0 +1,332 @@
...
+
+  const semantics::Scope *scope{nullptr}; // scope from front-end
+  std::list<FunctionStatement> funStmts;  // begin/end pair
+  std::list<Evaluation> evals;            // statements
+  std::list<FunctionLikeUnit> funcs;      // internal procedures
```
DavidTruby: See previous comment about `std::list`. If the reason for using `std::list` is that these need push and pop from both the back and front then `std::deque` may be a better choice. However in this PR I only see peek at the front and back and push at the back, in which case `std::vector` is probably more appropriate.

include/flang/lower/ASTBuilder.h
```diff
@@ -0,0 +1,332 @@
...
+
+private:
+  template <typename A>
+  const A *isA() {
```
DavidTruby: asA might be a better name for this than isA (isA in LLVM just does a boolean check it doesn't do a cast)

include/flang/lower/ASTBuilder.h
```diff
@@ -0,0 +1,332 @@
...
+         std::is_same_v<A, parser::CompilerDirective> ||
+         std::is_same_v<A, parser::OpenMPConstruct> ||
+         std::is_same_v<A, parser::OmpEndLoopDirective>;
+}
+
+/// Function-like units can contains lists of evaluations.  These can be
+/// (simple) statements or constructs, where a construct contains its own
+/// evaluations.
+struct Evaluation {
+  using EvalVariant = std::variant<
```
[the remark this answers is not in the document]
jeanPerier: Thanks to the templatized aspect of the parse-tree, if a statement is added to the parse-tree and not here, the functions of ASTBuilder.cpp that visit the parse-tree will most likely crash at f18 compile time because it is not possible to create the related `Evaluation` ctor.
So I am not sure it is worth moving this to parse-tree.h, especially since it is very specific to the AST code.
Do you see other use cases for this variant type that would benefit from moving it to parse-tree.h ?

include/flang/lower/ASTBuilder.h
```diff
@@ -0,0 +1,332 @@
...
+      // action statements
+      const parser::AllocateStmt *, const parser::AssignmentStmt *,
```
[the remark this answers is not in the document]
jeanPerier: Namespace abbreviations were only used in implementation files as a convenience for the implementer. Public headers do not use them.

include/flang/lower/ASTBuilder.h
```diff
@@ -0,0 +1,332 @@
...
+      const parser::EndForallStmt *>;
+
+  Evaluation() = delete;
+  Evaluation(const Evaluation &) = default;
+
+  /// General ctor
+  template <typename A>
+  Evaluation(const A &a, const parser::CharBlock &pos,
+             const std::optional<parser::Label> &lab, const ParentType &p)
+      : u{&a}, parent{p}, pos{pos}, lab{lab} {}
```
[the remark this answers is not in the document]
jeanPerier: Sure, I do not think there was a deep reason for that argument order, I have updated it.

include/flang/lower/ASTBuilder.h
```diff
@@ -0,0 +1,332 @@
...
+
+  /// Construct ctor
+  template <typename A>
+  Evaluation(const A &a, const ParentType &parent) : u{&a}, parent{parent} {
+    static_assert(AST::isConstruct<A>(), "must be a construct");
+  }
+
+  /// is `A` executable (an action statement or compiler generated)?
+  template <typename A>
+  constexpr static bool isAction(const A &a) {
```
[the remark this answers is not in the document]
jeanPerier: The name "action" actually matches Fortran R515 "action-stmt" which this function is intended to identify.
I think `isExectuable` might be confused with "executable-construct".
Why do you think `isExecutable` might be a better fit ?

include/flang/lower/ASTBuilder.h
jeanPerier: Yes, they are used in the bridge when lowering the Fortran program structure to FIR. Unfortunately this code will only come much later in the review process.

include/flang/lower/ASTBuilder.h
jeanPerier: Thanks, I have fixed that.

lib/lower/ASTBuilder.cpp
jeanPerier: Thanks for pointing to this file, I have updated this code to use it (I had to make the `parser::ParseTreeDumper::GetNodeName` static on the way).

lib/lower/ASTBuilder.cpp
jeanPerier: Changed.

lib/lower/ASTBuilder.cpp
```diff
@@ -0,0 +1,783 @@
...
+
+template <typename A>
+constexpr bool hasErrLabel(const A &stmt) {
+  if constexpr (std::is_same_v<A, Pa::ReadStmt> ||
+                std::is_same_v<A, Pa::WriteStmt>) {
+    for (const auto &control : stmt.controls) {
+      if (std::holds_alternative<Pa::ErrLabel>(control.u)) {
+        return true;
+      }
+    }
```
DavidTruby: nit suggestion: 
```
return std::any_of(std::begin(stmt.controls), std::end(stmt.controls), 
  [](const auto& control) { return std::holds_alternative<Pa::ErrLabel>(control.u); });
```

include/flang/lower/ASTBuilder.h
```diff
@@ -0,0 +1,332 @@
...
+  FunctionLikeUnit(const parser::FunctionSubprogram &f,
+                   const ParentType &parent);
+  FunctionLikeUnit(const parser::SubroutineSubprogram &f,
+                   const ParentType &parent);
+  FunctionLikeUnit(const parser::SeparateModuleSubprogram &f,
+                   const ParentType &parent);
+
+  bool isMainProgram() {
+    return std::get_if<const parser::Statement<parser::EndProgramStmt> *>(
+        &funStmts.back());
```
tskeith: `std::holds_alternative` would be clearer as it returns a bool.

include/flang/lower/ASTBuilder.h
```diff
@@ -0,0 +1,332 @@
...
+  }
+  const parser::FunctionStmt *isFunction() {
```
tskeith: It is surprising to have functions that return non-bool start with "is", especially with `bool isMainProgram` right above. `asFunction` might be a better name.

include/flang/lower/ASTBuilder.h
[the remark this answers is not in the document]
kiranchandramohan: Thanks for pointing to the standard.
I was going by the comment given above which says that the check is about whether A is _executable_ and the definition is action statement + compiler generated statement.
Later the definition of isActionStmt function has checks for both isAction and isGenerated.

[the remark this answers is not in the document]
sscalpone: Perhaps this function is better named isActionStmt and the other isActionOrGenerated?

include/flang/lower/ASTBuilder.h
jeanPerier: It will be used in the "Add bridge from AST to FIR" mentioned here: http://lists.llvm.org/pipermail/flang-dev/2020-January/000162.html

include/flang/lower/ASTBuilder.h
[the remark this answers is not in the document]
kiranchandramohan: Ideally, the classification of action statements,  constructs, construct statements and other statements should be in parse-tree.h and we should be able to say here that an evaluation is action-stmt | constructs | construct-smts | other-stmts | compiler-generated-stmts. But i guess the language currently does not provide for that? 
The only thing specific to here is compiler-generated-stmt(?).

lib/lower/ASTBuilder.cpp
DavidTruby: Or even better because this happens multiple times, add:
```
auto isError = [](const auto& v) { return std::holds_alternative<Pa::ErrLabel>(v.u); });
```
before the if constexpr chain and then use
```
return std::any_of(std::begin(stmt.controls), std::end(stmt.controls), isError);
```
in each if constexpr branch.

include/flang/lower/ASTBuilder.h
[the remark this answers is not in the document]
jeanPerier: I see your point. I will try a bit here to see if I can extract this list automatically from the parse-tree definition like https://github.com/flang-compiler/f18/blob/0c3c39d30e3f166a6a1303337c5fd7eead720fd0/include/flang/parser/parse-tree.h#L466
It will require a bit of templates to get rid of the `common::Indirection`.

.clang-format
schweitzpgi: The second reason `fir` is grouped with `mlir` which is grouped with `llvm` is that `fir` is a dialect of `mlir` and makes no sense without `mlir` (which transitively makes no sense without `llvm`). Conversely, the FIR dialect has no dependence on the f18 front-end and could be targeted by other Fortran front-ends.

Agreed with David that debating .clang-format issues should be waged in its own forum.

lib/lower/ASTBuilder.cpp
schweitzpgi: This follows the same principle as to why England, London, and Epping Upland don't appear on a map in the same sized font. Let me explain further.

If one applied the "type everything all the way out" rule universally, we would write something like
```
   Physics::ideal_gases::pressure *
      Physics::ideal_gases::volume =
   Physics::mass::number_of_molecules *
      Physics::constant::Avogadros_number *
      Physics::ideal_gases::temperature
```

Of course, one could argue there are intermediate steps that are just as good, such as
```
   using Physics;
   ideal_gases::pressure * ideal_gases::volume =
      mass::number_of_molecules * constant::Avogadros_number *
      ideal_gases::temperature
```
and so forth. (This leads to a trap. For example, names might be reused many times. Use of the identifier `Expression` might be defined in `Fortran::parser`, `Fortran::evaluate`, `Fortran::semantics`, `mlir`, and `llvm`, this making comprehension of a bare `Expression` more difficult.)

Now, if we look in a copy of "Intro to Physics", we will see the above formula expressed as
```
   Ideal Gas Law:
      P V = n R T
```

Why would the author/editor/publisher present this in such a terse form when we all should assume that long names are better or, at least, just as readable?

The last form, as presented to someone new to the subject, increases understanding, comprehension, and memory retention. In short [pun intended], it's brevity within a given context makes it far more "readable." (There is science behind this. Shapes, pattern matching, relative salience, visual scanning, short-term memory, short-term to long-term memory transfer. Call it how the human brain works.) We know that students will long remember the formula "PV=nRT" decades after their one (and perhaps only) Physics class and longer than the remember the value of Avogadro's number or Boltzmann's constant.

klausler: Why not use `P` then rather than `Pa` for `parser`, if this actually is a good idea?

include/flang/lower/ASTBuilder.h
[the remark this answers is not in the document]
schweitzpgi: I don't think everything should be or needs to be in `parse-tree.h`.  That file is already very large and expensive at compile-time (which was a criticism already made elsewhere).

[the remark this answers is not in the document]
klausler: If it doesn't relate to the representation of Fortran *syntax*, it shouldn't be in `parse-tree.h`.

include/flang/lower/ASTBuilder.h
schweitzpgi: The implementation of some aspects of the code has evolved over time.

include/flang/lower/ASTBuilder.h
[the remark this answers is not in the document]
kiranchandramohan: Thanks @jeanPerier for pointing out ActionStmt in parse-tree.h. It will be great if it can be extracted. But if it is difficult or adds to compile-time overheads then feel free to ignore.
@schweitzpgi , @klausler Is there another place or is your point that this location is fine?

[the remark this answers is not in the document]
schweitzpgi: @kiranchandramohan I think it's fine here. I say this with the caveat that this code is actively being worked on as I type.

This data structure is, for now, a transitory (and very thin) adapter mapping that only exists to simplify lowering. The future may have other ideas eventually.

lib/lower/ASTBuilder.cpp
schweitzpgi: At an earlier time, I had a collision where two namespaces trivially mapped to the same alias letter. We could shorten things to one letter whenever possible, though the current convention is used ubiquitously throughout the project, most of which is _not_ this PR.

klausler: I would prefer spelled-out short namespace names like `parser`.  They don't have to be decoded by the reader.

include/flang/lower/ASTBuilder.h
```diff
@@ -0,0 +1,332 @@
...
+    if (cstr) {
+      cstr->setBranches();
+    }
+  }
+
+  EvalVariant u;
+  ParentType parent;
+  parser::CharBlock pos;
+  std::optional<parser::Label> lab;
+  std::list<Evaluation> *subs{nullptr}; // construct sub-statements
```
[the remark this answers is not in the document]
schweitzpgi: Most Evaluations do not contain a sub-list. The pointer omits calling the std::list constructor as tax on every node in the AST.

include/flang/lower/ASTBuilder.h
schweitzpgi: By the way, the parser uses `std::list`, so the project as a whole hasn't shied away from using that class.

This code is work-in-progress. This PR should not be taken as a complete enumeration of the requirements as to what this data structure may eventually constitute or how it will ultimately be used. (It's already changed, fwiw.) There actually was code that did insertions of nodes into the `list` structure at one point.

We'll make a note that some performance tuning might be in order here.

include/flang/lower/ASTBuilder.h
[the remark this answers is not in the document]
sscalpone: Is a unique_ptr suitable here?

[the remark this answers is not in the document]
DavidTruby: The constructors for `std::list` and `std::vector` are both empty (and trivially inlineable as they are templates), at least in both `libc++` and `libstdc++`. This is so that people can use empty containers in cases like this without issues.

.clang-format
jeanPerier: As requested, discussion moved to: https://github.com/flang-compiler/f18/pull/945#issuecomment-580184803
Please comment directly there if you wish to reply to this thread.

include/flang/lower/ASTBuilder.h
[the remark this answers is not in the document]
jeanPerier: @kiranchandramohan, I have looked a bit into extracting from the parse-tree  as mentioned above and I think it would be more hassle than having this structure here.

Rational:
Action statements would be ok to extract from the parse-tree `ActionStmt`, but the rest would be much harder. Specifically, the "construct statements" is a subset of the child nodes of `ExecutableConstruct` nodes. This subset does not contain nodes that do not matter to lowering such as `BlockSpecificationPart` because all the related information is inside the scopes at that point. So a manual selection of this subset has to made somewhere. Since, this selection is related to lowering and not to the syntax as @klausler mentioned, then I think it is best for it to live here now.

**For all these reasons, I propose a "won't do" resolution to this conversation. If that is acceptable to you, can you mark this conversation as resolved ?**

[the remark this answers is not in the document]
kiranchandramohan: Thanks @jeanPerier for having a look. Just to complete the discussion, if a person is writing a tool and want to go over all the Action, construct and other statements then that person will have to create all this information by himself or include the flang/lower/ASTBuilder.h class. It would have been better to include a file from the parser directory.

I am fine with your resolution.
I don't see an option to resolve. I think it is limited to the author or people with write access. Please mark as resolved.

include/flang/lower/ASTBuilder.h
jeanPerier: Regarding the `std::list` vs `std::vector`, there is another a reason for actually using a list right now; the address of elements inside the lists are taken.  See:

- `FunctionLikeUnit::evals` elements address are taken here: https://github.com/jeanPerier/f18/blob/10e01a4e505fe6f874dffe4d5f08d2503a695576/lib/lower/ASTBuilder.cpp#L308

- `ModuleLikeUnit::funcs` elements address are taken here: https://github.com/jeanPerier/f18/blob/10e01a4e505fe6f874dffe4d5f08d2503a695576/lib/lower/ASTBuilder.cpp#L323

- `FunctionLikeUnit::funcs` elements are indirectly taken (ASTBuilder::funclist points to funcs, address of element of *funclist are taken) at:
https://github.com/jeanPerier/f18/blob/10e01a4e505fe6f874dffe4d5f08d2503a695576/lib/lower/ASTBuilder.cpp#L291

- `Program::units` elements address here: 
https://github.com/jeanPerier/f18/blob/10e01a4e505fe6f874dffe4d5f08d2503a695576/lib/lower/ASTBuilder.cpp#L323

Because `std::vector` element addresses may be [invalidated after insertion](https://en.cppreference.com/w/cpp/container#Iterator_invalidation), that would be a risky choice here.

include/flang/lower/ASTBuilder.h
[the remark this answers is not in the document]
schweitzpgi: @DavidTruby Thanks for pointing that out.

Logically, most `Evaluations` have no body. It makes no sense. One solution would be to use a subclass here. Another would be to use `std::optional<>`. The pointer-to solution here is, effectively, the `std::optional` solution used to work around other issues. The least desirable option, in my opinion, would be to have an list everywhere even if the run-time cost is zero since it would entail two distinct logical states. (1) This `Evaluation` contains other `Evaluation`s but not in this instance. (2) This `Evaluation` can absolutely, positively never contain other `Evaluation`s.

include/flang/lower/ASTBuilder.h
schweitzpgi: The intention is really that of a getter.

include/flang/lower/ASTBuilder.h
DavidTruby: If the addresses are taken and we might resize and invalidate those addresses, `std::deque` can be used instead of `std::list` as that has the same guarantee that addresses will not be invalidated after resizing.

lib/lower/ASTBuilder.cpp
```diff
@@ -0,0 +1,783 @@
...
+    evallist.emplace_back(eval);
+  }
+
+  /// pop the current list and return to the last Evaluation list
+  void popEval() {
+    assert(funclist && "not in a function");
+    evallist.pop_back();
+  }
+
+  AST::Program *pgm;
```
klausler: Either a `unique_ptr` or reference would be safer here.

lib/lower/ASTBuilder.cpp
jeanPerier: Abbreviations have been removed.

lib/lower/ASTBuilder.cpp
```diff
@@ -0,0 +1,783 @@
...
+  std::list<AST::FunctionLikeUnit> *funclist{nullptr};
```
klausler: The extra level of indirection here, what does it mean?  Is there a semantic difference in this data structure between a pointer to an empty list and a null pointer?  If not, just use a list and avoid the confusion about semantics and ownership.

lib/lower/ASTBuilder.cpp
```diff
@@ -0,0 +1,783 @@
...
+  std::vector<std::list<AST::Evaluation> *> evallist;
```
klausler: Same issue here -- the extra level of indirection might not be necessary.

lib/lower/ASTBuilder.cpp
jeanPerier: This was actually freed in the bridge code that will come in the "Add bridge from AST to FIR" was mentioned here: http://lists.llvm.org/pipermail/flang-dev/2020-January/000162.html

However, a smart pointer sounds a good idea here thanks.
`ASTBuilder::pgm` is now an `std::unique_ptr` and `createAST` returns it.

lib/lower/ASTBuilder.cpp
```diff
@@ -0,0 +1,783 @@
...
+    for (const auto &spec : std::get<std::list<Pa::InquireSpec>>(stmt.u)) {
+      if (std::holds_alternative<Pa::ErrLabel>(spec.u)) {
+        return true;
+      }
+    }
+  }
+  return false;
+}
+
+template <typename A>
```
klausler: Could be combined with the previous and following template member function by generalizing over the types (`ErrLabel`, `EorLabel`, &c.) of the control list specifiers.

include/flang/lower/ASTBuilder.h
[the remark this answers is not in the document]
jeanPerier: OK, I see your point. @sscalpone suggestion was applied.

include/flang/lower/ASTBuilder.h
jeanPerier: Thanks, updated to your suggestion.

include/flang/lower/ASTBuilder.h
jeanPerier: Following renaming was done:
`isFunction` -> `getFunction`
`isSubroutine` -> `getSubroutine`
`isMPSubp` -> `getMPSubp`

lib/lower/ASTBuilder.cpp
```diff
@@ -0,0 +1,783 @@
...
+      return true;
+    }
+  }
+  return false;
+}
+
+/// Determine if `callStmt` has alternate returns and if so set `e` to be the
+/// origin of a switch-like control flow
+void altRet(AST::Evaluation &e, const Pa::CallStmt *callStmt,
+            AST::Evaluation *cstr) {
```
klausler: Here and in following functions, if the `cstr` argument pointer can't be null, please use a reference instead.

lib/lower/ASTBuilder.cpp
```diff
@@ -0,0 +1,783 @@
...
+    if (e.isConstruct()) {
+      annotateEvalListCFG(*e.getConstructEvals(), &e);
+      // assume that the entry and exit are both possible branch targets
+      nextIsTarget = true;
+    }
+    if (e.isActionStmt() && e.lab.has_value()) {
+      e.isTarget = true;
+    }
+    std::visit(
+        Co::visitors{
```
klausler: These items could be combined.

include/flang/lower/ASTBuilder.h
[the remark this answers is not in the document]
jeanPerier: Updated to a `std::unique_ptr`. It keeps the the state indication mentioned by @schweitzpgi (different form an empty list).

include/flang/lower/ASTBuilder.h
jeanPerier: The name was updated to `getA`.

lib/lower/ASTBuilder.cpp
jeanPerier: Thanks for the suggestion, it seemed to me a good idea so I did do that.
@klausler, what did you down-voted that ?

include/flang/lower/ASTBuilder.h
```diff
@@ -0,0 +1,332 @@
...
+  std::list<Units> &getUnits() { return units; }
+
+private:
+  std::list<Units> units;
+};
+
+} // namespace AST
+
+/// Create an AST from the parse tree
+AST::Program *createAST(const parser::Program &root);
```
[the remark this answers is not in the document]
jeanPerier: That was updated while you commented, sorry for the bad update timing.

lib/lower/ASTBuilder.cpp
```diff
@@ -0,0 +1,783 @@
...
+/// the bridge to one such instantiation.
+class ASTBuilder {
+public:
+  ASTBuilder() {
+    pgm = new AST::Program;
+    parents.push_back(pgm);
+  }
+
+  /// Get the result
+  AST::Program *result() { return pgm; }
```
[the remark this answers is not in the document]
jeanPerier: That was updated while you commented to `std::unique_ptr`.

lib/lower/ASTBuilder.cpp
jeanPerier: That was updated while you commented.

lib/lower/ASTBuilder.cpp
jeanPerier: Here, the pointer to a list is important, ASTbuilder does not own the list of function it is working on, it points to `FunctionLikeUnit::funcs` or `ModuleLikeUnit::funcs` while building the AST.

lib/lower/ASTBuilder.cpp
jeanPerier: Same answer, the ASTbuilder does not own the lists.

lib/lower/ASTBuilder.cpp
klausler: The range-based `for` idiom is more clear than the cluttered call to the iterator-based "algorithm".

lib/lower/ASTBuilder.cpp
klausler: Can the pointer be null?

jeanPerier: Yes, this state is actually used here: https://github.com/flang-compiler/f18/blob/e6345f5c3ae6e521fb8994e501999a036cfb98a9/lib/lower/ASTBuilder.cpp#L349
It should be null when entering external subprograms but not when entering module and internal subprograms.

include/flang/lower/ASTBuilder.h
schweitzpgi: `std::list` was used intentionally. The reason for that was to allow O(1) insertions anywhere in the structure to simplify lowering.

lib/lower/ASTBuilder.cpp
klausler: Please note that in a comment; thanks.

include/flang/lower/ASTBuilder.h
```diff
@@ -0,0 +1,332 @@
+//===-- include/flang/lower/AstBuilder.h ------------------------*- C++ -*-===//
```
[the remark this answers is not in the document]
jeanPerier: Sorry for that, I wish clang-format could do that for us.... Fixed.

include/flang/lower/ASTBuilder.h
```diff
@@ -0,0 +1,332 @@
...
+//
+// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
+
+#ifndef FORTRAN_LOWER_AST_BUILDER_H_
+#define FORTRAN_LOWER_AST_BUILDER_H_
```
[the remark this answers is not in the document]
jeanPerier: What name would you suggest then ? `ParseTreeIndex` ?

include/flang/lower/ASTBuilder.h
```diff
@@ -0,0 +1,332 @@
...
+
+struct Evaluation;
+struct Program;
+struct ModuleLikeUnit;
+struct FunctionLikeUnit;
+
+using ParentType =
+    std::variant<Program *, ModuleLikeUnit *, FunctionLikeUnit *, Evaluation *>;
+
+enum class CFGAnnotation {
```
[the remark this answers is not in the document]
jeanPerier: I have added comments.

include/flang/lower/ASTBuilder.h
```diff
@@ -0,0 +1,332 @@
...
+               const parser::SubroutineSubprogram *, const parser::Module *,
+               const parser::Submodule *,
+               const parser::SeparateModuleSubprogram *,
+               const parser::BlockData *>
+      p;
+  ParentType parent;
+};
+
+/// Function-like units have similar structure. They all can contain executable
+/// statements.
```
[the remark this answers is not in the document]
jeanPerier: Right, these are stored in `FunctionLikeUnit::funcs`: https://github.com/flang-compiler/f18/blob/4a73dec857715a18da053599182b041c61414c51/include/flang/lower/ASTBuilder.h#L283
I have updated the comment to:
> Function-like units have similar structure. They all can contain executable statements as well as other function-like units (internal procedures and function statements).

lib/lower/ASTBuilder.cpp
```diff
@@ -0,0 +1,783 @@
+//===-- lib/lower/AstBuilder.cc -------------------------------------------===//
```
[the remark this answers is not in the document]
jeanPerier: Thanks, fixed.

lib/lower/ASTBuilder.cpp
```diff
@@ -0,0 +1,783 @@
...
+//
+//===----------------------------------------------------------------------===//
+
+#include "flang/lower/ASTBuilder.h"
+#include "flang/parser/dump-parse-tree.h"
+#include "flang/parser/parse-tree-visitor.h"
+#include <cassert>
+#include <utility>
+
+/// Build a light-weight AST to help with lowering to FIR.  The AST will
```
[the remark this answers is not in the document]
jeanPerier: Good point, done.

lib/lower/ASTBuilder.cpp
jeanPerier: Comment added above `funclist` member declaration:
 > funclist points to FunctionLikeUnit::funcs list (resp. ModuleLikeUnit::funcs) when building a FunctionLikeUnit (resp. ModuleLikeUnit) to store internal procedure (resp. module procedures). Otherwise (e.g. when building the top level Program), it is null.

lib/lower/ASTBuilder.cpp
jeanPerier: Comment added:
> evallist is a stack of pointer to FunctionLikeUnit::evals (or Evaluation::subs) that are being build.

lib/lower/ASTBuilder.cpp
```diff
@@ -0,0 +1,836 @@
...
+  void Post(const parser::Submodule &) { exitModule(); }
+
+  // Function like
+
+  bool Pre(const parser::MainProgram &node) { return enterFunc(node); }
+  bool Pre(const parser::FunctionSubprogram &node) { return enterFunc(node); }
+  bool Pre(const parser::SubroutineSubprogram &node) { return enterFunc(node); }
+  bool Pre(const parser::SeparateModuleSubprogram &node) {
+    return enterFunc(node);
+  }
```
kiranchandramohan: Here, it seems the same function is called. Isn't it possible to write a single Pre function for all these?
Same question for the many following Post functions below.

lib/lower/ASTBuilder.cpp
jeanPerier: `cstr` can be null see: https://github.com/flang-compiler/f18/blob/16ab9dacd8993694ace6c1c85dd8a4763ba89c21/lib/lower/ASTBuilder.cpp#L649
The makes a differences here:
https://github.com/flang-compiler/f18/blob/16ab9dacd8993694ace6c1c85dd8a4763ba89c21/include/flang/lower/ASTBuilder.h#L224

This is because `cstr` is the current construct and at the top level of a procedure/program there is no such "current construct".

I have added comments saying that.

include/flang/lower/ASTBuilder.h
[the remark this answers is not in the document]
klausler: It's a summary or overview of the executable parts of the parsed source file, so maybe something like `ExecutionPartsIndex`.  I'm sure you can come up with something, if you agree that `AST` would be problematic.

include/flang/lower/ASTBuilder.h
```diff
@@ -0,0 +1,332 @@
...
+
+namespace Fortran::lower {
+namespace AST {
+
+struct Evaluation;
+struct Program;
+struct ModuleLikeUnit;
+struct FunctionLikeUnit;
+
+using ParentType =
```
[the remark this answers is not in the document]
jeanPerier: They are non-nullable, I think `std::reference_wrapper` and such in variants makes visit/get... a bit more painful, but I agree some pointer safety would be good. So I've made `ParentType` a class and protected the pointers by with a constructor taking reference and making the variant member `const`.

include/flang/lower/ASTBuilder.h
```diff
@@ -0,0 +1,332 @@
...
+  Terminate
+};
+
+/// Compiler-generated jump
+///
+/// This is used to convert implicit control-flow edges to explicit form in the
+/// decorated AST
+struct CGJump {
+  CGJump(Evaluation *to) : target{to} {}
+  Evaluation *target{nullptr};
```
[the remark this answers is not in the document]
jeanPerier: Made it a reference.

include/flang/lower/ASTBuilder.h
```diff
@@ -0,0 +1,332 @@
...
+};
+
+/// is `A` a construct (or directive)?
+template <typename A>
+constexpr static bool isConstruct() {
+  return std::is_same_v<A, parser::AssociateConstruct> ||
```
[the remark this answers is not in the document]
jeanPerier: Thanks for pointing to this, I have split the `EvalVariant` and use that here.

include/flang/lower/ASTBuilder.h
```diff
@@ -0,0 +1,332 @@
...
+         std::is_same_v<A, parser::OpenMPConstruct> ||
+         std::is_same_v<A, parser::OmpEndLoopDirective>;
+}
+
+/// Function-like units can contains lists of evaluations.  These can be
+/// (simple) statements or constructs, where a construct contains its own
+/// evaluations.
+struct Evaluation {
+  using EvalVariant = std::variant<
+      // action statements
```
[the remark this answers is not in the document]
jeanPerier: 1. I have split the variant as suggested.
2. As for `ParentType` I think putting reference wrapper inside variant made it more cumbersome to handle them. Since this is a member type, I it is protected it with the constructor that takes references. I made the pointer constant and added getters/visitors working on normal reference to the parse-tree to hide pointers on the user side.

include/flang/lower/ASTBuilder.h
```diff
@@ -0,0 +1,332 @@
...
+  }
+
+  /// is `A` executable (an action statement or compiler generated)?
+  template <typename A>
+  constexpr static bool isAction(const A &a) {
+    return !AST::isConstruct<A>() && !isOther(a);
+  }
+
+  /// is `A` a compiler-generated evaluation?
+  template <typename A>
```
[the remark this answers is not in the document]
jeanPerier: Changed as suggested.

include/flang/lower/ASTBuilder.h
```diff
@@ -0,0 +1,332 @@
...
+  bool containsBranches{false}; // construct contains branches
+};
+
+/// A program is a list of program units.
+/// These units can be function like, module like, or block data
+struct ProgramUnit {
+  template <typename A>
+  ProgramUnit(A *ptr, const ParentType &parent) : p{ptr}, parent{parent} {}
+
+  std::variant<const parser::MainProgram *, const parser::FunctionSubprogram *,
```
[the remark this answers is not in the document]
jeanPerier: Same as for the other parse-tree pointers, they are inside a struct that can ensure they are not null and hide the pointer aspects. I made the member a constant to strengthen the guarantee it will never be null given the ctors take references.

include/flang/lower/ASTBuilder.h
```diff
@@ -0,0 +1,332 @@
...
+    return isA<parser::FunctionStmt>();
+  }
+  const parser::SubroutineStmt *isSubroutine() {
+    return isA<parser::SubroutineStmt>();
+  }
+  const parser::MpSubprogramStmt *isMPSubp() {
+    return isA<parser::MpSubprogramStmt>();
+  }
+
+  const semantics::Scope *scope{nullptr}; // scope from front-end
```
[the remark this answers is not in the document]
jeanPerier: It is actually not used anymore, so I have removed it. If and when it is used again, I agree a reference is a safer choice and should be possible.

lib/lower/ASTBuilder.cpp
jeanPerier: It is possible do to so, but as I answered to Peter, we do not want to condense the implementation too much based on today's pattern because we think the explicit Pre/Post makes it easy to reason about and work with for now.

lib/lower/ASTBuilder.cpp
```diff
@@ -0,0 +1,783 @@
...
+  void Post(const Pa::Statement<Co::Indirection<Pa::NamelistStmt>> &s) {
+    addEval(makeEvalIndirect(s));
+  }
+
+  //
+  // Construct statements
+  //
+
+  void Post(const Pa::Statement<Pa::AssociateStmt> &s) {
```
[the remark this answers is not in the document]
jeanPerier: I agree this could be condensed, but this might make it harder to work with and right now Val is also working on this file so I do not want to lock him up if he needs to modify these.

lib/lower/ASTBuilder.cpp
```diff
@@ -0,0 +1,783 @@
...
+          return AST::Evaluation{x, parents.back(), s.source, s.label};
+        },
+        s.statement.u));
+  }
+
+  //
+  // Constructs (enter and exit)
+  //
+
+  bool Pre(const Pa::AssociateConstruct &c) { return enterConstruct(c); }
```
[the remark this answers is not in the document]
jeanPerier: Same answer, it is easy to reason about and update explicit Pre/Post while lowering of more complex structure is progressing. We would rather reduce the repetition when we are confident the patterns are stable.

lib/lower/ASTBuilder.cpp
jeanPerier: Same answer as for other comments about combining code with templates, I agree they could, I think it also makes them easier to reason about right now.

lib/lower/ASTBuilder.cpp
jeanPerier: Same answer as for other comments about combining code with templates, I agree they could, I think it also makes them easier to reason about right now.

include/flang/lower/ASTBuilder.h
```diff
@@ -0,0 +1,332 @@
...
+  ParentType parent;
+  parser::CharBlock pos;
+  std::optional<parser::Label> lab;
+  std::list<Evaluation> *subs{nullptr}; // construct sub-statements
+  CFGAnnotation cfg{CFGAnnotation::None};
+  bool isTarget{false};         // this evaluation is a control target
+  bool containsBranches{false}; // construct contains branches
+};
+
+/// A program is a list of program units.
```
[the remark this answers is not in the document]
jeanPerier: I have explored your suggestion a bit, I see the following differences:
1. The "AST" unit class have a pointer to the "end" nodes, the `ProgramTree` does not.
2. One of the "AST" goal is to provide pointers to parent nodes which the `ProgramTree` does not provide. 
3. The `ProgramTree` has a single type for its nodes, that are split in 4 types in the "AST".

I agree they provide a similar view of the parse-tree, `ProgramTree` could be used if modified to handle point 1. and 2. above (Point 2. would require making `ProgramTree` nodes address stable). Point 3. is a bit more annoying because the type split of the nodes in the "AST" provides a compile time safety when going through the AST to lower.

lib/lower/ASTBuilder.cpp
klausler: I disagree; a template would ease reasoning about the code by making it clear that all these cases have identical handling.

include/flang/lower/ASTBuilder.h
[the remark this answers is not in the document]
klausler: Could semantics be changed to use the "AST" in place of the program tree and save a traversal of the parse tree?

include/flang/lower/ASTBuilder.h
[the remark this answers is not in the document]
jeanPerier: It was renamed to the Pre-FIR tree (PFT).

include/flang/lower/ASTBuilder.h
```diff
@@ -0,0 +1,332 @@
...
+      cstr->setBranches();
+    }
+  }
+
+  EvalVariant u;
+  ParentType parent;
+  parser::CharBlock pos;
+  std::optional<parser::Label> lab;
+  std::list<Evaluation> *subs{nullptr}; // construct sub-statements
+  CFGAnnotation cfg{CFGAnnotation::None};
```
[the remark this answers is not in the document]
jeanPerier: This member stores information related to the impact of the parse-tree node on the control-flow graph.

lib/lower/ASTBuilder.cpp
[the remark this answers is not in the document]
jeanPerier: After discussing with Val, I have condensed this as suggested.

lib/lower/ASTBuilder.cpp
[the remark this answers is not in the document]
jeanPerier: Eventually condensed as suggested.

lib/lower/ASTBuilder.cpp
jeanPerier: Eventually templatized as suggested.

lib/lower/ASTBuilder.cpp
jeanPerier: Eventually condensed as suggested.

lib/lower/ASTBuilder.cpp
jeanPerier: Eventually condensed as suggested.

lib/lower/PFTBuilder.cpp
```diff
@@ -0,0 +1,627 @@
...
+                   },
+                   [&](PFT::FunctionLikeUnit &func) {
+                     dumpFunctionLikeUnit(outputStream, func);
+                     for (auto &func : func.funcs) {
+                       dumpFunctionLikeUnit(outputStream, func);
+                     }
+                   },
+                   [&](PFT::ModuleLikeUnit &unit) {
+                     for (auto &func : unit.funcs) {
+                       dumpFunctionLikeUnit(outputStream, func);
```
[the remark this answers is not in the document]
jeanPerier: Thanks, there was an issue here. I am working on tests for the pre-fir tree and will ensure this is fixed.

include/flang/lower/PFTBuilder.h
```diff
@@ -0,0 +1,376 @@
...
+    return getA<parser::FunctionStmt>();
+  }
+  const parser::SubroutineStmt *getSubroutine() {
+    return getA<parser::SubroutineStmt>();
+  }
+  const parser::MpSubprogramStmt *getMPSubp() {
+    return getA<parser::MpSubprogramStmt>();
+  }
+
+  std::list<FunctionStatement> funStmts; // begin/end pair
```
tskeith: Does funStmts just contain an optional begin statement and an optional end statement? This seems like an unusual way to represent that. Why not separate members for the begin and end? Then the begin data member would only be able to contain one of the begin statements and similarly for end.

include/flang/lower/PFTBuilder.h
```diff
@@ -0,0 +1,376 @@
...
+/// Function-like units can contains lists of evaluations.  These can be
+/// (simple) statements or constructs, where a construct contains its own
+/// evaluations.
+struct Evaluation {
+  using EvalTuple = common::CombineTuples<ActionStmts, OtherStmts, Constructs,
+                                          ConstructStmts>;
+
+  /// Hide non-nullable pointers to the parse-tree node.
+  template <typename A>
+  using MakeRefType = const A *const;
```
tskeith: Is the second `const` necessary?

lib/lower/PFTBuilder.cpp
```diff
@@ -0,0 +1,627 @@
...
+  return walker.result();
+}
+
+void annotateControl(PFT::Program &pft) {
+  for (auto &unit : pft.getUnits()) {
+    std::visit(common::visitors{
+                   [](PFT::BlockDataUnit &) {},
+                   [](PFT::FunctionLikeUnit &func) {
+                     annotateFuncCFG(func);
+                     for (auto &statement : func.funcs) {
```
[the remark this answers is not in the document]
jeanPerier: Fixed, thanks for pointing that out.

lib/lower/PFTBuilder.cpp
```diff
@@ -0,0 +1,627 @@
...
+                       annotateFuncCFG(statement);
+                     }
+                   },
+                   [](PFT::ModuleLikeUnit &unit) {
+                     for (auto &func : unit.funcs) {
+                       annotateFuncCFG(func);
```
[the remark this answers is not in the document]
jeanPerier: Thanks, fixed by handling this inside `annotateFuncCFG`.

lib/lower/PFTBuilder.cpp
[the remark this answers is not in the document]
jeanPerier: OK, that is now fixed.

jeanPerier:
Regarding @kiranchandramohan comments:
> Can you add a few tests to demonstrate and test the functionality? It seems you have a pretty-printer/dumpAST function which can be used for this purpose.

I have added LIT tests to demonstrate that the Pre FIR tree captures the program structure as intended. I added an option to `f18` driver to dump the PFT. Beware, this will bring the first real link-time dependency to LLVM libraries.

include/flang/lower/PFTBuilder.h
jeanPerier: That seems reasonable to me, I have forwarded the question to Val.

include/flang/lower/PFTBuilder.h
jeanPerier: Functionally no, but it gives the insurance that no one can change the pointers after the constructors (which are enforcing that the pointers are not `nullptr`). Since I do not foresee use cases of re-defining the pointer to the parse-tree, I added this second `const`. Do you see a reason not to put it ?

tskeith:
Compiling f18.cpp I'm getting errors like this (clang 9 on macos):
```
In file included from ../../tools/f18/f18.cpp:14:
In file included from ../../include/flang/lower/PFTBuilder.h:14:
In file included from /Users/tsk/llvm/f18-llvm/llvm/include/llvm/Support/raw_ostream.h:16:
In file included from /Users/tsk/llvm/f18-llvm/llvm/include/llvm/ADT/SmallVector.h:17:
In file included from /Users/tsk/llvm/f18-llvm/llvm/include/llvm/Support/AlignOf.h:16:
/Users/tsk/llvm/f18-llvm/llvm/include/llvm/Support/Compiler.h:571:49: error: unused parameter 'Size' [-Werror,-Wunused-parameter]
inline void deallocate_buffer(void *Ptr, size_t Size, size_t Alignment) {
                                                ^
```
This is deallocate_buffer:
```
inline void deallocate_buffer(void *Ptr, size_t Size, size_t Alignment) {
  ::operator delete(Ptr
#ifdef __cpp_sized_deallocation
                    ,
                    Size
#endif
#ifdef __cpp_aligned_new
                    ,
                    std::align_val_t(Alignment)
#endif
  );
}
```
So you may or may not get unused parameter errors depending on the #defines in effect.

lib/lower/CMakeLists.txt
```diff
@@ -0,0 +1,15 @@
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wno-error")
```
tskeith: Can the warnings be fixed or suppressed?

include/flang/lower/PFTBuilder.h
vdonaldson: Changes here are ok with me.  That's high level code that shouldn't affect detail code.

tskeith:
For what it's worth, I was able to build and pass the tests on macos using clang 9 and LLVM built with clang 9.

I wasn't able to get it to work with gcc: f18 built with gcc doesn't link against LLVM built with clang, and I wasn't able to build LLVM with gcc.

lib/lower/CMakeLists.txt
jeanPerier: I cannot fixed those easily inside LLVM headers. Alos, I could certainly not fix them in released LLVM versions. However, I found a way to tell compilers to ignore warnings from headers depending on which directory they live in: adding `SYSTEM` to the related `include_directories`. It worked in my tests with clang 8 and gcc 8.3 on linux.

The drone integration are still failing with clang, but this now has to do with ZLIB linking issues and not unused parameter warnings anymore.

lib/lower/ASTBuilder.cpp
DavidTruby: I don't really agree that the range-for is more clear, I think in the any_of case it's really easy to see that you're trying to do a boolean check on the entire vector. However as I said in my suggestion I don't feel particularly strongly about it, I just thought this read a little better.

lib/lower/CMakeLists.txt
tskeith: With that change you can remove the -Wno-error flag here.

include/flang/lower/PFTBuilder.h
jeanPerier: OK, I have changed it to an optional field `beginStmt` and a mandatory field `endStmt`. Same for `ModuleLikeUnit`, except the `beginStmt` is not optional there.

lib/lower/CMakeLists.txt
jeanPerier: Good point, thanks. Done.

include/flang/lower/PFTBuilder.h
```diff
@@ -317,15 +317,20 @@ struct FunctionLikeUnit : public ProgramUnit {
...
   }
 
-  std::list<FunctionStatement> funStmts; // begin/end pair
-  EvaluationCollection evals;            // statements
-  std::list<FunctionLikeUnit> funcs;     // internal procedures
+  /// Anonymous programs do not have a begin statement
+  std::optional<FunctionStatement> beginStmt;
+  FunctionStatement endStmt;
+  EvaluationCollection evals;        // statements
+  std::list<FunctionLikeUnit> funcs; // internal procedures
```
tskeith: Interesting change on these two lines.

include/flang/lower/PFTBuilder.h
```diff
@@ -0,0 +1,394 @@
...
+/// data structure may <em>not</em> be changed between the construction of the
+/// PFT and all of its uses.
+///
+/// The PFT captures a structured view of the program.  The program is a list of
+/// units.  Function like units will contain lists of evaluations.  Evaluations
+/// are either statements or constructs, where a construct contains a list of
+/// evaluations. The resulting PFT structure can then be used to create FIR.
+
+namespace Fortran::lower {
+namespace PFT {
```
tskeith: Namespaces should be all lowercase like "llvm" and "clang". Otherwise references to things like `PFT::Program` look like PFT is a class.

sscalpone:
Perhaps change the title of the PR to match the new name.

tools/f18/f18.cpp
```diff
@@ -475,6 +487,9 @@ int main(int argc, char *const argv[]) {
       options.needProvenanceRangeToCharBlockMappings = true;
     } else if (arg == "-fdebug-dump-parse-tree") {
       driver.dumpParseTree = true;
+    } else if (arg == "-fdebug-pre-fir-tree") {
+      driver.dumpPreFirTree = true;
+    } else if (arg == "-fdebug-resolve-names") {
```
kiranchandramohan: Was this accidentally added?

test-lit/lower/pre-fir-tree01.f90
```diff
@@ -0,0 +1,129 @@
...
+
+! CHECK: Subroutine foo
+subroutine foo()
+  ! CHECK: <<DoConstruct>>
+  ! CHECK: NonLabelDoStmt
+  do i=1,5
+    ! CHECK: PrintStmt
+    print *, "hey"
+    ! CHECK: <<DoConstruct>>
+    do j=1,5
```
kiranchandramohan: Missed a check for dostmt?

jeanPerier: Yes, thanks. Note that I did not always add checks for construct If/Do statements when I was only testing other statements that I put inside. But I have fixed it here since there is a check for the end statements.

tools/f18/f18.cpp
jeanPerier: Yes, thanks for catching that. Fixed.

jeanPerier:
> Thanks @jeanPerier for making available the tests.
> 
> Is there a test for CGJump?

No because the code that is intended to introduce them comes in a later patch (the PR with the bridge mentioned in an earlier thread). By the way, `CGjump` may be deleted in the FIR development fork, so it may be removed from here in the later PR with bridge. And if it is still there, related tests will come with this PR.

tskeith:
> For what it's worth, I was able to build and pass the tests on macos using clang 9 and LLVM built with clang 9.
> 
> I wasn't able to get it to work with gcc: f18 built with gcc doesn't link against LLVM built with clang, and I wasn't able to build LLVM with gcc.

I've found that I can use the same build of LLVM (compiled with clang) for both gcc and clang builds of f18 by adding this to the gcc cmake command:
`-DCMAKE_EXE_LINKER_FLAGS="-L$CLANG/lib -lc++"`

include/flang/lower/PFTBuilder.h
jeanPerier: Discussion about this clang-format side-effect moved here: https://github.com/flang-compiler/f18/pull/945#issuecomment-585856610

include/flang/lower/PFTBuilder.h
jeanPerier: `PFT` namespace was changed to `pft`.

PR #961, opened 2020-01-28
Fix another bug checking simple contiguity

tskeith:
The test still wasn't correct for structure components. If the last
part-ref is a non-array or a single array element, but the whole
ArrayRef has non-zero rank, it is not contiguous. Otherwise, if there
are subscripts on the last part-ref they can be checked normally.

### DIFF
--- a/test/semantics/assign03.f90
+++ b/test/semantics/assign03.f90
@@ -142,12 +142,22 @@ contains
     end type
     type(t), target :: x
     type(t), target :: y(10,10)
+    integer :: v(10)
     p(1:16) => x%a
+    p(1:8) => x%a(:,3:4)
     p(1:1) => x%b  ! We treat scalars as simply contiguous
+    p(1:1) => x%a(1,1)
+    p(1:1) => y(1,1)%a(1,1)
+    p(1:1) => y(:,1)%a(1,1)  ! Rank 1 RHS
     !ERROR: Pointer bounds remapping target must have rank 1 or be simply contiguous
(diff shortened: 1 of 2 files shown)


tskeith:
Let me know if you can think of any other cases that should be tested. I'm still not real confident in this solution.

test/semantics/assign03.f90
```diff
@@ -143,18 +143,25 @@ subroutine s10
     type(t), target :: x
     type(t), target :: y(10,10)
     p(1:16) => x%a
+    p(1:8) => x%a(:,3:4)
     p(1:1) => x%b  ! We treat scalars as simply contiguous
+    p(1:1) => x%a(1,1)
+    p(1:1) => y(1,1)%a(1,1)
+    p(1:1) => y(:,1)%a(1,1)  ! Rank 1 RHS
```
klausler: This case is not simply contiguous.  See 9.5.4; the final *part-ref* here isn't the one with rank > 0.

As I understand this part of the standard, there really can't be anything to the right of the *part-ref* with rank in a simply contiguous *data-ref* (no components, subscripted or otherwise; no complex components; no substrings).  And other constraints prevent multiple *part-ref*s from having rank > 0.

tskeith:
This is a new solution based on our discussion. CheckSubscripts now returns enough information to be able to detect the trivially contiguous case (rank 0 base and all scalar subscripts).

I also added some tests to verify that vector subscripts are handled correctly.

PR #962, opened 2020-02-03
[OpenMP] Predetermined rules for loop index variables

ichoyjx:
This refers to three rules in OpenMP 4.5 Spec 2.15.1.1:
  * The loop iteration variable(s) in the associated do-loop(s) of a do,
    parallel do, taskloop, or distribute construct is (are) private.
  * The loop iteration variable in the associated do-loop of a simd
    construct with just one associated do-loop is linear with a linear-step
    that is the increment of the associated do-loop.
  * The loop iteration variables in the associated do-loops of a simd
    construct with multiple associated do-loops are lastprivate.

A simple example:
```
implicit none
  integer :: N = 1024
  integer i, j, k
  !$omp parallel do collapse(3)
  do i=1, N  <- i is private
     do j=1, N  <- j is private
        do k=1, N  <- k is private
        enddo
     enddo
  enddo
end
```

If `collapse` clause is not present, the associated do-loop for construct
`parallel do` is only `i` loop. With `collapse(n)`, `i`, `j`, and `k` are
all associated do-loops and the loop index variables are private to the
OpenMP construct:

```
implicit none
 !DEF: /MainProgram1/n ObjectEntity INTEGER(4)
 integer :: n = 1024
 !DEF: /MainProgram1/i ObjectEntity INTEGER(4)
 !DEF: /MainProgram1/j ObjectEntity INTEGER(4)
 !DEF: /MainProgram1/k ObjectEntity INTEGER(4)
 integer i, j, k
!$omp parallel do  collapse(3)
 !DEF: /MainProgram1/Block1/i (OmpPrivate) HostAssoc INTEGER(4)
 !REF: /MainProgram1/n
 do i=1,n
  !DEF: /MainProgram1/Block1/j (OmpPrivate) HostAssoc INTEGER(4)
  !REF: /MainProgram1/n
  do j=1,n
   !DEF: /MainProgram1/Block1/k (OmpPrivate) HostAssoc INTEGER(4)
   !REF: /MainProgram1/n
   do k=1,n
   end do
  end do
 end do
end program
```

This implementation assumes that the structural checks for do-loops
are done at this point, for example the `n` in `collapse(n)` should
be no more than the number of actual perfectly nested do-loops, etc..

### DIFF
--- a/lib/semantics/resolve-names.cpp
+++ b/lib/semantics/resolve-names.cpp
@@ -1237,14 +1237,6 @@ private:
   }
   std::size_t GetAssociatedLoopLevelFromClauses(const parser::OmpClauseList &);
 
-  template<typename T>
-  std::optional<std::int64_t> GetClauseIntValue(const T &x) {
-    if (AnalyzeExpr(context_, x.v)) {
-      return GetIntValue(x.v);
-    }
-    return std::nullopt;
-  }
-
   Symbol &MakeAssocSymbol(const SourceName &name, Symbol &prev) {
     const auto pair{
         GetContext().scope.try_emplace(name, Attrs{}, HostAssocDetails{prev})};
@@ -6054,26 +6046,25 @@ std::size_t OmpAttributeVisitor::GetAssociatedLoopLevelFromClauses(
     if (const auto *orderedClause{
             std::get_if<parser::OmpClause::Ordered>(&clause.u)}) {
       if (const auto v{
-              GetClauseIntValue<parser::OmpClause::Ordered>(*orderedClause)}) {
+              evaluate::ToInt64(resolver_.EvaluateIntExpr(orderedClause->v))}) {
         orderedLevel = *v;
       }
     }
     if (const auto *collapseClause{
             std::get_if<parser::OmpClause::Collapse>(&clause.u)}) {
-      if (const auto v{GetClauseIntValue<parser::OmpClause::Collapse>(
-              *collapseClause)}) {
+      if (const auto v{evaluate::ToInt64(
+              resolver_.EvaluateIntExpr(collapseClause->v))}) {
         collapseLevel = *v;
       }
     }
   }
 
-  if (orderedLevel &&
-      (!collapseLevel || (collapseLevel && orderedLevel >= collapseLevel))) {
+  if (orderedLevel && (!collapseLevel || orderedLevel >= collapseLevel)) {
     return orderedLevel;
   } else if (!orderedLevel && collapseLevel) {
     return collapseLevel;
-  }
-  return 1; // default is outermost loop
+  }  // orderedLevel < collapseLevel is an error handled in structural checks
+  return 1;  // default is outermost loop
 }
 
 // 2.15.1.1 Data-sharing Attribute Rules - Predetermined
@@ -6085,11 +6076,10 @@ std::size_t OmpAttributeVisitor::GetAssociatedLoopLevelFromClauses(
 //   - The loop iteration variables in the associated do-loops of a simd
 //     construct with multiple associated do-loops are lastprivate.
 //
-// This assumes that the do-loops association for collapse/ordered clause has
-// been performed (the number of nested do-loops >= n).
+// TODO: This assumes that the do-loops association for collapse/ordered
+//       clause has been performed (the number of nested do-loops >= n).
 void OmpAttributeVisitor::PrivatizeAssociatedLoopIndex(
     const parser::OpenMPLoopConstruct &x) {
-  using Bounds = parser::LoopControl::Bounds;
   std::size_t level{GetContext().associatedLoopLevel};
   Symbol::Flag ivDSA{Symbol::Flag::OmpPrivate};
   if (simdSet.test(GetContext().directive)) {
@@ -6104,16 +6094,19 @@ void OmpAttributeVisitor::PrivatizeAssociatedLoopIndex(
   for (const parser::DoConstruct *loop{&*outer}; loop && level > 0; --level) {
     // go through all the nested do-loops and resolve index variables
     auto &loopControl{loop->GetLoopControl().value()};
+    using Bounds = parser::LoopControl::Bounds;
     const Bounds &bounds{std::get<Bounds>(loopControl.u)};
     const parser::Name &iv{bounds.name.thing};
-    auto *symbol{ResolveOmp(iv, ivDSA)};
-    iv.symbol = symbol; // adjust the symbol within region
-    AddToContextObjectWithDSA(*symbol, ivDSA);
+    if (auto *symbol{ResolveOmp(iv, ivDSA)}) {
+      iv.symbol = symbol;  // adjust the symbol within region
+      AddToContextObjectWithDSA(*symbol, ivDSA);
+    }
 
     const auto &block{std::get<parser::Block>(loop->t)};
     const auto it{block.begin()};
-    loop = GetDoConstructIf(*it);
+    loop = it != block.end() ? GetDoConstructIf(*it) : nullptr;
   }
+  CHECK(level == 0);
 }
 
 bool OmpAttributeVisitor::Pre(const parser::OpenMPSectionsConstruct &x) {
--- a/lib/semantics/resolve-names.cpp
+++ b/lib/semantics/resolve-names.cpp
@@ -1148,6 +1148,11 @@ public:
   template<typename A> bool Pre(const A &) { return true; }
   template<typename A> void Post(const A &) {}
 
+  bool Pre(const parser::SpecificationPart &x) {
+    Walk(std::get<std::list<parser::OpenMPDeclarativeConstruct>>(x.t));
+    return false;
+  }
+
   bool Pre(const parser::OpenMPBlockConstruct &);
   void Post(const parser::OpenMPBlockConstruct &) { PopContext(); }
   void Post(const parser::OmpBeginBlockDirective &) {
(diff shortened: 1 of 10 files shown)


lib/semantics/resolve-names.cpp
```diff
@@ -1226,6 +1232,18 @@ class OmpAttributeVisitor {
     auto it{GetContext().objectWithDSA.find(&symbol)};
     return it != GetContext().objectWithDSA.end();
   }
+  void SetContextAssociatedLoopLevel(std::size_t level) {
+    GetContext().associatedLoopLevel = level;
+  }
+  std::size_t GetAssociatedLoopLevelFromClauses(const parser::OmpClauseList &);
+
+  template<typename T>
+  std::optional<std::int64_t> GetClauseIntValue(const T &x) {
```
sscalpone: Is there a similar routine somewhere else that does this?

ichoyjx: I searched around and didn't find one.

lib/semantics/resolve-names.cpp
```diff
@@ -5954,20 +5987,135 @@ bool OmpAttributeVisitor::Pre(const parser::OpenMPLoopConstruct &x) {
...
+    auto &loopControl{loop->GetLoopControl().value()};
+    const Bounds &bounds{std::get<Bounds>(loopControl.u)};
+    const parser::Name &iv{bounds.name.thing};
+    auto *symbol{ResolveOmp(iv, ivDSA)};
+    iv.symbol = symbol; // adjust the symbol within region
+    AddToContextObjectWithDSA(*symbol, ivDSA);
+
+    const auto &block{std::get<parser::Block>(loop->t)};
+    const auto it{block.begin()};
+    loop = GetDoConstructIf(*it);
```
jeanPerier: Can't the block be empty ? It may deserve a `CHECK(it != block.end())` if you are sure it can't.

lib/semantics/resolve-names.cpp
```diff
@@ -5954,20 +5987,135 @@ bool OmpAttributeVisitor::Pre(const parser::OpenMPLoopConstruct &x) {
...
+    const parser::OmpClauseList &x) {
+  std::size_t orderedLevel{0};
+  std::size_t collapseLevel{0};
+  for (const auto &clause : x.v) {
+    if (const auto *orderedClause{
+            std::get_if<parser::OmpClause::Ordered>(&clause.u)}) {
+      if (const auto v{
+              GetClauseIntValue<parser::OmpClause::Ordered>(*orderedClause)}) {
+        orderedLevel = *v;
+      }
```
jeanPerier: Is it normal to not be able to compute the ordered clause expression at compile time when `n` was specified ? OpenMP standard 4.5 says "The parameter of the ordered clause must be a constant" (2.7.1 page 59).
Same question can be asked for collapse value.
(Maybe that is fine and already reported by the expression analysis, I did not check)

lib/semantics/resolve-names.cpp
```diff
@@ -5954,20 +5987,135 @@ bool OmpAttributeVisitor::Pre(const parser::OpenMPLoopConstruct &x) {
...
+            std::get_if<parser::OmpClause::Collapse>(&clause.u)}) {
+      if (const auto v{GetClauseIntValue<parser::OmpClause::Collapse>(
+              *collapseClause)}) {
+        collapseLevel = *v;
+      }
+    }
+  }
+
+  if (orderedLevel &&
+      (!collapseLevel || (collapseLevel && orderedLevel >= collapseLevel))) {
```
jeanPerier: The `collapseLevel` in `(collapseLevel && orderedLevel >= collapseLevel)` is redundant.

lib/semantics/resolve-names.cpp
```diff
@@ -5954,20 +5987,135 @@ bool OmpAttributeVisitor::Pre(const parser::OpenMPLoopConstruct &x) {
...
+//   - The loop iteration variable(s) in the associated do-loop(s) of a do,
+//     parallel do, taskloop, or distribute construct is (are) private.
+//   - The loop iteration variable in the associated do-loop of a simd construct
+//     with just one associated do-loop is linear with a linear-step that is the
+//     increment of the associated do-loop.
+//   - The loop iteration variables in the associated do-loops of a simd
+//     construct with multiple associated do-loops are lastprivate.
+//
+// This assumes that the do-loops association for collapse/ordered clause has
+// been performed (the number of nested do-loops >= n).
```
jeanPerier: Is this check a TODO or just a note ? I would made the TODO explicit if it is one and it's not tracked somewhere else.

lib/semantics/resolve-names.cpp
```diff
@@ -5954,20 +5987,135 @@ bool OmpAttributeVisitor::Pre(const parser::OpenMPLoopConstruct &x) {
...
+    // go through all the nested do-loops and resolve index variables
+    auto &loopControl{loop->GetLoopControl().value()};
+    const Bounds &bounds{std::get<Bounds>(loopControl.u)};
+    const parser::Name &iv{bounds.name.thing};
+    auto *symbol{ResolveOmp(iv, ivDSA)};
+    iv.symbol = symbol; // adjust the symbol within region
+    AddToContextObjectWithDSA(*symbol, ivDSA);
+
+    const auto &block{std::get<parser::Block>(loop->t)};
+    const auto it{block.begin()};
```
jeanPerier: Is it possible to have OMP directives on the inner loops that would break the assumption that the next inner do loop is the first element of the block list of statements ?

lib/semantics/resolve-names.cpp
```diff
@@ -5954,20 +5987,135 @@ bool OmpAttributeVisitor::Pre(const parser::OpenMPLoopConstruct &x) {
...
+    loop = GetDoConstructIf(*it);
+  }
```
jeanPerier: Should you add a `CHECK(!level)` after your loop to enforce the assumption that loop nesting checks have already been done ?

lib/semantics/resolve-names.cpp
```diff
@@ -5954,20 +5987,135 @@ bool OmpAttributeVisitor::Pre(const parser::OpenMPLoopConstruct &x) {
...
+
+  auto &outer{std::get<std::optional<parser::DoConstruct>>(x.t)};
+  for (const parser::DoConstruct *loop{&*outer}; loop && level > 0; --level) {
+    // go through all the nested do-loops and resolve index variables
+    auto &loopControl{loop->GetLoopControl().value()};
+    const Bounds &bounds{std::get<Bounds>(loopControl.u)};
+    const parser::Name &iv{bounds.name.thing};
+    auto *symbol{ResolveOmp(iv, ivDSA)};
+    iv.symbol = symbol; // adjust the symbol within region
+    AddToContextObjectWithDSA(*symbol, ivDSA);
```
jeanPerier: Depending on whether this is an error recovery or not, consider adding a `CHECK(symbol)` or an `if(symbol) {..}`.

ichoyjx: Nice catch, I forgot it again... I will add a check.

lib/semantics/resolve-names.cpp
klausler: Why not `ToInt64(EvaluateIntExpr(x))`?

lib/semantics/resolve-names.cpp
ichoyjx: It's a TODO & note I guess, it's currently in `OmpChecker` but part of the `OmpChecker` needs to be moved before NR. I will add a `TODO` in the comments. thx.

klausler: What is "NR"?

ichoyjx: > What is "NR"?

Name Resolution.

lib/semantics/resolve-names.cpp
ichoyjx: Yes, it works too. I can remove this template function then.

lib/semantics/resolve-names.cpp
ichoyjx: They're both defined as `ScalarIntConstantExpr` and checked in Expr Analysis:
```
do.f90:4:30: error: Must be a constant value
    !$omp parallel do collapse(N)
                               ^
```

lib/semantics/resolve-names.cpp
ichoyjx: Yes. The Spec requires that all the associated do-loops should be perfectly nested (no code or OpenMP directive in between). Ideally, the OpenMP structural checker will be able to capture this kind of error prior to Name Resolution.

lib/semantics/resolve-names.cpp
klausler: ... or `CHECK(level == 0)`, which seems a little more clear to me in this context

lib/semantics/resolve-names.cpp
```diff
@@ -5954,20 +5979,136 @@ bool OmpAttributeVisitor::Pre(const parser::OpenMPLoopConstruct &x) {
...
+      if (const auto v{evaluate::ToInt64(
+              resolver_.EvaluateIntExpr(collapseClause->v))}) {
+        collapseLevel = *v;
+      }
+    }
+  }
+
+  if (orderedLevel && (!collapseLevel || orderedLevel >= collapseLevel)) {
+    return orderedLevel;
+  } else if (!orderedLevel && collapseLevel) {
```
jeanPerier: Sorry for not raising this yesterday, I think you could add a comment here that the case `orderedLevel < collapseLevel` is an error handled in omp structural checks.

ichoyjx: okay, I will.

PR #972, opened 2020-02-06
Fix issues comming from clang-10 warnings

jeanPerier:
Fix 3 warning from clang-10:

> error: unused variable 'gap' [-Werror,-Wunused-variable]

@klausler, I could not find what `gap` was intended for, so I've deleted it, but I am suspecting it's wrong and you had it there for a reason.

>error: definition of implicit copy assignment operator for 'SubprogramDetails' is deprecated because it has a user-declared copy constructor [-Werror, -Wdeprecated-copy]

@tskeith , it's a complaint about the rule of three/five not being respected. I don't get why `SubprogramDetails(const SubprogramDetails &that)` was made explicit but it is not copying some of the members, so I did not remove it. Is it really normal that some of the fields like `isInterface_` or `stmtFunction_` are not copied ?

>error: loop variable 'pair' of type [...] creates a copy from type [...] [-Werror,-Wrange-loop-construct]

The `auto` instead of `auto&` was creating copies in `for (const auto pair : symbols)` and such. That seems a legit complaint to me so I have added a `&`.

### DIFF
--- a/include/flang/semantics/symbol.h
+++ b/include/flang/semantics/symbol.h
@@ -52,10 +52,6 @@ private:
 
 class SubprogramDetails {
 public:
-  SubprogramDetails() {}
-  SubprogramDetails(const SubprogramDetails &that)
-    : dummyArgs_{that.dummyArgs_}, result_{that.result_} {}
-
   bool isFunction() const { return result_ != nullptr; }
   bool isInterface() const { return isInterface_; }
   void set_isInterface(bool value = true) { isInterface_ = value; }
--- a/lib/semantics/mod-file.cpp
+++ b/lib/semantics/mod-file.cpp
@@ -915,10 +915,10 @@ void SubprogramSymbolCollector::DoType(const DeclTypeSpec *type) {
       if (const DerivedTypeSpec * extends{typeSymbol.GetParentTypeSpec()}) {
         DoSymbol(extends->name(), extends->typeSymbol());
       }
-      for (const auto pair : derived->parameters()) {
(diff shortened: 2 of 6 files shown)


include/flang/semantics/symbol.h
```diff
@@ -55,6 +55,10 @@ class SubprogramDetails {
   SubprogramDetails() {}
   SubprogramDetails(const SubprogramDetails &that)
     : dummyArgs_{that.dummyArgs_}, result_{that.result_} {}
+  SubprogramDetails &operator=(const SubprogramDetails &) = delete;
+  ~SubprogramDetails() = default;
```
klausler: Was this explicitly defaulted destructor necessary?

tskeith:
I don't know of any reason for the explicit constructors for SubprogramDetails. I think you should just delete them.

include/flang/semantics/symbol.h
jeanPerier: It was not, thanks. I have removed all added constructor/destructor and the copy constructor after Tim's comment.

jeanPerier:
> I don't know of any reason for the explicit constructors for SubprogramDetails. I think you should just delete them.

Thanks for this feedback, looking at the commit that introduced it, I also cannot see a reason to have this ctor (Today, it is probably wrong as new members were added since that commit but the copy ctor not updated):
https://github.com/flang-compiler/f18/blob/44e4fb4b6cc107fad299e80046673804c15a2ac2/lib/semantics/symbol.h#L34

I have removed all the explicit constructors.

PR #974, opened 2020-02-07
Fix for typo in f18/test-lit/CMakeLists.txt 

Sameeranjoshi:
I think the flag(FLANG_TEST_DEPENDS) may be getting neglected and not checking for dependencies.

### DIFF
--- a/test-lit/CMakeLists.txt
+++ b/test-lit/CMakeLists.txt
@@ -22,7 +22,7 @@ set(FLANG_TEST_DEPENDS
 add_lit_testsuite(check-all "Running the Flang regression tests"
   ${CMAKE_CURRENT_BINARY_DIR}
   PARAMS ${FLANG_TEST_PARAMS}
-  DEPENDS ${FLANG_TEST_DEPENS}
+  DEPENDS ${FLANG_TEST_DEPENDS}
   )
 set_target_properties(check-all PROPERTIES FOLDER "Tests")
 


Sameeranjoshi:
Is this due to <path to llvm_install_location>/bin doesn't have llvm-lit, is the -DLLVM_INSTALL_UTILS=ON set in while building llvm?

tskeith:
With this change I get the following error from cmake:
```
CMake Error at /Users/tsk/llvm/f18-llvm/build/Release-clang-9/lib/cmake/llvm/AddLLVM.cmake:1546 (add_dependencies):
  The dependency target "flang" of target "check-all" does not exist.
Call Stack (most recent call first):
  /Users/tsk/llvm/f18-llvm/build/Release-clang-9/lib/cmake/llvm/AddLLVM.cmake:1567 (add_lit_target)
  test-lit/CMakeLists.txt:22 (add_lit_testsuite)
```
And a similar one for `llvm-lit`.

Does this require a change to the build instructions?

CarolineConcatto:
Very odd, I also have the message, but make check-all works just fine:

CarolineConcatto:
I haven't noticed this problem before, but I have it also. 
The reason I did not see it was becasue make check-all works.

Sameeranjoshi:
Thanks @CarolineConcatto @DavidTruby 
Closing issue as https://github.com/flang-compiler/f18/pull/1033 addresses the issue.

PR #976, opened 2020-02-07
[OpenMP] Predetermined rule for sequential loop index

ichoyjx:
This commit implements rule:
A loop iteration variable for a sequential loop in a parallel or
task generating construct is private in the innermost such construct
that encloses the loop.

A Simple example:
```
  i = -1                    <== Scope 0
  j = -1
  !$omp parallel            <== Scope 1
  print *,i,j      <-- both are shared (Scope 0)
  !$omp parallel            <== Scope 2
  print *,i,j      <-- a) i is shared (Scope 0), j is private (Scope 2)
  !$omp do                  <== Scope 3
  do i=1, 10       <-- i is private (Scope 3)
     do j=1, 10    <-- b) j is private (Scope 2, not 3!)
     enddo
  enddo
  print *,i,j      <-- c) i is shared (Scope 0), j is private (Scope 2)
  !$omp end parallel
  print *,i,j      <-- both are shared (Scope 0)
  !$omp end parallel
  print *,i,j      <-- both are shared (Scope 0)
end
```

The basic idea is when visiting the `DoConstruct` node within an OpenMP
construct, if the do-loop is not associated (like `i` loop is associated
with `!$omp do`) AND the do-loop is in the parallel/task generating
construct, resolve the loop index to be private to that innermost construct.

In the above example, `j` loop is not associated (then it is sequential) and
the innermost parallel/task generating construct that encloses the `j` loop
is the `parallel` construct marked with `<== Scope 2`, so `j` is private
to that construct. To do that, I also need to change the prototype of those
`ResolveOmp*` functions to allow specifiying the `scope` because the new
symbol for `j` should be created in Scope 2 and all the `symbol` field of
`Name j` in that `parallel` construct should be fixed, such as c).

### DIFF
--- a/lib/semantics/check-omp-structure.h
+++ b/lib/semantics/check-omp-structure.h
@@ -88,6 +88,8 @@ static constexpr OmpDirectiveSet simdSet{
     OmpDirective::TASKLOOP_SIMD,
     OmpDirective::TEAMS_DISTRIBUTE_PARALLEL_DO_SIMD,
     OmpDirective::TEAMS_DISTRIBUTE_SIMD};
+static constexpr OmpDirectiveSet taskGeneratingSet{
+    OmpDirectiveSet{OmpDirective::TASK} | taskloopSet};
 
 class OmpStructureChecker : public virtual BaseChecker {
 public:
--- a/lib/semantics/resolve-names.cpp
+++ b/lib/semantics/resolve-names.cpp
@@ -1178,19 +1178,19 @@ public:
   // 2.15.3 Data-Sharing Attribute Clauses
   void Post(const parser::OmpDefaultClause &);
   bool Pre(const parser::OmpClause::Shared &x) {
-    ResolveOmpObjectList(x.v, Symbol::Flag::OmpShared, currScope());
+    ResolveOmpObjectList(x.v, Symbol::Flag::OmpShared);
     return false;
   }
   bool Pre(const parser::OmpClause::Private &x) {
-    ResolveOmpObjectList(x.v, Symbol::Flag::OmpPrivate, currScope());
+    ResolveOmpObjectList(x.v, Symbol::Flag::OmpPrivate);
     return false;
   }
   bool Pre(const parser::OmpClause::Firstprivate &x) {
-    ResolveOmpObjectList(x.v, Symbol::Flag::OmpFirstPrivate, currScope());
+    ResolveOmpObjectList(x.v, Symbol::Flag::OmpFirstPrivate);
     return false;
   }
   bool Pre(const parser::OmpClause::Lastprivate &x) {
-    ResolveOmpObjectList(x.v, Symbol::Flag::OmpLastPrivate, currScope());
+    ResolveOmpObjectList(x.v, Symbol::Flag::OmpLastPrivate);
     return false;
   }
 
@@ -1288,9 +1288,8 @@ private:
   const parser::Name &GetLoopIndex(const parser::DoConstruct &);
   void ResolveSeqLoopIndexInParallelOrTaskConstruct(const parser::Name &);
 
-  void ResolveOmpObjectList(
-      const parser::OmpObjectList &, Symbol::Flag, Scope &scope);
-  void ResolveOmpObject(const parser::OmpObject &, Symbol::Flag, Scope &scope);
+  void ResolveOmpObjectList(const parser::OmpObjectList &, Symbol::Flag);
+  void ResolveOmpObject(const parser::OmpObject &, Symbol::Flag);
   Symbol *ResolveOmp(const parser::Name &, Symbol::Flag, Scope &);
   Symbol *ResolveOmp(Symbol &, Symbol::Flag, Scope &);
   Symbol *ResolveOmpCommonBlockName(const parser::Name *);
@@ -1392,6 +1391,7 @@ private:
   void FinishSpecificationParts(const ProgramTree &);
   void FinishDerivedTypeInstantiation(Scope &);
   void ResolveExecutionParts(const ProgramTree &);
+  void ResolveOmpParts(const parser::ProgramUnit &);
 };
 
 // ImplicitRules implementation
@@ -5782,7 +5782,7 @@ bool ResolveNamesVisitor::Pre(const parser::ProgramUnit &x) {
   ResolveSpecificationParts(root);
   FinishSpecificationParts(root);
   ResolveExecutionParts(root);
-  OmpAttributeVisitor{context(), *this}.Walk(x);
+  ResolveOmpParts(x);
   return false;
 }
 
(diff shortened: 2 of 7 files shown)


ichoyjx:
> Everything builds, tests, and looks good to me. I hesitate to approve since I don't really understand the changes.

Thanks Pete!!

lib/semantics/resolve-names.cpp
```diff
@@ -6217,19 +6274,20 @@ Symbol *OmpAttributeVisitor::ResolveOmpCommonBlockName(
 }
 
 void OmpAttributeVisitor::ResolveOmpObjectList(
-    const parser::OmpObjectList &ompObjectList, Symbol::Flag ompFlag) {
+    const parser::OmpObjectList &ompObjectList, Symbol::Flag ompFlag,
+    Scope &scope) {
```
tskeith: Why the change to pass in scope when it is always currScope() anyway?

lib/semantics/check-omp-structure.h
```diff
@@ -88,6 +88,8 @@ static constexpr OmpDirectiveSet simdSet{
     OmpDirective::TASKLOOP_SIMD,
     OmpDirective::TEAMS_DISTRIBUTE_PARALLEL_DO_SIMD,
     OmpDirective::TEAMS_DISTRIBUTE_SIMD};
+static constexpr OmpDirectiveSet taskGeneratingSet{
+    OmpDirectiveSet{OmpDirective::TASK} | taskloopSet};
```
tskeith: It looks like taskGeneratingSet and similar sets can be defined in the .cpp file instead of the .h file.

lib/semantics/resolve-names.cpp
```diff
@@ -6051,6 +6063,53 @@ bool OmpAttributeVisitor::Pre(const parser::OpenMPLoopConstruct &x) {
...
+  for (auto it{ompContext_.rbegin()}; it != ompContext_.rend(); ++it) {
+    if (parallelSet.test(it->directive) ||
+        taskGeneratingSet.test(it->directive)) {
+      symbol = ResolveOmp(iv, Symbol::Flag::OmpPrivate, it->scope);
+      targetIt = it;
+      break;
+    }
+  }
+  targetIt++;
+  if (symbol) {
```
tskeith: The increment should be inside the `if` because that is where `targetIt` is used. And `targetIt` doesn't need to be initialized.

lib/semantics/resolve-names.cpp
```diff
@@ -6051,6 +6062,53 @@ bool OmpAttributeVisitor::Pre(const parser::OpenMPLoopConstruct &x) {
...
+      AddToContextObjectWithDSA(*symbol, Symbol::Flag::OmpPrivate, *it);
+    }
+  }
+}
+
+// 2.15.1.1 Data-sharing Attribute Rules - Predetermined
+//   - A loop iteration variable for a sequential loop in a parallel
+//     or task generating construct is private in the innermost such
+//     construct that encloses the loop
+bool OmpAttributeVisitor::Pre(const parser::DoConstruct &x) {
```
jeanPerier: How the "sequential" aspect is determined here ? Should it be tested that the `parser::DoConstruct` belongs to a `parser::OpenMPLoopConstruct` ?

lib/semantics/resolve-names.cpp
```diff
@@ -6051,6 +6062,53 @@ bool OmpAttributeVisitor::Pre(const parser::OpenMPLoopConstruct &x) {
...
+  }
+  if (symbol) {
+    targetIt++;
+    symbol->set(Symbol::Flag::OmpPreDetermined);
+    iv.symbol = symbol;  // adjust the symbol within region
+    for (auto it{ompContext_.rbegin()}; it != targetIt; ++it) {
+      AddToContextObjectWithDSA(*symbol, Symbol::Flag::OmpPrivate, *it);
+    }
+  }
+}
```
jeanPerier: Don't you also need to fix all `parse::Name` symbols that were visited before it was determined their symbols is actually private ?
You may need a second pass here on all the statements inside the inner parallel or task construct. If so, I am not sure though what would be the best/easier between doing so every-time you need to fix a symbol, or make two passes overall, one to determine what symbol needs to be fixed in which scope, and another one later to fix all the relate `parser::Name` symbols.

ichoyjx: Yes, it's a problem & I've been thinking about the solutions since last week. Similar to the comment you made for `omp-symbol08.f90`, this is the issue listed as TODO item a) in this PR description. Two potential solutions:
   * two passes (then I need to make sure that each OpenMP msg is emitted only for once)
   * maintain a list or map (or some other DSes) for all the `Names` processed and go back to fix those

I think that I lean towards the second one.

lib/semantics/resolve-names.cpp
ichoyjx: No. The `sequential` here means that do-loop is within OpenMP construct but not associated with `OpenMPLoopConstruct`. Like:
```
!$omp parallel do
do i = 1, 10   <-- associated
   do j = 1, 10   <-- sequential
    ...
   end do
enddo
```

jeanPerier: How do you make the difference between loops that are associated to `OpenMPLoopConstruct` or not here ?

lib/semantics/resolve-names.cpp
jeanPerier: Sorry, missed the TODO in PR description. The map/list/vector/... of `Name`s may be an interesting option indeed.

lib/semantics/resolve-names.cpp
ichoyjx: Through line 6102 below, if the loop index is already determined as `OmpPredetermined`, then ignore.

ichoyjx:
Note that the latest commit added the second `Walk` of `OmpAttributeVisitor` if there is no error during the first `Walk`, this resolves the issue a).

lib/semantics/resolve-names.cpp
```diff
@@ -6076,6 +6095,53 @@ bool OmpAttributeVisitor::Pre(const parser::OpenMPLoopConstruct &x) {
...
+  std::vector<OmpContext>::reverse_iterator targetIt;
+  for (auto it{ompContext_.rbegin()}; it != ompContext_.rend(); ++it) {
+    if (parallelSet.test(it->directive) ||
+        taskGeneratingSet.test(it->directive)) {
+      symbol = ResolveOmp(iv, Symbol::Flag::OmpPrivate, it->scope);
+      targetIt = it;
+      break;
+    }
+  }
+  if (symbol) {
```
tskeith: I think this is a better way to write this:
```
  auto targetIt{ompContext_.rbegin()};
  for (;; ++targetIt) {
    if (targetIt == ompContext_.rend()) {
      return;
    }
    if (parallelSet.test(targetIt->directive) ||
        taskGeneratingSet.test(targetIt->directive)) {
      break;
    }
  }
  if (Symbol *symbol{ResolveOmp(iv, Symbol::Flag::OmpPrivate, targetIt->scope)}) {
```

lib/semantics/resolve-names.cpp
```diff
@@ -5771,6 +5782,14 @@ bool ResolveNamesVisitor::Pre(const parser::ProgramUnit &x) {
...
   ResolveExecutionParts(root);
   OmpAttributeVisitor{context(), *this}.Walk(x);
+  if (!context().AnyFatalError()) {
+    // The data-sharing attribute of the loop iteration variable for a
+    // sequential loop (2.15.1.1) can only be determined when visiting
+    // the corresponding DoConstruct, a second walk is to adjust the
+    // symbols for all the data-refs of that loop iteration variable
+    // prior to the DoConstruct.
+    OmpAttributeVisitor{context(), *this}.Walk(x);
+  }
```
tskeith: Can you extract this into a separate function so that this function shows the high-level order of processing a program unit?

lib/semantics/resolve-names.cpp
ichoyjx: thank you for the detailed suggestion! I've made the changes.

jeanPerier:
LGTM

PR #978, opened 2020-02-10
Semantic checks for constraints on types

psteinfeld:
I implemented and added tests for constraints C703, C704, C705, C706,
and C796.  In some cases, the code and/or test already existed, and all
I did was add a notation indicating the associated constraint.

### DIFF
--- a/lib/semantics/check-declarations.cpp
+++ b/lib/semantics/check-declarations.cpp
@@ -495,7 +495,7 @@ void CheckHelper::CheckDerivedType(
   }
   if (const DeclTypeSpec * parent{FindParentTypeSpec(symbol)}) {
     const DerivedTypeSpec *parentDerived{parent->AsDerived()};
-    if (!IsExtensibleType(parentDerived)) {
+    if (!IsExtensibleType(parentDerived)) {  // C705
       messages_.Say("The parent type is not extensible"_err_en_US);
     }
     if (!symbol.attrs().test(Attr::ABSTRACT) && parentDerived &&
--- a/lib/semantics/resolve-names.cpp
+++ b/lib/semantics/resolve-names.cpp
@@ -343,6 +343,7 @@ protected:
   }
   KindExpr GetKindParamExpr(
       TypeCategory, const std::optional<parser::KindSelector> &);
+  void CheckForAbstractType(const Symbol &typeSymbol);
 
 private:
   State state_;
@@ -731,7 +732,9 @@ public:
   void Post(const parser::LengthSelector &);
   bool Pre(const parser::KindParam &);
   bool Pre(const parser::DeclarationTypeSpec::Type &);
+  void Post(const parser::DeclarationTypeSpec::Type &);
   bool Pre(const parser::DeclarationTypeSpec::Class &);
+  void Post(const parser::DeclarationTypeSpec::Class &);
   bool Pre(const parser::DeclarationTypeSpec::Record &);
   void Post(const parser::DerivedTypeSpec &);
   bool Pre(const parser::DerivedTypeDef &);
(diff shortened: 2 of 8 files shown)


test/semantics/resolve70.f90
```diff
@@ -0,0 +1,58 @@
...
+  type :: extensible
+  end type
+  type, bind(c) :: inextensible
+  end type
+
+  ! This one's OK
+  class(extensible) :: y
+
+  !ERROR: A declaration that uses the CLASS keyword must specify an extensible derived type.  'inextensible' is not extensible
+  class(inextensible) :: x
```
tskeith: How about:
`Non-extensible derived type 'inextensible' may not be used with CLASS keyword

lib/semantics/check-declarations.cpp
```diff
@@ -495,7 +495,7 @@ void CheckHelper::CheckDerivedType(
   }
   if (const DeclTypeSpec * parent{FindParentTypeSpec(symbol)}) {
     const DerivedTypeSpec *parentDerived{parent->AsDerived()};
-    if (!IsExtensibleType(parentDerived)) {
+    if (!IsExtensibleType(parentDerived)) {  // c705
```
tskeith: c705 -> C705 for consistency.

lib/semantics/resolve-names.cpp
```diff
@@ -3287,11 +3294,29 @@ bool DeclarationVisitor::Pre(const parser::DeclarationTypeSpec::Type &) {
...
+
 bool DeclarationVisitor::Pre(const parser::DeclarationTypeSpec::Class &) {
   SetDeclTypeSpecCategory(DeclTypeSpec::Category::ClassDerived);
   return true;
 }
 
+void DeclarationVisitor::Post(
+    const parser::DeclarationTypeSpec::Class &parsedClass) {
+  const auto &typeName{std::get<parser::Name>(parsedClass.derived.t)};
+  const std::optional<DerivedTypeSpec> spec{ResolveDerivedType(typeName)};
```
klausler: Consider structuring this code with the declaration of `spec` (which need not be `const` and could be `auto') as the predicate of an `if` statement.

lib/semantics/resolve-names.cpp
```diff
@@ -1612,6 +1613,12 @@ void DeclTypeSpecVisitor::Post(
 void DeclTypeSpecVisitor::MakeNumericType(TypeCategory category, int kind) {
   SetDeclTypeSpec(context().MakeNumericType(category, kind));
 }
+ 
+void DeclTypeSpecVisitor::CheckForAbstractType(const Symbol &typeSymbol) {
```
klausler: If this function is only called once, I don't think it provides much value over its inline textual expansion.

test/semantics/resolve70.f90
psteinfeld: Good suggestion.  I'll change it.

lib/semantics/check-declarations.cpp
psteinfeld: My bad.  I'll change it.

lib/semantics/resolve-names.cpp
psteinfeld: Thanks, Peter.  Will do.

lib/semantics/resolve-names.cpp
psteinfeld: Agreed.  It's called twice, once in the `Post` function for `parser::TypeSpec` and again in the `Post` function for `parser::DeclarationTypeSpec`.  It just felt wrong to copy the code, so I created a separate function.

Issue #963, opened 2020-02-04
[LLVMify F18] Compiler module folders should have capitalised names


DavidTruby:
Paths such as `lib/parser`, `lib/semantics` etc in clang and llvm have capitalized folder names (`lib/Parser` and so on). 
We should match this for consistency with the rest of the LLVM project.

CarolineConcatto:
Does anyone know why git mv and git add in the same file is seeing by git as a file deleted in the old place and a new one created in the new place?

kiranchandramohan:
Was this example for illustration or did you see the issue on this file (Fortran-features.cpp)?
Are you renaming individual files or the directory as a whole? 
For me, git mv lib/common lib/Common worked and it showed all files in the directory as renames.

I don't know why this happens. But i have seen the same issue in one of Luke's patches.

CarolineConcatto:
I see on Fortran-Features, but also in others as well. I did git move for the entire folder, and after that git add. The problem happens when I do git add in the file. If it was changed git sees as a new file for some reason.

kiranchandramohan:
OK. 

Are you doing a 'git add' because you are modifying the contents of the file? 

You might have already seen this, online help seems to suggest the following,
1) git stores renames as addition and deletion internally. It has some function to check whether a rename has happened and it does that by comparing the old file and the new file. If the changes are below a threshold it will detect the rename operation otherwise it will not.
2) Do the rename and modify (add) in separate commits to overcome this issue.

CarolineConcatto:
Yes, git add after modifying. A little bit annoying that I need to do in 2 commits, but let do it then in order to not mess with git history of everyone.
Thanks, Kiran.

kiranchandramohan:
Check with @peterwaller-arm whether this issue will re-appear when the commits are squashed before committing or when using peter's script for submission.

CarolineConcatto:
I have done PR for this task:
https://github.com/flang-compiler/f18/pull/980

CarolineConcatto:
Question: should the namespaces also need to be changed?
Like:
namespace Fortran::common {
namespace Fortran::decimal {

DavidTruby:
> 
> 
> Question: should the namespaces also need to be changed?
> Like:
> namespace Fortran::common {
> namespace Fortran::decimal {

In clang and llvm the namespace names are lower case, so we should follow that

PR #980, opened 2020-02-11
[LLVMify F18] Compiler module folders should have capitalised names

CarolineConcatto:
This patch renames the modules in f18 to use a capital letter in the
module name[1]
Firstly we create new folders with the capitalised name of the modules
and then we do git mv for these new folders. Besides that, many files were
modified to point and include the names of these modules.

The task is very easy, the only problem is that it will mess with git
history thinking that new files were created, and lots of them were deleted.
So to reduce to chaos, this task will be split into two commits.
The first one creates and moves the files inside f18. The second commit
will update the includes in the files that need to be updated.

This first commit renames the folder inside:
->include/flang/[common | decimal | evaluate | parser | semantics]
->lib/[common | decimal | evaluate | parser | semantics]
->test/[decimal | evaluate | preprocessing | runtime | semantics]
->test-lit/driver
to capitalise names

This is achieved by doing git mv for each module, as shown below:
   $for i in lib/common/*; do git mv $i lib/Common; done

[1]https://github.com/flang-compiler/f18/issues/963


CarolineConcatto:
Question: should the namespaces also need to be changed?
Like:
namespace Fortran::common {
namespace Fortran::decimal {

CarolineConcatto:
This branch needs to rebase with origin/master to remove the conflict in documentation/ImplementingASemanticCheck.md. I did try that but when I was trying to do git push it was asking to use force($git push myfork HEAD --force)

CarolineConcatto:
Ok, after talking with @DavidTruby I think it was safe to do a git pull --rebase and a git push force.
The push has only the rebase with master and no changes in the code.

documentation/ImplementingASemanticCheck.md
```diff
@@ -216,12 +216,12 @@ find all of the `evaluate::ActualArgument` nodes.
 
 Note that the compiler has multiple types called `Expr`.  One is in the
 `parser` namespace.  `parser::Expr` is defined in the file
-`include/flang/parser/parse-tree.h`.  It represents a parsed expression that
+`include/flang/parser/Parse-tree.h`.  It represents a parsed expression that
```
RichBarton-Arm: That looks wrong to me.

lib/Evaluate/characteristics.cpp
```diff
@@ -1,21 +1,21 @@
...
-#include "flang/evaluate/check-expression.h"
-#include "flang/evaluate/fold.h"
-#include "flang/evaluate/intrinsics.h"
-#include "flang/evaluate/tools.h"
-#include "flang/evaluate/type.h"
-#include "flang/parser/message.h"
-#include "flang/semantics/scope.h"
-#include "flang/semantics/symbol.h"
+#include "flang/Evaluate/characteristics.h"
+#include "flang/Common/indirection.h"
```
RichBarton-Arm: These two are the wrong way around. Could fix at the same time?

CarolineConcatto:
There is some problem with the merge.  
Mainly with the file test/semantics/resolve70.f90.

lib/Evaluate/characteristics.cpp
CarolineConcatto: This behaviour is really odd. 
The problem I believe could come from the git-rebase. 
Because this does not make any sense to replace totally different names.

CarolineConcatto: @RichBarton-Arm  which files?
#include "flang/evaluate/characteristics.h" .    #include "flang/Evaluate/characteristics.h"$
#include "flang/common/indirection.h" .           #include "flang/Common/indirection.h"
#include "flang/evaluate/check-expression.h" . #include "flang/Evaluate/check-expression.h"
#include "flang/evaluate/fold.h" .                       #include "flang/Evaluate/fold.h"
#include "flang/evaluate/intrinsics.h"                   #include "flang/Evaluate/intrinsics.h"
#include "flang/evaluate/tools.h"                      #include "flang/Evaluate/tools.h"
#include "flang/evaluate/type.h"                    #include "flang/Evaluate/type.h"
#include "flang/parser/message.h"                #include "flang/Parser/message.h"
#include "flang/semantics/scope.h"               #include "flang/Semantics/scope.h"
#include "flang/semantics/symbol.h"              #include "flang/Semantics/symbol.h"

CarolineConcatto:
@RichBarton-Arm  I had this conversation with @peterwaller-arm before deciding what to do.
I am going to paste AS IS here our conversation. Hope @peterwaller-arm doesn't mind
Hey Peter,
I have a problem with git history, maybe you could help.
This ticket here is to Capitalise the modules if f18.(https://github.com/flang-compiler/f18/issues/963)
I do git mv and after that git add in the files that needed to be updated.
git add in the modified files in the old folder shows these files as deleted and the files in the new folder as new. This only happens to the files that were modified.
The suggestion online is to do two commits one with only git mv and a second commit git add in the modified files. 
Question is: I will probably be asked to squash the 2 commits. 
Will the problem described now be shown again if I do git squash? 

Peters answer:
No, squashing will do the right thing. I recommend doing the changes as two commits initially - that's a good suggestion from whoever said it.
I'm not sure exactly what you did from your description. It sounds like you're moving files and modifying them.

DavidTruby:
> Used the patented 'eye tool' on this one and only spotted a couple of things.
> 
> I don't think splitting this into two in the way that you have done is right because I think the first commit won't build without the second one. I get wanting to split this up but that shouldn't be at the expense of making each commit leave the repo in a buildable state. We should squash the commits before we merge the PR.

The reason for splitting this up isn't that we want to, but git just simply doesn't understand that a move has happened if you do both in one commit, and there doesn't seem to be a way to make it understand that other than splitting the commits. So we have a choice between a rogue commit that doesn't build or a broken history because git sees this as every file in the repo being deleted and new files being added. I personally think the latter is worse behaviour than the former!

lib/Evaluate/characteristics.cpp
RichBarton-Arm: Sorry - I mean that the rest of the includes seem to be in alphabetical order except for "flang/Common/indirection.h", which looks like it should be above "flang/Evaluate/characteristics.h". Clearly a bug in the original code though, so happy for this patch not to address it.

documentation/ImplementingASemanticCheck.md
CarolineConcatto: The only time I did not use sed, I got it wrong. Thanks.

lib/Evaluate/characteristics.cpp
DavidTruby: This should be caught by clang-format but in some places clang-format hasn't fully been run on every file. This should be fixed in the clang-format alignment patch at #945

CarolineConcatto:
[**Out-of-topic]**I noticed that there is no header text inside all .cpp and .h files inside f18/test.
Is this correct?

documentation/ImplementingASemanticCheck.md
CarolineConcatto: Done

lib/Evaluate/characteristics.cpp
tskeith: characteristics.h is first because this is characteristics.cpp, so the order is right.

DavidTruby: Ah I didn't spot that. Yes having actually looked at it rather than just reading the comments the order looks correct to me too.

RichBarton-Arm:
> 
> 
> [**Out-of-topic]**I noticed that there is no header text inside all .cpp and .h files inside f18/test.
> Is this correct?

I think that follows the pattern in other LLVM projects too. Sometimes the tests are very small files, so they would mostly be header. Don't think that needs to be part of this patch anyhow.

RichBarton-Arm:
@sscalpone we think this is ready to land. We need to squash before merging. Does anyone else want to review this before we squash?

sscalpone:
@CarolineConcatto @RichBarton-Arm Looks good.  Please resolve conflicts & squash.

sscalpone:
@DavidTruby There are still conflict to resolve.

include/flang/Semantics/symbol.h
```diff
@@ -10,7 +10,6 @@
 #define FORTRAN_SEMANTICS_SYMBOL_H_
 
 #include "type.h"
-#include "flang/common/Fortran.h"
```
CarolineConcatto: So odd that this is not updated in the branch. Fixing that

runtime/environment.h
```diff
@@ -9,7 +9,8 @@
 #ifndef FORTRAN_RUNTIME_ENVIRONMENT_H_
 #define FORTRAN_RUNTIME_ENVIRONMENT_H_
 
-#include "flang/decimal/decimal.h"
+#include "flang/Common/Fortran.h"
```
CarolineConcatto: This line does not exist anymore in master

runtime/numeric-output.h
```diff
@@ -20,7 +20,8 @@
 
 #include "format.h"
 #include "io-stmt.h"
-#include "flang/decimal/decimal.h"
+#include "flang/Common/unsigned-const-division.h"
```
CarolineConcatto: another line that that needs to be removed

include/flang/Semantics/symbol.h
CarolineConcatto: Done!

tools/f18/f18.cpp
```diff
@@ -8,22 +8,22 @@
...
-#include "flang/parser/parse-tree-visitor.h"
-#include "flang/parser/parse-tree.h"
-#include "flang/parser/parsing.h"
-#include "flang/parser/provenance.h"
-#include "flang/parser/unparse.h"
-#include "flang/semantics/expression.h"
-#include "flang/semantics/semantics.h"
-#include "flang/semantics/unparse-with-symbols.h"
 #include "llvm/Support/raw_ostream.h"
+#include "flang/Common/Fortran-features.h"
```
kiranchandramohan: Should the llvm file above come at the end of the listing?

CarolineConcatto: Yes, it should. I have no idea why it is at the end.

runtime/environment.h
CarolineConcatto: Done!

runtime/numeric-output.h
CarolineConcatto: Done!

tools/f18/f18.cpp
CarolineConcatto: Done!

PR #981, opened 2020-02-11
Semantic check for C708

psteinfeld:
An entity declared with the CLASS keyword shall be a dummy argument or have the ALLOCATABLE or POINTER attribute.

### DIFF
--- a/lib/semantics/check-declarations.cpp
+++ b/lib/semantics/check-declarations.cpp
@@ -371,6 +371,14 @@ void CheckHelper::CheckObjectEntity(
       }
     }
   }
+  if (const DeclTypeSpec * type{details.type()}) {  // C708
+    if (type->IsPolymorphic() &&
+        !(IsAllocatableOrPointer(symbol) || symbol.IsDummy())) {
+      messages_.Say("CLASS entity '%s' must be a dummy argument or have "
+                    "ALLOCATABLE or POINTER attribute"_err_en_US,
+          symbol.name());
+    }
+  }
 }
 
 // The six different kinds of array-specs:
--- a/lib/semantics/resolve-names.cpp
+++ b/lib/semantics/resolve-names.cpp
@@ -4314,12 +4314,8 @@ Symbol *DeclarationVisitor::DeclareLocalEntity(const parser::Name &name) {
   if (!PassesLocalityChecks(name, prev)) {
     return nullptr;
   }
-  name.symbol = nullptr;
-  Symbol &symbol{DeclareEntity<ObjectEntityDetails>(name, {})};
-  if (auto *type{prev.GetType()}) {
-    symbol.SetType(*type);
-    symbol.set(Symbol::Flag::Implicit, prev.test(Symbol::Flag::Implicit));
-  }
+  Symbol &symbol{MakeSymbol(name, HostAssocDetails{prev})};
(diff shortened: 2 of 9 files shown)


psteinfeld:
> 
> 
> Could this constraint be enforced in `check-declarations.cc` rather than in name resolution? If so, that's probably a better place for it.

Quite possibly.  I'll take a look.

psteinfeld:
> 
> 
> > Could this constraint be enforced in `check-declarations.cc` rather than in name resolution? If so, that's probably a better place for it.
> 
> Quite possibly. I'll take a look.

As suggested, I've moved the check to check-declarations.cpp.

@klausler, can you please take another look?

lib/semantics/check-declarations.cpp
```diff
@@ -295,6 +295,10 @@ void CheckHelper::CheckValue(
 
 void CheckHelper::CheckObjectEntity(
     const Symbol &symbol, const ObjectEntityDetails &details) {
+  if (symbol.test(Symbol::Flag::LocalityLocal) ||
```
klausler: What happens when new kinds of compiler-created symbols are invented?

For future-proofing, consider defining a member function or predicate on `Symbol` that determines whether the symbol is compiler-created, copying this Boolean expression into that function, and using it here.

lib/semantics/check-declarations.cpp
```diff
@@ -371,6 +375,14 @@ void CheckHelper::CheckObjectEntity(
       }
     }
   }
+  if (const DeclTypeSpec * type{details.type()}) {  // C708
+    if (type->IsPolymorphic() && !symbol.attrs().test(Attr::POINTER) &&
```
klausler: Consider applying DeMorgan's law for a more intuitive expression: `type->IsPolymorphic() && !(x || y || z)`

lib/semantics/check-declarations.cpp
psteinfeld: Yes.  I had that same thought after I pushed this change.  I'll do it.  But first I'll see if there are other instances where the compiler creates entities.

lib/semantics/check-declarations.cpp
psteinfeld: Thanks, Peter.  Will do.

tskeith: Using `IsAllocatableOrPointer` from tools.h would help to simplify it too.

lib/semantics/check-declarations.cpp
tskeith: It seems to me these should be HostAssocDetails symbols rather than new object symbols. Then we wouldn't be seeing them here at all.

lib/semantics/check-declarations.cpp
psteinfeld: Thanks, Tim.  I'll use it.

lib/semantics/check-declarations.cpp
psteinfeld: Thanks, Tim.  I'll look into doing that.

psteinfeld: I've pushed a new commit where I created host associated symbols for names that appear in LOCAL and LOCAL_INIT locality-specs.

PR #983, opened 2020-02-13
Add zlib to drone files so that linking LLVM works.

DavidTruby:


sscalpone:
@DavidTruby Soon f18 will need to build from a modern copy of llvm, probably from the tip, because of the dependency on MLIR.

PR #985, opened 2020-02-15
Semantic checks for C709, C710, and C711

psteinfeld:
C709 An assumed-type entity shall be a dummy data object that does not
have the ALLOCATABLE, CODIMENSION, INTENT (OUT), POINTER, or VALUE
attribute and is not an explicit-shape array.

C710 An assumed-type variable name shall not appear in a designator or
expression except as an actual argument corresponding to a dummy
argument that is assumed-type, or as the first argument to the intrinsic
function IS_CONTIGUOUS, LBOUND, PRESENT, RANK, SHAPE, SIZE, or UBOUND,
or the function C_LOC from the intrinsic module ISO_C_BINDING.

C711 An assumed-type actual argument that corresponds to an assumed-rank
dummy argument shall be assumed-shape or assumed-rank.

For C709 I added code to check-declarations.cpp.  In the process, I had
to change several function interfaces from `const Symbol` to `Symbol` so
that I could avoid multiple error messages for the same source line.  I
created the test resolve72.f90 to check it.

C710 was already checked, but I added a notation in the source.

For C711 I added code to check-call.cpp and the test call15.f90.

### DIFF
--- a/lib/semantics/check-call.cpp
+++ b/lib/semantics/check-call.cpp
@@ -610,11 +610,23 @@ static void CheckExplicitInterfaceArg(evaluate::ActualArgument &arg,
                 messages.Say(
                     "Actual argument is not a variable or typed expression"_err_en_US);
               }
-            } else if (!object.type.type().IsAssumedType()) {
+            } else {
               const Symbol &assumed{DEREF(arg.GetAssumedTypeDummy())};
-              messages.Say(
-                  "Assumed-type TYPE(*) '%s' may be associated only with an assumed-TYPE(*) %s"_err_en_US,
-                  assumed.name(), dummyName);
+              if (!object.type.type().IsAssumedType()) {
+                messages.Say(
+                    "Assumed-type TYPE(*) '%s' may be associated only with an"
+                    " assumed-TYPE(*) %s"_err_en_US,
+                    assumed.name(), dummyName);
+              } else if (const auto *details{
+                             assumed.detailsIf<ObjectEntityDetails>()}) {
+                if (!(details->IsAssumedShape() || details->IsAssumedRank())) {
+                  messages.Say(  // C711
+                      "Assumed-type TYPE(*) '%s' must be either assumed "
+                      "shape or assumed rank to be associated with TYPE(*) "
+                      "%s"_err_en_US,
+                      assumed.name(), dummyName);
+                }
+              }
             }
           },
           [&](const characteristics::DummyProcedure &proc) {
--- a/lib/semantics/check-declarations.cpp
+++ b/lib/semantics/check-declarations.cpp
@@ -59,6 +59,7 @@ private:
   void CheckObjectEntity(const Symbol &, const ObjectEntityDetails &);
   void CheckArraySpec(const Symbol &, const ArraySpec &);
   void CheckProcEntity(const Symbol &, const ProcEntityDetails &);
+  void CheckAssumedTypeEntity(const Symbol &, const ObjectEntityDetails &);
   void CheckDerivedType(const Symbol &, const DerivedTypeDetails &);
   void CheckGeneric(const Symbol &, const GenericDetails &);
   std::optional<std::vector<Procedure>> Characterize(const SymbolVector &);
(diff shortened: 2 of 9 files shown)


lib/semantics/check-call.cpp
```diff
@@ -587,6 +587,7 @@ static void CheckExplicitInterfaceArg(evaluate::ActualArgument &arg,
     const characteristics::DummyArgument &dummy,
     const characteristics::Procedure &proc, evaluate::FoldingContext &context,
     const Scope *scope) {
+  //  std::cout << "In CheckExplicitInterfaceArg\n";
```
klausler: ?

lib/semantics/check-call.cpp
```diff
@@ -610,11 +611,26 @@ static void CheckExplicitInterfaceArg(evaluate::ActualArgument &arg,
...
-            } else if (!object.type.type().IsAssumedType()) {
+            } else {
               const Symbol &assumed{DEREF(arg.GetAssumedTypeDummy())};
-              messages.Say(
-                  "Assumed-type TYPE(*) '%s' may be associated only with an assumed-TYPE(*) %s"_err_en_US,
-                  assumed.name(), dummyName);
+              if (!object.type.type().IsAssumedType()) {
+                messages.Say(
+                    "Assumed-type TYPE(*) '%s' may be associated only with an"
+                    " assumed-TYPE(*) %s"_err_en_US,
```
klausler: The second part of the message looks garbled.

lib/semantics/check-call.cpp
```diff
@@ -610,11 +611,26 @@ static void CheckExplicitInterfaceArg(evaluate::ActualArgument &arg,
...
+                    assumed.name(), dummyName);
+              } else {
+                if (const auto *details{
```
klausler: Could be merged with previous line.

lib/semantics/check-declarations.cpp
```diff
@@ -38,7 +38,7 @@ class CheckHelper {
   }
   void Check(const ArraySpec &);
   void Check(const DeclTypeSpec &, bool canHaveAssumedTypeParameters);
-  void Check(const Symbol &);
+  void Check(Symbol &);
```
klausler: Is there an unavoidable reason for removing the `const` qualification on these references?

lib/semantics/check-call.cpp
psteinfeld: Oops.  Thanks for catching this.  I'll remove it.

lib/semantics/check-call.cpp
psteinfeld: It does look a little wacky.  But I didn't change the message, I only changed
the string formatting to fit into 80 columns.  Here's a sample of how an actual message appears:
```                           
x.f90:4:3: error: Assumed-type TYPE(*) 'var' may be associated only with an assumed-TYPE(*) dummy argument 'arg='
```
If you think the message should be changed, let me know.

lib/semantics/check-call.cpp
psteinfeld: Thanks.  I'll fix it.

lib/semantics/check-declarations.cpp
psteinfeld: I was able to avoid the multiple messages on the same line by distinguishing between polymorphic types and assumed-type types.  This let me restore the `const`ness of the `Symbol` arguments.  I pushed a commit with these changes.

lib/semantics/check-call.cpp
klausler: Can you please just fix the message?

psteinfeld: Will do.  The obvious way to clarify this message is to add the phrase `actual argument` to the start of the message.  This would make the message read as follows:
```
Assumed-type TYPE(*) actual argument 'var' may be  associated only with an assumed-TYPE(*) dummy argument 'arg='
```
But it also seems wrong to me that the name of the dummy argument as an equal sign appended to the end of it.  This equal sign ends up appearing in many error messages associated with dummy arguments, so I assumed that it must make some kind of sense.  But it doesn't make sense to me.  Do you have any insight on why the message contains the equal sign?

klausler: I give up.  Leave it alone.  I'll take care of it.

lib/semantics/check-declarations.cpp
psteinfeld: OK.  Changes for all other comments have already been pushed.  Let me know if want any other changes.

lib/semantics/check-call.cpp
```diff
@@ -610,11 +610,24 @@ static void CheckExplicitInterfaceArg(evaluate::ActualArgument &arg,
...
-                  "Assumed-type TYPE(*) '%s' may be associated only with an assumed-TYPE(*) %s"_err_en_US,
-                  assumed.name(), dummyName);
+              if (!object.type.type().IsAssumedType()) {
+                messages.Say(
+                    "Assumed-type TYPE(*) '%s' may be associated only with an"
+                    " assumed-TYPE(*) %s"_err_en_US,
+                    assumed.name(), dummyName);
+              } else if (const auto *details{
+                             assumed
+                                 .detailsIf<ObjectEntityDetails>()}) {  // C711
```
tskeith: If you could find a different place for the comment (e.g. on the Say statement), clang-format wouldn't make such a big mess of this.

lib/semantics/check-declarations.cpp
```diff
@@ -293,11 +294,56 @@ void CheckHelper::CheckValue(
   }
 }
 
+void CheckHelper::CheckAssumedTypeEntity(  // C709
+    const Symbol &symbol, const ObjectEntityDetails &details) {
+  if (const DeclTypeSpec * type{symbol.GetType()};
+      type && type->category() == DeclTypeSpec::TypeStar) {
+    if (!symbol.IsDummy()) {
+      messages_.Say("Assumed-type entity '%s' must be a dummy "
+                    "argument"_err_en_US,
```
tskeith: This message fits on one line without splitting it.

psteinfeld: It looks to me like the line is 82 characters long if I merge the strings.  I'll leave it as is.

lib/semantics/check-call.cpp
psteinfeld: Thanks, Tim.  I'll move it.

lib/semantics/check-declarations.cpp
tskeith: If you merge the strings and reformat it fits.

psteinfeld: Ahh...  Thanks!

psteinfeld: There are some other similar strings.  I'll fix them up, too.

PR #987, opened 2020-02-17
Clause checks for combined constructs

DavidTruby:
This includes a refactor of the existing combined construct checks
that were present, as well as adding the remaining combined constructs
that had not been implemented yet.

### DIFF
--- a/include/flang/Common/enum-set.h
+++ b/include/flang/Common/enum-set.h
@@ -111,6 +111,15 @@ public:
     return result;
   }
 
+  constexpr EnumSet operator+(enumerationType v) const {
+    EnumSet result = *this | EnumSet{v};
+    return result;
+  }
+  constexpr EnumSet operator-(enumerationType v) const {
+    EnumSet result = *this ^ EnumSet { v };
+    return result;
+  }
+
   constexpr bool operator==(const EnumSet &that) const {
     return bitset_ == that.bitset_;
   }
--- a/lib/Semantics/check-omp-structure.cpp
+++ b/lib/Semantics/check-omp-structure.cpp
@@ -13,6 +13,52 @@
 
 namespace Fortran::semantics {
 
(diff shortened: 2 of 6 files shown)


test/semantics/omp-clause-validity01.f90
```diff
@@ -445,7 +445,6 @@
   enddo
   !$omp end taskloop simd
 
-  !ERROR: REDUCTION clause is not allowed on the TASKLOOP SIMD directive
   !$omp taskloop simd reduction(+:a)
```
ichoyjx: No reduction clause can be specified. (4.5 Spec P92)

lib/semantics/check-omp-structure.cpp
```diff
@@ -409,6 +556,15 @@ void OmpStructureChecker::Enter(const parser::OpenMPBlockConstruct &x) {
   case parser::OmpBlockDirective::Directive::Master:
     PushContext(beginDir.source, OmpDirective::MASTER);
     break;
+    // 2.11.5 target-parallel-clause -> target-clause |
+    //                                  parallel-clause
+  case parser::OmpBlockDirective::Directive::TargetParallel: {
+    PushContext(beginDir.source, OmpDirective::TARGET_PARALLEL);
+    SetContextAllowed(
+        (targetAllowedClauses | parallelAllowedClauses) - OmpClause::COPYIN);
```
ichoyjx: Could you add a small test for `copyin` clause? Thx.

test/semantics/omp-clause-validity01.f90
DavidTruby: Interesting. The reduction clause actually _can_ be specified in 5.0, but not in 4.5. Which is our "source of truth" here? I assume 4.5, in which case I'll add this limitation back in.

DavidTruby:
@ichoyjx Sorry for letting this rot for so long! I've added some more tests including the one you requested, and I've added back the restriction on reduction clauses on taskloop simd (and the relevant test).

PR #988, opened 2020-02-17
Added documentation explaining the use of std::list

DavidTruby:


DavidTruby:
I'm not sure about the wording for this so I am open to suggestions.

documentation/Parsing.md
```diff
@@ -164,7 +164,13 @@ Parse tree entities should be viewed as values, not objects; their
 addresses should not be abused for purposes of identification.  They are
 assembled with C++ move semantics during parse tree construction.
 Their default and copy constructors are deliberately deleted in their
-declarations.
+declarations. 
+
+The `std::list<>` data type is used pervasively throughout the parse tree
```
sscalpone: I suggest:

The `std::list<>` data type is used in the parse tree to reliably store pointers to other relevant entries in the tree.  Since the tree lists are moved and spliced at certain points `std::list<>` provides the necessary guarantee of
the stability of these pointers.

sscalpone:
Thanks for adding this info. Perhaps it belongs in the style guide because it isn't about parsing.

DavidTruby:
> Thanks for adding this info. Perhaps it belongs in the style guide because it isn't about parsing.

I put it here as the preceding paragraph is about the use of `std::variant` and `std::tuple` for the parse tree, so I thought this section was the best place to put information about parse tree data structures

DavidTruby:
@sscalpone done!

PR #989, opened 2020-02-19
Add FORALL checking to DoChecker

tskeith:
FORALL statements and constructs require a lot of the same checking
as DO CONCURRENT, so do the checks in DoChecker so that code can be
shared where possible. This requires some reorganization there.
Remove code from AssignmentChecker that did some of these checks.

Change names that contain `DoVar` or `DoVariable` to `IndexVar` to
reflect the fact that they may be DO or FORALL index variables.
Distinguish between the two when necessary with enum `IndexVarKind`.

Change some messages that referred to "concurrent-header" or
"concurrent-control" to specifically say "DO CONCURRENT" or "FORALL".


tskeith:
> I think its great to share all this logic. Why not renaming `DoChecker` to `DoAndForallChecker` or something similar. I know it is a bit ugly, but looking at the `DoChecker` header file, it is not clear if it is entering FORALL statements in order to check DO construct constraints or if it is also checking FORALL constraints independently from Do statements. I understand it is doing the latest from your PR and implementation, so having FORALL in the checker name seems fairer for this construct. I can live with it if you think it does not bring anything to rename the class.

You're right, it should have Forall in the name. I've changed it to `DoForallChecker`.

PR #992, opened 2020-02-19
Data Statement - Implementing Semantic Checks for Constraints C874-C8…

anchu-rajendran:
…87 #991

C882,C884: Implemented and testcase added, C883(already implemented):testcase added, C885-C887(already implemented): added testcase for data-repeat; currently it does not work for data-stmt-constant

### DIFF
--- a/lib/semantics/check-data.cpp
+++ b/lib/semantics/check-data.cpp
@@ -1,0 +1,50 @@
+//===-- lib/semantics/check-data.cpp --------------------------------------===//
+//
+// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
+
+#include "check-data.h"
+
+namespace Fortran::semantics {
+
+void DataChecker::Leave(const parser::DataStmtConstant &dataConst) {
+  if (auto *structure{
+          std::get_if<parser::StructureConstructor>(&dataConst.u)}) {
+    for (const auto &component :
+        std::get<std::list<parser::ComponentSpec>>(structure->t)) {
+      const parser::Expr &parsedExpr{
+          std::get<parser::ComponentDataSource>(component.t).v.value()};
+      if (const auto *expr{GetExpr(parsedExpr)}) {
+        if (!evaluate::IsConstantExpr(*expr)) {  // C884
+          context_.Say(parsedExpr.source,
+              "Structure constructor in data value must be a constant expression"_err_en_US);
+        }
+      }
+    }
+  }
+  // TODO: C886 and C887 for data-stmt-constant
+}
+
+// TODO: C874-C881
+
+void DataChecker::Leave(const parser::DataStmtRepeat &dataRepeat) {
+  if (const auto *designator{parser::Unwrap<parser::Designator>(dataRepeat)}) {
+    if (auto *dataRef{std::get_if<parser::DataRef>(&designator->u)}) {
+      evaluate::ExpressionAnalyzer exprAnalyzer{context_};
+      if (MaybeExpr checked{exprAnalyzer.Analyze(*dataRef)}) {
+        auto expr{
+            evaluate::Fold(context_.foldingContext(), std::move(checked))};
+        if (auto i64{ToInt64(expr)}) {
+          if (*i64 < 0) {  // C882
+            context_.Say(designator->source,
+                "Repeat count for data value must not be negative"_err_en_US);
+          }
+        }
+      }
+    }
+  }
+}
+}
--- a/lib/semantics/check-data.h
+++ b/lib/semantics/check-data.h
@@ -1,0 +1,28 @@
+//===-------lib/semantics/check-data.h ------------------------------------===//
+//
+// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
+
+#ifndef FORTRAN_SEMANTICS_CHECK_DATA_H_
+#define FORTRAN_SEMANTICS_CHECK_DATA_H_
+
+#include "flang/parser/parse-tree.h"
+#include "flang/parser/tools.h"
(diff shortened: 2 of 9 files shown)


anchu-rajendran:
I am not able to provide a ConstantSubobject  to data-object. Consider the following code snippet:
`integer, parameter :: digits(5) = ( /11,22,33,44,55/ )
integer::x
DATA x /digits(1)/`
throws the error "digits is not a derived type". Checking the standards, I do not see this as an error. Please correct me if I am wrong.

lib/semantics/check-data.cpp
```diff
@@ -0,0 +1,52 @@
...
+  if (auto *structure{
+          std::get_if<parser::StructureConstructor>(&dataConst.u)}) {
+    for (const auto &component :
+        std::get<std::list<parser::ComponentSpec>>(structure->t)) {
+      const parser::Expr &parsedExpr{
+          std::get<parser::ComponentDataSource>(component.t).v.value()};
+      if (const auto *expr{GetExpr(parsedExpr)}) {
+        if (!evaluate::IsConstantExpr(*expr)) {  // C884
+          context_.Say(parsedExpr.source,
+              "Structure constructor in DATA value should be a constant expression"_err_en_US);
```
tskeith: I think "must be" is better than  "should be" here and in the next message. That is more consistent with other messages and makes it clear it is a requirement, not just a suggestion.

lib/semantics/check-data.cpp
```diff
@@ -0,0 +1,52 @@
...
+  // TODO: C886 and C887 for data-stmt-constant
+}
+
+// TODO: C874-C881
+
+void DataChecker::Leave(const parser::DataStmtRepeat &dataRepeat) {
+  if (auto *repeatVal{std::get_if<
+          parser::Scalar<parser::Integer<parser::ConstantSubobject>>>(
+          &dataRepeat.u)}) {
+    const parser::Designator &designator{repeatVal->thing.thing.thing.value()};
```
tskeith: You can simplify these 4 lines to:
`if (const auto *designator{parser::Unwrap<parser::Designator>(dataRepeat)}) {`

`parser::Unwrap` takes care of descending through the constraint wrappers and Indirection.

tskeith:
> I am not able to provide a ConstantSubobject to data-stmt-constant.
> Consider the following code snippet:
> `integer, parameter :: digits(5) = ( /11,22,33,44,55/ )`
> `integer::x`
> `DATA x /digits(1)/`
> It throws the error "digits is not a derived type". Checking the standards, I do not see this as an error. Please correct me if I am wrong.

You are right. The parser can't tell if `digits(1)` is a constant subobject or a structure constructor so it is parsing it as the latter. After name resolution we know  that `digits` is an array so we need to rewrite the parse tree to fix it. I will take a look.

klausler:
> I am not able to provide a ConstantSubobject to data-stmt-constant.
> Consider the following code snippet:
> `integer, parameter :: digits(5) = ( /11,22,33,44,55/ )`
> `integer::x`
> `DATA x /digits(1)/`
> It throws the error "digits is not a derived type". Checking the standards, I do not see this as an error. Please correct me if I am wrong.

It is not an error.  The Fortran language is famously ambiguous, and there are productions in the language that cannot be parsed accurately without recourse to a symbol table -- and f18 does not construct a symbol table during parsing, but instead defers all semantic analysis.  Consequently, ambiguities like this get parsed as one of the possible alternatives, and we use a little pass after name resolution that repairs the parse tree before the rest of semantic analysis.  This pass will need to be extended to look at this case (x(1) as a data-stmt-value can be either scalar-constant-subobject or a structure-constructor).

lib/semantics/check-data.h
```diff
@@ -0,0 +1,30 @@
...
+class DataChecker : public virtual BaseChecker {
+public:
+  DataChecker(SemanticsContext &context)
+    : context_{context}, foldingContext_{context_.foldingContext()} {}
+  void Leave(const parser::DataStmtRepeat &);
+  void Leave(const parser::DataStmtConstant &);
+
+private:
+  SemanticsContext &context_;
+  evaluate::FoldingContext &foldingContext_;
```
klausler: There's a `FoldingContext` data member in `SemanticsContext` that you can use directly, so you don't need the reference here or the inclusion of `fold.h` above.

anchu-rajendran: Thank you for the feedback. I will remove `foldingContext_`

lib/semantics/check-data.cpp
anchu-rajendran: Thank you for the feedback! I will make the changes mentioned there.

anchu-rajendran:
Thank you for the clarification on the ambiguous case. 
I have another doubt:
In parse-tree.h,  the only member of `struct DataStmtRepeat` is ` std::variant<IntLiteralConstant, Scalar<Integer<NamedConstant>>, Scalar<Integer<ConstantSubobject>>> u` . 
I see that named constants are represented as  `Scalar<Integer<ConstantSubobject>` and not `Scalar<Integer<NamedConstant>>`. Should this also be changed (so the named constants are represented as `Scalar<Integer<NamedConstant>>`) ? Or, shall this remain as a redundant rule for parsing? The current semantic check are implemented treating NamedConstant as `Scalar<Integer<ConstantSubobject>`

klausler:
> Thank you for the clarification on the ambiguous case.
> I have another doubt:
> In parse-tree.h, the only member of `struct DataStmtRepeat` is ` std::variant<IntLiteralConstant, Scalar<Integer<NamedConstant>>, Scalar<Integer<ConstantSubobject>>> u` .
> I see that named constants are represented as `Scalar<Integer<ConstantSubobject>` and not `Scalar<Integer<NamedConstant>>`. Should this also be changed (so the named constants are represented as `Scalar<Integer<NamedConstant>>`) ? Or, shall this remain as a redundant rule for parsing? The current semantic check are implemented treating NamedConstant as `Scalar<Integer<ConstantSubobject>`

That's another example of ambiguity in the language spec (though it's harmless apart from the needless complexity), since any `named-constant` can also be considered as an instance of a `constant-subobject` (because a `name` can be a `designator`).  If you can remove the `named-constant` from the grammar and parse tree, it would probably yield a less confusing compiler.

sscalpone:
@anchu-rajendran I will merge this PR after you squash your commits into a single commit.

anchu-rajendran:
> @anchu-rajendran I will merge this PR after you squash your commits into a single commit.

Done. Thank you!

PR #993, opened 2020-02-19
Replace module writer posix file handling with llvm file handling.

DavidTruby:
Note: the file handling interface has changed non-trivially between LLVM 9 and LLVM 10/HEAD. Since we are intending to become a direct part of LLVM, this patch uses the new interface and so introduces a dependency on LLVM HEAD.


DavidTruby:
This doesn't appear to work on the CI, but it works fine for me locally. I'll test it on another machine to see what's going on

DavidTruby:
> Should this routine convert from using errno to std::error_code?

Yes, probably. I'll replace that too

DavidTruby:
Seems I was a little hasty here: since we don't rewrite the module unless it's changed, nothing was getting rewritten so with a clean build directory this doesn't work yet.

Although, that does demonstrate that the part where it checks if the module has changed is working fine!

lib/semantics/mod-file.cpp
```diff
@@ -730,9 +726,9 @@ static bool VerifyHeader(const char *content, std::size_t len) {
 }
 
 static std::size_t GetFileSize(const std::string &path) {
-  struct stat statbuf;
-  if (stat(path.c_str(), &statbuf) == 0) {
-    return static_cast<std::size_t>(statbuf.st_size);
+  std::size_t result;
+  if (llvm::sys::fs::file_size(path, result)) {
```
tskeith: This doesn't compile on macos because `size_t` is `unsigned long` and the 2nd argument of file_size is `uint64_t` which is `unsigned long long`. Declaring result as `uint64_t` fixes it.

lib/semantics/mod-file.cpp
```diff
@@ -618,53 +615,50 @@ std::ostream &PutLower(std::ostream &os, const std::string &str) {
...
+      dot < length && path[dot] == '.' ? path.substr(dot + 1) : ""};
   CHECK(length > suffix.length() &&
       path.substr(length - suffix.length()) == suffix);
-  auto tempPath{path.substr(0, length - suffix.length()) + "XXXXXX" + suffix};
-  int fd{mkstemps(&tempPath[0], suffix.length())};
-  auto mask{umask(0777)};
-  umask(mask);
-  chmod(tempPath.c_str(), 0666 & ~mask);  // temp is created with mode 0600
-  return Temp{fd, tempPath};
+  auto prefix = path.substr(0, length - suffix.length());
```
tskeith: Please use braces for initialization.

lib/semantics/mod-file.cpp
```diff
@@ -618,53 +615,50 @@ std::ostream &PutLower(std::ostream &os, const std::string &str) {
...
+  llvm::sys::fs::file_t fd;
+  llvm::SmallString<16> tempPath;
+  if (std::error_code err = llvm::sys::fs::createUniqueFile(prefix + "%%%%%%" + suffix, fd, tempPath))
+    return err;
```
tskeith: Please put braces around the body of the if statement.

lib/semantics/mod-file.cpp
```diff
@@ -618,53 +615,50 @@ std::ostream &PutLower(std::ostream &os, const std::string &str) {
...
-  return 0;
+  if (FileContentsMatch(path, header, contents))
+    return {};
+  llvm::ErrorOr<Temp> temp{MkTemp(path)};
+  if (!temp)
+    return temp.getError();
+  llvm::raw_fd_ostream writer(temp->fd, /*shouldClose=*/false);
+  writer << header;
+  writer << contents;
+  writer.flush();
```
tskeith: Can these writes fail? If so the error should be reported.

lib/semantics/mod-file.cpp
klausler: And around the initialization in the `if` statement.

lib/semantics/mod-file.cpp
```diff
@@ -675,27 +669,29 @@ static bool FileContentsMatch(const std::string &path,
...
+      fs::openNativeFile(path, fs::CD_OpenExisting, fs::FA_Read, fs::OF_None);
+  if (!efile) {
     return false;
   }
+  fs::file_t fd = efile.get();
   constexpr std::size_t bufSize{4096};
-  std::string buffer(bufSize, '\0');
-  if (read(fd, &buffer[0], hsize) != static_cast<ssize_t>(hsize) ||
+  std::vector<char> buffer(bufSize, '\0');
+  if (auto er = fs::readNativeFile(fd, buffer); !er || er.get() != hsize ||
```
klausler: braces

lib/semantics/mod-file.cpp
```diff
@@ -675,27 +669,29 @@ static bool FileContentsMatch(const std::string &path,
...
       std::memcmp(&buffer[0], &header[0], hsize) != 0) {
```
tskeith: I think this would be more readable if the declaration of `er` and the `if` were separated.

lib/semantics/mod-file.cpp
```diff
@@ -729,10 +725,10 @@ static bool VerifyHeader(const char *content, std::size_t len) {
   return expectSum == actualSum;
 }
 
-static std::size_t GetFileSize(const std::string &path) {
-  struct stat statbuf;
-  if (stat(path.c_str(), &statbuf) == 0) {
-    return static_cast<std::size_t>(statbuf.st_size);
+static std::uint64_t GetFileSize(const std::string &path) {
+  std::uint64_t result;
```
klausler: Should be initialized to forestall warnings.

DavidTruby:
> It doesn't appear to avoid writing the .mod file when the contents have changed. This is what I'm seeing:
> 
> ```
> $ echo "module m; end" > m.f90
> $ f18 -fparse-only t.f90
> $ stat -f '%Sm' m.mod
> Feb 20 11:21:47 2020
> $ f18 -fparse-only t.f90
> $ stat -f '%Sm' m.mod
> Feb 20 11:21:50 2020
> ```
> 
> The mtime shouldn't have changed.

I've rewritten the FileContentsMatch function, it works correctly for me now. Can you check if it works at your end?

tskeith:
> I've rewritten the FileContentsMatch function, it works correctly for me now. Can you check if it works at your end?

Yes, it works for me now. Thanks.

sscalpone:
@DavidTruby Would you please add a test that checks that the file modification time is appropriately modified and preserved?

DavidTruby:
> Would you please add a test that checks that the file modification time is appropriately modified and preserved?

I'll have to add a flag to the driver and the semantics parser to print info from the module writer so that this info can be checked by lit, as there's no reliable way to get the mtime at a high enough resolution to check it in a test. Is that acceptable?

tskeith:
> > Would you please add a test that checks that the file modification time is appropriately modified and preserved?
> 
> I'll have to add a flag to the driver and the semantics parser to print info from the module writer so that this info can be checked by lit, as there's no reliable way to get the mtime at a high enough resolution to check it in a test. Is that acceptable?

Can you use a python script to perform the test? That seems simpler.

sscalpone:
> > > Would you please add a test that checks that the file modification time is appropriately modified and preserved?
> > 
> > 
> > I'll have to add a flag to the driver and the semantics parser to print info from the module writer so that this info can be checked by lit, as there's no reliable way to get the mtime at a high enough resolution to check it in a test. Is that acceptable?
> 
> Can you use a python script to perform the test? That seems simpler.

Have a command-line option to display an info msg would be nice and potentially useful in other situations, whatever they may be.

DavidTruby:
> Can you use a python script to perform the test? That seems simpler.

I tried this first, but python's getmtime function only has a reported resolution of one second, and it doesn't seem to actually report even with that resolution correctly.

I think the debug information could be useful anyway, so I've added it in. We might want to make the flag a more general debug flag than I've added, I can change the name easily.

test-lit/semantics/mod-file-rewriter.f90
```diff
@@ -0,0 +1,9 @@
+! RUN: rm -r %t.dir && mkdir %t.dir && cd %t.dir
```
tskeith: The `rm` can fail if the directory isn't there, so this should be `rm -rf ...`

test-lit/semantics/mod-file-rewriter.f90
```diff
@@ -0,0 +1,9 @@
...
+! RUN: %f18 -fparse-only -fdebug-module-writer %s 2>&1 | FileCheck %s --check-prefix CHECK_FIRST_WRITE
+! RUN: %f18 -fparse-only -fdebug-module-writer %s 2>&1 | FileCheck %s --check-prefix CHECK_SECOND_WRITE
+
+module m
+end module m
+
+! CHECK_FIRST_WRITE: Processing module {{.*}}.mod: module written
+! CHECK_SECOND_WRITE: Processing module {{.*}}.mod: module unchanged, not writing
```
tskeith: A more thorough test would be to compile an initial version of `m`, then make a change that should result in the same .mod file and verify it is not rewritten. E.g. change:
```
real :: x(10)
```
to:
```
dimension :: x(10)
public :: x
real :: x
```
Then make a change that should result in a change (e.g. `public` to `private`) and verify it is rewritten.

DavidTruby: Thanks for the suggestion, I've changed the test to do exactly this

DavidTruby:
@sscalpone I've rebased and squash this, it should be ready for merge now.
Thanks!

PR #1009, opened 2020-02-20
Add more checks on WHERE and FORALL

tskeith:
Check that masks and LHS of assignments in WHERE statements and
constructs have consistent shapes. They must all have the same rank and
any extents that are compile-time constants must match.

Emit a warning for assignments in FORALL statements and constructs where
the LHS does not reference each of the index variables.

This makes AssignmentChecker need some Enter/Leave functions that
conflict with those in DoForallChecker so create StatementSemanticsPass3
and move AssignmentChecker there. Future checkers now have the option of
being in that pass.

### DIFF
--- a/lib/semantics/assignment.cpp
+++ b/lib/semantics/assignment.cpp
@@ -29,192 +29,60 @@ using namespace Fortran::parser::literals;
 
 namespace Fortran::semantics {
 
-using ControlExpr = evaluate::Expr<evaluate::SubscriptInteger>;
-using MaskExpr = evaluate::Expr<evaluate::LogicalResult>;
-
-// The context tracks some number of active FORALL statements/constructs
-// and some number of active WHERE statements/constructs.  WHERE can nest
-// in FORALL but not vice versa.  Pointer assignments are allowed in
-// FORALL but not in WHERE.  These constraints are manifest in the grammar
-// and don't need to be rechecked here, since errors cannot appear in the
-// parse tree.
-struct Control {
-  Symbol *name;
-  ControlExpr lower, upper, step;
-};
-
-struct ForallContext {
-  explicit ForallContext(const ForallContext *that) : outer{that} {}
-
-  const ForallContext *outer{nullptr};
-  std::optional<parser::CharBlock> constructName;
-  std::vector<Control> control;
-  std::optional<MaskExpr> maskExpr;
-  std::set<parser::CharBlock> activeNames;
-};
-
-struct WhereContext {
-  WhereContext(MaskExpr &&x, const WhereContext *o, const ForallContext *f)
-    : outer{o}, forall{f}, thisMaskExpr{std::move(x)} {}
-  const WhereContext *outer{nullptr};
-  const ForallContext *forall{nullptr};  // innermost enclosing FORALL
-  std::optional<parser::CharBlock> constructName;
-  MaskExpr thisMaskExpr;  // independent of outer WHERE, if any
-  MaskExpr cumulativeMaskExpr{thisMaskExpr};
-};
-
 class AssignmentContext {
 public:
-  explicit AssignmentContext(SemanticsContext &c) : context_{c} {}
-  AssignmentContext(const AssignmentContext &c, WhereContext &w)
-    : context_{c.context_}, where_{&w} {}
-  AssignmentContext(const AssignmentContext &c, ForallContext &f)
-    : context_{c.context_}, forall_{&f} {}
-
+  explicit AssignmentContext(SemanticsContext &context) : context_{context} {}
+  AssignmentContext(AssignmentContext &&) = default;
+  AssignmentContext(const AssignmentContext &) = delete;
   bool operator==(const AssignmentContext &x) const { return this == &x; }
 
+  template<typename A> void PushWhereContext(const A &);
+  void PopWhereContext();
   void Analyze(const parser::AssignmentStmt &);
   void Analyze(const parser::PointerAssignmentStmt &);
-  void Analyze(const parser::WhereStmt &);
-  void Analyze(const parser::WhereConstruct &);
-  void Analyze(const parser::ForallConstruct &);
-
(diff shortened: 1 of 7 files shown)


lib/semantics/assignment.cpp
```diff
@@ -29,194 +29,79 @@ using namespace Fortran::parser::literals;
...
-  void Analyze(const parser::MaskedElsewhereStmt &);
-  void Analyze(const parser::WhereConstruct::Elsewhere &);
-
   void CheckForPureContext(const SomeExpr &lhs, const SomeExpr &rhs,
       parser::CharBlock rhsSource, bool isPointerAssignment);
-
-  MaskExpr GetMask(const parser::LogicalExpr &, bool defaultValue = true);
-
+  void CheckShape(parser::CharBlock, const SomeExpr *);
+  void CheckForallIndexesUsed(parser::CharBlock, const SomeExpr &);
```
klausler: "indices" is the preferred plural of "index" in mathematical contexts.

lib/semantics/assignment.cpp
```diff
@@ -333,14 +218,68 @@ void AssignmentContext::CheckForPureContext(const SomeExpr &lhs,
...
-    mask = evaluate::ConvertTo(mask, common::Clone(DEREF(logical)));
+// 10.2.3.1(2) The masks and LHS of assignments must all have the same shape
+void AssignmentContext::CheckShape(parser::CharBlock at, const SomeExpr *expr) {
+  if (auto shape{evaluate::GetShape(foldingContext(), expr)}) {
+    std::size_t size{shape->size()};
+    if (whereDepth_ == 0) {
+      whereExtents_.resize(size);
+    } else if (whereExtents_.size() != size) {
+      Say(at,
+          "Must have rank %jd to match prior mask or assignment of"
```
klausler: `%zd` is for `size_t`; `%jd` is for `intmax_t`.

lib/semantics/assignment.cpp
```diff
@@ -333,14 +218,68 @@ void AssignmentContext::CheckForPureContext(const SomeExpr &lhs,
...
+          whereExtents_.size());
+      return;
+    }
+    for (std::size_t i{0}; i < size; ++i) {
+      if (std::optional<std::int64_t> extent{evaluate::ToInt64((*shape)[i])}) {
+        if (!whereExtents_[i]) {
+          whereExtents_[i] = *extent;
+        } else if (*whereExtents_[i] != *extent) {
+          Say(at,
+              "Dimension %d must have extent %jd to match prior mask or"
```
klausler: `%jd` is for `intmax_t`, which we should not assume is the same integral type as `int64_t`.

lib/semantics/semantics.cpp
```diff
@@ -111,18 +111,20 @@ template<typename... C> class SemanticsVisitor : public virtual C... {
...
 using StatementSemanticsPass1 = ExprChecker;
 using StatementSemanticsPass2 = SemanticsVisitor<  //
-    AllocateChecker, ArithmeticIfStmtChecker, AssignmentChecker, CoarrayChecker,
-    DeallocateChecker, DoForallChecker, IfStmtChecker, IoChecker,
-    NullifyChecker, OmpStructureChecker, PurityChecker, ReturnStmtChecker,
-    StopChecker>;
+    AllocateChecker, ArithmeticIfStmtChecker, CoarrayChecker, DeallocateChecker,
+    DoForallChecker, IfStmtChecker, IoChecker, NullifyChecker,
+    OmpStructureChecker, PurityChecker, ReturnStmtChecker, StopChecker>;
+using StatementSemanticsPass3 = SemanticsVisitor<AssignmentChecker>;
```
klausler: Adding a third pass over the parse tree seems expensive if we're just getting warning messages for obsolescent `FORALL` statements in return (is that the case?).  Can we avoid a third pass, or perhaps instantiate a localized one that only operates as a second pass over `FORALL` bodies when they appear (which is rare)?

tskeith: My assumption is that we will eventually need a third pass anyway as more checkers are added. If that's not the case, I agree this is overkill just for these FORALL checks.

In this case the conflict is only because I want to know when an assignment or concurrent-header is inside a FORALL vs. a DO CONCURRENT. SemanticsContext maintains a stack of constructs we are currently nested inside. That would be good enough except for the case of a FORALL statement, which wouldn't appear in the stack. We could get it to work by treating a FORALL statement like a construct (we would probably want to do the same with IF statements). Or perhaps rewrite FORALL statements into FORALL constructs early on so we only have that case to worry about.

klausler: However this works out, it would be counterintuitive if non-construct `IF`, `FORALL`, &/or `WHERE` statements were not represented in the context used for semantic analysis of their single controlled statements.

lib/semantics/assignment.cpp
tskeith: I will change it if you have a strong preference. I chose "Indexes" because irregular plurals make it harder to search the code for terms like "index".

lib/semantics/assignment.cpp
tskeith: So in this case we have to cast to `intmax_t`, right?

lib/semantics/assignment.cpp
tskeith: OK

lib/semantics/assignment.cpp
klausler: "index" is not incorrect, so leave it alone if that's your preference.

lib/semantics/assignment.cpp
klausler: That's what I typically do, yes.  It's ugly but it should be future-proof.

sscalpone: Or pass a string.  Do we have an itoa?

lib/semantics/semantics.cpp
tskeith: I've moved the check into check-do-forall.cpp where we were already collect FORALL index variables anyway. This allows us to get by with only two semantics passes.

PR #1014, opened 2020-02-22
Remove use of std::set::merge

tskeith:
Some version of clang that we are building with don't have
std::set::merge, even though it is part of C++17. Work around that by
using std::set::insert until we can count on merge being available
everywhere.

### DIFF
--- a/lib/Semantics/check-do-forall.cpp
+++ b/lib/Semantics/check-do-forall.cpp
@@ -763,13 +763,27 @@ private:
           common::visitors{
               [&](const evaluate::Assignment::BoundsSpec &spec) {
                 for (const auto &bound : spec) {
+// TODO: this is working around missing std::set::merge in some versions of
+// clang that we are building with
+#ifdef __clang__
+                  auto boundSymbols{evaluate::CollectSymbols(bound)};
+                  symbols.insert(boundSymbols.begin(), boundSymbols.end());
+#else
                   symbols.merge(evaluate::CollectSymbols(bound));
+#endif
                 }
               },
               [&](const evaluate::Assignment::BoundsRemapping &remapping) {
                 for (const auto &bounds : remapping) {
+#ifdef __clang__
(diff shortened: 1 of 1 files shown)


tskeith:
> I think that this is a case for conditional code so that the work-around only gets applied in cases where it's necessary, while the general case uses the standard API. Eventually the work-around case can be removed.

I don't know how to identify the cases where the work-around is needed. Some builds of clang 9 work and some don't, depending on the include files that are used.

jeanPerier:
> > I think that this is a case for conditional code so that the work-around only gets applied in cases where it's necessary, while the general case uses the standard API. Eventually the work-around case can be removed.
> 
> I don't know how to identify the cases where the work-around is needed. Some builds of clang 9 work and some don't, depending on the include files that are used.

You could use CMake + `try_compile` to achieve this:
- Write a small c++ file using ste::merge, place it under `cmake/test-set-merge.cpp`.
- Use `try_compile` in CMake on this test file and define a macro `CXX_HAS_SET_MERGE` based on this test
- Use the macro where applicable

Pros of doing this: when all supported compiler should have this, we can have depreciation period before removing the macro and so on. I would only deploy all this if there is a performance hit of using `insert` vs `merge` though.

tskeith:
> Pros of doing this: when all supported compiler should have this, we can have depreciation period before removing the macro and so on. I would only deploy all this if there is a performance hit of using `insert` vs `merge` though.

Thanks for the idea -- that's a handy technique. I agree with you that we should use this only if there is a performance concern (and I'm pretty sure there isn't). It would add too much complexity for a temporary work-around.

But I'm not sure how to proceed. If we have to support building with C++ environments that don't have std::set::merge, I would propose keeping my change as-is. Otherwise, don't make any change. But then how do we describe the compiler requirements to someone who wants to build f18?

klausler:
Can you make the change dependent on clang, independent of its version?

tskeith:
> Can you make the change dependent on clang, independent of its version?

Yes, I've put it in ifdef __clang__.

PR #1023, opened 2020-02-25
Allow for access-stmt before namelist-stmt

### LINKED ISSUE #1022: Having a namelist attribute specified before the namelist definition generates an error
```
module m
  integer :: x
  public :: NLIST
  NAMELIST /NLIST/ x
end module m
```

gfortran, ifort, nagfor all allow this.

tskeith:
You can declare a name in an access statement and then declare it as
a namelist group name. We weren't allowing that because we didn't
convert a symbol with UnknownDetails to one with NamelistDetails.

Fixes #1022.

### DIFF
--- a/test/Semantics/resolve40.f90
+++ b/test/Semantics/resolve40.f90
@@ -72,3 +72,19 @@ subroutine s9
   !ERROR: 'i' is already declared in this scoping unit
   data ((x(i,i),i=1,2),i=1,2)/4*0.0/
 end
+
+module m10
+  integer :: x
+  public :: nl
+  namelist /nl/ x
+end
+
+subroutine s11
+  integer :: nl2
+  !ERROR: 'nl2' is already declared in this scoping unit
+  namelist /nl2/x
+  namelist /nl3/x
+  !ERROR: 'nl3' is already declared in this scoping unit
+  integer :: nl3
+  nl2 = 1
+end
--- a/lib/Semantics/resolve-names.cpp
+++ b/lib/Semantics/resolve-names.cpp
@@ -2957,7 +2957,7 @@ void DeclarationVisitor::Post(const parser::EntityDecl &x) {
     if (ConvertToObjectEntity(symbol)) {
       Initialization(name, *init, false);
     }
-  } else if (attrs.test(Attr::PARAMETER)) { // C882, C883
+  } else if (attrs.test(Attr::PARAMETER)) {  // C882, C883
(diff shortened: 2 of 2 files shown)


test/Semantics/resolve40.f90
```diff
@@ -72,3 +72,9 @@ subroutine s9
   !ERROR: 'i' is already declared in this scoping unit
   data ((x(i,i),i=1,2),i=1,2)/4*0.0/
 end
+
+module m10
```
klausler: Please ensure that we've got at least one error test that has a namelist name conflicting with a variable (or other symbol), in both orders of declaration.

vjayathirtha-nv:
I tried the test case @klausler suggested.
```
module m
  integer, private :: x
  public :: INLIST
  NAMELIST /INLIST/ x
  integer :: INLIST
end module m
```
Here's the error I see 
```
/home/vjayath/nml4.f90:5:14: error: 'nlist' is already declared in this scoping unit
    integer :: NLIST
               ^^^^^
/home/vjayath/nml4.f90:3:13: Previous declaration of 'nlist'
    public :: NLIST
              ^^^^^
f18: semantic errors in /home/vjayath/nml4.f90
```
Should the info message be pointing to line no. 4 instead of 3? If a programmer has no idea the `NLIST` namelist has be declared, having an access statement seems perfectly legal.

Apart from that, the change looks good to me.

test/Semantics/resolve40.f90
tskeith: OK, I've add those tests.

tskeith:
> I tried the test case @klausler suggested.
> 
> ```
> module m
>   integer, private :: x
>   public :: INLIST
>   NAMELIST /INLIST/ x
>   integer :: INLIST
> end module m
> ```
> 
> Here's the error I see
> 
> ```
> /home/vjayath/nml4.f90:5:14: error: 'nlist' is already declared in this scoping unit
>     integer :: NLIST
>                ^^^^^
> /home/vjayath/nml4.f90:3:13: Previous declaration of 'nlist'
>     public :: NLIST
>               ^^^^^
> f18: semantic errors in /home/vjayath/nml4.f90
> ```
> 
> Should the info message be pointing to line no. 4 instead of 3? If a programmer has no idea the `NLIST` namelist has been declared, having an access statement seems perfectly legal.

I've fixed it so that the name of the symbol is the occurrence of "NLIST" at line 4.

PR #1024, opened 2020-02-26
Fix misparsed structure constructor in data stmt

tskeith:
In a data statement like `data x / a(1) /`, `a(1)` may be an array
element or a structure constructor. It is parsed as an array element
so if it turns out `a` is a derived type it must be rewritten as a
strucutre constructor.

### DIFF
--- a/test/Semantics/data01.f90
+++ b/test/Semantics/data01.f90
@@ -37,12 +37,25 @@ subroutine CheckRepeat
   DATA myName%age / digits(myAge) * 35 /
 end
 
-!subroutine CheckValue
-!  use m1
-!  !C883
-!  !ERROR: Derived type 'persn' not found
-!  DATA myname / persn(2, 'Abcd Efgh') /
-!  !C884
-!  !ERROR: Structure constructor in data value must be a constant expression
-!  DATA myname / person(myAge, 'Abcd Ijkl') /
-!end
+subroutine CheckValue
+  use m1
+  !OK: constant structure constructor
+  data myname / person(1, 'Abcd Ijkl') /
+  !C883
+  !ERROR: Must have INTEGER type, but is CHARACTER(1)
+  data myname / persn(2, 'Abcd Efgh') /
+  !C884
+  !ERROR: Structure constructor in data value must be a constant expression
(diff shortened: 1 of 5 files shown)


tskeith:
@anchu-rajendran, this fixes the problem you mentioned in PR #992.

test/Semantics/data01.f90
```diff
@@ -37,12 +37,25 @@ subroutine CheckRepeat
...
-!  DATA myname / person(myAge, 'Abcd Ijkl') /
-!end
+subroutine CheckValue
+  use m1
+  !OK: constant structure constructor
+  data myname / person(1, 'Abcd Ijkl') /
+  !C883
+  !ERROR: Must have INTEGER type, but is CHARACTER(1)
+  data myname / persn(2, 'Abcd Efgh') /
+  !C884
```
anchu-rajendran: @tskeith, Is there a change in error message emitted? This does not seem to be a valid error message to me. It was emitting something like "Derived type 'persn' not found".

anchu-rajendran:
> @anchu-rajendran, this fixes the problem you mentioned in PR #992.

Thank you! I will take a look and try to learn how this is handled.

test/Semantics/data01.f90
tskeith: The problem is that the compiler can't tell if `persn` was meant to be an array or a derived type. It now parses it as an array element reference and changes it to a derived type constructor if it is the name of a known derived type. Since it's not, it leaves it as an array element reference, so the error is that a character can't be used as an array index.

We might be able to improve the error message by reporting that it neither an array nor a derived type.

anchu-rajendran: I find that as a better way if possible. That can cover the cases where the constant part of any ConstantSubobject is not defined. If that requires any extra implementation in Semantic part, I can take a look at that.

tskeith:
> Attached the tests related to these issue. I am commenting here because I found them looking at your PR, but apart from issue 3, they are unrelated to this PR. Should I move all of this in a new issue or do you want to have a look at issue 3 here ?

Thanks, Jean. I'll take a look and decide if I want to fix those in this PR or a new one. You don't need to create an issue.

Issue #997, opened 2020-02-20
Port F18 semantics/test_errors.sh to equivalent to Clang -verify


RichBarton-Arm:
After #941 is completed, F18 will contain a directory called test containing regression tests for the F18 compiler driven by the LLVM lit testing framework. LLVM projects tend also to use FileCheck for such tests whilst the F18 tests will still use custom bash scripts instead. During the review for #934 we identified that FileCheck may not be the best option for porting the tests that use semantics/test_errors.sh. Instead we proposed that if Flang had an equivalent feature to Clang's -verify option then test_errors.sh tests would port well to that.

This task is to investigate if a flang -verify option is feasible and if so implement it and port semantics/test_errors.sh to using it. If it is not feasible, then investigate what a FileCheck version of semantics/test_errors.sh would look like so we can consider that option again.

Note that the rest of the test scripts are expected to port to using FileCheck under #996

sscalpone:
What is the advantage to using an internal -verify implementation vs an external script?

Issue #996, opened 2020-02-20
Port F18 lit tests to using FileCheck or suitable LLVM-alternative


RichBarton-Arm:
After #941 is completed, F18 will contain a directory called test containing regression tests for the F18 compiler driven by the LLVM lit testing framework. LLVM projects tend also to use FileCheck for such tests whilst the F18 tests will still use custom bash scripts instead.

This task is to evaluate in turn each of the custom scripts and consider if they can be replaced with FileCheck and rewrite the tests accordingly.

The list of test scripts are:
- semantics/test_modfile.sh
- semantics/test_symbols.sh
- semantics/test_any.sh
- semantics/common.sh
- evaluate/test_folding.sh

The tests in semantics that use test_errors.sh will be covered under #996 

For each script, the porting order of preference would be:
1. Port to FileCheck
2. Port to some other LLVM-like testing utility or feature
3. Rewrite the bash script to python

RichBarton-Arm:
Ivan Zechev ported the rest of the tests over to Python - see https://reviews.llvm.org/rGdc453dcf760e6d558da3a4d7fff301baa5f37aba for last commit in the series.

Issue #995, opened 2020-02-20
Port F18 unittests to gtest


RichBarton-Arm:
After #941 is complete, F18 will have a directory unittests that contain tests written in C++ that wrap various internals of F18's runtime or shared utilities in a test harness and test them. This is supported by some custom test harness code.

LLVM projects use the gtest framework for tests of this type. See https://github.com/llvm/llvm-project/tree/master/clang/unittests for an example of this.

The F18 tests need to be ported to also use this framework.

Sameeranjoshi:
Hi @RichBarton-Arm is someone working on this currently?
I was in need of at least a basic framework for adding FIR gtest. By basic I mean the Cmake file changes, so we can run at least a single test.

Sameeranjoshi:
This ticket seems like we need a port of all unittests/ files into a gtest based framework, but given that there is no unittest framework using gtest, should we have that first?

RichBarton-Arm:
Hi @Sameeranjoshi 
Currently there is nobody working on this so it is up for grabs.
Adding the gtest framework with a single test would definitely be a good start. We could then work on migration across from ctest to gtest once we have a target to aim at.
Please feel free to claim it and move it to in progress.

Sameeranjoshi:
https://reviews.llvm.org/D80377

RichBarton-Arm:
@Sameeranjoshi has committed https://reviews.llvm.org/D80377 which puts the infrastructure in place. All new tests can be added as gtest tests. We now need to port the existing tests over to this format.

banach-space:
A fresh patch in this direction: https://reviews.llvm.org/D97403

Thank you @ashermancinelli!

banach-space:
https://reviews.llvm.org/D105315 concludes the porting of `flang/unittests/Runtime`. Massive thank you to @ashermancinelli for the enormous effort!

We still need to port `flang/unittests/Decimal` and `flang/unittests/Evaluate`. Please volunteer :)

PR #1027, opened 2020-02-26
Porting F18 tests to use LLVM lit without FileCheck

LukeIreland1:
#941 had diverged a fair amount, and I thought making a new PR would make things a lot cleaner and simpler.

This PR ports all Fortran tests, and is a precursor for porting the remaining unit tests (currently using ctest) to use gtest instead as per #995.

The majority of these Fortran tests will later be ported to FileCheck per #996, and the remaining error tests will be ported using a `flang -verify` equivalent per #997.


test/Evaluate/folding01.f90
```diff
@@ -1,3 +1,4 @@
+!RUN: %S/test_folding.sh %s %flang
```
DavidTruby: Might it be worth adding "tools" for these in lit.cfg.py so instead of doing %S/test_folding.sh we can do %folding? It might just be a little neater than having to specify the path to the script.

test/Semantics/critical04.f90
```diff
@@ -1,4 +1,5 @@
-! RUN: ${F18} -funparse-with-symbols %s 2>&1 | ${FileCheck} %s
+!RUN: %S/test_any.sh %s %flang
+! EXEC: ${F18} -funparse-with-symbols %s 2>&1 | ${FileCheck} %s
```
DavidTruby: What does EXEC: do here? I've not seen it used anywhere in LLVM before.
Should this say %f18 instead of ${F18}?

test/Evaluate/folding01.f90
LukeIreland1: Funny you should say that! I did this originally, but both @tskeith and @RichBarton-Arm said they prefer it like this (I think it was something to do with the neatness of the config file being preferable).

test/Semantics/critical04.f90
LukeIreland1: It's identical, but used to make their scripts work. test_any.sh uses identical runlines, so we replaced them with EXEC lines, and left its version of %f18 as it works without being changed.

test-lit/Semantics/common.sh
```diff
@@ -24,5 +24,16 @@ case $1 in
...
+then
+  read -A $temp <<< $src
+  temp=$temp[1]
+  echo $temp
+  temp=$(echo "$temp" | cut -f 1 -d '.')
+else
+  temp=$(echo "$src" | cut -f 1 -d '.')
+fi
+temp="Outputs/$temp"
+mkdir -p $temp
```
tskeith: In the other PR, weren't you using `%t` and letting lit manage the temp directory? That seems simpler.

test/Evaluate/folding01.f90
tskeith: Yes, I still prefer it like this.

tskeith:
I don't see how I can review this set of commits. I can't look at all of them together because it's a mix of hand-written and mechanical changes but if I look at them individually there are all kinds or extraneous file moves. Those moves don't need to be visible in the final commits.

test/Evaluate/folding01.f90
DavidTruby: Happy to go with the majority opinion on this one, disregard my comment

tskeith:
When I run `ninja check-all` I get an error because it's looking for `test-lit/lit.cfg.py`.

RichBarton-Arm:
> 
> 
> I don't see how I can review this set of commits. I can't look at all of them together because it's a mix of hand-written and mechanical changes but if I look at them individually there are all kinds or extraneous file moves. Those moves don't need to be visible in the final commits.

Understood. Luke took this feedback from the last review and for this one he separated the mechanical changes from the automated changes from the manual changes in separate commits so I recommend reviewing commit by commit.

 * The first patch in the series introduces the porting script and moves the test scripts to the new locations that the porting script expects them to be.
 * The second and third patches are hand-written changes to the existing tests with rationale
 * The fourth commit is simply the result of running the script on the remaining Fortran ctest tests and porting them to lit style.
 * The fifth and sixth patches are mechanical renames/deletions of files.

Hope that helps with the review. If you can propose another arrangement of patches that would make the review easier then @LukeIreland1 can rearrange accordingly.

test-lit/Semantics/common.sh
LukeIreland1: I'm not 100% sure, but either lit doesn't actually do the thing I need it to do (cleaning up most likely), or judging from that if statement, it's more likely to do with the fact that this script doesn't know how to handle the input that lit now gives it due to manual changes in order to make the tests pass with lit.

tskeith:
> Hope that helps with the review. If you can propose another arrangement of patches that would make the review easier then @LukeIreland1 can rearrange accordingly.

Here's what I said in the the other PR:
> That sounds good to me, though before approving I would like the see the commits in the form that they will be merged. I assume there will be no evidence of the porting script or test-lit. Just removing ctest configuration, adding lit configuration, and 1-line (or so) changes to each test source file. And maybe some changes to the shell scripts.

RichBarton-Arm:
@tskeith Some of the tests needed more effort to port than a one-line change so there has to be some hand-written changes in this PR. These changes are separated out in patches #2 and #3 for specific review. Additionally, some changes were needed to the testing scripts themselves to make them interact well with lit and these are in patch #1. The file moves and mechanical file moves with a one-line diff are in patches #4-#6.

> That sounds good to me, though before approving I would like the see the commits in the form that they will be merged. I assume there will be no evidence of the porting script or test-lit. Just removing ctest configuration, adding lit configuration, and 1-line (or so) changes to each test source file. And maybe some changes to the shell scripts.

Unless I misunderstand your request, isn't this the same as reviewing all commits together? https://github.com/flang-compiler/f18/pull/1027/files, modulo a commit removing the script at the end?

My recommendation is to review patches #1, #2 and #3 like normal code changes then review the rest as a single lump bearing in mind most of it is mechanical. Does that seem like a way forward?

LukeIreland1:
Rebased on top of master

test-lit/Semantics/common.sh
LukeIreland1: I can pass %t, but the shell scripts need to make the directory still within %t. Otherwise, they complain that the directory doesn't exist(which is true, because they're not looking at %t, something like %t/log instead), which lit can't handle as a result of this combination.

test-lit/Semantics/common.sh
```diff
@@ -11,18 +11,12 @@ function die {
...
   (/*) src=$1 ;;
   (*) src=$(dirname $0)/$1 ;;
 esac
 USER_OPTIONS=`sed -n 's/^ *! *OPTIONS: *//p' $src`
 echo $USER_OPTIONS
-temp=`mktemp -d ./tmp.XXXXXX`
-[[ $KEEP ]] || trap "rm -rf $temp" EXIT
+F18=$2
+temp=$3
+mkdir -p $temp
```
tskeith: This used to report an error if the wrong number of arguments is passed in. That seems like a good thing.

test-lit/Semantics/common.sh
DavidTruby: %t is a unique _nonexistent_ path that tests can use for whatever they want. So you still need to create it, lit doesn't do that for you. In this case I do think you could use mkdir %t here.

test/Preprocessing/pp001.F
```diff
@@ -1,3 +1,4 @@
+!RUN: %flang -E %s
```
tskeith: These files were tests before. I'm not sure this adds much because there is no checking of results.

LukeIreland1: They don't need to be checked in this case.

test-lit/Semantics/common.sh
LukeIreland1: The wrong number of arguments can't be passed in (so it's not needed) as far as the automated tests are concerned, I'm sure people manually using the tests will know what they're doing (but they shouldn't since new tests should be gtest or lit), and the main reason I took this out was to be able to add %t as an argument, and I couldn't otherwise.

tskeith:
> Unless I misunderstand your request, isn't this the same as reviewing all commits together? https://github.com/flang-compiler/f18/pull/1027/files, modulo a commit removing the script at the end?

No, I'm asking for all of the mechanical changes to be together so that the history does not contain extraneous moves to test-lit and then back again.

> My recommendation is to review patches #1, #2 and #3 like normal code changes then review the rest as a single lump bearing in mind most of it is mechanical. Does that seem like a way forward?

Yes, that helps for reviewing, but my request for the final form of the commits still applies.

The changes to CMakeLists.txt don't belong with the mechanical changes.

test/Preprocessing/pp001.F
tskeith: Sorry, I meant to write "these files were **not** tests before". So running them as if they are tests is a change. I'm wondering if there is any benefit to doing that.

LukeIreland1: I suppose not, but we thought to port them unless there is an objection since their addition implies they will be included as a test in future.

LukeIreland1:
> > Unless I misunderstand your request, isn't this the same as reviewing all commits together? https://github.com/flang-compiler/f18/pull/1027/files, modulo a commit removing the script at the end?
> 
> No, I'm asking for all of the mechanical changes to be together so that the history does not contain extraneous moves to test-lit and then back again.
> 

That's fine, 5 and 6 can be combined.
> > My recommendation is to review patches #1, #2 and #3 like normal code changes then review the rest as a single lump bearing in mind most of it is mechanical. Does that seem like a way forward?
> 
> Yes, that helps for reviewing, but my request for the final form of the commits still applies.
> 
> The changes to CMakeLists.txt don't belong with the mechanical changes.

Even if the change to CMakeLists.txt is purely mechanical (or purely as a result of mechanical changes)?

test/Semantics/common.sh
```diff
@@ -11,18 +11,12 @@ function die {
...
   (/*) src=$1 ;;
   (*) src=$(dirname $0)/$1 ;;
 esac
 USER_OPTIONS=`sed -n 's/^ *! *OPTIONS: *//p' $src`
 echo $USER_OPTIONS
-temp=`mktemp -d ./tmp.XXXXXX`
-[[ $KEEP ]] || trap "rm -rf $temp" EXIT
+F18=$2
+temp=$3
+mkdir -p $temp
```
RichBarton-Arm: github wants you to add a newline here.

unittests/Evaluate/CMakeLists.txt
```diff
@@ -134,15 +134,6 @@ target_link_libraries(folding-test
 )
 
 set(FOLDING_TESTS
```
RichBarton-Arm: The empty list should also be removed, as well as any references to it lower down the file.

RichBarton-Arm:
> 
> 
> > Unless I misunderstand your request, isn't this the same as reviewing all commits together? https://github.com/flang-compiler/f18/pull/1027/files, modulo a commit removing the script at the end?
> 
> No, I'm asking for all of the mechanical changes to be together so that the history does not contain extraneous moves to test-lit and then back again.
> 
> > My recommendation is to review patches #1, #2 and #3 like normal code changes then review the rest as a single lump bearing in mind most of it is mechanical. Does that seem like a way forward?
> 
> Yes, that helps for reviewing, but my request for the final form of the commits still applies.

Perhaps the way forward is to review this change as a patch series and agree we are happy with it, then for Luke to squash the series into two commits:
* the patch which fixes test/Semantics/altreturn{02,02}
* a single commit that does the rest - which would look like tests moving out of `test` into `unittests` with CMakeLists.txt changes; tests in `test` changing to use lit and accompanying CMakeLists changes and content moving over from `test-lit` and merging with content in `test`

We can do a final review on that patch at the time. Bear in mind that if more new ctest tests are added between Luke creating the PR and it getting approved then we'd need to create a new one as it will likely not merge properly. So we'll need to synchronise on that.

Does that sound ok to you?

> The changes to CMakeLists.txt don't belong with the mechanical changes.

We wrote the script to remove ported tests from the CMakeLists files in test as it goes along. This is why it ends up in the same patch. We found this was useful to ensure we didn't miss anything along the way. We can easily change the script behaviour and make the CMakeLists.txt changes manually in a separate patch in the series if you have a strong preference for that.

test/Preprocessing/pp001.F
tskeith: The third option is to leave them as they were, just examples that are not executed as tests.

If you think there is value in executing them to check there are no compilation errors, that's fine. But that should be made explicit in the PR (or a separate PR) saying that's what you're doing that and why. Don't bury it under all of the changes related to porting tests.

RichBarton-Arm: Sorry for not being more clear in the description.

Lit will emit an error for test files in the test directory (identified by file extension) that contain no runlines. So to port these tests to lit, we have three options:
 1. Add RUN lines to them (our chosen option)
 2. Explicitly exclude this directory from lit's searching process
 3. add a lit.local.cfg in this test directory that changes the file extensions associated with tests.
 4. Rename the file extensions (yuk!)

If we don't like any of these options we could leave them behind in the unittests directory with no change and revisit later or we could delete them altogether. 

To me, it seems pointless and unhelpful to have test files in the test directory that look like tests but are intentionally not executed. If we don't want to execute these then I think we should delete them. Strong view weakly held though. What would you be happy with?

tskeith: 1 and 2 are both fine with me, as is moving them to a different directory, say `examples`. I just want it to be an explicit decision that is made, not something that happens accidentally.

@klausler, what do you think we should do with the preprocessor tests?

test-lit/Semantics/omp-nested01.f90
```diff
@@ -1,3 +1,5 @@
+!RUN: %S/test_errors.sh %s %flang %t
+!XFAIL: *
```
tskeith: Why?

LukeIreland1: XFail? This test is commented out, so we ported it, but left it as a failing test.

LukeIreland1:
Rebased on top of master

tskeith:
> Perhaps the way forward is to review this change as a patch series and agree we are happy with it, then for Luke to squash the series into two commits:
> 
>     * the patch which fixes test/Semantics/altreturn{02,02}
> 
>     * a single commit that does the rest - which would look like tests moving out of `test` into `unittests` with CMakeLists.txt changes; tests in `test` changing to use lit and accompanying CMakeLists changes and content moving over from `test-lit` and merging with content in `test`

How about this set of commits:
- fix altreturn tests
- move tests to `unittests`
- add "RUN:" lines to tests that will be ported to lit (i.e. the change that no one will look at because it is a similar change to hundreds of files)
- changes to run tests with lit rather than ctest
- move tests from `test-lit` to `test`

LukeIreland1:
Rebased on top of master.

LukeIreland1:
@sscalpone @tskeith @klausler Do any of you have any further requests for this change? It's ready to merge other than that.

tskeith:
> @sscalpone @tskeith @klausler Do any of you have any further requests for this change? It's ready to merge other than that.

This isn't okay. The history of every test file will show it moving to test-lit and back to test again.

LukeIreland1:
> > @sscalpone @tskeith @klausler Do any of you have any further requests for this change? It's ready to merge other than that.
> 
> This isn't okay. The history of every test file will show it moving to test-lit and back to test again.

Then as @RichBarton-Arm suggested, the last 4 can be combined. Although, I'm not sure I understand why this is an issue.

Would you like me to combine them already, or is there something about a specific commit you want changed? For now, I can only think of deleting the old tests in the same commit as the move, which will make the move commit look like the porting commit, and make the series pointless.

LukeIreland1:
@tskeith Are those 4 commits any better? They can be squashed further.

RichBarton-Arm:
Regardless of the order, the commit messages need improving. Patch #2 is too terse. It should explain the rationale why these tests are moving. Patch #3 only covers the tests that you manually ported. You should explain what the general case is for the other tests - i.e. add RUN line, explain what the EXEC lines are, explain how XFAIL works. You also need to call out what we are doing for the preprocessor tests as @tskeith requested before. Perhaps pull those changes into a separate patch as well - but I'll go with @tskeith's opinion on that one. The final patch should be expanded again to explain why you are making these changes.

test/Semantics/test_modfile.sh
```diff
@@ -26,7 +26,7 @@ for src in "$@"; do
   ) > $actual_files
   expected_files=$(sed -n 's/^!Expect: \(.*\)/\1/p' $src | sort)
   extra_files=$(echo "$expected_files" | comm -23 $actual_files -)
-  if [[ ! -z "$extra_files" ]]; then
+  if [[ ! -z "$extra_files" && ! -z "$expected_files" ]]; then
```
tskeith: Why is this change needed?

tskeith:
The current set of commits looks okay, and I agree with @RichBarton-Arm about the comments.

Regarding the preprocessor tests, I think any change there should be discussed in a separate PR. The goal here should be preserving existing functionality while porting to run tests with lit. So for this PR just exclude the preprocessor tests from being run by lit.

test/Semantics/test_modfile.sh
LukeIreland1: It's not, I thought I was getting some bizarre errors regarding files coming up in the extra_files var, but it was an outdated compiler.

LukeIreland1: Okay, turns out I was mistaking this change for another. I removed this and getsymbols02.f90 passed, but when I cleared my cache it failed.

getsymbols02-a.f90 complains `Unexpected .mod files produced: mm2a.mod`, but that seems like valid behaviour to me. Therefore, since all generic tests that use modfiles, never have the produced modfile in their expected files (when it should), I added this to make sure those tests still pass.

test/Preprocessing/pp001.F
```diff
@@ -1,3 +1,4 @@
+! RUN: %flang -E %s
```
tskeith: This shouldn't be here as these tests aren't changing.

test/Semantics/test_modfile.sh
tskeith: That test is supposed to be run with test_any.sh, not test_modfile.sh.

test/Preprocessing/pp001.F
LukeIreland1: Okay, we left them ported, but stopped them from being run. I'll remove them altogether.

klausler: Please do not delete these test files, if that's indeed your intention.

test/Preprocessing/lit.local.cfg.py
```diff
@@ -0,0 +1,16 @@
...
+import sys
+
+import lit.formats
+from lit.llvm import llvm_config
+
+# Configuration file for the 'lit' test runner.
+
+# testFormat: The test format to use to interpret tests.
+config.test_format = lit.formats.ShTest(execute_external=False)
+
```
RichBarton-Arm: Correct me if I am wrong, but I don't think anything above this line is needed for the intended purpose of this file which is simply to hide these potential test input files from lit's search algorithms. All you need is the line setting `config.suffixes`

test/Preprocessing/lit.local.cfg.py
```diff
@@ -0,0 +1,16 @@
...
+# suffixes: A list of file extensions to treat as test files.
```
RichBarton-Arm: I think you should leave a comment here explaining why you have added this file with this line. I would reference your new Issue that you created for eventually enabling these tests in this comment.

test/lit.cfg.py
```diff
@@ -50,7 +50,7 @@
 config.test_source_root = os.path.dirname(__file__)
 
 # test_exec_root: The root path where tests should be run.
-config.test_exec_root = os.path.join(config.flang_obj_root, 'test-lit')
+config.test_exec_root = os.path.join(config.flang_obj_root, 'test')
```
tskeith: It looks like `test_exec_root` is set up above.

LukeIreland1:
> Does there need to be a change to delete the test-lit directory too? When I download this PR locally I still have a test-lit directory containing a **pycache** directory.

There's no pycache directory in https://github.com/flang-compiler/f18/pull/1027/files, so I can only suggest `git clean -ffdx`

kiranchandramohan:
@DavidTruby @RichBarton-Arm Please add this to the projects page and remove PR941.

RichBarton-Arm:
@kiranchandramohan done

LukeIreland1:
Rebased on top of master. Annoyingly missed a new Fortran regression test as I wasn't expecting any at this stage, but has been resolved now.

test/Semantics/common.sh
```diff
@@ -11,18 +11,12 @@ function die {
   exit 1
 }
 
-case $# in
-  (1) ;;
-  (2) F18=$2 ;;
-  (*) echo "Usage: $(basename $0) <fortran-source> [<f18-executable>]"; exit 1
-esac
-[[ -z ${F18+x} ]] && die "Path to f18 must be second argument or in F18 environment variable"
-[[ ! -f $F18 ]] && die "f18 executable not found: $F18"
```
RichBarton-Arm: Why remove this line? Seems ok to me to move this to after line 20.

test/Semantics/common.sh
```diff
@@ -11,18 +11,12 @@ function die {
   exit 1
 }
 
-case $# in
-  (1) ;;
-  (2) F18=$2 ;;
-  (*) echo "Usage: $(basename $0) <fortran-source> [<f18-executable>]"; exit 1
```
RichBarton-Arm: Seems like you could preserve this usage message also with a similar bit of code to that in test_folding.sh?

test/Evaluate/test_folding.sh
```diff
@@ -19,23 +19,18 @@
...
 #     compiler could not fold the expression.
 
-PATH=/usr/bin:/bin
-srcdir=$(dirname $0)
-F18CC=${F18:-../../../tools/f18/bin/f18}
-CMD="$F18CC -fdebug-dump-symbols -fparse-only"
+CMD="$2 -fdebug-dump-symbols -fparse-only"
 
 if [[ $# < 1 ]]; then
   echo "Usage: $0 <fortran-source> [-pgmath=<true/false>]"
```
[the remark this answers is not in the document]
LukeIreland1: What's wrong about it? I'm confused.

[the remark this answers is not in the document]
RichBarton-Arm: I mean this usage string is not right. This usage string says the interface is $1 = source, $2 = optional flag to enable pgmath testing. The new interface is "1 = source, $2 = compiler binary $3 = temp directory. So the Usage string needs updating.

We also need to think about how to re-enable the -pgmath=true functionality in the new setup. I don't think (maybe wrong) that there is a general facility in lit for passing options to the RUN line. So I think you need to add something to the config and set an env var to pass through.

test/Semantics/common.sh
```diff
@@ -10,19 +10,18 @@ function die {
   echo "$(basename $0): $*" >&2
   exit 1
 }
+if [[ $# < 1 ]]; then
+  echo "Usage: $(basename $0) <fortran-source> [<f18-executable>] <temp test dir>"
```
RichBarton-Arm: The [ ] syntax in the usage string is meant to indicate an optional argument. In the new interface, the F18 executable is non-optional so these should be removed. I think all the arguments are non-optional in the new interface.

test/Evaluate/test_folding.sh
```diff
@@ -19,26 +19,21 @@
...
 
-PATH=/usr/bin:/bin
-srcdir=$(dirname $0)
-F18CC=${F18:-../../../tools/f18/bin/f18}
-CMD="$F18CC -fdebug-dump-symbols -fparse-only"
+CMD="$2 -fdebug-dump-symbols -fparse-only"
 
 if [[ $# < 1 ]]; then
-  echo "Usage: $0 <fortran-source> [-pgmath=<true/false>]"
+  echo "Usage: $0 <fortran-source> [<f18-executable>] <temp test dir>"
```
RichBarton-Arm: Same comment as on common.sh. <f18-executable> is no longer an optional argument, so lose the [] notation.

test/Semantics/common.sh
```diff
@@ -10,19 +10,18 @@ function die {
   echo "$(basename $0): $*" >&2
   exit 1
 }
+if [[ $# < 1 ]]; then
```
RichBarton-Arm: This logic is wrong. This means that only an invocation with no arguments will cause the usage string to be printed. Given that now our interface mandates 3 arguments it would be friendlier to print the usage string if less than 3 arguments are given.

test/Evaluate/test_folding.sh
```diff
@@ -19,26 +19,21 @@
 #   - test_x is not folded (it is neither .true. nor .false.). This means the
 #     compiler could not fold the expression.
 
-PATH=/usr/bin:/bin
-srcdir=$(dirname $0)
-F18CC=${F18:-../../../tools/f18/bin/f18}
-CMD="$F18CC -fdebug-dump-symbols -fparse-only"
+CMD="$2 -fdebug-dump-symbols -fparse-only"
 
 if [[ $# < 1 ]]; then
```
RichBarton-Arm: Same comment as common.sh. The minimum number of arguments has changed to 3, so we should print usage for less than 3.

test/Preprocessing/lit.local.cfg.py
```diff
@@ -0,0 +1,12 @@
+# -*- Python -*-
+
+import os
+import platform
+import sys
+
+import lit.formats
+from lit.llvm import llvm_config
+
+# Added this lit config file to prevent lit from discovering these tests
```
RichBarton-Arm: Pedantry, but the presence of the file itself does not do this. It is the assignment to config.suffixes that does it. Change the comment to explain that.

test/Preprocessing/lit.local.cfg.py
```diff
@@ -0,0 +1,12 @@
+# -*- Python -*-
+
+import os
```
RichBarton-Arm: Do you still need these imports? (Genuine question)

test/lit.cfg.py
```diff
@@ -72,7 +64,11 @@
 tools = [ToolSubst('%flang', command=FindTool('flang'), unresolved='fatal'),
          ToolSubst('%f18', command=FindTool('f18'), unresolved='fatal'),
          ToolSubst('%f18_with_includes', command=FindTool('f18'),
-         extra_args=[flang_includes], unresolved='fatal')]
+                   extra_args=[flang_includes], unresolved='fatal')]
```
RichBarton-Arm: This change in indentation is not needed

unittests/Evaluate/CMakeLists.txt
```diff
@@ -140,11 +140,3 @@ add_test(Real real-test)
 add_test(RESHAPE reshape-test)
 add_test(ISO-binding ISO-Fortran-binding-test)
 add_test(folding folding-test)
-
```
RichBarton-Arm: This should be part of patch #1

test/lit.cfg.py
```diff
@@ -50,17 +42,18 @@
...
 # Tweak the PATH to include the tools dir.
 llvm_config.with_environment('PATH', config.flang_tools_dir, append_path=True)
 llvm_config.with_environment('PATH', config.llvm_tools_dir, append_path=True)
 
 # For builds with FIR, set path for tco and enable related tests
-if config.flang_llvm_tools_dir != "" :
-  config.available_features.add('fir')
-  if config.llvm_tools_dir != config.flang_llvm_tools_dir :
-    llvm_config.with_environment('PATH', config.flang_llvm_tools_dir, append_path=True)
+if config.flang_llvm_tools_dir != "":
```
RichBarton-Arm: Again, please don't make random formatting changes that are unrelated to the patch.

test/CMakeLists.txt
```diff
@@ -1,12 +1,30 @@
...
+
+set(FLANG_TEST_DEPENDS
+  f18
+)
+
+if (LINK_WITH_FIR)
+  list(APPEND FLANG_TEST_DEPENDS tco)
+endif()
+
+add_lit_testsuite(check-all "Running the Flang regression tests"
```
schweitzpgi: This is a bug and only works when compiling f18 standalone. LLVM already has a cmake target of "check-all" and this will conflict with it.

DavidTruby: I've fixed this already in the cmake rework PR. Check-all will still exist as a if flang is built out of tree running only flang tests, and the check-flang target will always exist and always only run flang tests.

PR #1031, opened 2020-02-27
Semantic checks for C712 through C727

psteinfeld:
I've updated the compiler and test source with references to the constraints at
the points where they were enforced and tested.  Many of these were already
implemented and required no code change.  A few constraint checks were both
implemented and tested, and I only added references to the constraint
numbers in the compiler source and tests.  Here are the things I had to
implement:

Constraint C716 states that, in a REAL constant, if both a kind-param and an
exponent letter appear, the exponent letter must be 'E'.

Constraints C715 and C719 require that a KIND value be actually implemented.

Constraint C722 requires that functions that return assumed-length character
types are external.

Constraint C726 disallows assumed lenght charater types for dummy arguments and
return types.

### DIFF
--- a/lib/Semantics/expression.cpp
+++ b/lib/Semantics/expression.cpp
@@ -500,10 +500,10 @@ MaybeExpr ExpressionAnalyzer::Analyze(const parser::RealLiteralConstant &x) {
   // Use a local message context around the real literal for better
   // provenance on any messages.
   auto restorer{GetContextualMessages().SetLocation(x.real.source)};
-  // If a kind parameter appears, it defines the kind of the literal and any
-  // letter used in an exponent part (e.g., the 'E' in "6.02214E+23")
-  // should agree.  In the absence of an explicit kind parameter, any exponent
-  // letter determines the kind.  Otherwise, defaults apply.
+  // If a kind parameter appears, it defines the kind of the literal and the
+  // letter used in an exponent part must be 'E' (e.g., the 'E' in
+  // "6.02214E+23").  In the absence of an explicit kind parameter, any
+  // exponent letter determines the kind.  Otherwise, defaults apply.
   auto &defaults{context_.defaultKinds()};
   int defaultKind{defaults.GetDefaultKind(TypeCategory::Real)};
   const char *end{x.real.source.end()};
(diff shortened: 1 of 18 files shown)


lib/Semantics/resolve-names.cpp
```diff
@@ -3271,6 +3272,13 @@ void DeclarationVisitor::Post(const parser::IntrinsicTypeSpec::Character &) {
 }
 void DeclarationVisitor::Post(const parser::CharSelector::LengthAndKind &x) {
   charInfo_.kind = EvaluateSubscriptIntExpr(x.kind);
+  std::optional<std::int64_t> intKind{ToInt64(charInfo_.kind)};
+  if (intKind &&
+      !evaluate::IsValidKindOfIntrinsicType(
+          TypeCategory::Character, *intKind)) {  // C715, C719
+    Say(currStmtSource().value(),
+        "KIND value (%jd) not valid for CHARACTER"_err_en_US, *intKind);
```
tskeith: I learned in PR #1009 that `%jd` is for `std::intmax_t`, which isn't necessarily the same as `std::int64_t`. So casting to `std::intmax_t` is safest.

lib/Semantics/tools.cpp
```diff
@@ -196,6 +196,29 @@ bool IsDummy(const Symbol &symbol) {
   }
 }
 
+bool IsStmtFunction(const Symbol &symbol) {
+  const auto *subprogram{symbol.detailsIf<SubprogramDetails>()};
+  if (subprogram && subprogram->stmtFunction()) {
+    return true;
+  }
+  return false;
```
tskeith:
```suggestion
  return subprogram && subprogram->stmtFunction();
```

lib/Semantics/expression.cpp
```diff
@@ -525,14 +525,13 @@ MaybeExpr ExpressionAnalyzer::Analyze(const parser::RealLiteralConstant &x) {
     defaultKind = *letterKind;
   }
   auto kind{AnalyzeKindParam(x.kind, defaultKind)};
-  if (letterKind && kind != *letterKind && expoLetter != 'e') {
-    Say("Explicit kind parameter on real constant disagrees with "
-        "exponent letter '%c'"_en_US,
-        expoLetter);
+  if (x.kind && letterKind && expoLetter != 'e') {  // C716
```
klausler: I didn't make this case a fatal error because it's benign.

lib/Semantics/tools.cpp
psteinfeld: Thanks, Tim.  I'll change it.

lib/Semantics/resolve-names.cpp
psteinfeld: Thanks.  I'll change it.

lib/Semantics/expression.cpp
psteinfeld: Ahh...  In that case, I think that the original code is better.  I'll change it back.

PR #1033, opened 2020-02-27
Change README to refer to LLVM_BUILD_DIR.

DavidTruby:
LLVM_INSTALL_TOOLS doesn't seem to install llvm-lit anymore. However
pointing to the cmake file in the build directory works fine, and lit
and FileCheck will be picked up correctly this way.

### DIFF
--- a/test-lit/CMakeLists.txt
+++ b/test-lit/CMakeLists.txt
@@ -14,17 +14,12 @@ set(FLANG_TEST_PARAMS
   flang_site_config=${CMAKE_CURRENT_BINARY_DIR}/lit.site.cfg.py)
 
 set(FLANG_TEST_DEPENDS
-  flang
   f18
-  llvm-lit
-  FileCheck
-  count
-  not
 )
 add_lit_testsuite(check-all "Running the Flang regression tests"
   ${CMAKE_CURRENT_BINARY_DIR}
   PARAMS ${FLANG_TEST_PARAMS}
-  DEPENDS ${FLANG_TEST_DEPENS}
+  DEPENDS ${FLANG_TEST_DEPENDS}
   )
 set_target_properties(check-all PROPERTIES FOLDER "Tests")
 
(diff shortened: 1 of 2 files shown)


DavidTruby:
I've bundled in a fix for #974 to this, as the new build instructions and the fix of those depends need to go hand in hand.

test-lit/CMakeLists.txt
```diff
@@ -14,17 +14,12 @@ set(FLANG_TEST_PARAMS
   flang_site_config=${CMAKE_CURRENT_BINARY_DIR}/lit.site.cfg.py)
 
 set(FLANG_TEST_DEPENDS
-  flang
   f18
-  llvm-lit
-  FileCheck
-  count
-  not
```
DavidTruby: We will want to add these back in once #969 lands, but right now these aren't targets that we can depend on because they're not present as targets in our own cmake call.

RichBarton-Arm:
@sscalpone  or @tskeith are you happy to merge this for David?

tskeith:
It looks like these instructions are going to conflict with the ones in PR #1035 which require an LLVM install.

DavidTruby:
Ok, I think there's a bug here (or more than one) on the LLVM side. Surely either the AddMLIR.cmake file should be in the build directory or llvm-lit should be installed when -DLLVM_INSTALL_TOOLS is on. I'm somewhat inclined to say both should be the case personally. Let me investigate why llvm-lit isn't installed when I think it should be (and am sure it used to be) first.

tskeith:
> Ok, I think there's a bug here (or more than one) on the LLVM side. Surely either the AddMLIR.cmake file should be in the build directory or llvm-lit should be installed when -DLLVM_INSTALL_UTILS is on. I'm somewhat inclined to say both should be the case personally.

I agree on both counts.

> Let me investigate why llvm-lit isn't installed when I think it should be (and am sure it used to be) first.

That sounds good. I think we should assume we should build against an LLVM install, and until this is fixed work around the missing llvm-lit by setting LLVM_EXTERNAL_LIT or manually installing llvm-lit by copying it into the install.

PR #1035, opened 2020-02-27
RFC: FIR Merge - PR2: Add Fortran IR (FIR) MLIR dialect implementation

jeanPerier:
Add the FIR library that defines and implements the MLIR dialect used in flang.  FIR is defined and documented inside FIROps.td in this commit.

The following documents can help to understand how MLIR dialects must be defined:
- [MLIR LangRef](https://github.com/llvm/llvm-project/blob/master/mlir/docs/LangRef.md)  for general information about MLIR.
- [MLIR Table-driven Operation Definition Specification](https://github.com/llvm/llvm-project/blob/master/mlir/docs/OpDefinitions.md) can help you go through `FIROps.td`, `FIROpsSupport.h`, and `FIROps.h`.
- [MLIR Attributes and Types Definition](https://github.com/llvm/llvm-project/blob/master/mlir/docs/DefiningAttributesAndTypes.md) can help you go through `FIRAttr.h` and `FIRType.h`.

This patch adds the tool `tco` that processes FIR input files and drives transformations passes. The tco tool is used for testing. In this patch, the tco tool is to implement tests for parsing, printing, verifying, and round-tripping FIR.

Note:
This commit it a feature-based split of the work done in the FIR experimental branch. The related work log can be found in the commits between https://github.com/schweitzpgi/f18/commit/742edde572bd74d77cf7d447132ccf0949187fce and https://github.com/schweitzpgi/f18/commit/2ff55242126d86061f4fed9ef7b59d3636b5fd0b.

RFC: http://lists.llvm.org/pipermail/flang-dev/2020-January/000162.html


include/flang/Optimizer/Dialect/FIROps.td
```diff
@@ -0,0 +1,2779 @@
...
+  }]>;
+
+def fir_OneResultOpBuilder : OpBuilder<
+  "Builder *, OperationState &result, Type resultType,"
+  "ArrayRef<Value> operands, ArrayRef<NamedAttribute> attributes = {}",
+  [{
+    if (resultType)
+      result.addTypes(resultType);
+    result.addOperands(operands);
+    for (auto namedAttr : attributes)
```
River707: I'm pretty sure you can just do `result.addAttributes(attributes);`

include/flang/Optimizer/Dialect/FIROps.td
```diff
@@ -0,0 +1,2779 @@
...
+    10, and each element has type `!fir.type<T>`.
+  }];
+
+  let verifier = [{
+    if (auto ref = getType().dyn_cast<fir::ReferenceType>())
+      return emitOpError("undefined values of type !fir.ref not allowed");
+    return mlir::success();
+  }];
+
+  let parser = [{
```
River707: Use assemblyFormat?

include/flang/Optimizer/Dialect/FIROps.td
```diff
@@ -0,0 +1,2779 @@
...
+class fir_SwitchTerminatorOp<string mnemonic, list<OpTrait> traits = []> :
+    fir_Op<mnemonic, !listconcat(traits, [Terminator])> {
+  let arguments = (ins Variadic<AnyType>:$args);
+
+  let results = (outs);
+
+  let builders = [OpBuilder<
+    "Builder *, OperationState &result, Value selector,"
+    "ArrayRef<Value> properOperands, ArrayRef<Block *> destinations,"
+    "ArrayRef<ArrayRef<Value>> operands = {},"
```
River707: Generally `ValueRange` is a better parameter than `ArrayRef<Value>` to use for builders, as it can be constructed from operand/result/block/etc. ranges without constructing a vector.

include/flang/Optimizer/Dialect/FIROps.td
```diff
@@ -0,0 +1,2779 @@
...
+    "ArrayRef<NamedAttribute> attributes = {}",
+    [{
+      result.addOperands(selector);
+      result.addOperands(properOperands);
+      for (auto kvp : llvm::zip(destinations, operands)) {
+        result.addSuccessor(std::get<0>(kvp), std::get<1>(kvp));
+      }
+      for (auto namedAttr : attributes) {
+        result.addAttribute(namedAttr.first, namedAttr.second);
```
River707: Same as above, `result.addAttributes(attributes)` ?

include/flang/Optimizer/Dialect/FIROps.td
```diff
@@ -0,0 +1,2779 @@
...
+  );
+}
+
+class fir_AllocatableOp<string mnemonic, list<OpTrait> traits =[]> :
+    fir_AllocatableBaseOp<mnemonic, !listconcat([NoSideEffect], traits)>,
+    fir_TwoBuilders<fir_AllocateOpBuilder, fir_NamedAllocateOpBuilder>,
+    Arguments<(ins TypeAttr:$in_type, Variadic<AnyIntegerType>:$args)> {
+
+  let parser = [{
+    mlir::Type intype;
```
River707: Could you please move all of these large code blocks to a .cpp source file? It is much easier to reason about/format/refactor code when it is in a .cpp file. It also makes the .td file much cleaner.

include/flang/Optimizer/Dialect/FIROps.td
```diff
@@ -0,0 +1,2779 @@
...
+    unboxed memory reference. Returns a memory reference. When computing the
+    coordinate of an array element, the rank of the array must be known and
+    the number of indexing expressions must equal the rank of the array.
+
+    This operation will apply the access map from a boxed value implicitly.
+
+    Unlike LLVM's GEP instruction, one cannot stride over the outermost
+    reference; therefore, the leading 0 index must be omitted.
+
+      %i = ... : index
```
River707: If you use ```mlir code blocks, you get syntax highlighting.

include/flang/Optimizer/Dialect/FIROps.td
```diff
@@ -0,0 +1,2779 @@
...
+
+  let assemblyFormat = [{
+    operands attr-dict `:` functional-type(operands, results)
+  }];
+
+  let verifier = [{
+    // Recovering a LEN type parameter only makes sense from a boxed value
+    for (auto co : coor())
+      if (auto *s = co.getDefiningOp())
+        if (dyn_cast_or_null<LenParamIndexOp>(s)) {
```
River707: You've already checked `s` for null, so you just need dyn_cast. Alternatively, you could merge the two if checks and continue using dyn_cast_or_null.

include/flang/Optimizer/Dialect/FIROps.td
```diff
@@ -0,0 +1,2779 @@
...
+
+  let results = (outs fir_DimsType);
+
+  let assemblyFormat = [{
+    operands attr-dict `:` functional-type(operands, results)
+  }];
+
+  let verifier = [{
+    auto size = triples().size();
+    if (size < 1 && size <= 16 * 3)
```
River707: I don't quite parse this, it seems like `rhs` would always be true if `lhs` is true.

include/flang/Optimizer/Dialect/FIROps.td
```diff
@@ -0,0 +1,2779 @@
...
+              "int64_t lowerBound, int64_t upperBound, int64_t step = 1">,
+    OpBuilder<"mlir::Builder *builder, OperationState &result,"
+              "mlir::Value lowerBound, mlir::Value upperBound,"
+              "ArrayRef<mlir::Value> step = {}">
+  ];
+
+  let parser = "return parseLoopOp(parser, result);";
+
+  let printer = [{
+    p << getOperationName() << ' ' << getI
```
River707: nit:  Optional parse methods are a bit weird. I've found that it makes it much easier to read if you wrap them with `succeeded`/`failed`.

River707: Any reason not to have this attribute in the arguments list?

River707: Seems like this can use the `assemblyFormat` field.

include/flang/Optimizer/Dialect/FIRType.h
```diff
@@ -0,0 +1,357 @@
...
+    : public mlir::Type::TypeBase<BoxType, mlir::Type, detail::BoxTypeStorage> {
+public:
+  using Base::Base;
+  static BoxType get(mlir::Type eleTy, mlir::AffineMapAttr map = {});
+  static bool kindof(unsigned kind) { return kind == TypeKind::FIR_BOX; }
+  mlir::Type getEleTy() const;
+  mlir::AffineMapAttr getLayoutMap() const;
+
+  static mlir::LogicalResult
+  verifyConstructionInvariants(const mlir::AttributeStorage *, mlir::Type eleTy,
```
River707: The first param of these methods should be an `mlir::Location`.

lib/Optimizer/Dialect/FIRAttr.cpp
```diff
@@ -0,0 +1,236 @@
...
+  llvm::StringRef attrName;
+  if (parser.parseKeyword(&attrName)) {
+    parser.emitError(loc, "expected an attribute name");
+    return {};
+  }
+
+  if (attrName == ExactTypeAttr::getAttrName()) {
+    mlir::Type type;
+    if (parser.parseLess() || parser.parseType(type) || parser.parseGreater())
+      parser.emitError(loc, "expected a type");
```
River707: Seems like you are missing some returns for these error cases. Lack of coverage?

lib/Optimizer/Dialect/FIRAttr.cpp
```diff
@@ -0,0 +1,236 @@
...
+    return LowerBoundAttr::get(dialect->getContext());
+  if (attrName == UpperBoundAttr::getAttrName())
+    return UpperBoundAttr::get(dialect->getContext());
+  if (attrName == ClosedIntervalAttr::getAttrName())
+    return ClosedIntervalAttr::get(dialect->getContext());
+  if (attrName == RealAttr::getAttrName())
+    return parseFirRealAttr(dialect, parser, type);
+
+  llvm::Twine msg{"unknown FIR attribute: "};
+  parser.emitError(loc, msg.concat(attrName));
```
River707:
```c++
parser.emitError("unknown FIR attribute: ") << attrName;
```

?

tools/tco/tco.cpp
```diff
@@ -0,0 +1,119 @@
...
+
+cl::opt<std::string> outputFilename("o", cl::desc("Specify output filename"),
+                                    cl::value_desc("filename"), cl::init("-"));
+
+cl::opt<bool> emitFir("emit-fir", cl::desc("Parse and pretty-print the input"),
+                      cl::init(false));
+
+void printModuleBody(mlir::ModuleOp mod, raw_ostream &output) {
+  // don't output the terminator bogo-op
+  auto e{--mod.end()};
```
River707: Note: Block has a utility for this `Block::without_terminator`.

schweitzpgi:
> Hi Jean, I hope you don't mind but I left some comments. I still haven't been able to look through all of it though.

Thanks for the helpful tips, River. Much appreciated.

include/flang/Optimizer/Dialect/FIRType.h
schweitzpgi: For some reason, a recent rebase to the tip of mono broke all these `verifyConstructionInvariants` functions. The parameter types were changed to sneak these past the compilers.

include/flang/Optimizer/Dialect/FIROps.td
schweitzpgi: Does `assemblyFormat` support parsing Regions?

include/flang/Optimizer/Dialect/FIROps.td
schweitzpgi: Sleep deprivation. Fixed.

lib/Optimizer/Dialect/FIRAttr.cpp
schweitzpgi: Fixed returns. We need to add negative tests.

include/flang/Optimizer/Dialect/FIROps.td
schweitzpgi: Sorry. I'm not sure which attribute you're referring to here.

tools/tco/tco.cpp
schweitzpgi: Thanks. I was looking for that. :)

include/flang/Optimizer/Dialect/FIROps.td
River707: Not yet, but it is on the list of stuff to add.

River707: `constant`. I don't see where it is added to an arguments list. There are a few other ops in a similar situation. Adding them to the argument list would also allow for using the assemblyFormat field.

include/flang/Optimizer/Dialect/FIRType.h
River707: Yeah, that was my fault. The original parameter was an `Optional<mlir::Location`, but I changed it to `mlir::Location`. A raw attribute storage is passed instead of a proper `mlir::Location` to avoid cyclic includes in the base storage uniquer class. I really need to add a static assert of some kind to ensure that clients still use Location.

include/flang/Optimizer/Dialect/FIROps.td
schweitzpgi: Right. I'm wondering if such a NFC might be rolled up in another merge though to reduce any noise. (This is already a large PR.)

include/flang/Optimizer/Dialect/FIROps.td
schweitzpgi: (Obvious now. Browser window goof?)

This one is tricky (and related to the conversation we were having on discourse). We want to be able to read in floating-point constants which don't correspond to the standard dialect supported semantics. I'll try mucking around with it a bit more.

include/flang/Optimizer/Dialect/FIRType.h
schweitzpgi: Ah. OK, let me try that...

tskeith:
Can you update the build instructions? My LLVM build doesn't have AddMLIR.cmake but the install does. But the install doesn't have llvm-lit.

tskeith:
The tests fail for me with: `tco: command not found`

test-lit/Fir/fir-ops.fir
```diff
@@ -0,0 +1,458 @@
...
+
+// CHECK-LABEL: func @nop()
+func @nop()
+
+// CHECK-LABEL: func @get_func() -> (() -> ())
+func @get_func() -> (() -> ())
+
+// CHECK-LABEL: @instructions
+func @instructions() {
+  // CHECK: %[[A0:.*]] = fir.alloca !fir.array<10xi32>
```
River707: Not sure if you all are aware, but there is a script to help generate CHECK lines:
https://github.com/llvm/llvm-project/blob/master/mlir/utils/generate-test-checks.py

include/flang/Optimizer/Support/KindMapping.h
```diff
@@ -0,0 +1,89 @@
...
+/// 'Half', 'Float', 'Double', 'X86_FP80', or 'FP128' (for 'r' and 'c').
+///
+/// If LLVM adds support for new floating-point types, the final list should be
+/// extended.
+class KindMapping {
+public:
+  using KindTy = unsigned;
+  using Bitsize = unsigned;
+  using LLVMTypeID = llvm::Type::TypeID;
+  using MatchResult = llvm::Optional<bool>;
```
River707: What is this supposed to be used for? Is this just for parsing? Would it be better to do something like `mlir::ParseResult`?

include/flang/Optimizer/Support/KindMapping.h
```diff
@@ -0,0 +1,89 @@
...
+
+  /// Get the float semantics of !fir.real<kind>
+  const llvm::fltSemantics &getFloatSemantics(KindTy kind);
+
+private:
+  MatchResult badMapString(llvm::Twine const &ptr);
+  MatchResult parse(llvm::StringRef kindMap);
+
+  mlir::MLIRContext *context;
+  std::map<char, std::map<KindTy, Bitsize>> intMap;
```
River707: Do these need to be sorted? Would it be better to use an unordered container instead?

include/flang/Optimizer/Dialect/FIRType.h
```diff
@@ -0,0 +1,357 @@
...
+  static bool kindof(unsigned kind) { return kind == TypeKind::FIR_HEAP; }
+
+  mlir::Type getEleTy() const;
+
+  static mlir::LogicalResult
+  verifyConstructionInvariants(const mlir::AttributeStorage *,
+                               mlir::Type eleTy);
+};
+
+class LenType
```
River707: Is there a reason why some of these types have comments and others don't?

lib/Optimizer/Dialect/FIRType.cpp
```diff
@@ -0,0 +1,1288 @@
...
+
+// Field
+
+FieldType fir::FieldType::get(mlir::MLIRContext *ctxt, KindTy) {
+  return Base::get(ctxt, FIR_FIELD, 0);
+}
+
+// Len
+
+LenType fir::LenType::get(mlir::MLIRContext *ctxt, KindTy) {
```
River707: Why does this have a KindTy parameter if it is never used?

lib/Optimizer/Dialect/FIRType.cpp
```diff
@@ -0,0 +1,1288 @@
...
+    const SequenceType::Shape &shape, mlir::Type eleTy,
+    mlir::AffineMapAttr map) {
+  // DIMENSION attribute can only be applied to an intrinsic or record type
+  if (eleTy.isa<BoxType>() || eleTy.isa<BoxCharType>() ||
+      eleTy.isa<BoxProcType>() || eleTy.isa<DimsType>() ||
+      eleTy.isa<FieldType>() || eleTy.isa<LenType>() || eleTy.isa<HeapType>() ||
+      eleTy.isa<PointerType>() || eleTy.isa<ReferenceType>() ||
+      eleTy.isa<TypeDescType>() || eleTy.isa<SequenceType>()) {
+    auto context = attributeStorage->getDialect().getContext();
+    mlir::emitError(mlir::UnknownLoc::get(context),
```
River707: emitError returns a diagnostic, which converts to `mlir::failure` automatically, so you can just return that directly.

lib/Optimizer/Dialect/FIRType.cpp
```diff
@@ -0,0 +1,1288 @@
...
+  }
+  return mlir::success();
+}
+
+// compare if two shapes are equivalent
+bool fir::operator==(const SequenceType::Shape &sh_1,
+                     const SequenceType::Shape &sh_2) {
+  if (sh_1.size() != sh_2.size())
+    return false;
+  for (std::size_t i = 0, e{sh_1.size()}; i != e; ++i)
```
River707: Why the explicit std::size_t? Also why not simply `e = sh_1.size()`?

lib/Optimizer/Dialect/FIRType.cpp
```diff
@@ -0,0 +1,1288 @@
...
+  return Base::get(ctxt, FIR_FIELD, 0);
+}
+
+// Len
+
+LenType fir::LenType::get(mlir::MLIRContext *ctxt, KindTy) {
+  return Base::get(ctxt, FIR_LEN, 0);
+}
+
+// LOGICAL
```
River707: Shameless plug, but I've found the following comment block style to make chunking files much more readable:
https://github.com/llvm/llvm-project/blob/c3595d1069277b4ab0df49d7139b6f1bbc94f21c/mlir/lib/IR/Block.cpp#L15

lib/Optimizer/Dialect/FIROps.cpp
```diff
@@ -0,0 +1,573 @@
...
+mlir::FuncOp createFuncOp(mlir::Location loc, mlir::ModuleOp module,
+                          StringRef name, mlir::FunctionType type,
+                          llvm::ArrayRef<mlir::NamedAttribute> attrs) {
+  if (auto f = module.lookupSymbol<mlir::FuncOp>(name))
+    return f;
+  mlir::OpBuilder modBuilder(module.getBodyRegion());
+  return modBuilder.create<mlir::FuncOp>(loc, name, type, attrs);
+}
+
+GlobalOp createGlobalOp(mlir::Location loc, mlir::ModuleOp module,
```
River707: LLVM style is to explicitly qualify functions:
https://llvm.org/docs/CodingStandards.html#use-namespace-qualifiers-to-implement-previously-declared-functions

LLVM style is also to have `using namespace`, and to limit anonymous namespaces to just classes:
https://llvm.org/docs/CodingStandards.html#anonymous-namespaces

lib/Optimizer/Dialect/FIROps.cpp
```diff
@@ -0,0 +1,573 @@
...
+    return mlir::success();
+  }
+  return mlir::failure();
+}
+
+// LoopOp
+
+void LoopOp::build(mlir::Builder *builder, mlir::OperationState &result,
+                   int64_t lowerBound, int64_t upperBound, int64_t step) {
+  assert(false && "not implemented");
```
River707: Can we move this to a later patch when it is needed?

lib/Optimizer/Dialect/FIROps.cpp
```diff
@@ -0,0 +1,573 @@
...
+    return r.getEleTy();
+  if (auto r = ref.dyn_cast_or_null<PointerType>())
+    return r.getEleTy();
+  if (auto r = ref.dyn_cast_or_null<HeapType>())
+    return r.getEleTy();
+  return {};
+}
+
+mlir::ParseResult LoadOp::getElementOf(mlir::Type &ele, mlir::Type ref) {
+  if (mlir::Type r = elementTypeOf(ref)) {
```
River707:
```
if ((ele = ...))
  return ...
return ...
```
?

lib/Optimizer/Dialect/FIROps.cpp
```diff
@@ -0,0 +1,573 @@
...
+void GlobalOp::appendInitialValue(mlir::Operation *op) {
+  auto &block = getBlock();
+  block.getOperations().insert(block.end(), op);
+}
+
+// LoadOp
+
+/// Get the element type of a reference like type; otherwise null
+mlir::Type elementTypeOf(mlir::Type ref) {
+  if (auto r = ref.dyn_cast_or_null<ReferenceType>())
```
River707: You could use mlir::TypeSwitch to simplify some of this:
```
TypeSwitch<mlir::Type>(ref)
  .Case<HeapType, PointerType, ReferenceType>([](auto type) { return type.getEleType(); })
  .Default([](mlir::Type) { return {}; });
```

lib/Optimizer/Dialect/FIROps.cpp
```diff
@@ -0,0 +1,573 @@
...
+                          mlir::TypeAttr inty) {
+  result.addAttribute("in_type", inty);
+  result.addTypes(TypeDescType::get(inty.getValue()));
+}
+
+// GlobalOp
+
+void GlobalOp::appendInitialValue(mlir::Operation *op) {
+  auto &block = getBlock();
+  block.getOperations().insert(block.end(), op);
```
River707: .getOperations().push_back(op)
?

lib/Optimizer/Dialect/FIROps.cpp
```diff
@@ -0,0 +1,573 @@
...
+void printCallOp(mlir::OpAsmPrinter &p, fir::CallOp &op) {
+  auto callee = op.callee();
+  bool isDirect = callee.hasValue();
+  p << op.getOperationName() << ' ';
+  if (isDirect)
+    p << callee.getValue();
+  else
+    p << op.getOperand(0);
+  p << '(';
+  p.printOperands(llvm::drop_begin(op.getOperands(), isDirect ? 0 : 1));
```
River707: You can do `p << ... << op.getOperands().drop_begin(...)`

lib/Optimizer/Dialect/FIROps.cpp
```diff
@@ -0,0 +1,573 @@
...
+      intype.dyn_cast<PointerType>() || intype.dyn_cast<FunctionType>())
+    return {};
+  return HeapType::get(intype);
+}
+
+// BoxDimsOp
+
+/// Get the result types packed in a tuple tuple
+mlir::Type BoxDimsOp::getTupleType() {
+  llvm::SmallVector<mlir::Type, 3> triple{
```
River707: nit: std::array<mlir::Type, 3> ?

lib/Optimizer/Dialect/FIROps.cpp
```diff
@@ -0,0 +1,573 @@
...
+}
+
+mlir::Type AllocMemOp::getRefTy(mlir::Type ty) { return HeapType::get(ty); }
+
+/// Create a legal heap reference as return type
+mlir::Type AllocMemOp::wrapResultType(mlir::Type intype) {
+  // Fortran semantics: C852 an entity cannot be both ALLOCATABLE and POINTER
+  // 8.5.3 note 1 prohibits ALLOCATABLE procedures as well
+  // FIR semantics: one may not allocate a memory reference value
+  if (intype.dyn_cast<ReferenceType>() || intype.dyn_cast<HeapType>() ||
```
River707: Please use isa if you don't need the result.

lib/Optimizer/Dialect/FIROps.cpp
```diff
@@ -0,0 +1,573 @@
...
+// AllocaOp
+
+mlir::Type AllocaOp::getAllocatedType() {
+  return getType().cast<ReferenceType>().getEleTy();
+}
+
+/// Create a legal memory reference as return type
+mlir::Type AllocaOp::wrapResultType(mlir::Type intype) {
+  // FIR semantics: memory references to memory references are disallowed
+  if (intype.dyn_cast<ReferenceType>())
```
River707: Same here.

lib/Optimizer/Dialect/FIROps.cpp
```diff
@@ -0,0 +1,573 @@
...
+  if (auto st = inType.dyn_cast<fir::SequenceType>()) {
+    auto shape = st.getShape();
+    if (shape.size() == 0)
+      return true;
+    for (auto ext : shape)
+      if (ext < 0)
+        return true;
+  } else if (auto rt = inType.dyn_cast<fir::RecordType>()) {
+    // don't recurse if we're already visiting this one
+    for (auto name : visited)
```
River707: llvm::is_contained(...)?

lib/Optimizer/Dialect/FIRAttr.cpp
```diff
@@ -0,0 +1,236 @@
...
+    os << fir::LowerBoundAttr::getAttrName();
+  } else if (attr.dyn_cast_or_null<fir::UpperBoundAttr>()) {
+    os << fir::UpperBoundAttr::getAttrName();
+  } else if (auto a = attr.dyn_cast_or_null<fir::RealAttr>()) {
+    os << fir::RealAttr::getAttrName() << '<' << a.getFKind() << ", i x";
+    llvm::SmallString<40> ss;
+    a.getValue().bitcastToAPInt().toStringUnsigned(ss, 16);
+    os << ss << '>';
+  } else {
+    assert(false && "attribute pretty-printer is not implemented");
```
River707: nit: llvm_unreachable?

jeanPerier:
> Can you update the build instructions? My LLVM build doesn't have AddMLIR.cmake but the install does. But the install doesn't have llvm-lit.

> The tests fail for me with: tco: command not found

Thanks for testing, I have updated the README and also changed the cmake files so that `tco` is always built when `LINK_WITH_FIR=On` is set (which is default).

test-lit/Fir/fir-ops.fir
schweitzpgi: Thanks. I'm making a note of this.

include/flang/Optimizer/Support/KindMapping.h
schweitzpgi: Right. This code doesn't use much mlir support code. It looks like we could easily use `mlir::OptionalParseResult` here though.

include/flang/Optimizer/Support/KindMapping.h
schweitzpgi: I believe fast lookup is the only property really sought.

include/flang/Optimizer/Dialect/FIRType.h
schweitzpgi: Thanks. I'll document these.

lib/Optimizer/Dialect/FIRType.cpp
schweitzpgi: As I recall, there was some issue I had with uniquing these singleton types. I used a dummy KindTy to work around the issue at the time. We should investigate if that problem was fixed.

lib/Optimizer/Dialect/FIROps.cpp
schweitzpgi: Well, it is a triple. :)

lib/Optimizer/Dialect/FIROps.cpp
schweitzpgi: Tried this, but the compiler asks if I meant `drop_back` instead of `drop_begin`.

River707: Ahh, sorry I forgot that the naming is different: drop_front:
https://github.com/llvm/llvm-project/blob/0b6abe428164c38c7dec6c539e454cfd671067a4/mlir/include/mlir/Support/STLExtras.h#L255

Most ranges in MLIR use the above as a base, so most of the ArrayRef functionality is available automatically.

lib/Optimizer/Dialect/FIROps.cpp
schweitzpgi: Ongoing activity.

lib/Optimizer/Dialect/FIRType.cpp
```diff
@@ -0,0 +1,1294 @@
...
+
+  static unsigned hashKey(const KeyTy &key) { return llvm::hash_combine(key); }
+
+  bool operator==(const KeyTy &key) const { return key == getElementType(); }
+
+  static BoxProcTypeStorage *construct(mlir::TypeStorageAllocator &allocator,
+                                       mlir::Type eleTy) {
+    assert(eleTy && "element type is null");
+    auto *storage = allocator.allocate<BoxProcTypeStorage>();
+    return new (storage) BoxProcTypeStorage{eleTy};
```
DavidTruby: Suggestion: 
`return std::allocator_traits<decltype(allocator)>::construct<BoxProcTypeStorage>(allocator, eleTy);`

include/fir/.clang-format
```diff
@@ -1,2 +0,0 @@
-BasedOnStyle: LLVM
-AlwaysBreakTemplateDeclarations: Yes
```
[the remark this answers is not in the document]
jeanPerier: No, I had added all the FIR related .clang-format in the previous PR. But in the meantime, this directory was deleted and replaced by `include/flang/Optimizer/Dialect`.

include/flang/Optimizer/Dialect/FIROps.h
```diff
@@ -0,0 +1,76 @@
...
+  AlwaysTrue,
+  // Number of predicates.
+  NumPredicates
+};
+
+void buildCmpFOp(mlir::Builder *builder, mlir::OperationState &result,
+                 CmpFPredicate predicate, mlir::Value lhs, mlir::Value rhs);
+void buildCmpCOp(mlir::Builder *builder, mlir::OperationState &result,
+                 CmpFPredicate predicate, mlir::Value lhs, mlir::Value rhs);
+unsigned getCaseArgumentOffset(ArrayRef<mlir::Attribute> cases, unsigned dest);
```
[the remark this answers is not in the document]
jeanPerier: It came from `using namespace mlir;` This using directive is required because of the generated `FIROps.h.inc` that does not use llvm and mlir namespace qualifier. I have qualified the `ArrayRef` you have mentioned and lowered the using directive to only impact the `FIROps.h.inc`  include.

include/flang/Optimizer/Dialect/FIROps.h
```diff
@@ -0,0 +1,76 @@
...
+
+#ifndef OPTIMIZER_DIALECT_FIROPS_H
+#define OPTIMIZER_DIALECT_FIROPS_H
+
+#include "mlir/IR/Builders.h"
+#include "mlir/IR/OpDefinition.h"
+#include "mlir/IR/OpImplementation.h"
+#include "mlir/IR/SymbolTable.h"
+#include "llvm/ADT/ArrayRef.h"
+#include "llvm/ADT/StringRef.h"
```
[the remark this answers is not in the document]
jeanPerier: Yes, they are used in the generated `FIROps.h.inc` that is included here and does not handle dependencies on its own.

include/flang/Optimizer/Dialect/FIROpsSupport.h
```diff
@@ -0,0 +1,73 @@
...
+  return false;
+}
+
+/// Get or create a FuncOp in a module.
+///
+/// If `module` already contains FuncOp `name`, it is returned. Otherwise, a new
+/// FuncOp is created, and that new FuncOp is returned.
+mlir::FuncOp createFuncOp(mlir::Location loc, mlir::ModuleOp module,
+                          llvm::StringRef name, mlir::FunctionType type,
+                          llvm::ArrayRef<mlir::NamedAttribute> attrs = {});
```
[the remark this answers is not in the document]
jeanPerier: They are accessible in both `FIROPs.h` and `mlir/Dialect/StandardOps/IR/Ops.h`. `StringRef` and `ArrayRef` are very common in all FIR/MLIR sources and MLIR also never re-include these headers at every level (see for instance https://github.com/llvm/llvm-project/blob/69d757c0e8ffc5b49fda10df38e470a56d616ef4/mlir/include/mlir/Dialect/StandardOps/IR/Ops.h#L256 in the header I just mentioned.)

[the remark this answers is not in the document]
schweitzpgi: Yes, MLIR seems to follow the LLVM coding standard with respect to redundantly including header files.

https://llvm.org/docs/CodingStandards.html#minimal-list-of-includes

> you can include them either directly or indirectly through another header file.

include/flang/Optimizer/Dialect/FIRType.h
```diff
@@ -0,0 +1,397 @@
...
+
+// Intrinsic types
+
+/// Model of the Fortran CHARACTER intrinsic type, including the KIND type
+/// parameter. The model does not include a LEN type parameter. A CharacterType
+/// is thus the type of a single character value.
+class CharacterType
+    : public mlir::Type::TypeBase<CharacterType, mlir::Type,
+                                  detail::CharacterTypeStorage>,
+      public IntrinsicTypeMixin<CharacterType, TypeKind::FIR_CHARACTER> {
```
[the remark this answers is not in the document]
jeanPerier: No, the whole point `IntrinsicTypeMixin` is to share the implement of functions `kindof` and `getId`, these functions need to be public.

include/flang/Optimizer/Dialect/FIROps.h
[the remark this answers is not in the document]
schweitzpgi: MLIR code often has implicit using on the `llvm` namespace. It's pretty ubiquitous, in fact.

include/flang/Optimizer/Dialect/FIRType.h
```diff
@@ -0,0 +1,397 @@
...
+bool isa_std_type(mlir::Type t);
+
+/// Is `t` any of the FIR dialect or Standard dialect types?
+bool isa_fir_or_std_type(mlir::Type t);
+
+/// Is `t` a FIR dialect type that implies a memory (de)reference?
+bool isaMemref(mlir::Type t);
+
+/// Is `t` a FIR dialect aggregate type?
+bool isanAggregate(mlir::Type t);
```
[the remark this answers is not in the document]
schweitzpgi: `isa_foo` is one of those odd cases in LLVM that does use underscores. e.g., `isa_or_null` and `isa_and_nonnull`.  It's not unique. There are some others.

That said, we can flip a coin and make these cases consistent here.

include/flang/Optimizer/Dialect/FIRType.h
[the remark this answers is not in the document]
DavidTruby: Yes, I see, I don't think private inheritance quite does what I thought it did!

include/flang/Optimizer/Support/KindMapping.h
```diff
@@ -0,0 +1,90 @@
...
+  /// Get the float semantics of !fir.real<kind>
+  const llvm::fltSemantics &getFloatSemantics(KindTy kind);
+
+private:
+  MatchResult badMapString(const llvm::Twine &ptr);
+  MatchResult parse(llvm::StringRef kindMap);
+
+  mlir::MLIRContext *context;
+  std::unordered_map<char, std::unordered_map<KindTy, Bitsize>> intMap;
+  std::unordered_map<char, std::unordered_map<KindTy, LLVMTypeID>> floatMap;
```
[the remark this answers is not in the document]
schweitzpgi: Thanks, yes. It was already on my list to replace it again.

lib/Optimizer/CMakeLists.txt
```diff
@@ -0,0 +1,5 @@
+# Sources generated by tablegen have unused parameters.
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wno-unused-parameter")
```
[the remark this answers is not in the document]
schweitzpgi: We rebase to the tip of LLVM and MLIR very frequently, so a lot of header file changes come and go. We need to be able to tolerate those changes without breaking our builds (-Werror), and unused parameters warnings do crop up fairly often.

include/flang/Optimizer/Dialect/CMakeLists.txt
```diff
@@ -0,0 +1,4 @@
+set(LLVM_TARGET_DEFINITIONS FIROps.td)
+mlir_tablegen(FIROps.h.inc -gen-op-decls)
+mlir_tablegen(FIROps.cpp.inc -gen-op-defs)
+add_public_tablegen_target(FIROpsIncGen)
```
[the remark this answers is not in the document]
schweitzpgi: Thank you for the suggestion. It does not work in my environment.

include/flang/Optimizer/Dialect/FIROps.td
```diff
@@ -0,0 +1,2844 @@
...
+  let builders = [
+    OpBuilder<"mlir::Builder *builder, OperationState &result,"
+              "mlir::Value lowerBound, mlir::Value upperBound,"
+              "ValueRange step = {}, ArrayRef<NamedAttribute> attributes = {}">
+  ];
+
+  let parser = "return parseLoopOp(parser, result);";
+
+  let printer = [{
+    p << getOperationName()
```
kiranchandramohan: Isn't LLVM's view that it will not reassociate floating point without the reassoc attribute? Does MLIR have any difference in treatment?

lib/Optimizer/Dialect/FIRType.cpp
```diff
@@ -0,0 +1,1294 @@
...
+bool isa_std_type(mlir::Type t) {
+  return inbounds(t.getKind(), mlir::Type::FIRST_STANDARD_TYPE,
+                  mlir::Type::LAST_STANDARD_TYPE);
+}
+
+bool isa_fir_or_std_type(mlir::Type t) {
+  return isa_fir_type(t) || isa_std_type(t);
+}
+
+bool isa_memref(mlir::Type t) {
```
kiranchandramohan: Is this isa_fir_memref?

lib/Optimizer/CMakeLists.txt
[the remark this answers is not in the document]
DavidTruby: Oh yes I see, my idea wouldn't work for files that include files from FIR, is that right?

include/flang/Optimizer/Dialect/CMakeLists.txt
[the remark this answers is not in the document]
DavidTruby: Having looked at it it looks like that function assumes you are inside the MLIR repo so that would be why it doesn't work.

lib/Optimizer/Dialect/FIROps.cpp
```diff
@@ -0,0 +1,571 @@
...
+/// or contains an abstract sequence type
+static bool verifyInType(mlir::Type inType,
+                         llvm::SmallVectorImpl<llvm::StringRef> &visited) {
+  if (auto st = inType.dyn_cast<fir::SequenceType>()) {
+    auto shape = st.getShape();
+    if (shape.size() == 0)
+      return true;
+    for (auto ext : shape)
+      if (ext < 0)
+        return true;
```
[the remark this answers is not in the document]
jeanPerier: Thanks for the suggestion, I find the original code quit straightforward and more readable here.

lib/Optimizer/Dialect/FIROps.cpp
```diff
@@ -0,0 +1,571 @@
...
+  return mlir::success();
+}
+
+fir::LoopOp fir::getForInductionVarOwner(mlir::Value val) {
+  auto ivArg = val.dyn_cast<mlir::BlockArgument>();
+  if (!ivArg)
+    return {};
+  assert(ivArg.getOwner() && "unlinked block argument");
+  auto *containingInst = ivArg.getOwner()->getParentOp();
+  return dyn_cast_or_null<fir::LoopOp>(containingInst);
```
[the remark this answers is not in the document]
jeanPerier: `fir::LoopOp` class is not a pointer, it is a class, but dyn_cast_or_null<T> can take T* arguments and then return T*.
See: https://github.com/llvm/llvm-project/blob/1bd6123b781120c9190b9ba58b900cdcb718cdd1/llvm/include/llvm/Support/Casting.h#L361

lib/Optimizer/Dialect/FIRType.cpp
jeanPerier: Thanks for the suggestion, I find the two liner less intimidating.

lib/Optimizer/Support/KindMapping.cpp
```diff
@@ -0,0 +1,240 @@
...
+
+static MatchResult parseColon(const char *&ptr) {
+  return parseSingleChar<':'>(ptr);
+}
+
+static MatchResult parseComma(const char *&ptr) {
+  return parseSingleChar<','>(ptr);
+}
+
+static MatchResult parseInt(unsigned &result, const char *&ptr) {
```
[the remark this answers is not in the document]
schweitzpgi: It's an ongoing issue that we've written code months before the evolving MLIR code adds or exposes similar functionality.

lib/Optimizer/Dialect/FIROps.cpp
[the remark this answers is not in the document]
DavidTruby: That's why I'm confused, the return statement has type LoopOp* but the function has return type LoopOp. I can't make that compile in my head.

lib/Optimizer/Support/KindMapping.cpp
```diff
@@ -0,0 +1,240 @@
...
+    return llvm::APFloat::IEEEhalf();
+  case LLVMTypeID::FloatTyID:
+    return llvm::APFloat::IEEEsingle();
+  case LLVMTypeID::DoubleTyID:
+    return llvm::APFloat::IEEEdouble();
+  case LLVMTypeID::X86_FP80TyID:
+    return llvm::APFloat::x87DoubleExtended();
+  case LLVMTypeID::FP128TyID:
+    return llvm::APFloat::IEEEquad();
+  case LLVMTypeID::PPC_FP128TyID:
```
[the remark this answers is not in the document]
schweitzpgi: Thanks. It should be in parseTypeID. It's not part of the defaults because the front-end does not support it.

lib/Optimizer/Dialect/FIROps.cpp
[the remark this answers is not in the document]
schweitzpgi: The LoopOp is a specific wrapper that holds a pointer to the Operation. The cast is promoting the Operation to that wrapper or returning an empty wrapper.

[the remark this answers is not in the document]
DavidTruby: Sorry to keep questioning this, I'm sure I must be misunderstanding something. Does LoopOp have a constructor that takes a pointer to another LoopOp as a parameter then?

include/flang/Optimizer/Dialect/CMakeLists.txt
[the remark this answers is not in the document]
schweitzpgi: Darn. Well, that would explain it.

lib/Optimizer/Dialect/FIROps.cpp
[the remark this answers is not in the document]
jeanPerier: Ok, sorry for the wrong answer earlier. Here is my understanding of why this works:

- `containingInst` is an `mlir::Operation*`. See: https://github.com/llvm/llvm-project/blob/fcfd4fb686b83a5c8c7ba2c3bd4992ff24e83870/mlir/include/mlir/IR/OpDefinition.h#L109-L111

- All MLIR operations are using the curiously recursive template pattern and inherit from the `mlir::Op` class. This class provides a conversion operator from `mlir::Operation*` pointers that is intended to do the magic you see according to the related comment:
See: https://github.com/llvm/llvm-project/blob/fcfd4fb686b83a5c8c7ba2c3bd4992ff24e83870/mlir/include/mlir/IR/OpDefinition.h#L1038

- There is also a ctor form nullptr:
https://github.com/llvm/llvm-project/blob/fcfd4fb686b83a5c8c7ba2c3bd4992ff24e83870/mlir/include/mlir/IR/OpDefinition.h#L1034

Under the hood, all operations that are handled as value really just are a wrapper around a pointer. Note that this pointer can be null, in which case the operation cast to the `false` boolean when checked in if and asserts. 

You can have a look into the generated `include/flang/Optimizer/Dialect/FIROps.h.inc` in your build directory to see how `LoopOp` is actually defined.

lib/Optimizer/CMakeLists.txt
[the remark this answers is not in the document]
schweitzpgi: We do want the header files in Optimizer/* to compile without warnings. But, as LLVM is more permissive and the Optimizer library uses a good bit of LLVM, we want to be relaxed by default much like LLVM itself.

lib/Optimizer/Dialect/FIROps.cpp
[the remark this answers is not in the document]
River707: Essentially yes. The `*Op` classes are effectively nice "views" into an `Operation`, they hold no state other than an `Operation*`(1). To enable casting an `Operation` to one of these `*Op` classes, we provide custom cast implementations for llvm::isa/cast/dyn_cast to pick up on(2).


[1]: https://mlir.llvm.org/docs/Tutorials/Toy/Ch-2/#op-vs-operation-using-mlir-operations
[2]: https://github.com/llvm/llvm-project/blob/133db44996a80db957360a0b727479d4a8d4d1a2/mlir/include/mlir/IR/Operation.h#L685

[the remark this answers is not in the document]
DavidTruby: Ok got it, thanks! Sorry to have kept poking at this I just thought it looked odd that we were going from LoopOp* to LoopOp during the return of this function. I guess I'll get used to this pattern when working with mlir

include/flang/Optimizer/Dialect/FIROps.h
```diff
@@ -0,0 +1,76 @@
...
+
+using namespace mlir;
+
+namespace fir {
+
+class FirEndOp;
+class LoopOp;
+
+enum class CmpFPredicate {
```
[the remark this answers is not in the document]
schweitzpgi: I'll give this a shot. I seem to vaguely recall it used to be buried somewhere and not fully exposed.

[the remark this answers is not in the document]
River707: Note that I just recently(yesterday) moved the definition to tablegen:
https://github.com/llvm/llvm-project/commit/c10896682d0bb457b9d77fdd753ed9e7e6806db1#diff-cc3306bb0c567bb00f8a4a43b3145858

include/flang/Optimizer/Dialect/FIROps.td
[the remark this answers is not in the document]
schweitzpgi: Yes. That's correct.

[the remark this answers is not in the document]
schweitzpgi: It's not documented.
There is a type (3 actually) for referencing memory.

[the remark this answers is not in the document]
schweitzpgi: Noted.

schweitzpgi: Yes. I believe that is the case.
Well, this pertains to FIR specifically. So we want to respect the Fortran rules of when and where we can reassociate explicitly.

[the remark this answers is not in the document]
schweitzpgi: For the Fortran use case, no. The programmer may specify an ordering to avoid overflows, for example, This is meant to explicitly enforce the determinism specified by the Fortran programmer, while allowing aggressive optimizations in the default case.

test-lit/Fir/fir-ops.fir
```diff
@@ -0,0 +1,458 @@
...
+// CHECK-LABEL: func @dvd4() -> !fir.type<derived4(p:i8){f:f32}>
+func @dvd1() -> !fir.type<derived1>
+func @dvd2() -> !fir.type<derived2(p:i32)>
+func @dvd3() -> !fir.type<derived3{f:f32}>
+func @dvd4() -> !fir.type<derived4(p:i8){f:f32}>
+
+// FIR array types
+// CHECK-LABEL: func @arr1() -> !fir.array<10xf32>
+// CHECK-LABEL: func @arr2() -> !fir.array<10x10xf32>
+// CHECK-LABEL: func @arr3() -> !fir.array<?xf32>
```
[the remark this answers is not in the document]
schweitzpgi: Yes. This started out as one big test. Needs some cleanup.

include/flang/Optimizer/Dialect/FIROps.h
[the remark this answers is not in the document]
schweitzpgi: Thanks for the heads-up. I'll rebase and make sure this hangs together. :)

include/flang/Optimizer/Dialect/FIROps.td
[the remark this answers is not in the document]
kiranchandramohan: OK. But will there will be verification for valid conversions and also for lowering to LLVM dialect, right?
I was specifically asking whether the fir references can be converted to standard memrefs. Is there currently an operation which does that?

include/flang/Optimizer/Dialect/FIROps.td
sscalpone: @jeanPerier If this request is being deferred for now, please add it to the list of tasks for the FIR integration.

include/flang/Optimizer/Dialect/FIROps.td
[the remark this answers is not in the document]
schweitzpgi: Yes, We will define the restrictions and codify them in the verifier.

The plan will be to utilize standard passes including lowering to standard dialect when it is appropriate. However, doing such conversions is not in the scope of this PR, no.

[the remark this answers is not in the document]
kiranchandramohan: OK. I should have clarified, I was not asking for it as part of this patch. But this question was for my information.
I have more questions relating to what will be a suitable type for use along with the OpenMP operations. Will it be standard memrefs or fir references? But i guess i have to discuss this with you separately and not here.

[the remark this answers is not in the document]
schweitzpgi: That would be great and I'm happy to discuss with you. (But, yes, I am trying to keep the PR focused. Thanks!)

test-lit/Fir/fir-ops.fir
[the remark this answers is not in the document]
jeanPerier: Thanks for spotting that, I cleaned that up.

lib/Optimizer/Dialect/FIRType.cpp
```diff
@@ -0,0 +1,1288 @@
...
+}
+
+mlir::LogicalResult
+fir::HeapType::verifyConstructionInvariants(const mlir::AttributeStorage *,
+                                            mlir::Type eleTy) {
+  if (eleTy.isa<BoxType>() || eleTy.isa<BoxCharType>() ||
+      eleTy.isa<BoxProcType>() || eleTy.isa<DimsType>() ||
+      eleTy.isa<FieldType>() || eleTy.isa<LenType>() || eleTy.isa<HeapType>() ||
+      eleTy.isa<PointerType>() || eleTy.isa<ReferenceType>() ||
+      eleTy.isa<TypeDescType>())
```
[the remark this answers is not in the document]
jeanPerier: Good point, thanks. Done.

include/flang/Optimizer/Dialect/FIROps.h
```diff
@@ -0,0 +1,76 @@
...
+  UGT,
+  UGE,
+  ULT,
+  ULE,
+  UNE,
+  // Any unordered
+  UNO,
+  // Always true
+  AlwaysTrue,
+  // Number of predicates.
```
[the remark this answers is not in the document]
jeanPerier: This enum was removed to use the one from StandardOps, so this comment is not there anymore.

include/flang/Optimizer/Dialect/FIRType.h
```diff
@@ -0,0 +1,397 @@
...
+bool isa_fir_type(mlir::Type t);
+
+/// Is `t` any of the Standard dialect types?
+bool isa_std_type(mlir::Type t);
+
+/// Is `t` any of the FIR dialect or Standard dialect types?
+bool isa_fir_or_std_type(mlir::Type t);
+
+/// Is `t` a FIR dialect type that implies a memory (de)reference?
+bool isaMemref(mlir::Type t);
```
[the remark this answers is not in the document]
jeanPerier: Changed to `isa_ref_type` to avoid ambiguity with MLIR memref.

lib/Optimizer/Dialect/FIRDialect.cpp
```diff
@@ -0,0 +1,77 @@
...
+#include "mlir/Dialect/LLVMIR/LLVMDialect.h"
+#include "mlir/IR/StandardTypes.h"
+#include "mlir/Transforms/SideEffectsInterface.h"
+
+using namespace fir;
+
+namespace {
+
+template <typename A>
+void selectBuild(mlir::OpBuilder *builder, mlir::OperationState *result,
```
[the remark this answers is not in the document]
jeanPerier: Not used anymore, removed.

include/flang/Optimizer/Dialect/FIROps.td
schweitzpgi: We'll move some code around in a subsequent PR.

lib/Optimizer/Dialect/FIRType.cpp
schweitzpgi: Noting this for a later PR.

include/flang/Optimizer/Dialect/FIROps.td
jeanPerier: Added on list of task.

lib/Optimizer/Support/KindMapping.cpp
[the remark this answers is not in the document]
jeanPerier: Will not be solved in this PR. Added on list of task.

include/flang/Optimizer/Dialect/FIROps.td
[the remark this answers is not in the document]
jeanPerier: Will not be solved in this PR. Added on list of task.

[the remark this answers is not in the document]
jeanPerier: Will not be solved in this PR. Added on list of task.

lib/Optimizer/Dialect/FIRType.cpp
jeanPerier: It turned out it is not needed anymore. KindTy parameter was removed.

README.md
```diff
@@ -93,6 +93,55 @@ LLVM_INSTALL_UTILS=On
...
+git clone --depth=1 -b f18 https://github.com/flang-compiler/f18-llvm-project.git
+mkdir build
+mkdir install
+cd build
+cmake ../f18-llvm-project/llvm -DCMAKE_BUILD_TYPE=Release \
+    -DLLVM_ENABLE_PROJECTS=mlir -DCMAKE_CXX_STANDARD=17 \
+    -DLLVM_INSTALL_UTILS=On \
+    -DCMAKE_INSTALL_PREFIX=../install
+make -j
+make install
```
psteinfeld: I tried these instructions on my Windows laptop using Windows Subsystem for Linux with Ubuntu installed and using GNU 8.3.0 as my compiler.  The "make -j" command caused my laptop to crash.  Presumably, it ran out of memory.  "make" without the "-j" command worked.  I did not try using an option like "make -j 8".  I also was able to build using Ninja rather than make.  This seemed to work better.  To do this, I just added the option "-G Ninja" at the front of the cmake command.  Specifically, my first line was 
```
cmake -G Ninja ../f18-llvm-project/llvm -DCMAKE_BUILD_TYPE=Release \
```
The command to do the build is then just
```
ninja
```
To do the install, it's
```
ninja install
```
This seemed to work better than using make.  Of course, it requires that you have Ninja installed on your build machine.

include/flang/Optimizer/Dialect/FIROps.td
jeanPerier: Kept as it is in this PR. Added to the list of point to track.

README.md
jeanPerier: Thanks for testing the instructions ! I have removed the `-j`, that was a bit wild indeed. The examples are given with make to be portable and understandable by a wide audience. It is also consistent with other example (e.g. in `Build F18`). 
People familiar will builds will know how to replace this with Ninja and to set-up there `-j` according to their machine.

jeanPerier:
> Why is the LangRef not part of this? If this is autogen then can you add a pointer to how to access that?
Good point, it was not part of this PR because it require `add_mlir_function` to generate it but as mentioned in an earlier comment of @DavidTruby we cannot use it (it seems to only works for dialect whose sources live inside MLIR sources). So, I have replicated the part to generate the `FIRLangRef.md` and updated the README.md

jeanPerier:
> Is there no representation for co-arrays in FIR now?

That's correct. We want to gain more experience with both FIR and co-arrays before deciding how they should be handled in FIR.

jeanPerier:
Thanks all for your reviews !
We should have addressed all your comments so far (either solved, rejected, or agreed to solve them in later PRs as improvement). Please do not hesitate replying to this if you feel some of your comments were forgotten. Otherwise you are welcome to approve this PR so we can move to the next patches !

lib/Optimizer/Support/KindMapping.cpp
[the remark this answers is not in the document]
schweitzpgi: I looked at this a little more. I don't think we will pursue this one at this point. The "parser" here has nothing to do with MLIR syntax but is parsing a string that encodes the mapping of an arbitrary Fortran front-end's kind values to LLVM supported types. For example, in Bob's Fortran front-end, the type `INTEGER(KIND=747)` might correspond to an integer with 51 bits, an `i51` in LLVM IR. This kind mapping micro-language allows different front-ends (or potentially users) to configure these mappings at their choosing. Conceptually, it's like a data layout string in LLVM.

jeanPerier:
> Thanks for addressing the review comments. LGTM.
> 
> It will great if you can,
> 
> 1. Update the commit message with details of how to generate the documentation.
> 2. Spend some time to change or replace the original FIR investigation documentation
>    https://github.com/flang-compiler/f18/blob/master/documentation/Investigating-FIR-as-an-MLIR-dialect.md
> 3. Convey in the commit message what exactly is there in this patch and what is not. For e.g.,
>    -> contains the definition, printing, parsing and verification of operations.
>    -> is the verification of all operations there in this patch?
>    -> Co-arrays are not there.

Thanks @kiranchandramohan , I have removed https://github.com/flang-compiler/f18/blob/master/documentation/Investigating-FIR-as-an-MLIR-dialect.md because it was an early-design document that had a purpose when we started looking at FIR but is now outdated, thanks for pointing to it. The info inside is replaced by FIRLangRef.md that can be generated and new documents that will come with the next patches.

When squashing and rebasing for the merge, I will change the commit to:

```
Add Fortran IR (FIR) MLIR dialect implementation

Adds FIR library that implements an MLIR dialect to which Fortran
parse-tree will be lowered to.

FIR is defined and documented inside FIROps.td added in this commit.
It is possible to generate a more readable description FIRLangRef.md
from FIROps.td following the related instructions added to the README.md
by this commit.

This patch adds FIR definition and implementation that allow parsing,
printing, and verifying FIR. FIR transformations and lowering to Standard
and LLVM dialects are not part of this patch. The FIR verifiers are verifying
the basic properties of FIR operations in order to provide a sufficient
frame for lowering. Verifiers for more advanced FIR properties can be added
as needed.

Coarrays are not covered by FIR defined in this patch.

This patch also adds tco tool that is meant to process FIR input files and
drives transformations on it. The tco tool is used for testing.
In this patch, it is only used to demonstrate parsing/verifying/
and dumping FIR with round-trip tests.



Note:
This commit does not reflect an actual work log, it is a feature-based split of the
changes done in the FIR experimental branch. The related work log can be found in the
commits between:

https://github.com/schweitzpgi/f18/commit/742edde572bd74d77cf7d447132ccf0949187fce
and
https://github.com/schweitzpgi/f18/commit/2ff55242126d86061f4fed9ef7b59d3636b5fd0b

Changes on top of these original commits were made during this patch review.
```

jeanPerier:
I have pushed an update to adapt with the latest changes in MLIR (since end-of last week, this PR was could not be build with LLVM head due to MLIR interface changes). The `f18` branch of `f18-llvm-project` LLVM fork mentioned in the `README` was updated accordingly, but the current head from llvm-project also works with FIR at the time of this comment.

The other commit I pushed addresses @River707 comment that suggested moving big code pieces from the .td to a .cpp file (the comment was deferred, but the need to rebase with latest LLVM head gave us time to to it).

jeanPerier:
I have squashed and rebased to prepare merging, if you still need the commit history for the comments, you can find it here: https://github.com/flang-compiler/f18/commits/d0a7dabd6941a6725436e1c37ccc6e2b9aa0d6fb

DavidTruby:
F18 doesn't build for me since this merge, I get the following error:

```
/home/davtru01/f18/include/flang/Optimizer/Dialect/FIROps.td:17:9: error: Could not find include file 'mlir/Interfaces/ControlFlowInterfaces.td'
include "mlir/Interfaces/ControlFlowInterfaces.td"
```

I have mlir installed with LLVM and it finds AddMLIR.cmake correctly.

psteinfeld:
Similar to @DavidTruby, after pulling in these change, I cannot build.  I was able to build successfully a couple of days ago before these changes were merged into master.  My build environment is using GNU 8.3.0 and my own built version of LLVM, from the latest sources as of about a week ago, which is version 11.0.0.

The build fails very quickly.  Here's a fragment of the log file, starting at the beginning:
```
[1/190] Building FIROps.h.inc...
FAILED: include/flang/Optimizer/Dialect/FIROps.h.inc 
cd /mnt/c/GitHub/f18/master/Release && /mnt/c/GitHub/f18/fir/install/bin/mlir-tblgen -gen-op-decls -I -I -I /mnt/c/GitHub/f18/master/include/flang/Optimizer/Dialect -I /mnt/c/GitHub/f18/fir/install/include /mnt/c/GitHub/f18/master/include/flang/Optimizer/Dialect/FIROps.td --write-if-changed -o include/flang/Optimizer/Dialect/FIROps.h.inc -d include/flang/Optimizer/Dialect/FIROps.h.inc.d
/mnt/c/GitHub/f18/master/include/flang/Optimizer/Dialect/FIROps.td:17:9: error: Could not find include file 'mlir/Interfaces/ControlFlowInterfaces.td'
include "mlir/Interfaces/ControlFlowInterfaces.td"
        ^
/mnt/c/GitHub/f18/master/include/flang/Optimizer/Dialect/FIROps.td:17:9: error: Unexpected input at top level
include "mlir/Interfaces/ControlFlowInterfaces.td"
        ^
[2/190] Building FIROps.cpp.inc...
FAILED: include/flang/Optimizer/Dialect/FIROps.cpp.inc 
cd /mnt/c/GitHub/f18/master/Release && /mnt/c/GitHub/f18/fir/install/bin/mlir-tblgen -gen-op-defs -I -I -I /mnt/c/GitHub/f18/master/include/flang/Optimizer/Dialect -I /mnt/c/GitHub/f18/fir/install/include /mnt/c/GitHub/f18/master/include/flang/Optimizer/Dialect/FIROps.td --write-if-changed -o include/flang/Optimizer/Dialect/FIROps.cpp.inc -d include/flang/Optimizer/Dialect/FIROps.cpp.inc.d
/mnt/c/GitHub/f18/master/include/flang/Optimizer/Dialect/FIROps.td:17:9: error: Could not find include file 'mlir/Interfaces/ControlFlowInterfaces.td'
include "mlir/Interfaces/ControlFlowInterfaces.td"
        ^
/mnt/c/GitHub/f18/master/include/flang/Optimizer/Dialect/FIROps.td:17:9: error: Unexpected input at top level
include "mlir/Interfaces/ControlFlowInterfaces.td"
        ^
[3/190] Building CXX object lib/Common/CMakeFiles/FortranCommon.dir/Fortran.cpp.o
   ...
```

DavidTruby:
@psteinfeld, @jeanPerier mentioned to me that MLIR's interface changed on Tuesday. So you need a build of LLVM from yesterday or today for this to work correctly. Once I built LLVM from scratch from a new git clone, it worked fine.

psteinfeld:
> 
> 
> @psteinfeld, @jeanPerier mentioned to me that MLIR's interface changed on Tuesday. So you need a build of LLVM from yesterday or today for this to work correctly. Once I built LLVM from scratch from a new git clone, it worked fine.

Thanks!  I'll give it a try.

psteinfeld:
> 
> 
> > @psteinfeld, @jeanPerier mentioned to me that MLIR's interface changed on Tuesday. So you need a build of LLVM from yesterday or today for this to work correctly. Once I built LLVM from scratch from a new git clone, it worked fine.
> 
> Thanks! I'll give it a try.

I cleaned out my local LLVM build, rebuilt the latest from scratch, and now everything works.  Thanks for your help!

Note that my first attempt was to simple grab the latest sources from GitHub, run cmake, and rebuild.  That wasn't good enough.  I had to delete the build area and do everything from scratch.

How frequently should I expect to need to rebuild LLVM?

DavidTruby:
Once the new cmake changes land and we are part of the llvm monorepo, the easiest way of managing things should be to build alongside llvm and we shouldn't have these issues. See #1045 for more info!

psteinfeld:
> 
> 
> Once the new cmake changes land and we are part of the llvm monorepo, the easiest way of managing things should be to build alongside llvm and we shouldn't have these issues. See #1045 for more info!

Thanks, David!  You're up past my bed time.

PR #1037, opened 2020-02-27
Add semantic checks C8104, C8105. Add tests for C8103, C8104, C8105

vjayathirtha-nv:


vjayathirtha-nv:
> Can this be in check-declarations.cpp?

That was my initial choice, but in `check-declarations.cpp` we don't have access to source location information to point the error at the right place. So I had to make a separate checker.

lib/Semantics/check-namelist.cpp
```diff
@@ -0,0 +1,39 @@
...
+  for (auto &x : nmlStmt.v) {
+    if (auto *nml{std::get<parser::Name>(x.t).symbol}) {
+      for (const auto &nmlObjName : std::get<std::list<parser::Name>>(x.t)) {
+        auto *nmlObjSymbol{nmlObjName.symbol};
+        if (nmlObjSymbol && nmlObjSymbol->has<ObjectEntityDetails>()) {
+          const auto *symDetails{
+              std::get_if<ObjectEntityDetails>(&nmlObjSymbol->details())};
+          if (symDetails && symDetails->IsAssumedSize()) {  // C8104
+            context_.Say(nmlObjName.source,
+                "A namelist group object '%s' must not be assumed-size"_err_en_US,
```
psteinfeld: This line is greater than 80 characters.  Not everyone does this, but I like to keep source lines within 80 columns.

lib/Semantics/check-namelist.cpp
```diff
@@ -0,0 +1,40 @@
...
+
+#include "check-namelist.h"
+
+namespace Fortran::semantics {
+
+void NamelistChecker::Leave(const parser::NamelistStmt &nmlStmt) {
+  for (auto &x : nmlStmt.v) {
+    if (auto *nml{std::get<parser::Name>(x.t).symbol}) {
+      for (const auto &nmlObjName : std::get<std::list<parser::Name>>(x.t)) {
+        auto *nmlObjSymbol{nmlObjName.symbol};
```
tskeith: I think it's helpful to use `const auto` when you can rather than just `auto`. That applies to `x`, `nml`, and `nmlObjSymbol`.

lib/Semantics/semantics.cpp
```diff
@@ -14,6 +14,7 @@
 #include "check-arithmeticif.h"
 #include "check-coarray.h"
 #include "check-data.h"
+#include "check-namelist.h"
```
tskeith: This file needs to be clang-formatted, the includes aren't sorted.

PR #1039, opened 2020-02-28
Improve array element errors

tskeith:
When something is parsed as an array element it was sometimes intended
to be a function call or structure constructor. So if the base name is
not found the errors can be confusing. This is an attempt to improve
them.

When the subscript list is empty, it was probably meant to be a function
call, so report that the name is not a function.

If the base is a scalar but there are subscripts, report that it is not
an array.

### DIFF
--- a/lib/Semantics/expression.cpp
+++ b/lib/Semantics/expression.cpp
@@ -238,8 +238,10 @@ MaybeExpr ExpressionAnalyzer::CompleteSubscripts(ArrayRef &&ref) {
   if (subscripts == 0) {
     // nothing to check
   } else if (subscripts != symbolRank) {
-    Say("Reference to rank-%d object '%s' has %d subscripts"_err_en_US,
-        symbolRank, symbol.name(), subscripts);
+    if (symbolRank != 0) {
+      Say("Reference to rank-%d object '%s' has %d subscripts"_err_en_US,
+          symbolRank, symbol.name(), subscripts);
+    }
     return std::nullopt;
   } else if (Component * component{ref.base().UnwrapComponent()}) {
     int baseRank{component->base().Rank()};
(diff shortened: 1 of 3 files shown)


lib/Semantics/expression.cpp
```diff
@@ -870,16 +872,25 @@ std::vector<Subscript> ExpressionAnalyzer::AnalyzeSectionSubscripts(
...
       }
+    } else if (std::optional<DataRef> dataRef{
+                   ExtractDataRef(std::move(*baseExpr))}) {
+      return ApplySubscripts(
+          std::move(*dataRef), AnalyzeSectionSubscripts(ae.subscripts));
     } else {
       Say("Subscripts may be applied only to an object, component, or array constant"_err_en_US);
     }
   }
+  // error was reported: analyze subscripts without reporting more errors
```
psteinfeld: I don't understand this comment.  Do you mean that an error may been reported?

lib/Semantics/expression.cpp
```diff
@@ -238,8 +238,10 @@ MaybeExpr ExpressionAnalyzer::CompleteSubscripts(ArrayRef &&ref) {
   if (subscripts == 0) {
     // nothing to check
   } else if (subscripts != symbolRank) {
-    Say("Reference to rank-%d object '%s' has %d subscripts"_err_en_US,
-        symbolRank, symbol.name(), subscripts);
+    if (symbolRank != 0) {
+      Say("Reference to rank-%d object '%s' has %d subscripts"_err_en_US,
```
psteinfeld: It would be good to have a test case that produces this message, although it looks like there's never been a test case for this condition.

lib/Semantics/expression.cpp
tskeith: We only hit this line if an error was reported about the base of the array element. In that case we want to analyze the subscript expressions (because every expression should be analyzed) but not report any errors from that analysis because the might be confusing. E.g. in data01.f90, after "'persn' is not an array" you would get an error about 'Abcd Efgh' not being an integer (because array subscripts are required to be integers).

lib/Semantics/expression.cpp
tskeith: Good idea. There was "Reference to rank-0 object..." error that I removed, so I've added a couple of tests.

lib/Semantics/expression.cpp
```diff
@@ -238,8 +238,10 @@ MaybeExpr ExpressionAnalyzer::CompleteSubscripts(ArrayRef &&ref) {
...
+          symbolRank, symbol.name(), subscripts);
```
jeanPerier: I am a bit lost with the semantic check status for assumed-rank array, I have them in mind because I just reviewed #1010 .  The code above would still emit something weird for them:
```
function foo(b)
  integer :: foo
  integer :: b(..)
  foo = b(2, 2)
end function
```
`error: Reference to rank-1 object 'b' has 2 subscripts` is emitted because `symbol.Rank()` for assumed-rank returns 1.

If we are not yet enforcing `C838` then that may just be a TODO and please ignore my comment.

PR #1041, opened 2020-02-28
[Not for review] PR 1035 Drone.io test

jeanPerier:
I've lost commits right into PR 1035, so doing fake PR to test fix in the meantime.


jeanPerier:
Tests passed.

PR #1043, opened 2020-03-02
Link against zlib when LLVM does.

### LINKED ISSUE #1042: CMake Error for drone.ci due to target "ZLIB::ZLIB" not found
The drone.ci build fails for commits done to F18
Log:
https://cloud.drone.io/flang-compiler/f18/1176/1/2

DavidTruby:
If LLVM links against zlib we also need to do the same. Simply using find_package is enough to make the transitive dependencies that the LLVM libraries have on zlib link us against it correctly.

Fixes #1042

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -49,6 +49,10 @@ message(STATUS "Build Type: ${CMAKE_BUILD_TYPE}" )
 
 find_package(LLVM REQUIRED CONFIG)
 message(STATUS "Found LLVM ${LLVM_PACKAGE_VERSION} in ${LLVM_DIR}")
+# If LLVM links to zlib we need the imported targets so we can too.
+if(LLVM_ENABLE_ZLIB)
+  find_package(ZLIB REQUIRED)
+endif()
 
 list(APPEND CMAKE_MODULE_PATH ${LLVM_DIR})
 


DavidTruby:
If you want the old behaviour (not linking against zlib) then you can build LLVM with -DLLVM_ENABLE_ZLIB=Off. This patch won't import zlib if the LLVM it is building against didn't.

tskeith:
All I can say is that this builds okay for me (though I didn't have a problem before).

Someone who knows something about building with LLVM should review this. Maybe @schweitzpgi or @jeanPerier.

DavidTruby:
This was triggered by a patch added to LLVM over the weekend that adds zlib as a transitive dependency to LLVMSupport. We aren't (currently) actually calling any of the functions from LLVMSupport that actually call zlib, so we don't see linker errors if we don't do this (just the new cmake errors).
Strictly speaking though this is the correct thing to do as any function from LLVMSupport could choose to add a call to zlib at any time, and we may calls to functions that call into zlib without noticing. In either of these cases we will start seeing strange linker errors.

DavidTruby:
@tskeith @jeanPerier Can we merge this? It's still causing all our CI to fail as well as local builds with checkouts of llvm less than a week old.

PR #1046, opened 2020-03-03
Improve checking of structure constructor keywords

tskeith:
When a misparsed FunctionReference was converted to a StructureConstructor,
the keywords were not checked for accessibility. For example, in `t(i=1)`,
`i` must not be a PRIVATE component of derived type `t`.

The conversion happens in expression analysis so that where the accessibity
must be checked. So move `CheckAccessibleComponent` to `tools.h` so that it
can be shared by `resolve-names.cpp` and `expression.cpp`.

Also check that an access-spec can only appear in a module.

### DIFF
--- a/lib/Semantics/expression.cpp
+++ b/lib/Semantics/expression.cpp
@@ -1402,6 +1402,11 @@ MaybeExpr ExpressionAnalyzer::Analyze(
       }
     }
     if (symbol) {
+      if (const auto *currScope{context_.globalScope().FindScope(source)}) {
+        if (auto msg{CheckAccessibleComponent(*currScope, *symbol)}) {
+          Say(source, *msg);
+        }
+      }
       if (checkConflicts) {
         auto componentIter{
             std::find(components.begin(), components.end(), *symbol)};
@@ -1433,13 +1438,10 @@ MaybeExpr ExpressionAnalyzer::Analyze(
         } else if (symbol->has<semantics::ObjectEntityDetails>()) {
           // C1594(4)
           const auto &innermost{context_.FindScope(expr.source)};
-          if (const auto *pureProc{
-                  semantics::FindPureProcedureContaining(innermost)}) {
-            if (const Symbol *
-                pointer{semantics::FindPointerComponent(*symbol)}) {
+          if (const auto *pureProc{FindPureProcedureContaining(innermost)}) {
+            if (const Symbol * pointer{FindPointerComponent(*symbol)}) {
               if (const Symbol *
-                  object{semantics::FindExternallyVisibleObject(
-                      *value, *pureProc)}) {
+                  object{FindExternallyVisibleObject(*value, *pureProc)}) {
                 if (auto *msg{Say(expr.source,
                         "Externally visible object '%s' may not be "
                         "associated with pointer component '%s' in a "
--- a/lib/Semantics/resolve-names.cpp
+++ b/lib/Semantics/resolve-names.cpp
@@ -230,7 +230,6 @@ public:
   bool SetPassNameOn(Symbol &);
   bool SetBindNameOn(Symbol &);
   void Post(const parser::LanguageBindingSpec &);
-  bool Pre(const parser::AccessSpec &);
   bool Pre(const parser::IntentSpec &);
   bool Pre(const parser::Pass &);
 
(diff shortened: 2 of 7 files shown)


lib/Semantics/expression.cpp
```diff
@@ -1364,6 +1364,11 @@ MaybeExpr ExpressionAnalyzer::Analyze(
         Say(source,
             "Keyword '%s=' does not name a component of derived type '%s'"_err_en_US,
             source, typeName);
+      } else if (const auto *currScope{
+                     context_.globalScope().FindScope(source)}) {
+        if (auto msg{CheckAccessibleComponent(*currScope, *symbol)}) {
+          Say(source, *msg);
+        }
```
jeanPerier: I think this check would also work if the code was trying to initialized private components without keywords if you moved this check right after: https://github.com/flang-compiler/f18/blob/2d853b4b9e2ce4b41daa00efb56834be2b6f4276/lib/Semantics/expression.cpp#L1409

tskeith: Good idea, I've done that.

tskeith:
> Private components can't be initialized in a structure constructor, keyword or not. See C7102.
> 
> (EDIT: I mean they can't be initialized in scopes where the component is not accessible; in the module where the type is defined, and its descendants, they're fine.)

Right, at Jean's suggestion my second commit implements that.

lib/Semantics/resolve-names.cpp
```diff
@@ -3067,6 +3031,19 @@ void DeclarationVisitor::Post(const parser::EnumDef &) {
   enumerationState_ = EnumeratorState{};
 }
 
+bool DeclarationVisitor::Pre(const parser::AccessSpec &x) {
+  Attr attr{AccessSpecToAttr(x)};
+  const Scope &scope{
+      currScope().IsDerivedType() ? currScope().parent() : currScope()};
+  if (!scope.IsModule()) {
+    Say(currStmtSource().value(),
+        "%s attribute may only appear in the specification part of a module"_err_en_US,
```
klausler: module or submodule

lib/Semantics/resolve-names.cpp
```diff
@@ -5517,7 +5491,7 @@ bool ResolveNamesVisitor::SetProcFlag(
 
 bool ModuleVisitor::Pre(const parser::AccessStmt &x) {
   Attr accessAttr{AccessSpecToAttr(std::get<parser::AccessSpec>(x.t))};
-  if (currScope().kind() != Scope::Kind::Module) {
+  if (!currScope().IsModule()) {
     Say(currStmtSource().value(),
         "%s statement may only appear in the specification part of a module"_err_en_US,
```
klausler: or submodule

lib/Semantics/tools.cpp
```diff
@@ -956,6 +956,31 @@ bool IsPolymorphicAllocatable(const Symbol &symbol) {
   return IsAllocatable(symbol) && IsPolymorphic(symbol);
 }
 
+static const Scope *FindModuleAncestor(const Symbol &symbol) {
```
klausler: Might be useful elsewhere; consider adding to `tools.h`.

test/Semantics/resolve34.f90
```diff
@@ -91,3 +91,41 @@ subroutine s7
...
+    x = t(i1=2, i2=5)  !OK
+  end
+end
+subroutine s8
+  use m8
+  type(t) :: x
+  !ERROR: PRIVATE component 'i2' is only accessible within module 'm8'
+  x = t(2, 5)
+  !ERROR: PRIVATE component 'i2' is only accessible within module 'm8'
+  x = t(i1=2, i2=5)
```
klausler: `x = t(2,5)` should also be tested.

lib/Semantics/resolve-names.cpp
tskeith: That's what I thought too, because it makes perfect sense to allow it in a submodule. But:
```
C817 An access-spec shall appear only in the specification-part of a module.
C869 (R827) An access-stmt shall appear only in the specification-part of a module.
```
I'll add some comments referring to these constraints.

lib/Semantics/tools.cpp
tskeith: I moved it to tools.h and used it a couple of other places.

test/Semantics/resolve34.f90
tskeith: Done

PR #1047, opened 2020-03-03
[LLVMify F18] Replace the use std::ostream with LLVM streams llvm::ostream

CarolineConcatto:
[LLVMify F18] Replace the use std::ostream with LLVM streams llvm::ostream

This patch replaces the occurrence of std::ostream by llvm::raw_ostream,
because  LLVM Coding Standards[1] says that "All new code should use raw_ostream instead of ostream".[1]

As a consequence, this patch also replaces the use of:
   std:stringstream by llvm::raw_string_ostream
   std::ofstream by llvm::raw_fd_ostream
   std::cout by llvm::outs()   and
   std::cerr by llvm::errs()
where it is needed, because of the use of llvm::raw_ostream.

[1]https://llvm.org/docs/CodingStandards.html

### DIFF
--- a/documentation/ParserCombinators.md
+++ b/documentation/ParserCombinators.md
@@ -160,5 +160,5 @@ is built.  All of the following parsers consume characters acquired from
 
 ### Debugging Parser
 Last, a string literal `"..."_debug` denotes a parser that emits the string to
-`std::cerr` and succeeds.  It is useful for tracing while debugging a parser but should
+`llvm::errs` and succeeds.  It is useful for tracing while debugging a parser but should
 obviously not be committed for production code.
--- a/include/flang/Common/enum-set.h
+++ b/include/flang/Common/enum-set.h
@@ -16,11 +16,11 @@
 
 #include "constexpr-bitset.h"
 #include "idioms.h"
+#include "llvm/Support/raw_ostream.h"
 #include <bitset>
 #include <cstddef>
 #include <initializer_list>
 #include <optional>
-#include <ostream>
 #include <string>
 #include <type_traits>
 
@@ -199,8 +199,8 @@ public:
     }
   }
 
-  std::ostream &Dump(
-      std::ostream &o, std::string EnumToString(enumerationType)) const {
+  llvm::raw_ostream &Dump(
+      llvm::raw_ostream &o, std::string EnumToString(enumerationType)) const {
     char sep{'{'};
     IterateOverMembers([&](auto e) {
       o << sep << EnumToString(e);
--- a/include/flang/Evaluate/formatting.h
+++ b/include/flang/Evaluate/formatting.h
@@ -9,10 +9,10 @@
 #ifndef FORTRAN_EVALUATE_FORMATTING_H_
 #define FORTRAN_EVALUATE_FORMATTING_H_
 
-// It is inconvenient in C++ to have std::ostream::operator<<() as a direct
+// It is inconvenient in C++ to have llvm::raw_ostream::operator<<() as a direct
 // friend function of a class template with many instantiations, so the
 // various representational class templates in lib/Evaluate format themselves
-// via AsFortran(std::ostream &) member functions, which the operator<<()
+// via AsFortran(llvm::raw_ostream &) member functions, which the operator<<()
 // overload below will call.  Others have AsFortran() member functions that
 // return strings.
 //
@@ -20,31 +20,31 @@
 // representational class templates that need it, not by external clients.
 
 #include "flang/Common/indirection.h"
+#include "llvm/Support/raw_ostream.h"
 #include <optional>
-#include <ostream>
 #include <type_traits>
 
 namespace Fortran::evaluate {
 
 template<typename A>
-auto operator<<(std::ostream &o, const A &x) -> decltype(x.AsFortran(o)) {
+auto operator<<(llvm::raw_ostream &o, const A &x) -> decltype(x.AsFortran(o)) {
   return x.AsFortran(o);
 }
 
 template<typename A>
-auto operator<<(std::ostream &o, const A &x) -> decltype(o << x.AsFortran()) {
+auto operator<<(llvm::raw_ostream &o, const A &x) -> decltype(o << x.AsFortran()) {
   return o << x.AsFortran();
 }
 
 template<typename A, bool COPYABLE>
 auto operator<<(
-    std::ostream &o, const Fortran::common::Indirection<A, COPYABLE> &x)
+    llvm::raw_ostream &o, const Fortran::common::Indirection<A, COPYABLE> &x)
     -> decltype(o << x.value()) {
   return o << x.value();
 }
 
 template<typename A>
-auto operator<<(std::ostream &o, const std::optional<A> &x)
+auto operator<<(llvm::raw_ostream &o, const std::optional<A> &x)
     -> decltype(o << *x) {
   if (x) {
     o << *x;
--- a/include/flang/Parser/dump-parse-tree.h
+++ b/include/flang/Parser/dump-parse-tree.h
@@ -15,8 +15,7 @@
 #include "unparse.h"
 #include "flang/Common/idioms.h"
 #include "flang/Common/indirection.h"
-#include <ostream>
-#include <sstream>
+#include "llvm/Support/raw_ostream.h"
 #include <string>
 #include <type_traits>
 
@@ -37,7 +36,7 @@ struct HasSource<T, decltype((void)T::source, 0)> : std::true_type {};
 class ParseTreeDumper {
 public:
   explicit ParseTreeDumper(
-      std::ostream &out, const AnalyzedObjectsAsFortran *asFortran = nullptr)
+      llvm::raw_ostream &out, const AnalyzedObjectsAsFortran *asFortran = nullptr)
     : out_(out), asFortran_{asFortran} {}
 
   static constexpr const char *GetNodeName(const char *) { return "char *"; }
@@ -761,7 +760,8 @@ public:
 protected:
   // Return a Fortran representation of this node to include in the dump
   template<typename T> std::string AsFortran(const T &x) {
-    std::ostringstream ss;
+    std::string buf;
+    llvm::raw_string_ostream ss{buf};
     if constexpr (std::is_same_v<T, Expr>) {
       if (asFortran_ && x.typedExpr) {
         asFortran_->expr(ss, *x.typedExpr);
@@ -784,7 +784,7 @@ protected:
         std::is_same_v<T, std::int64_t> || std::is_same_v<T, std::uint64_t>) {
       ss << x;
     }
-    if (ss.tellp()) {
+    if (ss.tell()) {
       return ss.str();
     }
     if constexpr (std::is_same_v<T, Name> || HasSource<T>::value) {
@@ -830,13 +830,13 @@ protected:
 
 private:
   int indent_{0};
-  std::ostream &out_;
+  llvm::raw_ostream &out_;
   const AnalyzedObjectsAsFortran *const asFortran_;
   bool emptyline_{false};
 };
 
 template<typename T>
-void DumpTree(std::ostream &out, const T &x,
+void DumpTree(llvm::raw_ostream &out, const T &x,
     const AnalyzedObjectsAsFortran *asFortran = nullptr) {
   ParseTreeDumper dumper{out, asFortran};
   Walk(x, dumper);
--- a/include/flang/Parser/parsing.h
+++ b/include/flang/Parser/parsing.h
@@ -15,8 +15,8 @@
 #include "parse-tree.h"
 #include "provenance.h"
 #include "flang/Common/Fortran-features.h"
+#include "llvm/Support/raw_ostream.h"
 #include <optional>
-#include <ostream>
 #include <string>
 #include <utility>
 #include <vector>
@@ -50,19 +50,19 @@ public:
   std::optional<Program> &parseTree() { return parseTree_; }
 
   const SourceFile *Prescan(const std::string &path, Options);
-  void DumpCookedChars(std::ostream &) const;
-  void DumpProvenance(std::ostream &) const;
-  void DumpParsingLog(std::ostream &) const;
-  void Parse(std::ostream *debugOutput = nullptr);
+  void DumpCookedChars(llvm::raw_ostream &) const;
+  void DumpProvenance(llvm::raw_ostream &) const;
+  void DumpParsingLog(llvm::raw_ostream &) const;
+  void Parse(llvm::raw_ostream &debugOutput);
   void ClearLog();
 
-  void EmitMessage(std::ostream &o, const char *at, const std::string &message,
+  void EmitMessage(llvm::raw_ostream &o, const char *at, const std::string &message,
       bool echoSourceLine = false) const {
     cooked_.allSources().EmitMessage(
         o, cooked_.GetProvenanceRange(CharBlock(at)), message, echoSourceLine);
   }
 
-  bool ForTesting(std::string path, std::ostream &);
+  bool ForTesting(std::string path, llvm::raw_ostream &);
 
 private:
   Options options_;
--- a/include/flang/Parser/source.h
+++ b/include/flang/Parser/source.h
@@ -16,10 +16,13 @@
 
 #include "characters.h"
 #include <cstddef>
-#include <sstream>
 #include <string>
 #include <utility>
 #include <vector>
+
+namespace llvm {
+class raw_ostream;
+}
 
 namespace Fortran::parser {
 
@@ -44,8 +47,8 @@ public:
   std::size_t lines() const { return lineStart_.size(); }
   Encoding encoding() const { return encoding_; }
 
-  bool Open(std::string path, std::stringstream *error);
-  bool ReadStandardInput(std::stringstream *error);
+  bool Open(std::string path, llvm::raw_ostream &error);
+  bool ReadStandardInput(llvm::raw_ostream &error);
   void Close();
   SourcePosition FindOffsetLineAndColumn(std::size_t) const;
   std::size_t GetLineStartOffset(int lineNumber) const {
@@ -53,7 +56,7 @@ public:
   }
 
 private:
-  bool ReadFile(std::string errorPath, std::stringstream *error);
+  bool ReadFile(std::string errorPath, llvm::raw_ostream &error);
   void IdentifyPayload();
   void RecordLineStarts();
 
--- a/lib/Evaluate/formatting.cpp
+++ b/lib/Evaluate/formatting.cpp
@@ -14,11 +14,12 @@
 #include "flang/Evaluate/tools.h"
 #include "flang/Parser/characters.h"
 #include "flang/Semantics/symbol.h"
-#include <sstream>
+#include "llvm/Support/raw_ostream.h"
 
 namespace Fortran::evaluate {
 
-static void ShapeAsFortran(std::ostream &o, const ConstantSubscripts &shape) {
+static void ShapeAsFortran(
+    llvm::raw_ostream &o, const ConstantSubscripts &shape) {
   if (GetRank(shape) > 1) {
     o << ",shape=";
     char ch{'['};
@@ -31,7 +32,8 @@ static void ShapeAsFortran(std::ostream &o, const ConstantSubscripts &shape) {
 }
 
 template<typename RESULT, typename VALUE>
-std::ostream &ConstantBase<RESULT, VALUE>::AsFortran(std::ostream &o) const {
+llvm::raw_ostream &ConstantBase<RESULT, VALUE>::AsFortran(
+    llvm::raw_ostream &o) const {
   if (Rank() > 1) {
     o << "reshape(";
   }
@@ -71,8 +73,8 @@ std::ostream &ConstantBase<RESULT, VALUE>::AsFortran(std::ostream &o) const {
 }
 
 template<int KIND>
-std::ostream &Constant<Type<TypeCategory::Character, KIND>>::AsFortran(
-    std::ostream &o) const {
+llvm::raw_ostream &Constant<Type<TypeCategory::Character, KIND>>::AsFortran(
+    llvm::raw_ostream &o) const {
   if (Rank() > 1) {
     o << "reshape(";
   }
@@ -97,11 +99,12 @@ std::ostream &Constant<Type<TypeCategory::Character, KIND>>::AsFortran(
   return o;
 }
 
-std::ostream &ActualArgument::AssumedType::AsFortran(std::ostream &o) const {
+llvm::raw_ostream &ActualArgument::AssumedType::AsFortran(
+    llvm::raw_ostream &o) const {
   return o << symbol_->name().ToString();
 }
 
-std::ostream &ActualArgument::AsFortran(std::ostream &o) const {
+llvm::raw_ostream &ActualArgument::AsFortran(llvm::raw_ostream &o) const {
   if (keyword_) {
     o << keyword_->ToString() << '=';
   }
@@ -115,11 +118,11 @@ std::ostream &ActualArgument::AsFortran(std::ostream &o) const {
   }
 }
 
-std::ostream &SpecificIntrinsic::AsFortran(std::ostream &o) const {
+llvm::raw_ostream &SpecificIntrinsic::AsFortran(llvm::raw_ostream &o) const {
   return o << name;
 }
 
-std::ostream &ProcedureRef::AsFortran(std::ostream &o) const {
+llvm::raw_ostream &ProcedureRef::AsFortran(llvm::raw_ostream &o) const {
   for (const auto &arg : arguments_) {
     if (arg && arg->isPassedObject()) {
       arg->AsFortran(o) << '%';
@@ -306,7 +309,8 @@ static OperatorSpelling SpellOperator(const Relational<T> &x) {
 }
 
 template<typename D, typename R, typename... O>
-std::ostream &Operation<D, R, O...>::AsFortran(std::ostream &o) const {
+llvm::raw_ostream &Operation<D, R, O...>::AsFortran(
+    llvm::raw_ostream &o) const {
   Precedence lhsPrec{ToPrecedence(left())};
   OperatorSpelling spelling{SpellOperator(derived())};
   o << spelling.prefix;
@@ -337,7 +341,7 @@ std::ostream &Operation<D, R, O...>::AsFortran(std::ostream &o) const {
 }
 
 template<typename TO, TypeCategory FROMCAT>
-std::ostream &Convert<TO, FROMCAT>::AsFortran(std::ostream &o) const {
+llvm::raw_ostream &Convert<TO, FROMCAT>::AsFortran(llvm::raw_ostream &o) const {
   static_assert(TO::category == TypeCategory::Integer ||
           TO::category == TypeCategory::Real ||
           TO::category == TypeCategory::Character ||
@@ -355,21 +359,22 @@ std::ostream &Convert<TO, FROMCAT>::AsFortran(std::ostream &o) const {
   return o << ",kind=" << TO::kind << ')';
 }
 
-std::ostream &Relational<SomeType>::AsFortran(std::ostream &o) const {
+llvm::raw_ostream &Relational<SomeType>::AsFortran(llvm::raw_ostream &o) const {
   std::visit([&](const auto &rel) { rel.AsFortran(o); }, u);
   return o;
 }
 
 template<typename T>
-std::ostream &EmitArray(std::ostream &o, const Expr<T> &expr) {
+llvm::raw_ostream &EmitArray(llvm::raw_ostream &o, const Expr<T> &expr) {
   return expr.AsFortran(o);
 }
 
 template<typename T>
-std::ostream &EmitArray(std::ostream &, const ArrayConstructorValues<T> &);
+llvm::raw_ostream &EmitArray(
+    llvm::raw_ostream &, const ArrayConstructorValues<T> &);
 
 template<typename T>
-std::ostream &EmitArray(std::ostream &o, const ImpliedDo<T> &implDo) {
+llvm::raw_ostream &EmitArray(llvm::raw_ostream &o, const ImpliedDo<T> &implDo) {
   o << '(';
   EmitArray(o, implDo.values());
   o << ',' << ImpliedDoIndex::Result::AsFortran()
@@ -381,8 +386,8 @@ std::ostream &EmitArray(std::ostream &o, const ImpliedDo<T> &implDo) {
 }
 
 template<typename T>
-std::ostream &EmitArray(
-    std::ostream &o, const ArrayConstructorValues<T> &values) {
+llvm::raw_ostream &EmitArray(
+    llvm::raw_ostream &o, const ArrayConstructorValues<T> &values) {
   const char *sep{""};
   for (const auto &value : values) {
     o << sep;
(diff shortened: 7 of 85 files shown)


include/flang/Common/enum-set.h
```diff
@@ -16,11 +16,11 @@
 
 #include "constexpr-bitset.h"
 #include "idioms.h"
+#include "llvm/Support/raw_ostream.h"
```
DavidTruby: We can actually forward declare `llvm::raw_ostream` (unlike `std::ostream`) here.
So instead of including the header you can do:
```
namespace llvm {
class raw_ostream;
}
```
and then include the header in the cpp file instead for faster compile times.

This also applies for most of the headers below.

lib/Semantics/mod-file.h
```diff
@@ -33,17 +33,24 @@ class ModFileWriter {
...
+  llvm::raw_string_ostream useExtraAttrs_{useExtraAttrs_Buf};  // attrs added to used entity
+  llvm::raw_string_ostream decls_{decls_Buf};
+  llvm::raw_string_ostream contains_{contains_Buf};
 
   void WriteAll(const Scope &);
   void WriteOne(const Scope &);
   void Write(const Symbol &);
   std::string GetAsString(const Symbol &);
   void PutSymbols(const Scope &);
   void PutSymbol(std::stringstream &, const Symbol &);
```
DavidTruby: Can we remove this function?

lib/Semantics/mod-file.h
```diff
@@ -8,8 +8,8 @@
 
 #ifndef FORTRAN_SEMANTICS_MOD_FILE_H_
 #define FORTRAN_SEMANTICS_MOD_FILE_H_
-
 #include "flang/Semantics/attr.h"
+#include "llvm/Support/raw_ostream.h"
 #include <sstream>
```
DavidTruby: Can we remove this header?

lib/Semantics/scope.cpp
```diff
@@ -10,6 +10,7 @@
 #include "flang/Parser/characters.h"
 #include "flang/Semantics/symbol.h"
 #include "flang/Semantics/type.h"
+#include "llvm/Support/raw_ostream.h"
 #include <algorithm>
 #include <memory>
 #include <sstream>
```
DavidTruby: Can we remove this header?

lib/Semantics/tools.cpp
```diff
@@ -17,6 +17,7 @@
 #include "flang/Semantics/symbol.h"
 #include "flang/Semantics/tools.h"
 #include "flang/Semantics/type.h"
+#include "llvm/Support/raw_ostream.h"
 #include <algorithm>
 #include <set>
 #include <sstream>
```
DavidTruby: Can we remove this header?

runtime/descriptor.cpp
```diff
@@ -145,11 +146,13 @@ void Descriptor::Check() const {
...
 }
 
-std::ostream &Descriptor::Dump(std::ostream &o) const {
-  o << "Descriptor @ 0x" << std::hex << reinterpret_cast<std::intptr_t>(this)
-    << std::dec << ":\n";
-  o << "  base_addr 0x" << std::hex
-    << reinterpret_cast<std::intptr_t>(raw_.base_addr) << std::dec << '\n';
+llvm::raw_ostream &Descriptor::Dump(llvm::raw_ostream &o) const {
+ o << "Descriptor @ 0x"; 
+ o.llvm::raw_ostream::write_hex(reinterpret_cast<std::intptr_t>(this))
```
DavidTruby: Can this not just be o.write_hex? I don't know why the qualification would be necessary.
It also builds for me without the qualification.

test/Decimal/quick-sanity-test.cpp
```diff
@@ -14,11 +15,12 @@ union u {
...
 };
 
-std::ostream &failed(float x) {
+llvm::raw_ostream &failed(float x) {
   ++fails;
   union u u;
   u.x = x;
-  return std::cout << "FAIL: 0x" << std::hex << u.u << std::dec;
+  llvm::outs() << "FAIL: 0x";
+ return llvm::outs().write_hex(u.u);
```
DavidTruby: See here where it's used unqualified.

test/Evaluate/intrinsics.cpp
```diff
@@ -4,6 +4,7 @@
 #include "flang/Evaluate/expression.h"
 #include "flang/Evaluate/tools.h"
 #include "flang/Parser/provenance.h"
+#include "llvm/Support/raw_ostream.h"
 #include <initializer_list>
 #include <iostream>
```
DavidTruby: Use of this header is forbidden in LLVM. We should be able to remove it now?

test/Runtime/hello.cpp
```diff
@@ -2,6 +2,7 @@
 
 #include "../../runtime/descriptor.h"
 #include "../../runtime/io-api.h"
+#include "llvm/Support/raw_ostream.h"
 #include <cstring>
 #include <iostream>
```
DavidTruby: Ditto here.

include/flang/Common/enum-set.h
schweitzpgi: +1

lib/Evaluate/formatting.cpp
```diff
@@ -14,13 +14,13 @@
 #include "flang/Evaluate/tools.h"
 #include "flang/Parser/characters.h"
 #include "flang/Semantics/symbol.h"
-#include <sstream>
+#include "llvm/Support/raw_ostream.h"
```
schweitzpgi: Interesting that it included string stream.

lib/Semantics/mod-file.h
```diff
@@ -33,17 +33,24 @@ class ModFileWriter {
...
   SemanticsContext &context_;
-  std::stringstream uses_;
-  std::stringstream useExtraAttrs_;  // attrs added to used entity
-  std::stringstream decls_;
-  std::stringstream contains_;
+  // Buffer to use with raw_string_ostream
+  std::string uses_Buf;
+  std::string useExtraAttrs_Buf;
+  std::string decls_Buf;
+  std::string contains_Buf;
```
schweitzpgi: Do these buffers need to be part of this class? Or can they be provided by the calling context and destructed there?

runtime/descriptor.cpp
```diff
@@ -172,9 +175,9 @@ std::size_t DescriptorAddendum::SizeInBytes() const {
   return SizeInBytes(LenParameters());
 }
 
-std::ostream &DescriptorAddendum::Dump(std::ostream &o) const {
-  o << "  derivedType @ 0x" << std::hex
-    << reinterpret_cast<std::intptr_t>(derivedType_) << std::dec << '\n';
+llvm::raw_ostream &DescriptorAddendum::Dump(llvm::raw_ostream &o) const {
+  o << "  derivedType @ 0x"; o.llvm::raw_ostream::write_hex(
```
schweitzpgi: You don't need the namespace and class name here.
Also maybe needs a clang-format pass.

test/Evaluate/intrinsics.cpp
schweitzpgi: This isn't library code. It's a test program to generate tests.

test/Evaluate/real.cpp
```diff
@@ -398,7 +400,9 @@ void subsetTests(int pass, Rounding rounding, std::uint32_t opds) {
       ("%d IsInfinite(0x%jx)", pass, static_cast<std::intmax_t>(rj));
 
       static constexpr int kind{REAL::bits / 8};
-      std::stringstream ss, css;
+      std::string ssBuf, cssBuf;
+      llvm::raw_string_ostream ss {ssBuf};
+      llvm::raw_string_ostream css {cssBuf};
```
schweitzpgi: Make sure to run clang-format. (I believe it will squish out all whitespace around the braces.)

CarolineConcatto: So, I did not removed iostream from this file and others because it did not interact with the other files that I needed to change. This file has this function:

test/Evaluate/intrinsics.cpp
CarolineConcatto: This file has this method:
 void DoCall(std::optional<DynamicType> resultType = std::nullopt, int rank = 0, bool isElemental = false)
That uses std::cout, as the compiler did not complain about it I did not change. For that reason also that this is there.

CarolineConcatto: I can do another commit removing all "#include <iostream>" in this PR if you all think it is fine. 
ATM this PR removes all uses of std::ostream in the code. I can also remove all the std::cout.

include/flang/Common/enum-set.h
CarolineConcatto: Ok.

lib/Semantics/tools.cpp
CarolineConcatto: According to our talks it will need, because of this: 
 llvm::raw_string_ostream ss {buf};
ss << "node has not been analyzed:\n";

include/flang/Common/enum-set.h
CarolineConcatto: This header has this method llvm::raw_ostream &Dump implemented in enum-set, so the compile complains if I remove the header. 
It is ok to have header + namespace?
llvm::raw_ostream &Dump(
      llvm::raw_ostream &o, std::string EnumToString(enumerationType)) const {
    char sep{'{'};
    IterateOverMembers([&](auto e) {
      o << sep << EnumToString(e);
      sep = ',';
    });

DavidTruby: I think we should move this function definition to the corresponding cpp file instead and just leave the declaration here

lib/Semantics/tools.cpp
DavidTruby: I meant the sstream header here not the llvm one

lib/Semantics/mod-file.h
CarolineConcatto: This one is ok, as well as #include <sstream>.

lib/Semantics/scope.cpp
CarolineConcatto: I believe not because of this method:

test/Evaluate/intrinsics.cpp
CarolineConcatto: Another point is that I needed to all llvm because of this method:
void DoCall that uses AsFortran method and all AsFortran method was changed to use llvm:raw_ostream

include/flang/Common/enum-set.h
CarolineConcatto: Should we do this in This PR or should I make another one? There are more header files that were implemented this way? We need to merge this before merging with LLVM.

DavidTruby: I think this should be quite a small change to this PR so you should just do it here in my opinion

include/flang/Evaluate/formatting.h
```diff
@@ -9,44 +9,44 @@
...
 template<typename A>
-auto operator<<(std::ostream &o, const A &x) -> decltype(o << x.AsFortran()) {
+auto operator<<(llvm::raw_ostream &o, const A &x) -> decltype(o << x.AsFortran()) {
   return o << x.AsFortran();
 }
 
 template<typename A, bool COPYABLE>
 auto operator<<(
-    std::ostream &o, const Fortran::common::Indirection<A, COPYABLE> &x)
+    llvm::raw_ostream &o, const Fortran::common::Indirection<A, COPYABLE> &x)
```
CarolineConcatto: So this is another example where the header file implements a method.

include/flang/Parser/dump-parse-tree.h
```diff
@@ -761,7 +760,8 @@ class ParseTreeDumper {
 protected:
   // Return a Fortran representation of this node to include in the dump
   template<typename T> std::string AsFortran(const T &x) {
-    std::ostringstream ss;
+    std::string buf;
+    llvm::raw_string_ostream ss {buf};
```
CarolineConcatto: Another header file with a method implementation.

include/flang/Parser/parsing.h
```diff
@@ -50,19 +50,19 @@ class Parsing {
...
-  void DumpParsingLog(std::ostream &) const;
-  void Parse(std::ostream *debugOutput = nullptr);
+  void DumpCookedChars(llvm::raw_ostream &) const;
+  void DumpProvenance(llvm::raw_ostream &) const;
+  void DumpParsingLog(llvm::raw_ostream &) const;
+  void Parse(llvm::raw_ostream *debugOutput = nullptr);
   void ClearLog();
 
-  void EmitMessage(std::ostream &o, const char *at, const std::string &message,
+  void EmitMessage(llvm::raw_ostream &o, const char *at, const std::string &message,
```
CarolineConcatto: Another header file with a header Implementation

lib/Evaluate/formatting.cpp
CarolineConcatto: I believe that the reason was because of this function:
std::string ExpressionBase<RESULT>::AsFortran() const {
  std::ostringstream ss;
I've removed with the new commit.

runtime/descriptor.h
```diff
@@ -20,13 +20,16 @@
 
 #include "derived-type.h"
 #include "type-code.h"
-#include "flang/ISO_Fortran_binding.h"
```
CarolineConcatto: Adding it back.

CarolineConcatto: Removed the wrong header

test/Evaluate/CMakeLists.txt
```diff
@@ -17,6 +17,7 @@ add_executable(leading-zero-bit-count-test
 
 target_link_libraries(leading-zero-bit-count-test
   FortranEvaluateTesting
+  LLVMSupport
```
CarolineConcatto: I've added LLVMSupport in all the target_link_libraries that used FortranEvaluateTesting, because FortranEvaluateTesting used testing.cpp that used llvm::outs() and llvm::errs().

include/flang/Common/enum-set.h
CarolineConcatto: We think that these changes may lead to further changes. For now, I will leave this task for later as we have a deadline for this PR.

test/Evaluate/real.cpp
CarolineConcatto: Done

include/flang/Evaluate/formatting.h
CarolineConcatto: This task can be visited later when we have a merged with LLVM

include/flang/Parser/dump-parse-tree.h
CarolineConcatto: This task can be visited later when we have a merged with LLVM

include/flang/Parser/parsing.h
CarolineConcatto: This task can be visited later when we have a merged with LLVM

runtime/descriptor.h
CarolineConcatto: Done!

include/flang/Parser/dump-parse-tree.h
kiranchandramohan: Is there a formatting issue here?

include/flang/Parser/source.h
```diff
@@ -44,16 +47,16 @@ class SourceFile {
   std::size_t lines() const { return lineStart_.size(); }
   Encoding encoding() const { return encoding_; }
 
-  bool Open(std::string path, std::stringstream *error);
-  bool ReadStandardInput(std::stringstream *error);
+  bool Open(std::string path, llvm::raw_ostream *error);
+  bool ReadStandardInput(llvm::raw_ostream *error);
```
kiranchandramohan: Why is this not raw_string_ostream?

include/flang/Parser/source.h
```diff
@@ -44,16 +47,16 @@ class SourceFile {
...
   void Close();
   SourcePosition FindOffsetLineAndColumn(std::size_t) const;
   std::size_t GetLineStartOffset(int lineNumber) const {
     return lineStart_.at(lineNumber - 1);
   }
 
 private:
-  bool ReadFile(std::string errorPath, std::stringstream *error);
+  bool ReadFile(std::string errorPath, llvm::raw_ostream *error);
```
kiranchandramohan: raw_string_ostream?

lib/Parser/source.cpp
```diff
@@ -111,7 +112,7 @@ static std::size_t RemoveCarriageReturns(char *buffer, std::size_t bytes) {
   return wrote;
 }
 
-bool SourceFile::Open(std::string path, std::stringstream *error) {
+bool SourceFile::Open(std::string path, llvm::raw_ostream *error) {
```
kiranchandramohan: raw_string_ostream? below also.

lib/Semantics/expression.cpp
```diff
@@ -2502,7 +2503,7 @@ bool ExpressionAnalyzer::EnforceTypeConstraint(parser::CharBlock at,
     const MaybeExpr &result, TypeCategory category, bool defaultKind) {
   if (result) {
     if (auto type{result->GetType()}) {
-      if (type->category() != category) { // C885
+      if (type->category() != category) {  // C885
```
kiranchandramohan: additional space before comment?

lib/Semantics/mod-file.cpp
```diff
@@ -186,7 +193,7 @@ void ModFileWriter::PutSymbols(const Scope &scope) {
 // Emit a symbol to decls_, except for bindings in a derived type (type-bound
 // procedures, type-bound generics, final procedures) which go to typeBindings.
 void ModFileWriter::PutSymbol(
-    std::stringstream &typeBindings, const Symbol &symbol) {
+    llvm::raw_ostream &typeBindings, const Symbol &symbol) {
```
kiranchandramohan: raw_string_ostream?

lib/Semantics/mod-file.h
```diff
@@ -33,17 +36,23 @@ class ModFileWriter {
...
+  llvm::raw_string_ostream decls_{decls_Buf};
+  llvm::raw_string_ostream contains_{contains_Buf};
 
   void WriteAll(const Scope &);
   void WriteOne(const Scope &);
   void Write(const Symbol &);
   std::string GetAsString(const Symbol &);
   void PutSymbols(const Scope &);
-  void PutSymbol(std::stringstream &, const Symbol &);
+  void PutSymbol(llvm::raw_ostream &, const Symbol &);
```
kiranchandramohan: raw_string_ostream?

test/Evaluate/intrinsics.cpp
```diff
@@ -88,17 +88,17 @@ struct TestCall {
...
       sep = ',';
-      a->AsFortran(std::cout);
+      a->AsFortran(llvm::outs());
     }
     if (sep == '(') {
-      std::cout << '(';
+      llvm::outs() << '(';
     }
-    std::cout << ')' << std::endl;
+    llvm::outs() << ')' << '\n';
```
kiranchandramohan: should there be a flush?

test/Evaluate/ISO-Fortran-binding.cpp
```diff
@@ -71,13 +69,12 @@ static void AddNoiseToCdesc(CFI_cdesc_t *dv, CFI_rank_t rank) {
...
-            << " type: " << static_cast<int>(ty) << " elem_len: " << eLen
-            << " rank: " << static_cast<int>(rank) << " extent: " << std::hex
-            << reinterpret_cast<std::intptr_t>(eAddr) << std::endl
-            << std::dec;
+  llvm::outs() << " base_addr: ";
+  llvm::outs().write_hex(reinterpret_cast<std::intptr_t>(bAddr))
+      << " attribute: " << static_cast<int>(attr)
+      << " type: " << static_cast<int>(ty) << " elem_len: " << eLen
+      << " rank: " << static_cast<int>(rank) << " extent: ";
+  llvm::outs().write_hex(reinterpret_cast<std::intptr_t>(eAddr)) << '\n';
```
kiranchandramohan: should there be a flush?

include/flang/Parser/source.h
DavidTruby: There's no need for these functions to take the more specific type really, we may for example eventually want to call or test them with a different stream type.

They should be references rather than pointers though now that you point it out

include/flang/Parser/dump-parse-tree.h
DavidTruby: What's wrong with this formatting?

lib/Semantics/mod-file.h
```diff
@@ -33,17 +36,23 @@ class ModFileWriter {
 
 private:
   SemanticsContext &context_;
-  std::stringstream uses_;
-  std::stringstream useExtraAttrs_;  // attrs added to used entity
-  std::stringstream decls_;
-  std::stringstream contains_;
+  // Buffer to use with raw_string_ostream
+  std::string uses_Buf;
```
tskeith: Should be `usesBuf_`.

include/flang/Parser/dump-parse-tree.h
kiranchandramohan: Thre is a space between ss and the brace.

lib/Semantics/mod-file.cpp
```diff
@@ -157,13 +158,17 @@ std::string ModFileWriter::GetAsString(const Symbol &symbol) {
     all << ") " << symbol.name();
   }
   all << '\n' << uses_.str();
-  uses_.str(""s);
+  uses_.flush();
+  uses_Buf.clear();
```
kiranchandramohan: Can this be uses_.str().clear() ?

include/flang/Parser/source.h
kiranchandramohan: OK. If that is the case then please keep it consistent or provide justification on why raw_string_stream is used in functions Open and ReadStandardInput in Parser/provenance.cpp and TokenSequence::Put in Parser/token-sequence.cpp

include/flang/Parser/dump-parse-tree.h
CarolineConcatto: I did not run clang-format-7 on the header files. Only on the cpp.

test/Evaluate/ISO-Fortran-binding.cpp
CarolineConcatto: Ok, lesson learn. If there was a std::endl we need to replace by '\n' and flush().

lib/Semantics/mod-file.cpp
CarolineConcatto: I really dont know, because when I grep on llvm  there is only one use of something.str().clear. But lots of something.clear() for llvm::raw_string_stream.
@DavidTruby what do you think?

lib/Semantics/expression.cpp
CarolineConcatto: It was added by clang-format-7 for some reason.

lib/Parser/source.cpp
```diff
@@ -111,36 +112,36 @@ static std::size_t RemoveCarriageReturns(char *buffer, std::size_t bytes) {
...
-bool SourceFile::Open(std::string path, std::stringstream *error) {
+bool SourceFile::Open(std::string path, llvm::raw_ostream &error) {
   Close();
   path_ = path;
   std::string errorPath{"'"s + path + "'"};
   errno = 0;
   fileDescriptor_ = open(path.c_str(), O_RDONLY);
   if (fileDescriptor_ < 0) {
-    *error << "Could not open " << errorPath << ": " << std::strerror(errno);
+    error << "Could not open " << errorPath << ": " << std::strerror(errno);
```
kiranchandramohan: std::strerror -> llvm::sys::StrError? 
A few other occurences as well.

lib/Parser/parsing.cpp
```diff
@@ -94,13 +95,15 @@ void Parsing::DumpCookedChars(std::ostream &out) const {
...
+  cooked_.Dump(out);
+}
 
-void Parsing::DumpParsingLog(std::ostream &out) const {
+void Parsing::DumpParsingLog(llvm::raw_ostream &out) const {
   log_.Dump(out, cooked_);
 }
 
-void Parsing::Parse(std::ostream *out) {
+void Parsing::Parse(llvm::raw_ostream *out) {
```
kiranchandramohan: Is it possible to remove the pointer here also?

lib/Parser/source.cpp
CarolineConcatto: So, the compiler did not complain about that, and I believe it belongs to #include <cerrno>. Moreover, there is nothing in the LLVM Code style talking about this function.
But maybe you are right because I do see llvm::sys::StrError being used in llvm-project.
@DavidTruby ??

lib/Semantics/mod-file.cpp
CarolineConcatto: Talking with @DavidTruby it looks either is fine. But maybe the one proposed by Kiran in cleaner.

lib/Parser/parsing.cpp
CarolineConcatto: The problem with this function is its initialization in parsing.h
void Parsing::Parse(llvm::raw_ostream *out = nulltptr)
And it is initialized like that because in lib/Parser/parsing.cpp:130 it calls Parse without any parameter:

DavidTruby: Llvm has a raw_null_ostream that you can pass instead, and you can then remove the null pointer checking code and just always print to out as it will get discarded when printing go raw_null_ostream.

lib/Parser/source.cpp
CarolineConcatto: Ok, so it looks there is nothing asking to change, but the llvm one is thread-safe. So, for this reason(being thread-safe)[1], I am replacing std::strerror by  llvm::sys::StrError.
[1]https://llvm.org/doxygen/Errno_8cpp_source.html

lib/Parser/parsing.cpp
DavidTruby: What I'm suggesting is that you could default the parameter to llvm::raw_null_ostream instead of having it as a pointer and defaulting it to null. The current solution with the pointer is probably fine though.

lib/Parser/source.cpp
DavidTruby: llvm::sys::StrError has the same visible behaviour, but is thread safe where possible. I think we should use it over std::strerror for that reason.

CarolineConcatto: Done!

lib/Semantics/mod-file.cpp
CarolineConcatto: Done!

lib/Semantics/mod-file.h
CarolineConcatto: Done

test/Evaluate/ISO-Fortran-binding.cpp
CarolineConcatto: Done for this file and others as well.

lib/Semantics/mod-file.cpp
CarolineConcatto: We've decided to replace all std::stringstream  parameters  by llvm::raw_ostream ones.

lib/Semantics/expression.cpp
CarolineConcatto: I did not run clang-format in this file as it adds the extra space.

lib/Semantics/mod-file.h
CarolineConcatto: As far as I understand yes, it needs to be here. But  I double-check with @DavidTruby and he agrees.

CarolineConcatto:
@kiranchandramohan, if it is fine I rather do the message update when I do the squash for this PR.

kiranchandramohan:
Yes, it is fine.

lib/Semantics/mod-file.h
DavidTruby: If we don't have these here, the strings won't persist across member function calls, and these functions rely on that behaviour sometimes. Otherwise we could just have the streams themselves also not be part of the class.

lib/Semantics/mod-file.cpp
schweitzpgi: +1 for improved flexibility.

lib/Semantics/mod-file.h
schweitzpgi: To be clear, I wasn't suggesting the backing strings be constructed and destroyed in some interleaved way. Clearly they need to persist throughout the context for correct output. This seems like a fine place to put them as they'll have the same lifetime as the streams. On the other hand, the object that owns the streams had explicit ways to reset the buffers at one point, which suggested that a single writer object might be reused to generate multiple outputs.

CarolineConcatto:
Ok it has failed but it is not a problem with the PR, but memory problem.  2 runs are ok and 2 failed

kiranchandramohan:
@CarolineConcatto Thanks for pointing out that using the strerror from llvm in the runtime will introduce a dependency on the LLVM Runtime. You are correct, please remove the llvm function from the runtime and remove LLVMSupport from runtime/CMakeLists.txt.

CarolineConcatto:
@sscalpone I believe I've done the squash now!

sscalpone:
@CarolineConcatto Conflicts must be resolved.

runtime/descriptor.h
```diff
@@ -28,6 +28,10 @@
 #include <cstdio>
 #include <cstring>
 
+namespace llvm {
+class raw_ostream;
+}
+
```
isuruf: Should these be removed?

CarolineConcatto: Done!

tools/f18/f18.cpp
```diff
@@ -123,14 +125,14 @@ bool ParentProcess() {
...
     }
-    std::cerr << '\n';
+    llvm::errs() << '\n';
   }
   argv.push_back(nullptr);
   execvp(argv[0], &argv[0]);
-  std::cerr << "execvp(" << argv[0] << ") failed: " << std::strerror(errno)
-            << '\n';
+  llvm::errs() << "execvp(" << argv[0] << ") failed: " << llvm::sys::StrError(errno)
+               << '\n';
```
tskeith: It looks like this needs reformatting. Maybe other files too.

sscalpone:
A quick grep shows a few instances of #include <ostream>, e.g. message.h.

sscalpone:
Does the comment referencing ostream and cerr need adjustment?

tools/f18/f18.cpp
CarolineConcatto: This is clang-format-7! I can re-run to see if it will change.

CarolineConcatto:
Working on this again! Probably they were introduced by Friday. It would be nice if we could merge this, otherwise, I will need to re-do the works every-time there is a merge.

CarolineConcatto:
@sscalpon I've changed the message a little bit. I don't know if that is what you want, but let me know. I squashed also the patch and we should not have any conflict by now!

tools/f18/f18.cpp
CarolineConcatto: Done!

CarolineConcatto:
Memory problems again make it fail some tests, but the rest are ok!

CarolineConcatto:
Hey @sscalpone,
Sorry for me asking but, what dump.h?
I could not find any file called dump.h I could find this one:
include/flang/Parser/dump-parse-tree.h.
If I can remember, I needed to change because of AsFortran method. It uses stringstream and the compiler was complaining. I could not find any use of std:cerr in this file too. 
So I believe I am looking to the wrong file. Do you mind to point me to what file are you seeing the problem?

sscalpone:
Oops.

./tools/f18/dump.cpp:// reference and pointer, and for dumping to a provided ostream or cerr.

CarolineConcatto:
Just in case:
There are std::stringstream or #include <sstream>  in flatten.cpp. This file is not the compiler, is Peter's squash program. 
std::cout will be in the documentation/ImplementingASemanticCheck.md. Which I believe should not be changed because they explain how someone had implemented the semantic checks in the compiler.
std::cerr will be in documentation/ParserCombinators.md. Also another documentation file, that maybe needs change.

CarolineConcatto:
Hey @sscalpone I can change that message np.

documentation/ParserCombinators.md
```diff
@@ -160,5 +160,5 @@ is built.  All of the following parsers consume characters acquired from
 
 ### Debugging Parser
 Last, a string literal `"..."_debug` denotes a parser that emits the string to
-`std::cerr` and succeeds.  It is useful for tracing while debugging a parser but should
+`llv::errs` and succeeds.  It is useful for tracing while debugging a parser but should
```
sscalpone: Is this really llv?  Sorry I didn't look it up.

sscalpone:
@CarolineConcatto Conflicts need resolving.

sscalpone:
Is the comment in `tools/f18/dump.cpp` that mentions ostream still accurate?  I don't know if you can pass ostream to raw_ostream.

CarolineConcatto:
I believe the comment is about void Dump method, right?
It is overloaded. That method can dump raw_ostream or llvm:errs.

I was writing that this method dumps still dumps raw_ostream, that I would still leave as ostream. But by the confusion I've made, it is better to replace ostream by raw_ostream to avoid further confusion.

documentation/ParserCombinators.md
CarolineConcatto: Sorry @sscalpone. It looks like I cannot differentiate 2 l's from one l anymore. :(
I've looked too much for this.

CarolineConcatto: Done

sscalpone:
@CarolineConcatto There's another conflict :(

CarolineConcatto:
@sscalpone conflict solved!

CarolineConcatto:
Again 1 test configuration out of 4 failed. Patch is still good!

PR #1050, opened 2020-03-06
Perform definability checks on LHS of assignment

tskeith:
If the pure context check succeeds, call `WhyNotModifiable` to verify
the LHS can be modified.

Detect assignment to whole assumed-size array.

Change `IsVariable` to return false for a parameter or a component or
array reference whose base it a parameter.

When analyzing an assignment statement, report an error if the LHS is
a constant expression. Otherwise it might get folded and when we detect
the problem later the error will be confusing.

Fix IsImpliedShape and IsAssumedSize predicates in ArraySpec.

Fix C709 check in check-declarations.cpp.

### DIFF
--- a/lib/Evaluate/tools.cpp
+++ b/lib/Evaluate/tools.cpp
@@ -11,6 +11,7 @@
 #include "flang/Evaluate/characteristics.h"
 #include "flang/Evaluate/traverse.h"
 #include "flang/Parser/message.h"
+#include "flang/Semantics/tools.h"
 #include <algorithm>
 #include <variant>
 
@@ -37,7 +38,31 @@ Expr<SomeType> Parenthesize(Expr<SomeType> &&expr) {
       std::move(expr.u));
 }
 
+std::optional<DataRef> ExtractDataRef(const Substring &substring) {
+  return std::visit(
+      common::visitors{
+          [&](const DataRef &x) -> std::optional<DataRef> { return x; },
+          [&](const StaticDataObject::Pointer &) -> std::optional<DataRef> {
+            return std::nullopt;
+          },
+      },
+      substring.parent());
+}
+
 // IsVariable()
+
+auto IsVariableHelper::operator()(const Symbol &symbol) const -> Result {
+  return !symbol.attrs().test(semantics::Attr::PARAMETER);
+}
+auto IsVariableHelper::operator()(const Component &x) const -> Result {
+  return (*this)(x.base());
+}
+auto IsVariableHelper::operator()(const ArrayRef &x) const -> Result {
+  return (*this)(x.base());
+}
+auto IsVariableHelper::operator()(const Substring &x) const -> Result {
+  return (*this)(x.GetBaseObject());
+}
 auto IsVariableHelper::operator()(const ProcedureDesignator &x) const
     -> Result {
   const Symbol *symbol{x.GetSymbol()};
--- a/lib/Semantics/assignment.cpp
+++ b/lib/Semantics/assignment.cpp
@@ -66,10 +66,23 @@ void AssignmentContext::Analyze(const parser::AssignmentStmt &stmt) {
     const SomeExpr &rhs{assignment->rhs};
     auto lhsLoc{std::get<parser::Variable>(stmt.t).GetSource()};
     auto rhsLoc{std::get<parser::Expr>(stmt.t).source};
+    auto shape{evaluate::GetShape(foldingContext(), lhs)};
+    if (shape && !shape->empty() && !shape->back().has_value()) {  // C1014
+      Say(lhsLoc,
+          "Left-hand side of assignment may not be a whole assumed-size array"_err_en_US);
+    }
+    if (CheckForPureContext(lhs, rhs, rhsLoc, false)) {
+      const Scope &scope{context_.FindScope(lhsLoc)};
+      if (auto whyNot{WhyNotModifiable(lhsLoc, lhs, scope)}) {
+        if (auto *msg{Say(lhsLoc,
+                "Left-hand side of assignment is not modifiable"_err_en_US)}) {
+          msg->Attach(*whyNot);
+        }
+      }
+    }
     if (whereDepth_ > 0) {
       CheckShape(lhsLoc, &lhs);
     }
-    CheckForPureContext(lhs, rhs, rhsLoc, false);
   }
 }
 
@@ -169,7 +182,8 @@ bool AssignmentContext::CheckForPureContext(const SomeExpr &lhs,
         // ASSOCIATE(a=>x) -- check x, not a, for "a=..."
         base = dataRef ? &dataRef->GetFirstSymbol() : nullptr;
       }
-      if (!CheckDefinabilityInPureScope(messages, *base, scope, *pure)) {
+      if (base &&
+          !CheckDefinabilityInPureScope(messages, *base, scope, *pure)) {
         return false;
       }
     }
--- a/lib/Semantics/assignment.cpp
+++ b/lib/Semantics/assignment.cpp
@@ -43,7 +43,7 @@ public:
   void Analyze(const parser::ConcurrentControl &);
 
 private:
-  void CheckForPureContext(const SomeExpr &lhs, const SomeExpr &rhs,
+  bool CheckForPureContext(const SomeExpr &lhs, const SomeExpr &rhs,
       parser::CharBlock rhsSource, bool isPointerAssignment);
   void CheckShape(parser::CharBlock, const SomeExpr *);
   template<typename... A>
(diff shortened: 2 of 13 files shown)


lib/Semantics/assignment.cpp
```diff
@@ -43,7 +43,7 @@ class AssignmentContext {
   void Analyze(const parser::ConcurrentControl &);
 
 private:
-  void CheckForPureContext(const SomeExpr &lhs, const SomeExpr &rhs,
+  bool CheckForPureContext(const SomeExpr &lhs, const SomeExpr &rhs,
```
psteinfeld: This function name is not that evocative of its return value.  It would be nice to have a name that gave an indication of the meaning of the return result.  How about "AssignableInPureContext()"?

lib/Semantics/assignment.cpp
```diff
@@ -110,16 +125,18 @@ static const char *WhyBaseObjectIsSuspicious(
   }
 }
 
-// Checks C1594(1,2)
-void CheckDefinabilityInPureScope(parser::ContextualMessages &messages,
+// Checks C1594(1,2); false if check fails
+bool CheckDefinabilityInPureScope(parser::ContextualMessages &messages,
```
psteinfeld: Similarly, how about calling this "IsDefinableInPureScope()"?

lib/Semantics/assignment.cpp
```diff
@@ -135,21 +152,24 @@ static std::optional<std::string> GetPointerComponentDesignatorName(
   return std::nullopt;
 }
 
-// Checks C1594(5,6)
-void CheckCopyabilityInPureScope(parser::ContextualMessages &messages,
+// Checks C1594(5,6); false if check fails
+bool CheckCopyabilityInPureScope(parser::ContextualMessages &messages,
```
psteinfeld: Another case.

lib/Semantics/assignment.cpp
```diff
@@ -135,21 +152,24 @@ static std::optional<std::string> GetPointerComponentDesignatorName(
...
             base->name(), why, *pointer);
+        return false;
       }
     }
   }
+  return true;
 }
 
-void AssignmentContext::CheckForPureContext(const SomeExpr &lhs,
+bool AssignmentContext::CheckForPureContext(const SomeExpr &lhs,
```
psteinfeld: Another case.

lib/Semantics/assignment.cpp
```diff
@@ -176,29 +195,31 @@ void AssignmentContext::CheckForPureContext(const SomeExpr &lhs,
...
-                bad.BuildResultDesignatorName());
-          } else {
-            CheckCopyabilityInPureScope(messages, rhs, scope);
-          }
+    } else if (auto type{evaluate::DynamicType::From(lhs)}) {
+      // C1596 checks for polymorphic deallocation in a pure subprogram
+      // due to automatic reallocation on assignment
+      if (type->IsPolymorphic()) {
+        context_.Say(
+            "Deallocation of polymorphic object is not permitted in a pure subprogram"_err_en_US);
```
psteinfeld: This check was already there, but I don't see a test the produces this message.

lib/Semantics/tools.cpp
```diff
@@ -789,29 +789,23 @@ std::optional<parser::MessageFixedText> WhyNotModifiable(
...
-        if (auto maybeWhy{WhyNotModifiable(symbol, scope)}) {
-          return std::make_unique<parser::Message>(symbol.name(),
-              parser::MessageFormattedText{
-                  std::move(*maybeWhy), symbol.name()});
-        }
-      }
-    } else {
-      // reference to function returning POINTER
+  if (!evaluate::IsVariable(expr)) {
+    return parser::Message{at, "Expression is not a variable"_en_US};
```
psteinfeld: Thanks for improving these messages.  But I notice that there are no tests that produce either of them.

lib/Semantics/assignment.cpp
klausler: That's an adjective.

lib/Semantics/assignment.cpp
```diff
@@ -66,10 +66,25 @@ void AssignmentContext::Analyze(const parser::AssignmentStmt &stmt) {
     const SomeExpr &rhs{assignment->rhs};
     auto lhsLoc{std::get<parser::Variable>(stmt.t).GetSource()};
     auto rhsLoc{std::get<parser::Expr>(stmt.t).source};
+    if (const Symbol * symbol{UnwrapWholeSymbolDataRef(lhs)}) {
+      if (IsAssumedSizeArray(*symbol)) {  // C1014
```
klausler: This code catches `A=...` but will it miss `A(:)=...` or `A%X=...`?

lib/Semantics/assignment.cpp
klausler: It's not a pure predicate function; it has side effects.  `Check...` is fine.

test/Semantics/resolve72.f90
```diff
@@ -19,7 +19,7 @@ subroutine inner1(arg1, arg2, arg3, arg4, arg5, arg6, arg7, arg8)
         type(*), pointer :: arg6
         !ERROR: Assumed-type argument 'arg7' cannot have the VALUE attribute
         type(*), value :: arg7
-        !ERROR: Assumed-type argument 'arg8' must be assumed shape or assumed size array
+        !ERROR: Assumed-type argument 'arg8' may not be an explicit-shape array
```
klausler: I'm not sure that the new message is an improvement.  The existing message implies how the problem can be fixed.

tskeith:
> Your changes are failing call12.f90 semantic test (segfault, only with gcc). I do not think it has to do with my comment about suspicious deference though, it looks like a trickier issue.

My change to `expression.cpp` fixes this. The crash happens for me with gcc 9.2 but not 8.3 or clang 9.0. But with clang the address sanitizer reports a problem. I can't fully convince myself this is a program bug and not a compiler bug. @jeanPerier, @klausler: what do you think?

lib/Evaluate/tools.cpp
```diff
@@ -38,6 +38,16 @@ Expr<SomeType> Parenthesize(Expr<SomeType> &&expr) {
...
+
+auto IsVariableHelper::operator()(const Symbol &symbol) const -> Result {
+  return !symbol.attrs().test(semantics::Attr::PARAMETER);
+}
+auto IsVariableHelper::operator()(const Component &x) const -> Result {
+  return (*this)(x.base());
+}
+auto IsVariableHelper::operator()(const ArrayRef &x) const -> Result {
+  return (*this)(x.base());
+}
```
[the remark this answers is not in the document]
tskeith: Thanks for catching that, I've extended it to include Substring.

lib/Semantics/assignment.cpp
```diff
@@ -160,13 +180,12 @@ void AssignmentContext::CheckForPureContext(const SomeExpr &lhs,
...
-          CheckDefinabilityInPureScope(
-              messages, dataRef->GetFirstSymbol(), scope, *pure);
-        }
-      } else {
-        CheckDefinabilityInPureScope(messages, *base, scope, *pure);
+        auto dataRef{ExtractDataRef(assoc->expr())};
+        // ASSOCIATE(a=>x) -- check x, not a, for "a=..."
+        base = dataRef ? &dataRef->GetFirstSymbol() : nullptr;
+      }
+      if (!CheckDefinabilityInPureScope(messages, *base, scope, *pure)) {
```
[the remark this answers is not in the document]
tskeith: Fixed

lib/Semantics/assignment.cpp
tskeith: The main purpose of the function is to perform the check and emit error messages if it fails. The result is just to indicate that they passed so we can continue with other checks. So I think the current name is better.

lib/Semantics/tools.cpp
tskeith: These aren't errors so we don't have a way to test them currently. I believe that some of them are emitted.

test/Semantics/resolve72.f90
tskeith: OK, I've changed it back and added the "assumed rank" part.

jeanPerier:
> My change to expression.cpp fixes this. The crash happens for me with gcc 9.2 but not 8.3 or clang 9.0. But with clang the address sanitizer reports a problem. I can't fully convince myself this is a program bug and not a compiler bug. @jeanPerier, @klausler: what do you think?

I am not sure, I cannot point at what would be the program bug either, but  `*dataRef = std::move(component->base());` with optional, variants and CopyableIndirection involved touches enough C++ standard so that I am not sure there is not a tricky rule at play. In any-case, the fix is simpler to understand.

So, looks good to me !

klausler:
> > Your changes are failing call12.f90 semantic test (segfault, only with gcc). I do not think it has to do with my comment about suspicious deference though, it looks like a trickier issue.
> 
> My change to `expression.cpp` fixes this. The crash happens for me with gcc 9.2 but not 8.3 or clang 9.0. But with clang the address sanitizer reports a problem. I can't fully convince myself this is a program bug and not a compiler bug. @jeanPerier, @klausler: what do you think?

I'll take a look with clang; I can't build anything successfully with gcc any more.

Issue #966, opened 2020-02-04
[LLVMify F18] Use llvm_unreachable for unreachable code cases


DavidTruby:
`llvm_unreachable` expands to an assert in debug mode and a compiler optimization hint abstracted across the supported compilers in release mode. This allows additional compiler optimizations based on this assumption.
We should replace our uses of `assert(false)`, `DIE("unreachable")` and similar with `llvm_unreachable` to take advantage of this.

klausler:
There are no instances of `assert(false)` or `CHECK(false)` in f18.  There's 9 instances of `die("unreachable")` or `DIE("unreachable")`, all in the same file.  I recommend replacement with a macro, which can be defined alongside the macro `CRASH_NO_CASE`.

DavidTruby:
Why define our own macro when something suitable already exists elsewhere in the project?

klausler:
Because, as mentioned above, you'll probably also want to modify `CRASH_NO_CASE`.

kiranktp:
llvm_unreachable will always print the filename and the line number of the problematic location.

llvm_unreachable("..") is a drop in replacement for DIE("..").
so different cases like CRASH_NO_CASE, SWITCH_COVERS_ALL_CASES and CHECK can leverage llvm_unreachable(..) similar to DIE("..).

But Fortran::common::die accepts varargs. This honors the format specifiers. In all the case this function will not print the filename and the line number.
Ex: common::die("missing case to fold intrinsic function %s", name.c_str())
The error we get is:
fatal internal error: missing case to fold intrinsic function xyz

Should we replace this function also to use llvm_unreachable or can we have f18 specific implementation for this scenario?

Either ways, for the above case we need to have f18 specific implementation to honor format specifiers, as llvm_unreachable does not honor format specifier.
Please advice. Correct me if my understanding is incorrect.

DavidTruby:
Hi Kiran, thanks for looking at this. At a glance I think we can replace our die function with something like the following:
```cpp
template <typename Args...> 
[[noreturn]] void die(const char* c, Args... args) {
   std::string buf;
   llvm::raw_string_ostream ss {buf};
   buf << llvm::format(c, args...);
   llvm_unreachable(buf.str());
}
```
and then get rid of our DIE macro. Do you think something like this could work?

As regards the other point, I can't see why the file and line information wouldn't always be useful, so I don't think we need the distinction between the two.

kiranktp:
[...]

PR #1057, opened 2020-03-10
[LLVMify F18] Use llvm_unreachable for unreachable code cases #966

kiranktp:
Current implementation of DIE in f18 is exact replica of llvm_unreachable from llvm.
Since f18(flang) will be part of llvm mono repo very soon, it would be good to use  llvm_unreachable. 
In this commit, f18 specific DIE macro and its references are replaced with llvm_unreachable.

### DIFF
--- a/include/flang/Common/idioms.h
+++ b/include/flang/Common/idioms.h
@@ -23,6 +23,7 @@
 #error g++ >= 7.2 is required
 #endif
 
+#include "llvm/Support/ErrorHandling.h"
 #include <functional>
 #include <list>
 #include <optional>
@@ -63,10 +64,8 @@ template<typename... LAMBDAS> visitors(LAMBDAS... x)->visitors<LAMBDAS...>;
 // Calls std::fprintf(stderr, ...), then abort().
 [[noreturn]] void die(const char *, ...);
 
-#define DIE(x) Fortran::common::die(x " at " __FILE__ "(%d)", __LINE__)
-
 // For switch statement default: labels.
-#define CRASH_NO_CASE DIE("no case")
+#define CRASH_NO_CASE llvm_unreachable("no case")
 
 // clang-format off
 // For switch statements whose cases have return statements for
@@ -82,7 +81,7 @@ template<typename... LAMBDAS> visitors(LAMBDAS... x)->visitors<LAMBDAS...>;
 // For cheap assertions that should be applied in production.
 // To disable, compile with '-DCHECK=(void)'
 #ifndef CHECK
-#define CHECK(x) ((x) || (DIE("CHECK(" #x ") failed"), false))
+#define CHECK(x) ((x) || (llvm_unreachable("CHECK(" #x ") failed"), false))
 #endif
 
 // User-defined type traits that default to false:
--- a/include/flang/Common/enum-set.h
+++ b/include/flang/Common/enum-set.h
@@ -187,7 +187,8 @@ public:
           return {enumerator};
         }
       }
-      die("EnumSet::LeastElement(): no bit found in non-empty std::bitset");
+      llvm_unreachable(
+          "EnumSet::LeastElement(): no bit found in non-empty std::bitset");
     }
   }
 
--- a/lib/Common/CMakeLists.txt
+++ b/lib/Common/CMakeLists.txt
@@ -13,6 +13,10 @@ add_library(FortranCommon
   idioms.cpp
 )
 
+target_link_libraries(FortranCommon
+  LLVMSupport
+)
+
 install (TARGETS FortranCommon
   ARCHIVE DESTINATION lib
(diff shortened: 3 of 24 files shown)


include/flang/Common/idioms.h
```diff
@@ -23,6 +23,9 @@
 #error g++ >= 7.2 is required
 #endif
 
+#include "llvm/Support/ErrorHandling.h"
+#include "llvm/Support/Format.h"
+#include "llvm/Support/raw_ostream.h"
```
tskeith: This file is included in almost every source file in f18. Have you verified that adding these three includes (which include more files) don't have an impact on build speed. Can you get them out of this header?

include/flang/Common/idioms.h
```diff
@@ -60,13 +63,16 @@ template<typename... LAMBDAS> struct visitors : LAMBDAS... {
...
-// Calls std::fprintf(stderr, ...), then abort().
-[[noreturn]] void die(const char *, ...);
-
-#define DIE(x) Fortran::common::die(x " at " __FILE__ "(%d)", __LINE__)
+template <typename... Args> 
+[[noreturn]] void die(const char* c, Args ... args) {
+   std::string buf;
+   llvm::raw_string_ostream ss {buf};
+   ss << llvm::format(c, args...);
+   llvm_unreachable(ss.str().c_str());
```
tskeith: What file and line number is going to be reported here?

tskeith:
Can you add information to the commit message about why this change is desirable?

tskeith:
There is a reference to CHECK in CMakeLists.txt that needs to be fixed.

include/flang/Common/idioms.h
DavidTruby: Probably the file and line number of the header, I'm not sure what the right solution to pass the original file and line number forward would be.

kiranktp: As David responded. I overlooked the file name and line number. The file and line number will be always of idioms.h header.
Probably we can retain earlier implementation of die for cases with varargs (with format specifiers).

DavidTruby: I agree that that's the best thing to do for now. We can always come back to this later and have another look.

DavidTruby: Alternatively, we use the current DIE macro but call this new die function instead of the old implementation? That should also work?

kiranktp: With current implementation, DIE macro and all its usage has been replaced with llvm_unreachable. 
Old implementation of die is required in 8 places. We can leave the function and its usage untouched.

kiranktp:
> There is a reference to CHECK in CMakeLists.txt that needs to be fixed.
CHECK macro will be disabled for MinSizeRel build and the CHECK macro definition in idioms.h header is already guarded to avoid redefinition.
With the new change, CHECK will invoke llvm_unreachable instead of die.
What has to be fixed in this case?

tskeith:
> > There is a reference to CHECK in CMakeLists.txt that needs to be fixed.

> CHECK macro will be disabled for MinSizeRel build and the CHECK macro definition in idioms.h header is already guarded to avoid redefinition.
> With the new change, CHECK will invoke llvm_unreachable instead of die.
> What has to be fixed in this case?

Sorry, I thought you were removing CHECK and not just DIE. You can ignore this comment.

include/flang/Common/idioms.h
kiranktp: llvm/Support/ErrorHandling.h is needed for llvm_unreachable macro. 
idioms.h files is included in 55+ files and among this 50+ files invoke llvm_unreachable function.
I think this is the best place for this llvm header.
I have removed newer implementation of die function, so rest of the two llvm headers are not needed now.

kiranktp:
There are totally 60+ usage of die function (DIE macro). Among this only 48 usage was replaced with llvm_unreachable and rest will continue to use f18 implementation of die function. Since f18 implementation is available, it would be better to continue using the local implementation of die function rather than adding a dependency on llvm header and using llvm_unreachable.
@tim also has the same opinion that there is no benefit of partially replacing die function with llvm_unreachable.
I will go ahead and close this PR.

DavidTruby:
Are we letting the perfect be the enemy of the good here? I'm not sure why if 70% of the cases can be replaced by a better implementation we need to block that on the remaining 30% of cases; can't we merge that 70% and think about the remaining cases later?

sscalpone:
This change seems to conflate assertions with unreachable code. (The original code does too, but unlike llvm, the original code doesn't make a distinction between assertions and code that should not be executed.

In llvm, the preference is to use the system assert.  That makes debugging a release build a bit more difficult because asserts are disabled and the meaning of llvm_unreachable changes.

Therefore, I think it's better to have an indirection between the assert and llvm_unreachable, like flang_assert and flang_unreachable.

PR #1061, opened 2020-03-10
Extend shape analysis to cope better with ASSOCIATE construct entities

klausler:
This change fixes some recent regressions with spurious error messages about array assignments to assumed-size left-hand sides when the LHS was actually an `ASSOCIATE` construct entity.

### DIFF
--- a/lib/Evaluate/shape.cpp
+++ b/lib/Evaluate/shape.cpp
@@ -177,52 +177,84 @@ bool ContainsAnyImpliedDoIndex(const ExtentExpr &expr) {
   return MyVisitor{}(expr);
 }
 
-ExtentExpr GetLowerBound(
-    FoldingContext &context, const NamedEntity &base, int dimension) {
-  const Symbol &symbol{ResolveAssociations(base.GetLastSymbol())};
+// Determines lower bound on a dimension.  This can be other than 1 only
+// for a reference to a whole array object or component. (See LBOUND, 16.9.109).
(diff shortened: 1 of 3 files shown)


lib/Evaluate/shape.cpp
```diff
@@ -256,6 +256,13 @@ MaybeExtentExpr GetExtent(
         }
       }
     }
+  } else if (const auto *assoc{
+                 symbol.detailsIf<semantics::AssocEntityDetails>()}) {
+    if (auto shape{GetShape(context, assoc->expr())}) {
+      if (!shape->empty()) {
+        return std::move(shape->back());
```
tskeith: Doesn't this have to take into account dimension? E.g. `shape->at(dimension)`

klausler: Right.

lib/Evaluate/shape.cpp
```diff
@@ -316,6 +323,13 @@ MaybeExtentExpr GetUpperBound(
...
```
tskeith: dimension here too?

Does GetLowerBound also need similar code?

klausler: Yes, and no (lower bounds of associate construct entities that aren't whole variables will always be 1).

klausler: Well, that's not entirely true.  Associate construct entities that are associated with whole components can also have lower bounds other than 1.  And pointers, too.  I'll think more on this tomorrow.

PR #1070, opened 2020-03-13
First cut at runtime API for allocatable management

klausler:
As we did with runtime I/O APIs, here's an initial interface spec for allocatables encoded in the form of a header file of prototypes with lots of commentary.  I haven't tried to compile this header or use it in any other way; it's meant to generate discussion, (dis)approval, and (one hopes) better ideas or identification of missing capabilities.

As was the case with I/O, I'm trying to push complexity into the implementation of the runtime in order to simplify the form of the generated code, so that the API and compiler-generated code can be smaller and more stable over time; I'd rather we fix bugs with derived type reallocation and finalization in the runtime than in generated code, for example.

Pointers, actual argument temporaries, adjustables, &c. will use a similar API, probably with renaming for generality, but I'd like to get general agreement on the API for allocatables before trying to generalize further.  Assume that all of the names can change, they're not as important as their arguments.

### DIFF
--- a/runtime/allocatable.h
+++ b/runtime/allocatable.h
@@ -1,0 +1,115 @@
+//===-- runtime/allocatable.h -----------------------------------*- C++ -*-===//
+//
+// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
+
+// Defines APIs for Fortran runtime library support of code generated
+// to manipulate and query allocatable variables, dummy arguments, & components.
+#ifndef FORTRAN_RUNTIME_ALLOCATABLE_H_
+#define FORTRAN_RUNTIME_ALLOCATABLE_H_
+#include "descriptor.h"
+#include "entry-names.h"
+
+namespace Fortran::runtime {
+extern "C" {
+
+// Initializes the descriptor for an allocatable of intrinsic or derived type.
+// The incoming descriptor is treated as (and can be) uninitialized garbage.
+// Must be called for each allocatable variable as its scope comes into being.
+// The storage for the allocatable's descriptor must have already been
+// allocated to a size sufficient for the rank, corank, and type.
+// A descriptor must be initialized before being used for any purpose,
+// but needs reinitialization in a deallocated state only when there is
+// a change of type, rank, or corank.
+void RTNAME(AllocatableInitIntrinsic)(
+    Descriptor &, TypeCategory, int kind, int rank = 0, int corank = 0);
+void RTNAME(AllocatableInitCharacter)(Descriptor &, SubscriptValue length,
+    int kind = 1, int rank = 0, int corank = 0);
+void RTNAME(AllocatableInitDerived)(
+    Descriptor &, const DerivedType &, int rank = 0, int corank = 0);
+
+// Checks that an allocatable is not already allocated in statements
+// with STAT=.  Use this on a value descriptor before setting bounds or
+// type parameters.  Not necessary on a freshly initialized descriptor.
+// (If there's no STAT=, the error will be caught later anyway, but
+// this API allows the error to be caught before descriptor is modified.)
+// Return 0 on success (deallocated state), else the STAT= value.
+int RTNAME(AllocatableCheckAllocated)(Descriptor &,
+    Descriptor *errMsg = nullptr, const char *sourceFile = nullptr,
+    int sourceLine = 0);
+
+// For MOLD= allocation; sets bounds, cobounds, and length type
+// parameters from another descriptor.  The destination descriptor must
+// be initialized and deallocated.
+void RTNAME(AllocatableApplyMold)(Descriptor &, const Descriptor &mold);
+
+// Explicitly sets the bounds and length type parameters of an initialized
+// deallocated allocatable.
+void RTNAME(AllocatableSetBounds)(
+    Descriptor &, int zeroBasedDim, SubscriptValue lower, SubscriptValue upper);
+
+// The upper bound is ignored for the last codimension.
+void RTNAME(AllocatableSetCoBounds)(Descriptor &, int zeroBasedCoDim,
+    SubscriptValue lower, SubscriptValue upper = 0);
+
+// Length type parameters are indexed in declaration order; i.e., 0 is the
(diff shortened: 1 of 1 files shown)


runtime/allocatable.h
```diff
@@ -0,0 +1,88 @@
...
+// a change of type, rank, or corank.
+void RTNAME(AllocatableInitIntrinsic)(
+    Descriptor &, TypeCategory, int kind, int rank = 0, int corank = 0);
+void RTNAME(AllocatableInitCharacter)(Descriptor &, SubscriptValue length,
+    int kind = 1, int rank = 0, int corank = 0);
+void RTNAME(AllocatableInitDerived)(
+    Descriptor &, const DerivedType &, int rank = 0, int corank = 0);
+
+// For MOLD= and SOURCE= allocation with no (other) prior initialization.
+void RTNAME(AllocatableInitMold)(Descriptor &, const Descriptor &);
```
jeanPerier: `to` and `from` argument name would be valuable for this interface.

runtime/allocatable.h
```diff
@@ -0,0 +1,88 @@
...
+// declaration order; i.e., 0 is the first length type parameter in the
+// deepest base type.
+void RTNAME(AllocatableSetBounds)(
+    Descriptor &, int dim, SubscriptValue lower, SubscriptValue upper);
+void RTNAME(AllocatableSetDerivedLength)(
+    Descriptor &, int which, SubscriptValue);
+
+// The upper bound is ignored for the last codimension.
+void RTNAME(AllocatableSetCoBounds)(
+    Descriptor &, int codim, SubscriptValue lower, SubscriptValue upper = 0);
```
jeanPerier: Are `codim` (and `dim` in `AllocatableSetBounds`) indexed starting from 1 as the `DIM` argument in Fortran intrinsic ?

runtime/allocatable.h
```diff
@@ -0,0 +1,88 @@
...
+// initialized and its bounds and length type parameters set and must be
+// in a deallocated state.  For SOURCE= allocation of an uninitialized
+// descriptor, use AllocatableInitMold() before the first allocation.
+// On failure, if hasStat is true, returns a nonzero error code for
+// STAT= and (if present) fills in errMsg; if hasStat is false, the
+// image is terminated.  On success, leaves errMsg alone and returns zero.
+// Successfully allocated memory is initialized if the allocatable has a
+// derived type, and is always initialized by AllocatableAllocateSource().
+int RTNAME(AllocatableAllocate)(
+    Descriptor &, bool hasStat = false, Descriptor *errMsg = nullptr);
```
jeanPerier: Did you intend to leave the check that the allocatable is not already allocated to this function ?
The issue I see with this is that we may have already mutated the descriptor (to set the bounds, or type parameters...) and it case we realize now that it is already allocated and `hasStat` is true, we would have to restore the original information as per 9.7.6 point 5:
> “each allocate-object that was not successfully allocated or deallocated shall retain its previous allocation status or pointer association status

In general, though it might be safer to backup the descriptor when `STAT` is present in order to fulfill 9.7.6 p6. But if not, care may have to be taken here.

runtime/allocatable.h
```diff
@@ -0,0 +1,88 @@
...
+int RTNAME(AllocatableAllocateSource)(Descriptor &, const Descriptor &source,
+    bool hasStat = false, Descriptor *errMsg = nullptr);
```
jeanPerier: Would `AllocatableAllocate` and `AllocatableAllocateSource` be responsible to check that coarray allocation on other images succeeded or will it have to be done separably ?

runtime/allocatable.h
```diff
@@ -0,0 +1,88 @@
...
+
+// Explicitly sets the bounds and length type parameters of an initialized
+// deallocated allocatable.  (Not for use with MOLD= or SOURCE= in an
+// ALLOCATE statement; see below).  Length type parameters are indexed in
+// declaration order; i.e., 0 is the first length type parameter in the
+// deepest base type.
+void RTNAME(AllocatableSetBounds)(
+    Descriptor &, int dim, SubscriptValue lower, SubscriptValue upper);
+void RTNAME(AllocatableSetDerivedLength)(
+    Descriptor &, int which, SubscriptValue);
```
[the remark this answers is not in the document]
klausler: I don't know.  The compiler-generated code will have both of the values in hand, but the error handling should be in the runtime.  I think I should add another API for generated code to call with both values, for use in the case of an explicit (non-deferred) length type parameter.  This applies to both derived types and `CHARACTER`.

runtime/allocatable.h
```diff
@@ -0,0 +1,88 @@
...
+    int kind = 1, int rank = 0, int corank = 0);
+void RTNAME(AllocatableInitDerived)(
+    Descriptor &, const DerivedType &, int rank = 0, int corank = 0);
+
+// For MOLD= and SOURCE= allocation with no (other) prior initialization.
+void RTNAME(AllocatableInitMold)(Descriptor &, const Descriptor &);
+
+// Explicitly sets the bounds and length type parameters of an initialized
+// deallocated allocatable.  (Not for use with MOLD= or SOURCE= in an
+// ALLOCATE statement; see below).  Length type parameters are indexed in
```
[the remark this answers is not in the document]
klausler: Right, I'll fix the comment.  I know that this is true for SOURCE=, and I'll confirm that it's also permitted with MOLD=.

runtime/allocatable.h
klausler: Ok.

runtime/allocatable.h
klausler: No, I intend for them to be zero-based, which is the convention used elsewhere in the runtime APIs for descriptors.  Would you prefer 1-based indices?

jeanPerier: No, zero based is fine (actually easier since FIR is already zero based here).

runtime/allocatable.h
klausler: Maybe the APIs for setting bounds, length parameters, &c. should be capable of failing in the case of an allocated descriptor.  Or have a distinct API for checking for allocation for use in cases with STAT=; then the generated code can jump around the mutations.  Backing up a descriptor seems complicated to me.

runtime/allocatable.h
klausler: That check should be done in these runtime library routines, I think.  Synchronization, too.  Unless you think there's a need for compiler-generated code to interact with those actions; they seem to me to be messy details best left to the library.

Issue #908, opened 2020-01-06
Proposal: Replace content of flang-compiler/f18 with "Moved" message


peterwaller-arm:
This proposal is inspired by https://github.com/tensorflow/mlir/, which recently was merged into llvm-project.

### Proposal

Once the `flang-compiler/f18` repository has been merged into `llvm/llvm-project`, there should be one final commit on the `flang-compiler/f18` repository, which deletes the contents of the git repository (leaving alone the commit history, issues and pull requests).

This should make it clear if anyone makes the mistake of obtaining a clone of the old repository unintentionally.

As an added benefit, the README can include instructions for migrating commits in your fork to the new repository, as the [MLIR project does](tree/3722f0344c5d121f76ac73f3e652fc89e5fef799#migration).

Thanks to @DavidTruby for the suggestion.

Issue #876, opened 2019-12-18
Submission into the llvm-project


peterwaller-arm:
This issue tracks what needs to happen for the mono-repo push. I intend to keep this message up-to-date. Please post suggestions below or edit the body of this message as appropriate.

#### TODO

- [x] Implement first cut of history rewriting (#854)
- [x] Contact [llvm-dev]() and [flang-dev]() to express intent.
- [ ] Get green light.
- [ ] Finalize rewrite program.
- [ ] Decide date.

#### When the time comes:

- [ ] Freeze submissions to flang-compiler/f18
  - [ ] Have someone responsible for doing this
- [ ] Contact Tom Stellard and Mike Edwards to temporarily suspend commit emails.
- [ ] Generate rewritten history
  - [ ] Pull latest f18/master and llvm-project/master
  - [ ] Run rewriting program
- [ ] Final manual sanity checking
  * Check that the name of whoever did the rewrite does not end up on the commits where they do not belong.
  * Check that the key promise of "the trees of the merge commits are the same after rewrite" holds.
  * Eyeball the history, make sure it follows on from the right point.
- [ ] `git push llvm-project rewritten-history:master`
- [ ] Announce merge complete on llvm-dev and flang-dev.
- [ ] Contact Tom Stellard and Mike Edwards to unsuspend commit emails.
- Future submissions now must go into llvm-project.
  * flang-compiler/f18 remains frozen for accepting future pull requests.
  * Future commit messages referencing f18 issues and pull requests should use the notation `flang-compiler/f18#123` rather than simply `#123`.

RichBarton-Arm:
I think some additional pre-requisites are:
- [ ] Ensure all F18 developers have LLVM commit access
- [ ] Nominate code owners for various parts of the flang repo and commit a CODE_OWNERS file

And an additional step for when the time comes
- [ ] Close all outstanding PR's on old f18 repo. It is the responsibility of the PR owner to raise a new review in Phabricator to continue the review in the monorepo if the PR is still live.

peterwaller-arm:
Alex L posts a good argument for a merge commit on the mainline: http://lists.llvm.org/pipermail/llvm-dev/2019-December/137690.html

To summarize: this will prevent the builders from trying to build all 2,700 commits, instead they will just build the merge commit.

klausler:
How much time will pass between the freezing of the existing f18 repository on github and the resumption of development in llvm-project?

tskeith:
More pre-requisites:
[...]

PR #1072, opened 2020-03-13
Replace content of flang-compiler/f18 with "Moved" message

CarolineConcatto:
This patch replaces the content of flang-compiler-/f18 with a README.md file,as requested in task/issue[1].

The README  includes instructions for migrating commits from local fork to LLVM
and its new repository location.

Important: This patch will only be merged after the f18 was ported to LLVM
monorepo according to task/issue[2].

[1]https://github.com/flang-compiler/f18/issues/908
[2]https://github.com/flang-compiler/f18/issues/876


README.md
```diff
@@ -1,244 +1,36 @@
...
-  
-   Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
-   See https://llvm.org/LICENSE.txt for license information.
-   SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
-  
--->
+# 301 - Moved
 
-# F18
+F18 is now part of LLVM.
```
kiranchandramohan: Should you also say 
-> that is available in the flang directory
-> renamed and available as flang in LLVM.

CarolineConcatto: I believe I should only say something like this:

DavidTruby: This wording works for me

CarolineConcatto: Done!

sscalpone:
How did you test the migration processes that are outlined in the readme?

DavidTruby:
> How did you test the migration processes that are outlined in the readme?

I tested it using the flattened version of the repo I've been using to get ready for upstreaming. Once the CMake patch is merged I can re-flatten and push that out to my fork of llvm-project along with Eric's LLVM-side cmake changes. That should make it easy for people to test both the commands here and building F18/Flang in-tree with LLVM, as well as see what the flattened commit history will look like.

sscalpone:
How about reducing this PR to just the message?  It will be easier to review & keep correct.  Then, after the upstream move, we can do another PR to remove all of the files.

RichBarton-Arm:
I think there is a significant advantage in the commit history having the addition of the readme and the removal of the files as an atomic commit.

I don't think it will be hard to co-ordinate. If we can agree on the contents of the readme beforehand, David can update this PR after we push into LLVM and then someone with F18 commit access can merge it before commit access is shut off. There should be no risk of any changes taking place on the F18 side after we have merged into LLVM.

DavidTruby:
We also don't want the changed README going in to LLVM, so we have to do it all after we merge anyway

sscalpone:
My point is that we can (easily) approve the text of the message now whereas we cannot audit the list of files to be deleted until after we've moved things to LLVM.  My preference is to push README.md as soon as the code is moved to LLVM so there's no delay while we collect the list of files to be deleted.

With that, I'll defer to @CarolineConcatto on how to proceed.

CarolineConcatto:
@sscalpone either way is fine.
 The solution to having only one PR is good for git history. As it will be only one PR for deleting and letting a message.
 If you are worried about the files, @DavidTruby and I could work together to guarantee that all the files that went in LLVM were also in the F18 project. 
Moreover, we could still review the Readme and later see the files, and only merge once we are happy about both. Not a problem for me.
But you have the power to do the merge. So you have the word.

sscalpone:
It is ok if the PR combines the changes to the README and the deletion of all files.  We can finish the review after the PR after f18 is merged to llvm.org.  The text in the README is ok now.

PR #1073, opened 2020-03-13
Complete formatting of pointer assignments

klausler:
And move that formatting to formatting.cpp with the rest of AsFortran() member functions.


lib/Evaluate/formatting.cpp
```diff
@@ -718,6 +718,43 @@ std::ostream &DescriptorInquiry::AsFortran(std::ostream &o) const {
   }
 }
 
+std::ostream &Assignment::AsFortran(std::ostream &o) const {
```
CarolineConcatto: I believe that we want to remove std::ostream from f18 to be more align with llvm, corret?

PR #1076, opened 2020-03-17
Support forward references to sibling module procedures

klausler:
Enable additional instances of `ResolveNamesVisitor` to be created to recursively handle calls to unresolved functions in specification expressions in an "on demand" fashion.

And improve error message for procedure passed as invalid argument to an intrinsic.

### DIFF
--- a/include/flang/Semantics/symbol.h
+++ b/include/flang/Semantics/symbol.h
@@ -27,6 +27,7 @@ namespace Fortran::semantics {
 
 class Scope;
 class Symbol;
+class ProgramTree;
 
 using SymbolRef = common::Reference<const Symbol>;
 using SymbolVector = std::vector<SymbolRef>;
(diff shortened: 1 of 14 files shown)


include/flang/Semantics/symbol.h
```diff
@@ -91,12 +92,15 @@ ENUM_CLASS(SubprogramKind, Module, Internal)
 // type information.
 class SubprogramNameDetails {
 public:
-  SubprogramNameDetails(SubprogramKind kind) : kind_{kind} {}
+  SubprogramNameDetails(SubprogramKind kind, ProgramTree &node)
+    : kind_{kind}, node_{node} {}
   SubprogramNameDetails() = delete;
   SubprogramKind kind() const { return kind_; }
+  ProgramTree &node() const { return *node_; }
```
tskeith: `ProgramTree` nodes only exist during name resolution but `Symbol`s live longer. So I think this will be a dangling reference if it is called after name resolution.

klausler: Sure, but there shouldn't be any SubprogramNameDetails symbols after name resolution, right?

tskeith: Oh, right.

lib/Semantics/resolve-names.cpp
```diff
@@ -375,10 +382,9 @@ class ImplicitRulesVisitor : public DeclTypeSpecVisitor {
 protected:
   void BeginScope(const Scope &);
   void SetScope(const Scope &);
+  void SetImplicitRules(ImplicitRules &rules) { implicitRules_ = &rules; }
```
tskeith: This isn't used.

tskeith:
It would be good to have some tests.

lib/Semantics/resolve-names.cpp
klausler: Thanks (I didn't clean it out after it became needless).

klausler:
> It would be good to have some tests.

The code gets exercised by f90_correct test md102 and f18 test expr-errors02.f90, and I'll add one that's specific to forward-referenced module procedures, esp. with generics.

PR #1078, opened 2020-03-17
Do not emit a prefix for a default-kind character constant in AsFortran

klausler:


sscalpone:
Any reason besides making the output shorter & easier to read?

klausler:
pgf90 can't handle numeric kind prefixes on character literals, and it made binding names difficult for Tim to use in error messages.

PR #1084, opened 2020-03-20
Constraint checks C727 to C730 and most constraints related to attributes

psteinfeld:
The full list of checked constraints is C727, C728, C729, C730, C743, C755, C759, C778,
and C1543.

I added a function to tools.cpp to check to see if a symbol name is the name
of an intrinsic type.

The biggest change was to resolve-names.cpp to check to see if attributes were
either duplicated or in conflict with each other.  I changed all locations
where attributes were set to check for duplicates or conflicts.

I also added tests for all checks and annotated the tests and code with the
numbers of the constraints being tested/checked.


lib/Semantics/resolve-names.cpp
```diff
@@ -1548,26 +1554,70 @@ bool AttrsVisitor::SetBindNameOn(Symbol &symbol) {
...
+bool AttrsVisitor::IsDuplicateAttr(Attr attrName) {
+  if (attrs_->test(attrName)) {
+    Say(currStmtSource().value(),
+        "Attribute '%s' cannot be used more than once"_err_en_US,
+        AttrToString(attrName));
+    return true;
+  }
+  return false;
+}
+bool AttrsVisitor::HaveAttrConflict(Attr attrName, Attr attr1, Attr attr2) {
```
tskeith: This could use a comment explaining what the arguments are.

test/Semantics/resolve76.f90
```diff
@@ -1,30 +1,26 @@
...
-    b = 2.0
-    !ERROR: No intrinsic or user-defined ASSIGNMENT(=) matches operand types REAL(4) and LOGICAL(4)
-    b = .false.
-  end
-  module procedure f
-    f = .true.
-    !ERROR: No intrinsic or user-defined ASSIGNMENT(=) matches operand types LOGICAL(4) and REAL(4)
-    f = 1.0
-  end
-end
```
tskeith: What happened to the test previously in resolve76.f90?

test/Semantics/resolve81.f90
```diff
@@ -0,0 +1,64 @@
...
+  !ERROR: Attribute 'POINTER' cannot be used more than once
+  real, pointer, pointer :: realPtr
+  !ERROR: Attribute 'PROTECTED' cannot be used more than once
+  real, protected, protected :: realProt
+  !ERROR: Attribute 'SAVE' cannot be used more than once
+  real, save, save :: saveVar
+  !ERROR: Attribute 'TARGET' cannot be used more than once
+  real, target, target :: targetVar
+  !ERROR: Attribute 'VOLATILE' cannot be used more than once
+  real, volatile, volatile :: volatileVar
```
tskeith: There should be some tests where the duplicate attributes are not in the same statement.

lib/Semantics/tools.cpp
```diff
@@ -674,6 +674,18 @@ bool IsInitialized(const Symbol &symbol) {
   return false;
 }
 
+bool HasIntrinsicTypeName(const Symbol &symbol) {
+  static constexpr std::size_t intrinsicNameCount{6};
+  static const std::string intrinsicNames[intrinsicNameCount]{
+      "integer", "real", "doubleprecision", "complex", "character", "logical"};
```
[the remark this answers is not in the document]
klausler: Using `std::array` would be even nicer.  Or null termination, which I tend to use myself.

lib/Semantics/resolve-names.cpp
```diff
@@ -1548,26 +1554,70 @@ bool AttrsVisitor::SetBindNameOn(Symbol &symbol) {
...
+    }
   }
   return false;
 }
 
+// C730, C743, C755, C778, C1543 say no attribute or prefix repetitions
+bool AttrsVisitor::IsDuplicateAttr(Attr attrName) {
+  if (attrs_->test(attrName)) {
+    Say(currStmtSource().value(),
+        "Attribute '%s' cannot be used more than once"_err_en_US,
```
klausler: Why a fatal error?  It's not going to hurt anything.

lib/Semantics/tools.cpp
```diff
@@ -674,6 +674,18 @@ bool IsInitialized(const Symbol &symbol) {
...
+  for (std::size_t i{0}; i < intrinsicNameCount; ++i) {
+    if (symbol.name().ToString() == intrinsicNames[i]) {
+      return true;
+    }
+  }
+  return false;
+}
```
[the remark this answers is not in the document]
tskeith: I think this is the best we can do currently if we want to try to use the strings associated with TypeCategory:
```
  std::string name{symbol.name().ToString()};
  if (name == "doubleprecision") {
    return true;
  } else if (name == "derived") {
    return false;
  } else {
    for (int i{0}; i != common::TypeCategory_enumSize; ++i) {
      if (name == parser::ToLowerCaseLetters(EnumToString(TypeCategory{i}))) {
        return true;
      }
    }
    return false;
  }
```

lib/Semantics/tools.cpp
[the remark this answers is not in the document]
psteinfeld: Thanks, Tim.  I couldn't figure out how to do this.

lib/Semantics/tools.cpp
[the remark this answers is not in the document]
psteinfeld: Thanks, Tim.  I've recoded my utility function to use an ENUM_CLASS to list the intrinsic type names.  The best way I could figure out how to do this was to duplicate the names in the enumeration.  I'm not sure if this is an improvement or not.

lib/Semantics/resolve-names.cpp
```diff
@@ -242,10 +242,15 @@ class AttrsVisitor : public virtual BaseVisitor {
   bool Pre(const parser::IntentSpec &);
   bool Pre(const parser::Pass &);
 
+  bool IsDuplicateAttr(Attr);
+  bool HaveAttrConflict(Attr, Attr, Attr);
+  bool IsConflictingAttr(Attr);
```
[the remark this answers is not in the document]
psteinfeld: Thanks.  I'll make it so.

lib/Semantics/resolve-names.cpp
```diff
@@ -1548,26 +1554,70 @@ bool AttrsVisitor::SetBindNameOn(Symbol &symbol) {
 
 void AttrsVisitor::Post(const parser::LanguageBindingSpec &x) {
   CHECK(attrs_);
-  attrs_->set(Attr::BIND_C);
-  if (x.v) {
-    bindName_ = EvaluateExpr(*x.v);
+  if (attrs_->test(Attr::BIND_C)) {
```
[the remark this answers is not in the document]
psteinfeld: Gee...  I can't think of a good reason.  Thanks for catching this.  I'll change it.

lib/Semantics/resolve-names.cpp
psteinfeld: Thanks.  I'll supply one.

test/Semantics/resolve76.f90
psteinfeld: Thanks for catching this.  I mistakenly overwrote it when I merged from master into my branch.  I'll restore it.

test/Semantics/resolve81.f90
psteinfeld: Wow, I didn't even realize that this was possible.  I'll add some tests.

lib/Semantics/resolve-names.cpp
psteinfeld: Thanks for pointing this out.  I agree with you.  I just checked, and duplicate attributes don't seem to be noticed by pgf90, but they're fatal errors for both IBM and Intel.

I'll change the duplicates to warnings.

lib/Semantics/tools.cpp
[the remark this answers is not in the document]
psteinfeld: Thanks for the code snippet, Tim.  I like this better than what I came up with.

lib/Semantics/resolve-names.cpp
psteinfeld: Actually, some attributes such as CODIMENSION, EXTENDS, DIMENSION convey extra information.  It seems appropriate to me that duplicating these attributes should be an error.

klausler: EXTENDS is an optional part of the definition of a derived type, not an attribute of an entity.

I'm not saying that we should allow all possible duplicated attributes (besides the ones that the standard says we have to allow).

lib/Semantics/resolve-names.cpp
```diff
@@ -1583,12 +1580,15 @@ bool AttrsVisitor::Pre(const parser::Pass &x) {
...
     Say(currStmtSource().value(),
-        "Attribute '%s' cannot be used more than once"_err_en_US,
+        "Attribute '%s' cannot be used more than once"_en_US,
         AttrToString(attrName));
     return true;
   }
   return false;
 }
+
+// See if attrName voilates a constraint cause by a conflict.  attr1 and attr2
```
tskeith: voilates -> violates

test/Semantics/resolve77.f90
```diff
@@ -1,52 +1,37 @@
 ! RUN: %S/test_errors.sh %s %flang %t
-! Tests valid and invalid usage of forward references to procedures
-! in specification expressions.
```
tskeith: This looks like another merge problem.

psteinfeld: Ay yi yi.  Thanks for catching this.

lib/Semantics/resolve-names.cpp
psteinfeld: Fixed.

PR #1086, opened 2020-03-21
Semantics and test for ENTRY statements

klausler:
Checks the constraints and the feasible "shalls" in 15.6.2.6 for `ENTRY` statements, their dummy arguments, and their results.

Symbols for the entry points are created in the parent scope of the containing subprogram, and point to the scope of the containing subprogram.  Symbols for `ENTRY` results, when functions, are within the scope of the containing subprogram.

This PR checks semantics only; it doesn't do anything to represent sharing of dummy arguments, storage association of function results, &c.

### DIFF
--- a/include/flang/Semantics/symbol.h
+++ b/include/flang/Semantics/symbol.h
@@ -61,6 +61,9 @@ public:
   bool isFunction() const { return result_ != nullptr; }
   bool isInterface() const { return isInterface_; }
   void set_isInterface(bool value = true) { isInterface_ = value; }
+  Scope *entryScope() { return entryScope_; }
+  const Scope *entryScope() const { return entryScope_; }
+  void set_entryScope(Scope &scope) { entryScope_ = &scope; }
   MaybeExpr bindName() const { return bindName_; }
   void set_bindName(MaybeExpr &&expr) { bindName_ = std::move(expr); }
   const Symbol &result() const {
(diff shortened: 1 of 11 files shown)


klausler:
Testing against f90_correct shows that there's still some bugs to fix in this code, so updates are coming.  I'll comment again when all tests are passing.  Also, I need to figure out how to emit `ENTRY` points in module files.

klausler:
Ok, the tests are greatly extended to cover additional cases, and they pass; so do all of the cases in the PGI Fortran suite.  And I've figured out how I think `ENTRY` should interact with module files.

include/flang/Semantics/symbol.h
```diff
@@ -82,8 +85,10 @@ class SubprogramDetails {
   MaybeExpr bindName_;
   std::vector<Symbol *> dummyArgs_;  // nullptr -> alternate return indicator
   Symbol *result_{nullptr};
+  Scope *entryScope_{nullptr};  // if ENTRY, points to subprogram's scope
```
jeanPerier: I am a bit lost with the purpose of entry_scope() vs symbol->scope().
Is this to keep the property that `symbol->scope()->symbol() = identity` or does it have other purposes ?

klausler: `Symbol::scope()` is used by subprograms, derived types, modules, &c. to link symbols to the scopes that they own.  `ENTRY` point symbols don't have their own scopes, and they are not in the scope of the subprogram in which they appear.  For example,
```
subroutine subr
real x
entry ent
end
```
there's the global scope, which contains the symbol subr` that owns the scope that contains `x`.  The symbol `ent` is also in the global scope -- to its callers, it's no different from any other subroutine -- but it doesn't own a scope.  Instead, it links to the scope of `subr` via its `entryScope_` data member.

jeanPerier: Makes sens, thanks for explaining.

PR #1089, opened 2020-03-26
Semantics for SELECT CASE

klausler:


lib/Semantics/check-case.cpp
```diff
@@ -0,0 +1,244 @@
...
+              type->kind() == caseExprType_.kind())) {
+        if (auto converted{evaluate::Fold(context_.foldingContext(),
+                evaluate::ConvertToType(T::GetType(), common::Clone(*x->v)))}) {
+          x->v = std::move(converted);
+          if (auto value{evaluate::GetScalarConstantValue<T>(*x->v)}) {
+            return *value;
+          }
+        }
+        context_.Say(
+            expr.source, "CASE value must be a constant scalar"_err_en_US);
```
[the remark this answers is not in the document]
klausler: In this case, I don't think that it's really possible to reach this point -- other checks in expression semantics should have caught non-scalar and non-constant cases.  So I'm coughing up an error message in order to fail more gracefully than just dying.

lib/Semantics/check-case.cpp
```diff
@@ -0,0 +1,244 @@
...
+      if constexpr (T::category == TypeCategory::Logical) { // C1148
+        if (!iter->IsDefault() &&
+            (!iter->lower || !iter->upper || *iter->lower != *iter->upper)) {
+          context_.Say(iter->stmt.source,
+              "CASE range is not allowed for LOGICAL"_err_en_US);
+        }
+      }
+      if (iter->lower && iter->upper && *iter->lower > *iter->upper) {
+        context_.Say(iter->stmt.source,
+            "CASE %s has lower bound greater than upper bound"_en_US,
```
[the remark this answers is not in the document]
klausler: Will do.

lib/Semantics/check-case.cpp
```diff
@@ -0,0 +1,244 @@
...
+            evaluate::Constant<T>{*upper}.AsFortran(bs << ':');
+          }
+          bs << ')';
+        } else if (upper) {
+          evaluate::Constant<T>{*upper}.AsFortran(bs << "(:") << ')';
+        } else {
+          bs << "DEFAULT";
+        }
+      }
+      return result;
```
tskeith: You can avoid needing the extra block with `return bs.str();`

lib/Semantics/check-case.cpp
```diff
@@ -0,0 +1,244 @@
...
+            },
+            [&](const parser::Default &) { cases_.emplace_front(stmt); },
+        },
+        selector.u);
+  }
+
+  std::optional<Value> GetValue(const parser::CaseValue &caseValue) {
+    const auto &expr{caseValue.thing.thing.value()};
+    auto *x{expr.typedExpr.get()};
+    if (x && x->v) { // C1147
```
tskeith: Using `if (auto *x{GetExpr(caseValue)}) {` here might simplify things, but I'm not sure about that.

lib/Semantics/check-case.cpp
```diff
@@ -0,0 +1,244 @@
...
+  const auto &selectCaseStmt{
+      std::get<parser::Statement<parser::SelectCaseStmt>>(construct.t)};
+  const auto &selectCase{selectCaseStmt.statement};
+  const auto &selectExpr{
+      std::get<parser::Scalar<parser::Expr>>(selectCase.t).thing};
+  const auto *typedExpr{selectExpr.typedExpr.get()};
+  if (!typedExpr || !typedExpr->v) {
+    return; // expression semantics failed
+  }
+  const SomeExpr &expr{*typedExpr->v};
```
tskeith: This could be:
```
  const SomeExpr *expr{GetExpr(selectExpr)};
  if (!expr) {
    return; // expression semantics failed
  }
```

lib/Semantics/check-case.cpp
klausler: As written, I think it saves a copy.

lib/Semantics/check-case.cpp
klausler: Thanks.  An earlier version of this code needed to do more with `typedExpr` but now I should use `GetExpr()`.

lib/Semantics/check-case.cpp
klausler: Will try it out.

PR #1096, opened 2020-03-30
First draft: runtime API for CHARACTER

klausler:
Jean, here's a first cut at an API for generated code to use to implement `CHARACTER` concatenation and assignment.  It does not yet have interfaces for relational operators (comparisons).  I'm interested in hearing what you think of the overall approach before I flesh it out further.

The *.cpp files added to runtime/ are there to ensure that the corresponding header files compile cleanly.


klausler:
Added more reviewers.

PR #1097, opened 2020-03-31
Checks for constraints C731 through C740

psteinfeld:
In most cases, I just added the contraint names to the code and tests.

I implemented the following checks:
  - C736 A child type with a coarray ultimate component must have a parent with
    a coarray ultimate component.
  - C737 A child type with and EVENT_TYPE or LOCK_TYPE component must have a
    parent either which is EVENT_TYPE or LOCK_TYPE or a type with an EVENT_TYPE
    or LOCK_TYPE component.
  - C740 Sequence types must contain at least on component
  - C740 Data components of sequence types must either be of an intrinsic type
    or a sequenced derived type.

After implementing these checks, some tests had new errors unrelated to their
original purpose, so I fixed them.

### DIFF
--- a/lib/Semantics/check-declarations.cpp
+++ b/lib/Semantics/check-declarations.cpp
@@ -679,13 +679,14 @@ void CheckHelper::CheckSubprogram(
 
 void CheckHelper::CheckDerivedType(
     const Symbol &symbol, const DerivedTypeDetails &details) {
-  if (!symbol.scope()) {
+  const Scope *scope{symbol.scope()};
+  if (!scope) {
     CHECK(details.isForwardReferenced());
     return;
   }
-  CHECK(symbol.scope()->symbol() == &symbol);
-  CHECK(symbol.scope()->IsDerivedType());
-  if (symbol.attrs().test(Attr::ABSTRACT) &&
+  CHECK(scope->symbol() == &symbol);
+  CHECK(scope->IsDerivedType());
+  if (symbol.attrs().test(Attr::ABSTRACT) && // C734
       (symbol.attrs().test(Attr::BIND_C) || details.sequence())) {
     messages_.Say("An ABSTRACT derived type must be extensible"_err_en_US);
   }
@@ -699,13 +700,33 @@ void CheckHelper::CheckDerivedType(
       ScopeComponentIterator components{*parentDerived};
       for (const Symbol &component : components) {
         if (component.attrs().test(Attr::DEFERRED)) {
-          if (symbol.scope()->FindComponent(component.name()) == &component) {
+          if (scope->FindComponent(component.name()) == &component) {
             SayWithDeclaration(component,
                 "Non-ABSTRACT extension of ABSTRACT derived type '%s' lacks a binding for DEFERRED procedure '%s'"_err_en_US,
                 parentDerived->typeSymbol().name(), component.name());
           }
         }
       }
+    }
+    DerivedTypeSpec derived{symbol.name(), symbol};
+    derived.set_scope(*scope);
+    if (FindCoarrayUltimateComponent(derived) && // C736
+        !(parentDerived && FindCoarrayUltimateComponent(*parentDerived))) {
+      messages_.Say(
+          "Type '%s' has a coarray ultimate component so the type at the base "
+          "of its type extension chain ('%s') must be a type that has a "
+          "coarray ultimate component"_err_en_US,
+          symbol.name(), scope->GetDerivedTypeBase().GetSymbol()->name());
+    }
+    if (FindEventOrLockPotentialComponent(derived) && // C737
+        !(FindEventOrLockPotentialComponent(*parentDerived) ||
+            IsEventTypeOrLockType(parentDerived))) {
+      messages_.Say(
+          "Type '%s' has an EVENT_TYPE or LOCK_TYPE component, so the type "
+          "at the base of its type extension chain ('%s') must either have an "
+          "EVENT_TYPE or LOCK_TYPE component, or be EVENT_TYPE or "
+          "LOCK_TYPE"_err_en_US,
+          symbol.name(), scope->GetDerivedTypeBase().GetSymbol()->name());
     }
   }
   if (HasIntrinsicTypeName(symbol)) { // C729
@@ -1141,7 +1162,7 @@ void CheckHelper::CheckProcBinding(
   CHECK(dtScope.kind() == Scope::Kind::DerivedType);
   if (const Symbol * dtSymbol{dtScope.symbol()}) {
     if (symbol.attrs().test(Attr::DEFERRED)) {
-      if (!dtSymbol->attrs().test(Attr::ABSTRACT)) {
+      if (!dtSymbol->attrs().test(Attr::ABSTRACT)) { // C733
         SayWithDeclaration(*dtSymbol,
             "Procedure bound to non-ABSTRACT derived type '%s' may not be DEFERRED"_err_en_US,
             dtSymbol->name());
--- a/lib/Semantics/resolve-names.cpp
+++ b/lib/Semantics/resolve-names.cpp
@@ -3682,18 +3682,40 @@ bool DeclarationVisitor::Pre(const parser::DerivedTypeDef &x) {
     }
   }
   Walk(std::get<std::list<parser::Statement<parser::PrivateOrSequence>>>(x.t));
+  const auto &componentDefs{
+      std::get<std::list<parser::Statement<parser::ComponentDefStmt>>>(x.t)};
+  Walk(componentDefs);
   if (derivedTypeInfo_.sequence) {
     details.set_sequence(true);
-    if (derivedTypeInfo_.extends) {
+    if (componentDefs.empty()) { // C740
       Say(stmt.source,
-          "A sequence type may not have the EXTENDS attribute"_err_en_US); // C735
+          "A sequence type must have at least one component"_err_en_US);
     }
-    if (!details.paramNames().empty()) {
+    if (!details.paramNames().empty()) { // C740
       Say(stmt.source,
-          "A sequence type may not have type parameters"_err_en_US); // C740
+          "A sequence type may not have type parameters"_err_en_US);
+    }
+    if (derivedTypeInfo_.extends) { // C735
+      Say(stmt.source,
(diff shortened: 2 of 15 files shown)


lib/Semantics/tools.cpp
```diff
@@ -537,6 +537,18 @@ const DeclTypeSpec *FindParentTypeSpec(const Symbol &symbol) {
   return nullptr;
 }
 
+bool IsSequenceType(const DeclTypeSpec *type) {
+  if (!type) {
+    return false;
+  }
```
tskeith: It looks like `type` is never null so it could be a reference. And given that, it might make more sense for it to be a member function of `DeclTypeSpec`, along with `IsPolymorphic`, `IsNumeric`, etc.

lib/Semantics/check-declarations.cpp
```diff
@@ -707,6 +707,24 @@ void CheckHelper::CheckDerivedType(
...
+    }
+    if (FindEventOrLockPotentialComponent(derived) && // C737
+        !(FindEventOrLockPotentialComponent(*parentDerived) ||
+            IsEventTypeOrLockType(parentDerived))) {
+      messages_.Say(
+          "Since type '%s' has an EVENT_TYPE or LOCK_TYPE component, the type"
+          " it extends ('%s') must either be EVENT_TYPE or LOCK_TYPE or"
+          " contain an EVENT_TYPE or LOCK_TYPE component"_err_en_US,
+          symbol.name(), parentDerived->typeSymbol().name());
+    }
```
tskeith: These two errors are similar so they should use the same wording. Perhaps change the second one to:
"Type '%s' has an EVENT_TYPE or LOCK_TYPE component so the type it extends must...".
Also, "have" would be better than "contain" in the third line.

test/Semantics/resolve86.f90
```diff
@@ -0,0 +1,21 @@
...
+  end type goodChildType
+
+  type plainParent
+  end type plainParent
+
+  !ERROR: Type 'badchildtype' has a coarray ultimate component so the type it extends ('plainparent') must have a coarray ultimate component
+  type, extends(plainParent) :: badChildType
+    real, codimension[*] :: childField
+  end type badChildType
+end subroutine s
```
tskeith: It wouldn't hurt to have a test where the type has a coarray ultimate component that is not a coarray component (i.e. that verifies it still does the right thing in the "ultimate" case).

include/flang/Semantics/tools.h
```diff
@@ -49,6 +49,7 @@ const DeclTypeSpec *FindParentTypeSpec(const DerivedTypeSpec &);
 const DeclTypeSpec *FindParentTypeSpec(const DeclTypeSpec &);
 const DeclTypeSpec *FindParentTypeSpec(const Scope &);
 const DeclTypeSpec *FindParentTypeSpec(const Symbol &);
+bool IsSequenceType(const DeclTypeSpec *type);
```
klausler: You might want to make the argument a reference rather than a pointer, unless you expect that this new predicate would be meaningfully called with a null pointer.

lib/Semantics/check-declarations.cpp
```diff
@@ -707,6 +707,24 @@ void CheckHelper::CheckDerivedType(
...
+      messages_.Say(
+          "Type '%s' has a coarray ultimate component so the type it extends "
+          "('%s') must have a coarray ultimate component"_err_en_US,
+          symbol.name(), parentDerived->typeSymbol().name());
+    }
+    if (FindEventOrLockPotentialComponent(derived) && // C737
+        !(FindEventOrLockPotentialComponent(*parentDerived) ||
+            IsEventTypeOrLockType(parentDerived))) {
+      messages_.Say(
+          "Since type '%s' has an EVENT_TYPE or LOCK_TYPE component, the type"
```
klausler: The wording is strictly correct here but might be misleading; the parent type does not have to have an event or lock component itself, but can inherit one from its parent.  What's required by Fortran is that the ultimate parent type (the "oldest" ancestor, the one without an `EXTENDS`) must have (or be) an event or lock.

Suppose we have some non-extended derived type X, and that Y extends X, and that Z extends Y.  And suppose that Z contains an event or lock.  C737 requires that Y must have an event or lock, directly or via inheritence.  But if Y has an event or lock itself, then C737 applies to Y as well, and Z must have (or be) an event or lock.  If Y does not have an event or lock, then C737 on Z requires that X must have (or be) an event or lock.  So it always works out that the deepest parent derived type ends up having to have (or be) an event or lock, and we might as well make that clear in our implementation and our error messages, even if the Standard chooses to be gratuitously confusing.

This applies to coarray component checks too, I think.

(Please check my reasoning, maybe I've missed something.)

lib/Semantics/resolve-names.cpp
```diff
@@ -3682,18 +3682,40 @@ bool DeclarationVisitor::Pre(const parser::DerivedTypeDef &x) {
...
+          "A sequence type may not have the EXTENDS attribute"_err_en_US);
+    } else {
+      for (const auto &componentName : details.componentNames()) {
+        const Symbol *componentSymbol{scope.FindComponent(componentName)};
+        if (componentSymbol && componentSymbol->has<ObjectEntityDetails>()) {
+          const auto &componentDetails{
+              componentSymbol->get<ObjectEntityDetails>()};
+          const DeclTypeSpec *componentType{componentDetails.type()};
+          if (componentType && // C740
+              !(componentType->AsIntrinsic() ||
```
klausler: Would be more readable if you apply DeMorgan's law and make this a conjunction, I think.

lib/Semantics/resolve-names.cpp
```diff
@@ -4649,7 +4675,7 @@ std::optional<DerivedTypeSpec> DeclarationVisitor::ResolveDerivedType(
       details.set_isForwardReferenced();
       symbol->set_details(std::move(details));
     } else { // C883
-      Say(name, "Derived type '%s' not found"_err_en_US);
+      Say(name, "Derived type '%s' not found"_err_en_US); // C732
```
klausler: If we're truly checking both C732 and C883 correctly here, combine the comments.

lib/Semantics/tools.cpp
psteinfeld: Good idea.  I'll do it.

test/Semantics/resolve86.f90
psteinfeld: I don't understand.  The test includes the type `goodChildType` that has a coarray ultimate component and extends `coarrayParent` that also has a coarray ultimate component.  I'm not sure what you mean by "a coarray ultimate component that is not a coarray component".

include/flang/Semantics/tools.h
psteinfeld: Thanks, Peter.  I plan to follow Tim's suggestion to make this a member function of `DeclTypeSpec`.

lib/Semantics/resolve-names.cpp
psteinfeld: I tried to make the logic follow the constraint.  The error happens if the component is not an intrinsic or a sequence type.  For me, this makes the code correspond more directly to the error message that follows the boolean expression.  Let me know if you disagree, and I'll change it.

lib/Semantics/resolve-names.cpp
psteinfeld: Good point.  I'll combine them.

lib/Semantics/check-declarations.cpp
psteinfeld: Thanks, Peter.  I'll enhance the test to add a more deeply nested test case and check the behavior of other compilers.

klausler: I think that you want to discover the *most* deeply nested ancestor type and test that, rather than the immediate parent, and adjust the message to complain about the constraint in terms of the most deeply nested ancestor type.

test/Semantics/resolve86.f90
tskeith: I was thinking something like:
```
type, extends(plainParent) :: t1
  type(t2) :: x
end type
```
`t1` doesn't have a coarray component but this should fail if `t2` has a coarray ultimate component.

lib/Semantics/check-declarations.cpp
psteinfeld: Thanks, Tim.  I'll reword the message to make it more consistent, and, based on Peter's input below, more descriptive of the actual requirement.

lib/Semantics/check-declarations.cpp
psteinfeld: Also, I'll reword the message to be more descriptive of the actual requirement.

psteinfeld: Correct me if I'm wrong, but constraint C736, which applies to coarrays, requires that the parent type have a `coarray ultimate component`, whereas constraint C737, which applies to `LOCK_TYPE` and `EVENT_TYPE` states that the parent type must have a `potential subobject component`.  If I understand things correctly, this means that the parent must directly contain a coarray component.  This is the interpretation of GNU Fortran, which is the only compiler I could find that both supports coarrays and enforces C736 at all.

klausler: Direct components and ultimate components and potential subobject components are three distinct concepts.  See 7.5.1.  It is true that C736 is defined in terms of ultimate components and that C737 is defined in terms of potential subobject components.

In short: Direct components are those that are always *present* in every instance.  Ultimate components are the direct components, plus pointers and allocatables,, but not the contents of those pointers and allocatables.  Potential subobject components are the direct components, plus the potential subobject components of allocatables -- they are all of the storage *owned* by the instance.

psteinfeld: I take it all back.  It looks like your interpretation of coarrays is correct.  I'll update the error message and add a test.

lib/Semantics/resolve-names.cpp
klausler: I already expressed my opinion above.

test/Semantics/resolve86.f90
psteinfeld: Got it.  I'll add a test.

lib/Semantics/resolve-names.cpp
psteinfeld: Actually, I don't think that this check has anything to do with constraint C883.  I'll remove the reference.

PR #1098, opened 2020-04-03
[OpenMP] Avoid abort when collapse clause value is negative

clementval:
If the value in the `collapse` close is negative f18 abort without the correct error message. This PR change the `size_t` in name resolution to a `int64_t` and check appropriately for negative or zero before the privatization of induction variable. 
The correct error is then catch by the OpenMP structure check.


clementval:
Thanks for the review @ichoyjx

clementval:
@ichoyjx I close the Pull Request since the repo as move to the llvm monorepo. I'll open a new one in phabricator.

PR #1099, opened 2020-04-03
Complete semantics for ASSIGN

klausler:
Ensure name resolution of names in ASSIGN and assigned GOTO statements.

Semantic checking of ASSIGN and assigned GOTO statements.

Allow an assigned scalar default integer variable as a format.

Capitalize some error messages.

Two new tests (one has label resolution errors, one has other errors; split so that label resolution errors don't preclude the other checks).

### DIFF
--- a/lib/Semantics/resolve-names.cpp
+++ b/lib/Semantics/resolve-names.cpp
@@ -1387,6 +1387,8 @@ public:
   bool Pre(const parser::StmtFunctionStmt &);
   bool Pre(const parser::DefinedOpName &);
   bool Pre(const parser::ProgramUnit &);
+  void Post(const parser::AssignStmt &);
+  void Post(const parser::AssignedGotoStmt &);
 
   // These nodes should never be reached: they are handled in ProgramUnit
   bool Pre(const parser::MainProgram &) { DIE("unreachable"); }
@@ -6054,6 +6056,17 @@ bool ResolveNamesVisitor::Pre(const parser::DefinedOpName &x) {
     MakePlaceholder(name, MiscDetails::Kind::TypeBoundDefinedOp);
   }
   return false;
+}
+
+void ResolveNamesVisitor::Post(const parser::AssignStmt &x) {
+  if (auto *name{ResolveName(std::get<parser::Name>(x.t))}) {
+    ConvertToObjectEntity(DEREF(name->symbol));
+  }
+}
+void ResolveNamesVisitor::Post(const parser::AssignedGotoStmt &x) {
+  if (auto *name{ResolveName(std::get<parser::Name>(x.t))}) {
+    ConvertToObjectEntity(DEREF(name->symbol));
+  }
 }
 
 bool ResolveNamesVisitor::Pre(const parser::ProgramUnit &x) {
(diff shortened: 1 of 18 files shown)


lib/Semantics/resolve-names.cpp
```diff
@@ -6056,6 +6058,21 @@ bool ResolveNamesVisitor::Pre(const parser::DefinedOpName &x) {
   return false;
 }
 
+void ResolveNamesVisitor::Post(const parser::AssignStmt &x) {
+  if (auto *name{ResolveName(std::get<parser::Name>(x.t))}) {
+    if (Symbol * symbol{name->symbol}) {
+      ConvertToObjectEntity(*symbol);
+    }
```
tskeith, line 6065: I believe `ResolveName` doesn't return a name unless it's resolved, so these three lines could be replaced by:
```
ConvertToObjectEntity(DEREF(name->symbol));
```

test/Semantics/assign07.f90
```diff
@@ -0,0 +1,35 @@
...
+    assign 1 to lab ! ok
+    assign 1 to implicitlab1 ! ok
+    !ERROR: Label '666' was not found
+    assign 666 to lab
+    !ERROR: Label '2' was not found
+    assign 2 to lab
+    assign 4 to lab
+    if (n==1) goto lab ! ok
+    if (n==1) goto implicitlab2 ! ok
+    if (n==1) goto lab(1) ! ok
```
vjayathirtha-nv, line 22: You could have an example where the label has not been assigned to any one of the list of statement labels.
```
assign 10 to lab2
if (n==1) goto lab2(1,2)
.
.
10 <some executable instruction>
1 <another executable instruction>
2 <a third executable instruction>

```

vjayathirtha-nv: This should produce an error.

vjayathirtha-nv: Another test case. This should also produce an error.
```
assign 1 to lab
lab2 = lab
goto lab2
```

klausler: Why are these errors?

vjayathirtha-nv: I don't have the Fortran 90 doc with me but I found elsewhere these rules
```
- If a list of labels appears, the statement label assigned to the variable must be one of the labels in the list.
- The variable must have a statement label value assigned to it by an ASSIGN statement (not an arithmetic assignment statement) before the GO TO statement is executed.
```

klausler: Sure, but neither of these restrictions imposed on the programmer can be detected reliably as errors at compilation time without data-flow analysis.

vdonaldson: I too interpret these two cases as violating program requirements.
I find quite a lot of variation in both error checking and run time behavior across other compilers.

vjayathirtha-nv: I can agree with that.

vdonaldson: Single pass static analysis should suffice to diagnose these two errors.  (Expected code generation is to have a map of symbols --> assign sets.)

klausler: They do violate requirements on the program.  But I don't think that we'll be able to emit compile-time errors reliably without data-flow analysis, and we don't have a CFG until semantics; the best we could do would be warnings.  As these are deleted features, I don't want to spend a whole lot of programmer or compilation time on them.

vdonaldson: A simple scan would catch obvious violations such as these two cases.  However, these are program requirements, not constraint violations, and helping a programmer or porter to use this functionality doesn't seem much of a priority, even in comparison to other legacy functionality.

PR #1100, opened 2020-04-04
Check for constrains C741 through C749

psteinfeld:
Most of these checks were already implemented and I just added references to
them to the code and tests.

I implemented the check for C747 to not allow coarray components in derived
types that are of type C_PTR, C_FUNPTR, or type TEAM_TYPE.

I implemented the check for C748 that requires a data component whose type has
a coarray ultimate component to be a nonpointer, nonallocatable scalar and not
be a coarray.

I also fixed some unrelated tests that got new errors when I implemented these
new checks.

### DIFF
--- a/lib/Semantics/check-declarations.cpp
+++ b/lib/Semantics/check-declarations.cpp
@@ -385,12 +385,16 @@ void CheckHelper::CheckObjectEntity(
   symbolBeingChecked_ = nullptr;
   if (!details.coshape().empty()) {
     if (IsAllocatable(symbol)) {
-      if (!details.coshape().IsDeferredShape()) { // C827
+      if (!details.coshape().IsDeferredShape()) { // C746, C827
         messages_.Say(
             "ALLOCATABLE coarray must have a deferred coshape"_err_en_US);
       }
+    } else if (symbol.owner().IsDerivedType()) { // C746
+      messages_.Say(
+          "Coarray components must be ALLOCATABLE and have a deferred "
+          "coshape"_err_en_US);
     } else {
-      if (!details.coshape().IsAssumedSize()) { // C828
+      if (!details.coshape().IsAssumedSize()) { // C746, C828
         messages_.Say(
             "Non-ALLOCATABLE coarray must have an explicit coshape"_err_en_US);
       }
--- a/lib/Semantics/resolve-names.cpp
+++ b/lib/Semantics/resolve-names.cpp
@@ -3680,7 +3680,7 @@ bool DeclarationVisitor::Pre(const parser::DerivedTypeDef &x) {
     if (symbol->has<TypeParamDetails>() && !paramNames.count(name)) {
       SayDerivedType(name,
           "'%s' is not a type parameter of this derived type"_err_en_US,
-          currScope()); // C742
+          currScope()); // C741
     }
   }
   Walk(std::get<std::list<parser::Statement<parser::PrivateOrSequence>>>(x.t));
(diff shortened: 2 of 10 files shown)


test/Semantics/resolve88.f90
```diff
@@ -0,0 +1,73 @@
...
+
+  ! C746 If a coarray-spec appears, it shall be a deferred-coshape-spec-list and
+  ! the component shall have the ALLOCATABLE attribute.
+
+  type testCoArrayType
+    real, allocatable, codimension[:] :: allocatableField
+    !ERROR: Coarray components must be ALLOCATABLE and have a deferred coshape
+    real, codimension[:] :: deferredField
+    !ERROR: 'pointerfield' may not have the POINTER attribute because it is a coarray
+    !ERROR: Coarray components must be ALLOCATABLE and have a deferred coshape
```
tskeith, line 15: It would be nice if the wording of these two messages were similar. For example, the second could be:
`Component 'pointerfield' must have the ALLOCATABLE attribute because it is a coarray`

lib/Semantics/resolve-names.cpp
```diff
@@ -3821,14 +3821,36 @@ void DeclarationVisitor::Post(const parser::ComponentDecl &x) {
...
+            Say("A coarray component may not be of C_PTR or C_FUNPTR from "
+                "ISO_C_BINDING when an allocatable object is a "
+                "coarray"_err_en_US);
+          }
+        }
+      }
+      if (FindCoarrayUltimateComponent(*derived)) { // C748
+        if (attrs.HasAny({Attr::POINTER, Attr::ALLOCATABLE})) {
+          Say("A component whose type has a coarray ultimate component "
+              "cannot be a POINTER or ALLOCATABLE"_err_en_US);
```
tskeith, line 3847: We have mostly been using "may not" rather than "cannot".

lib/Semantics/check-declarations.cpp
```diff
@@ -385,12 +385,16 @@ void CheckHelper::CheckObjectEntity(
...
     if (IsAllocatable(symbol)) {
-      if (!details.coshape().IsDeferredShape()) { // C827
+      if (!details.coshape().IsDeferredShape()) { // C746, C827
         messages_.Say(
             "ALLOCATABLE coarray must have a deferred coshape"_err_en_US);
       }
+    } else if (symbol.owner().IsDerivedType()) { // C746
+      messages_.Say(
+          "Coarray components must be ALLOCATABLE and have a deferred "
+          "coshape"_err_en_US);
```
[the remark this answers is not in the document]
klausler: Or just combine the tests and use the second message, it's got all the information.

lib/Semantics/check-declarations.cpp
```diff
@@ -385,12 +385,16 @@ void CheckHelper::CheckObjectEntity(
...
     } else {
-      if (!details.coshape().IsAssumedSize()) { // C828
+      if (!details.coshape().IsAssumedSize()) { // C746, C828
```
klausler, line 397: C746 applies to components, but components are all handled earlier.

lib/Semantics/resolve-names.cpp
```diff
@@ -3821,14 +3821,36 @@ void DeclarationVisitor::Post(const parser::ComponentDecl &x) {
...
               "POINTER or ALLOCATABLE"_err_en_US);
         }
       }
+      if (!coarraySpec().empty()) { // C747
+        if (IsTeamType(derived)) {
+          Say("A coarray component may not be of type TEAM_TYPE from "
+              "ISO_FORTRAN_ENV"_err_en_US);
+        } else {
+          if (IsIsoCType(derived)) {
+            Say("A coarray component may not be of C_PTR or C_FUNPTR from "
```
klausler, line 3838: Missing the word "type", or needs rewording.

lib/Semantics/resolve-names.cpp
klausler: I also use "must not", but I shouldn't -- "may" is the correct modal verb here, since we're talking about what's allowed, not what might be possible.

You might as well capture the result of `FindCoarrayUltimateComponent` and use its name in the message and link to its declaration.

lib/Semantics/check-declarations.cpp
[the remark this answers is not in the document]
psteinfeld: Thanks, Tim.  I'll change the code to remove that phrase when it doesn't apply.

test/Semantics/allocate11.f90
```diff
@@ -5,19 +5,6 @@
 
 ! Rules I should know when working with coarrays and derived type:
 
-! C736: If EXTENDS appears and the type being defined has a coarray ultimate
-! component, its parent type shall have a coarray ultimate component.
```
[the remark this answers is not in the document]
psteinfeld: C736 is specific to the definition of a component of a derived type, and all of the tests in allocate11.f90 were allocate statements.  I added the test resolve86.f90 in my previous pull request to cover the case of defining a component that contains a coarray ultimate component without having a parent type that contains a coarray ultimate component.  This may have happened after you reviewed and approved that pull request.

test/Semantics/resolve88.f90
psteinfeld: Sounds good.  I'll change it.

lib/Semantics/resolve-names.cpp
psteinfeld: Thanks.  I'll fix it.

psteinfeld: I notice that there are many uses of the word "cannot" in resolve-names.cpp.  @tskeith, should I fix them all in this pull request?

tskeith: I think it makes more sense to change them in a separate pull request, but if you want to do it in this one that's okay too.

lib/Semantics/check-declarations.cpp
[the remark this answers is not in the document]
psteinfeld: @klausler, I think it's a little confusing to have the message tell the user that the declaration must have a deferred coshape when it already has one.  I plan to stick with Tim's suggestion.

lib/Semantics/check-declarations.cpp
psteinfeld: Good catch!  This also applies to the comment above.  I'll remove the spurious references.

lib/Semantics/resolve-names.cpp
psteinfeld: Thanks for catching this.  I'll fix it.

lib/Semantics/resolve-names.cpp
psteinfeld: I think I'll wait.

lib/Semantics/resolve-names.cpp
```diff
@@ -3835,20 +3835,25 @@ void DeclarationVisitor::Post(const parser::ComponentDecl &x) {
...
                 "ISO_C_BINDING when an allocatable object is a "
                 "coarray"_err_en_US);
           }
         }
       }
-      if (FindCoarrayUltimateComponent(*derived)) { // C748
+      if (auto it{FindCoarrayUltimateComponent(*derived)}) { // C748
+        Message declMsg{Say(it->name(),
+            "Type '%s' has coarray ultimate component '%s' declared here"_en_US,
+            declType->AsFortran(), it.BuildResultDesignatorName())};
```
tskeith, line 3847: Does this work correctly if the coarray component is declared in a module that is read in from a .mod file?

PR #1101, opened 2020-04-06
Add message formatting for std::int64_t

tskeith:
There is no printf formatting string for `std::int64_t`. Instead we have
to cast to `std::intmax_t` and use `%jd`. This change simplifies that by
automatically converting `std::int64_t` to `std::intmax_t `when formatting
messages.

This follows the pattern of `std::string` and `parser::CharBlock`, which are
formatted with `%s`.


tskeith:
> Might as well catch and convert `uint64_t` too, for that type and `size_t`, for the `%z` formats.

I've added a conversion from std::uint64_t to std::uintmax_t. We don't have any instances of that yet, but we may someday.

PR #1102, opened 2020-04-07
Add dependencies on LLVM utilities (FileCheck) for running flang tests

joker-eph:
Without this, `ninja check-flang` fails with:

llvm-project/llvm/utils/lit/lit/llvm/subst.py:134: fatal: Did not find FileCheck in...


joker-eph:
It works if you have a prebuilt LLVM or if you run `ninja check` and not `ninja check-flang` specifically.

PR #1104, opened 2020-04-08
Add dependency on libSemantics to libEvaluate to solve linker issues.

DavidTruby:
When building with LLVM HEAD, the unittests fail to link with ld.bfd
or ld.gold due to a linker ordering issue. Adding libSemantics as a
dependency to libEvaluate seems to fix that issue.

### DIFF
--- a/lib/Evaluate/CMakeLists.txt
+++ b/lib/Evaluate/CMakeLists.txt
@@ -32,6 +32,7 @@ target_compile_features(FortranEvaluate PUBLIC cxx_std_17)
 target_link_libraries(FortranEvaluate
   FortranCommon
   FortranDecimal
+  FortranSemantics
   FortranParser
 )
 


tskeith:
By the way, I was seeing the problem without explicitly requesting ld.bfd or ld.gold. Maybe it was using one of them by default, but I don't know how to tell. This was with GCC 8.4 on ubuntu 18.04.

DavidTruby:
BFD is the default on Ubuntu, unless its been changed on that machine or explicitly requested on the build line, so that would make sense.

bcfriesen:
Hi,

this patch works, I can build flang clang 11 dev build with no errors. Thanks @DavidTruby!

DavidTruby:
Committed directly into llvm-project: https://github.com/llvm/llvm-project/commit/a1726e65bc1fdbd2800839e48e68cb4a68cc7092

PR #1112, opened 2021-12-10
change master to main in branch references

jeffhammond:
https://github.com/llvm/llvm-project/tree/main/flang is the correct reference here.


p-costa:
duplicate of #1109

